PR #6, opened 2014-12-30
Add database and service model

robyoung:
This commit integrates SQLAlchemy and introduces a Service model.
This follows the Flask application layout described in Flask Web
Development [1].

The tests have been refactored into more traditional class based tests
because managing the state (for authorization and setting up the
database) was getting confusing with functions and decorators.
The Authorization header is provided by default and an option is
provided to switch this off for a test as authorised access is the
normal test case.

[1] https://github.com/miguelgrinberg/flasky


.ebextensions/dbmigrations.config
```diff
@@ -0,0 +1,3 @@
+container_commands:
+  upgrade:
+    command: "python application.py db upgrade"
```
TheDoubleK: I've created the DB using this command, but would be nice to add details of database setup to the README.  (Also... how do you run the tests?  The run_tests.sh script is not happy.)

config.py
```diff
@@ -6,6 +6,9 @@
 class Config:
     CONFIG_PROPERTY = "some_property"
```
TheDoubleK: We can probably lose this property?

robyoung: Yes, I'll remove it.

PR #8, opened 2015-01-14
	Adds most CRUD functionality to services resource

robyoung:
Adds views to list existing services, create new ones and update existing
ones. It does not add the ability to delete an existing service as this
is not yet needed.

In doing this a few things have changed.

I have changed the error handlers to return JSON and removed the error
handler html templates. I have done this because at the moment we are
building a JSON API, so our error responses should also be in JSON. I
have also replaced the root view with one that returns a JSON object
that links to the /services resource and removed it's template. We have
an intention to build a hypermedia API, part of that is clients being
able to discover the resources they can interact with on their own.

I have refactored the tests to pull application tests into their own
package and split tests on the /services resource out into their own
module. I've done this to better reflect the structure of the
application in the tests and because the prevous test module was getting
quite large. One strange thing I have done is to use a mixin to provide
some common tests to each endpoint that accepts JSON data. We will see
if this is a model that works well.

I have NOT done any validation of the services JSON itself. We have
another story going on to validate against a JSON schema.

Also refactors authentication so that it is required on all views rather than being added explicitly to each view. We require authentication on all views so adding it explicitly creates a potential for something to forget.


app/lib/authentication.py
```diff
@@ -4,20 +4,14 @@
...
-
-            if not incoming_token:
-                abort(401)
-            if not token_is_valid(incoming_token):
-                abort(403)
-
-        return view(*args, **kwargs)
-
-    return wrapped_view
+def requires_authentication():
```
TheDoubleK: :+1:  Yes I forgot to add this in my branch until it was very nearly finished.  Definitely makes sense to require auth for everything, certainly for the time being.

TheDoubleK:
The Python 3.4 Travis build hung for ten minutes and was terminated, though the 2.7 run was fine.

app/main/views.py
```diff
@@ -47,8 +52,67 @@ def get_iaas():
...
+
+
+def get_json_from_request():
+    if request.content_type != 'application/json':
+        abort(400, "Unexpected Content-Type, expecting 'application/json'")
+    data = request.get_json()
+    if data is None:
+        abort(400, "Invalid JSON; must be a valid JSON object")
+    if 'services' not in data:
+        abort(400, "Invalid JSON must have a 'services' key")
```
TheDoubleK: Are we requiring this for the POST as well as the responses? I will need to change the schemas if so, as they don't include the "services" wrapper.

robyoung:
Closed in favour of #10

PR #9, opened 2015-01-15
Validating json

TheDoubleK:
Adds endpoint that will validate POSTed JSON against schemas for the four lots.


app/main/views.py
```diff
@@ -3,6 +3,13 @@
 from . import main
 from ..lib.authentication import requires_authentication
 from ..models import Service
+from flask import abort
+from flask import make_response
+from flask import render_template
+from flask import request
```
robyoung: these can be collapsed into `from flask import abort, make_response, render_template, request`

schemata/g6-iaas-schema.json
```diff
@@ -0,0 +1,1028 @@
+{
```
robyoung: Hmm, I'm not sure about schemata. While strictly correct in terms of Greek roots it's not the common pluralisation in technical uses. For example, database schemas. See: http://english.stackexchange.com/questions/77764/plural-form-of-schema

tests/test_json_validation.py
```diff
@@ -0,0 +1,42 @@
...
+from jsonschema import validate
+from app.services.g6importService import validate_json
+
+
+def test_all_schemas_are_valid():
+    for file_name in os.listdir('schemata'):
+        if os.path.isfile('schemata/%s' % file_name):
+            print('Testing schema: %s' % file_name)
+            json_schema_file = open('schemata/%s' % file_name)
+            assert check_schema(json.load(json_schema_file))
```
robyoung: File access should be done in a context manager so that it gets closed correctly
```python
with open('schemata/%s' % file_name) as json_schema_file:
  assert check_schema(json.load(json_schema_file))
```

app/main/views.py
```diff
@@ -3,6 +3,13 @@
...
+
+from app import services
+from app.services import g6importService
```
robyoung: These imports are a bit wonkey. In other cases we import the function / class from the module rather than just the module. So, for consistency, I'd recommend the second import style. This will mean changing line https://github.com/alphagov/digitalmarketplace-api/pull/9/files#diff-8b9538e63b0d5c7fa84a61f3b8e7ec6dR80

app/services/g6importService.py
```diff
@@ -0,0 +1,62 @@
+import json
+import jsonschema
+from jsonschema import validate
```
robyoung: These imports are a bit strange. I'd recommend adding `ValidationError` to the second import and then try catch blocks later in the module.

tests/test_json_validation.py
```diff
@@ -0,0 +1,42 @@
+import os
+import json
+import jsonschema
+from jsonschema import validate
```
robyoung: Same point about imports here. I'd recommend adding `SchemaError` and `ValidationError` to the second line and changing the try catch blocks later in the module.

app/main/views.py
```diff
@@ -48,7 +59,26 @@ def get_iaas():
...
     return jsonify(id=id, data=service.data)
+
+
+# Test this locally by running the command:
+# curl -i -H "Content-Type: application/json" \
+# -H "Authorization: Bearer myToken" \
+# -X POST \
+# -d @example_listings/SSP-JSON-SCS.json \
+# 127.0.0.1:5000/g6/service/add
+@main.route('/g6/services', methods=['POST'])
```
robyoung: Function level documentation is usually done with [docstrings](https://www.python.org/dev/peps/pep-0257/)

app/services/g6importService.py
```diff
@@ -0,0 +1,62 @@
+import json
```
robyoung: Naming conventions in python tend towards `snake_case` rather than `camelCase`. For variable names this is just a convention for module names this is prescribed for module names. See: https://www.python.org/dev/peps/pep-0008/#prescriptive-naming-conventions

schemata/g6-iaas-schema.json
TheDoubleK: The Sqlalchemy package uses schemata :)
But yeah I felt like a bit of an arse when I made the folder in the first place and will change it.

robyoung:
Closed in favour of #10

PR #12, opened 2015-01-20
Import script

robyoung:
Make the import script parallel so that importing can happen a bit quicker.


scripts/import.py
```diff
@@ -23,13 +25,21 @@ def list_files(directory):
...
                     "content-type": "application/json",
                     "authorization": "Bearer {}".format(access_token),
                 })
-            print(response)
+            return response
+
+    pool = multiprocessing.Pool(10)
+
+    for result in pool.imap(put_file, list_files(listing_dir)):
+        print(result)
```
TheDoubleK: In the case of error responses (4xx, 5xx) it would be useful to print the filename of the failed upload.

TheDoubleK:
I'm having trouble running this against both staging and production:
Staging: `requests.exceptions.SSLError: [Errno 1] _ssl.c:510: error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed`

Prod: `requests.exceptions.SSLError: [Errno 1] _ssl.c:510: error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed`

minglis:
Merging this, note that the requirement to supply certs may be removed once we have real certs for all environments.

PR #13, opened 2015-01-20
Generate service IDs externally

robyoung:
We have decided to use the IDs generated in the SSP to identify the
services. For the moment we will not have a defined way for this API to
generate IDs itself. As such POSTing doesn't make sense. The import will
use PUT and put the SSP ID in the URL. So, changes include:

- POST /services endpoint now returns a 501 Not Implemented with an
  explanatory message.
- PUT /services/{service_id} can now create new services as well as
  update them.


app/main/views.py
```diff
@@ -87,25 +85,26 @@ def update_service(service_id):
     db.session.add(service)
     db.session.commit()
 
-    return jsonify(services=jsonify_service(service)), 200
+    return jsonify(services=jsonify_service(service)), http_status
```
TheDoubleK: A possible alternative to returning the whole JSON in the response might be to instead set the "Content-Location" header to the URL where the created/updated resource can be accessed.   e.g. a PUT to `/services/12345` would return 
`200 OK
Content-Location: /services/12345`

migrations/versions/4cb7869866dc_.py
```diff
@@ -0,0 +1,28 @@
...
+revision = '4cb7869866dc'
+down_revision = '8d2b9dcce09'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('services', sa.Column('service_id', sa.BigInteger(), nullable=True))
```
TheDoubleK: Service ID probably shouldn't be nullable?

migrations/versions/4cb7869866dc_.py
```diff
@@ -0,0 +1,28 @@
...
+    op.create_index(op.f('ix_services_service_id'), 'services', ['service_id'], unique=False)
```
TheDoubleK: ...and service_id should probably have unique=True?

migrations/versions/4cb7869866dc_.py
robyoung: Yes to this comment and the next one. Thanks.

app/main/views.py
robyoung: I think that works with a `Location` header for `POST` where you need to direct the client to a different URL but I don't agree that that either is ok for `PUT`. The resource we've set isn't located at a different URL, it's at the URL we've just interacted with.
I found [this link](https://www.subbu.org/blog/2008/10/location-vs-content-location) useful.

robyoung: If we don't want to send back loads of data for each `PUT` request we could just return a `204 No content` and similarly not return a response for `201 Created`.
see: http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.6

migrations/versions/4cb7869866dc_.py
robyoung: I'm going to squash this change in to the previous commit so that we don't either get an needless extra migration or a commit in the history that breaks the database.

app/main/views.py
TheDoubleK: I don't have a strong opinion on this either way - when we were going to be generating a new ID it made sense to me to return the created artefact, but I'm not sure now if clients will do anything with the response?  If not, then it seems just returning a status code might be all that's needed.

app/main/views.py
```diff
@@ -87,25 +85,26 @@ def update_service(service_id):
...
 
-    return jsonify(services=jsonify_service(service)), 200
+    return "", http_status
 
 
 @main.route('/services/<service_id>', methods=['GET'])
 def get_service(service_id):
-    service = Service.query.filter(Service.id == service_id).first_or_404()
+    service = Service.query.filter(Service.service_id == service_id)\
+                           .first_or_404()
```
TheDoubleK: I guess first_or_404 will do the job here, but we'll need something more sophisticated (i.e. select the one that is active or 404) once we have the possibility of multiple versions of a service being stored.

robyoung: Yea, I've not added any of that stuff in this pull request. I'm hoping we can hold off on that stuff until after launch.

TheDoubleK: Yep makes sense - we don't need it for launch.

TheDoubleK:
Should we also drop the 'services_id_seq' from the services table here?

TheDoubleK:
I'm going to merge this - I think we should drop 'services_id_seq' as part of a future story, maybe the same one where we add 'supplier_id' as a field.

PR #14, opened 2015-01-20
Add pagination to service listing

robyoung:
For review only, depends on #12 and #13. Please do not merge until those pull requests have been merged and this one has been rebased on them.

After discussion with @TheDoubleK and @minglis we have decided to put
the pagination links in the response under the 'links' section. This is
beacause it's more obvious to a user browsing the API than other
solutions such as Link headers.

Page size is not controllable by the user, it's simpler to just define
it and let them paginate.

### DIFF
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -1,5 +1,6 @@
 from flask import (jsonify, Response, abort, render_template,
-                   request, url_for)
+                   request)
+from flask import url_for as base_url_for
 
 from . import main
 from .. import db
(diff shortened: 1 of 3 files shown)


allait:
Are we sure about returning `{"href": null, "rel": "prev"}` vs not returning the link at all? I've looked at the GitHub API as an example, and they don't seem to return the prev link for page 1 (https://developer.github.com/v3/#json-p-callbacks), don't really know about others.

Other than that, LGTM :+1:

robyoung:
That's a fair point, I think I pulled it from a Spring HATEOAS example. I agree, less is more, if it doesn't make sense don't show it.

PR #15, opened 2015-01-26
Add new db fields

TheDoubleK:
In order to support a select-all-services-for-supplier API endpoint it will be useful to include supplier_id as an explicit field in the database. 

See: https://www.pivotaltracker.com/story/show/86911238


robyoung:
Should this have any tests?

robyoung:
Ooops, soz, didn't mean to close then.

robyoung:
Also, would it make sense to use SQLAlchemy's `default` and `onupdate` column attributes? That way the update / create behaviour is tied to the model and takes effect no matter how the model is updated / created.

See:
http://stackoverflow.com/questions/8277278/in-sqlalchemy-how-do-i-create-a-mysql-date-column#answer-8278639

TheDoubleK:
Tests... is that a rhetorical question?

I'll have a look at `default` and `onupdate` - now.

TheDoubleK:
The `onupdate` isn't working for me here - it doesn't seem to be captured by Alembic in the generated migration script.  It may be possible to add by hand, but I'm not sure it's necessary - if we think the PUT endpoint will be the way services get updated is there any harm in updating the datetime field here instead?

robyoung:
I think it should have at least tests that verifies that the correct value is being put into the `service_id` field and that the temporal fields are updated correctly. They seem pretty tiny and obvious now but it's more to catch regressions in the future.

robyoung:
I don't think it should be picked up by Albemic, I think it's just handled in the model. When you say it isn't working for you, does that mean it isn't updating the value when you save through the model?

The reason I'm nervous about having that logic in the view layer is that it's specific logic that we need to remember to implement in potentially multiple places. For example, when we introduce supplier as a model we will need the same logic in the view for that. If we ever introduce the ability to `PATCH` or update specific parts of resources (which seems reasonable given how large and complex they are) we would also need to implement the logic there as well.

robyoung:
if you rebase this on master now it should bring in the style fixes in the ES import script

TheDoubleK:
We will need to drop all service data from staging & prod databases before running the DB upgrade.

TheDoubleK:
> When you say it isn't working for you, does that mean it isn't updating the value when you save through the model?

Yes, exactly that - the updated_at is not updated when onupdate is specified in the model.

robyoung:
I have no idea, I've given it a try (for good while) and can't get it to work either. It does look like it is how other [people are doing it though](https://github.com/hasgeek/coaster/blob/master/coaster/sqlalchemy.py#L73). What about if we get a couple of tests in to verify the behaviour is working as expected, then, after launch we can look at refactoring how `updated_at` and `created_at` are set?

PR #20, opened 2015-01-29
Remove example service routes

robyoung:
These were a temporary addition for testing before we had loaded the
services. They are no longer needed.


TheDoubleK:
Should we remove the JSON files that were served too? Or might it be useful to have them there still for reference?

robyoung:
yes good call

PR #21, opened 2015-01-29
Add a status endpoint

robyoung:
Add a status endpoint that can be used by the ELB to determine whether
the app is healthy. I have chosen not to touch the database on this
status endpoint as that is shared between all EC2 instances and
therefore a slow database would make all instances look like they're
slow.

### DIFF
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -20,6 +20,11 @@ def index():
     ]), 200
 
 
+@main.route('/_status')
+def status():
+    return jsonify(status="ok")
+
+
 @main.route('/services/g6-scs-example')
 def get_scs():
     content = render_template('sample_static_responses/sample-scs.json')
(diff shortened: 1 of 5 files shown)


app/main/views.py
```diff
@@ -20,6 +20,11 @@ def index():
     ]), 200
 
 
+@main.route('/_status')
+def status():
+    return jsonify(status="ok")
```
TheDoubleK: Is there any more useful information we could add to this? Can we get at current build number? Or some info about the beanstalk instance/environment to identify exactly which app is up?

robyoung:
@allait just pointed out that this won't work as we have authentication on all routes, closing until I have a solution.

TheDoubleK:
Yep this works.  We can add more information to the response later if necessary.

PR #22, opened 2015-01-30
Add an error handler for 403 errors

robyoung:
Previously this was not implemented resulting in forbidden requests
responding with the default HTML response. I have added the
invalid token in the response so that requesters can validate that what
they think they sent is what is getting through. It is not a security
risk putting it in the response as they have just provided it to us,
however, we should be carefull not to log it.


TheDoubleK:
I love it when you can get an application to echo rude words back to you.

PR #26, opened 2015-02-06
Increase page size for services fetch to 100

TheDoubleK:


allait:
I feel like stubbing the `API_FETCH_SIZE` with a lower value in tests might be a good idea. I've run the `run_tests.sh` and it takes ~0.8s longer with this change, which is not that bad, but I think it's better to keep the test run time as low as possible right from the start.

Other than that the change looks good and listing 100 services per response seems like a good idea, especially for our ES export scripts.

allait:
Merging this now, we'll add the fetch size stubbing later

PR #29, opened 2015-02-16
Build sqlalchemy uri from rds

robyoung:
When running a Beanstalk environment the RDS credentials are passed to
the app via a series of environment variables. The SQLAlchemy
environment variable is not provided but it can be created from the RDS
variables. This is a backwards compatible change that will generate the
SQLAlchemy URI from the RDS environment variables if it has not been
explicitly set in the environment.

### DIFF
--- a/config.py
+++ b/config.py
@@ -1,4 +1,6 @@
 import os
+import re
+from .app.helpers import convert_to_boolean
 
 basedir = os.path.abspath(os.path.dirname(__file__))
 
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -27,6 +52,12 @@ class Live(Config):
     DEBUG = False
 
 
+def config_attrs(config):
+    """Returns config attributes from a Config object"""
+    p = re.compile('^[A-Z_]+$')
+    return filter(lambda attr: bool(p.match(attr)), dir(config))
```
allait: Have you decided against the `str.isupper` or is this outdated?

robyoung:
Closing after discussion with @allait, reasons include; It's nice not to be coupled to RDS, it's not needed now that we're not going with the eb command line tools.

PR #32, opened 2015-02-17
Disallow additional fields in service JSON schema

allait:
Ignores fields we add to the service response ourselves (`links` and `supplierName`) and returns 400 for everything else.

`links` and `supplierName` are not saved to the service.data JSON.


json_schemas/g6-iaas-schema.json
```diff
@@ -2,6 +2,7 @@
   "id": "G6 Submissions IaaS Schema",
   "$schema": "http://json-schema.org/schema#",
   "type": "object",
+  "additionalProperties": false,
```
TheDoubleK: :+1: 
Yes we should probably have had this in from the beginning.

PR #36, opened 2015-02-26
Mi adding g5 filter migrations

minglis:
Added the mappings for more data fields to allow for filtering of G6 data,

### DIFF
--- a/scripts/process-g6-into-elastic-search.py
+++ b/scripts/process-g6-into-elastic-search.py
@@ -169,21 +169,34 @@ def attributes(data):
     ###################
 
     ## multi selects
-    if "guaranteedResources" in data and data["guaranteedResources"]:
+    if "guaranteedResources" in data and data["guaranteedResources"] == True:
         attributes.append({"name" : "lot2q3",  "lot2q3" : "guaranteed"})
-    elif "guaranteedResources" in data and not data["guaranteedResources"]:
-        attributes.append({"name" : "lot2q3",  "lot2q3" : "non-guaranteed"})
+    elif "guaranteedResources" in data and data["guaranteedResources"] == False:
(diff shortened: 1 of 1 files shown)


scripts/process-g6-into-elastic-search.py
```diff
@@ -60,6 +60,173 @@ def category_name_to_id(name):
...
+    elif "persistentStorage" in data and not data["persistentStorage"]:
+        attributes.append({"name": "lot1q4",  "lot1q4": "nonpersistent"})
+
+    attributes.append(boolean_attribute("elasticCloud", "lot1q2",
+                                        attributes, data))
+
+    return attributes
+
+
+def boolean_attribute(g6_field_name, g5_field_name, attributes, data):
```
TheDoubleK: The 'attributes' parameter isn't used in this method - should it be?

scripts/process-g6-into-elastic-search.py
```diff
@@ -60,6 +60,173 @@ def category_name_to_id(name):
...
+        if "Uptime Institute Tier 4" in data["datacentreTier"]:
+            attributes.append({"name": "q32",  "q32": "tier4uptimeinstitute"})
+
+    # provising time is 1 Day etc in G6, but the question is
+    # "documented?" in G5 so any value provokes true
+    if "provisioningTime" in data:
+        attributes.append({"name": "q40",  "q40": "true"})
+
+    # ##################
+    # ## PaaS / IaaS ###
```
TheDoubleK: Isn't this just PaaS (lot 2), with the IaaS versions of the same questions below?  Should it be conditional on the lot of the current G6 service which of the two filters gets added, rather than adding filters for two different lots to a single service?

minglis: yep, will fix this.

PR #37, opened 2015-02-26
Remove duplicate attributes

TheDoubleK:
Some attributes (provisioningTime, guaranteedResources, persistentStorage) were duplicated in the g6 import script.

### DIFF
--- a/scripts/process-g6-into-elastic-search.py
+++ b/scripts/process-g6-into-elastic-search.py
@@ -164,33 +164,11 @@ def attributes(data):
         if "Uptime Institute Tier 4" in data["datacentreTier"]:
             attributes.append({"name": "q32", "q32": "tier4uptimeinstitute"})
 
-    # provising time is 1 Day etc in G6, but
+    # provisioning time is 1 Day etc in G6, but
     # the question is "documented?" in G5
     # so any value provokes true
     if "provisioningTime" in data:
         attributes.append({"name": "q40", "q40": "true"})
-
-    #
-    # PaaS
-    #
-
-    if "lot" in data and data["lot"] == "PaaS":
-        # multi selects
-        if ("guaranteedResources" in data and
-                data["guaranteedResources"] is True):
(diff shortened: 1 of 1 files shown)


scripts/process-g6-into-elastic-search.py
```diff
@@ -164,13 +164,31 @@ def attributes(data):
...
     # the question is "documented?" in G5
     # so any value provokes true
     if "provisioningTime" in data:
         attributes.append({"name": "q40", "q40": "true"})
 
     #
+    # IaaS
+    #
+
+    if "lot" in data and data["lot"] is "IaaS":
```
robyoung: just fyi really, could be rewritten as
```python
if data.get("lot") is "IaaS":
```

scripts/process-g6-into-elastic-search.py
```diff
@@ -164,13 +164,31 @@ def attributes(data):
...
+        # multi selects
+        if "guaranteedResources" in data and data["guaranteedResources"]:
+            attributes.append({"name": "lot1q3", "lot1q3": "guaranteed"})
+        elif "guaranteedResources" in data and not data["guaranteedResources"]:
+            attributes.append({"name": "lot1q3", "lot1q3": "nonguaranteed"})
+
+        if "persistentStorage" in data and data["persistentStorage"]:
+            attributes.append({"name": "lot1q4", "lot1q4": "persistent"})
+        elif "persistentStorage" in data and not data["persistentStorage"]:
```
robyoung: similarly with these ones
```python
elif not data.get('persistentStorage', True):
```

TheDoubleK:
I'm just tidying up the mess left by @minglis here, not rewriting the whole thing :)

scripts/process-g6-into-elastic-search.py
allait: This really should be `data.get("lot") == "IaaS"`, `is` only works here thanks to short string cache.

PR #38, opened 2015-02-27
Update to run on port 5000

TheDoubleK:
For development we will want to run multiple apps, so they should each bind to a different port number.
The default port is 5000 anyway, but we should state the port explicitly in the code which is why I've added it here.


allait:
LGTM :+1:

PR #39, opened 2015-03-03
Update run script

TheDoubleK:
Modified to only set the AUTH_TOKENS environment variable if it has not already been set.


TheDoubleK:
I think I've included this in next PR, closing while I check the diffs...

PR #41, opened 2015-03-03
Prefix env vars with DM_ and update run scripts

TheDoubleK:
Two things here:
1. All the other environment vars we use are prefixed with 'DM_'.  For consistency and ease of checking the current environment for DM apps I think it is worth prefixing all our environment vars with 'DM_'.
2. The run scripts should set required vars to sensible defaults if they haven't been set.

We'll need to change the Beanstalk setup to reflect the new name if this goes in.


robyoung:
:+1: on prefixing environment variables with `DM_`.

If we deploy this as is to the current AWS it will break things unless we have also added the `DM_` version of each environment variable. Have we considered how we manage releasing this change? I see there being two options; add the `DM_` prefixed variables to the deployment tools and then merge this or change this PR to fall back to the original environment variable names if the `DM_` prefixed ones aren't there.

TheDoubleK:
I think add the new DM_AUTH_TOKENS to the beanstalk config, leave the old one there for now, and then delete from Beanstalk once the code has been deployed.  I can add to the config now if we're happy to go that way.

robyoung:
Sure, sounds good. We need to add it to the running environments and to whatever tooling we have for creating new environments (ie https://github.com/alphagov/digitalmarketplace-aws).

TheDoubleK:
I've added the DM_AUTH_TOKENS on staging, but held off production for now as saving the config changes reboots the instances, which would mean some downtime for the prod API.  We should pick a good time to do this.

TheDoubleK:
OK I also bumped prod - no downtime as far as I could tell - AWS seems to manage it nicely, even if the environment is grey for a while during the update.

TheDoubleK:
Closing for now to implement newly agreed scheme for Env vars

PR #45, opened 2015-03-11
Mi data audit

minglis:
This story is to deliver the data audit story:

[https://www.pivotaltracker.com/story/show/89081984](https://www.pivotaltracker.com/story/show/89081984)

Basic premise is that there is a POST API endpoint which will allow us to update fields in the JSON data the describes a service. This action will have the side-effect creating an archived version of the previous state, such that we can discover what a service was at any point in time.

Story does the following things:
1) Changes to the database - **IF THIS NEEDS IT'S ON PULL REQUEST I'LL SPLIT IT**:
- adds **updated_by** and **updated_reason** columns to the Services table;
- populates the existing rows with default values for these new columns, which then are not-null
- adds a new table ServicesArchive

2) Changes to the PUT API
- Now will only allow a **single** PUT against a service to create the initial row.
- Any subsequent PUT against that service will result in a 409 Conflict response
- PUT does **not** make an archived entry
- requires an extra piece of JSON to describe who imported and why:

    { updated_details: { updated_by: '_username_', updated_reason: '_the whys_' }}

- import script populates this new JSON which current UNIX user and standard reason - no change to command line args

3) Changes to POST API
- POST API enabled
- Requires **updated_details** JSON as described above
- Takes second JSON block where any key value pair can be supplied

    {services: {key: value}}

- Value can be any valid JSON - objects, lists, strings, etc
- Logic is to load service from the database, update the JSON with the supplied values and then valid against the JSONParse schemas we use on import.
- Successful validation will lead to the creation of the archive row and the update of the *live* row.
- URL

    POST /services/{:service_id}

4) New API endpoints for the archive
- The ServiceArchive table is addressable through two new endpoings

To get an individual row from the archive:

    GET /service-archive/{:pk}

To get all rows for a service:

    GET /service-archive?service-id={:service-id}

5) Explorer page for API testing
- Added a new page

    GET /_explorer

- This page is a Bootstrap/JQuery top on the API endpoints for in-browser testing.
- You're all welcome to make it better!


quis:
:+1: Excellent pull request.

In `GET /service-archive/{:pk}`, what does `pk` stand for?

Can you update more than one key/value pair at once, ie do you give an `updated_reason` for each individual value or for the action of saving a "page" of edits?

robyoung:
I'm reviewing this at the moment and have a few comments. Could we hold off merging until I've got them together please.

scripts/import.py
```diff
@@ -64,6 +67,7 @@ def __call__(self, file_path):
                 "authorization": "Bearer {}".format(self.access_token),
             },
             verify=self.cert if self.cert else True)
+        print(response.text)
```
robyoung: Is this supposed to be here?

robyoung:
NITPICK: Some of the commit comment styling could be improved a bit. They don't display well in github. https://github.com/alphagov/styleguides/blob/master/git.md#commit-messages

.gitignore
```diff
@@ -2,3 +2,5 @@ venv
 .elasticbeanstalk/
 *.pyc
 __pycache__
+.idea/
```
allait: Can we keep these in the personal `.gitignore` since it's not really related to the project? @TheDoubleK has removed them from a few projects already.

app/main/views.py
```diff
@@ -99,6 +165,43 @@ def get_service(service_id):
...
+@main.route('/services-archive/<int:service_archive_id>', methods=['GET'])
+def get_archived_service(service_archive_id):
+    """
+    Retrieves a service from the service_archive by PK
+    :param service_archive_id:
+    :return: service
+    """
+
+    try:
+        int(service_archive_id)
```
allait: This should be already validated by the route's `<int: ...>`

app/main/views.py
```diff
@@ -99,6 +165,43 @@ def get_service(service_id):
...
+        abort(400, "Invalid service id supplied")
+
+    service = ServiceArchive.query.filter(
+        ServiceArchive.id == service_archive_id
+    ).first_or_404()
+
+    return jsonify(services=jsonify_service(service))
+
+
+def prepare_archived_service(service):
```
allait: This looks like it could be a `ServiceArchive.from_service(service)`

app/main/views.py
```diff
@@ -146,13 +249,13 @@ def pagination_links(pagination, endpoint, args):
     ]
 
 
-def get_json_from_request():
-    if request.content_type != 'application/json':
+def get_json_from_request(key):
```
allait: This is almost everywhere called as `get_json_from_request(key)[key]`, maybe it's worth renaming to e.g. `get_request_json_key` and returning `data[key]`

app/main/views.py
```diff
@@ -49,35 +49,101 @@ def list_services():
...
+        Uses existing JSON Parse routines for validation
+    """
+    service = Service.query.filter(
+        Service.service_id == service_id
+    ).first_or_404()
+
+    service_to_archive = prepare_archived_service(service)
+    update_json = get_json_from_request('update_details')['update_details']
+    validate_updater_json_or_400(update_json)
+    service_update = get_json_from_request('services')['services']
```
robyoung: I find this API a bit obscure. Earlier we'd discussed following (albeit rather weakly) the jsonapi approach. Here we're partially updating the services resource but we're calling the resource serviceUpdate in the payload which doesn't match how we're layout out our payloads elsewhere. I think the key should be services rather than serviceUpdate.

A much smaller point, but it had me a bit confused for a few seconds. I didn't initially realise that updater was just some additional metadata about the request itself. Maybe this kind of stuff should be collected together into a metadata section?

robyoung: This is effectively a PATCH request. I totally get that PATCH isn't a well supported HTTP method and therefore using POST against a specific resource URL for partial update is a good alternative. However, I think that the definition and pitfalls described by [RFC5789](http://tools.ietf.org/html/rfc5789) are useful. I think we should document our decision to take this route and commit to using it consistently.

app/main/views.py
```diff
@@ -49,35 +49,101 @@ def list_services():
...
+        db.session.commit()
+    except DatabaseError:
+        db.session.rollback()
+        abort(500, "Database error")
+
+    return jsonify(message="done"), 200
+
+
+@main.route('/services/<int:service_id>', methods=['PUT'])
+def import_service(service_id):
```
robyoung: This seems like a strange name, maybe `replace_service` or `overwrite_service`.

app/main/views.py
robyoung: I'm not sure this is the right pattern. It means that we do the first two checks repeatedly for each invocation (that may also include JSON deserialisation). Maybe pass keys in as an array and then use the returned dictionary? (I'm not convinced this is a better pattern).

```python
data = get_json_from_request(['updater', 'serviceUpdate'])
validate_updater_json_or_400(data['updater'])
service_update = data['service_update']
```

app/static/explorer.js
```diff
@@ -0,0 +1,103 @@
...
+
+    return headers;
+}
+
+function printHTMLResponse(jqXHR, obj) {
+    return "<i><h5>Response Status:</h5> " +  jqXHR.status +
+           "<p/><h5>Response Headers:</h5><pre>" + jqXHR.getAllResponseHeaders() +
+           "</pre><h5>Response Body:</h5><pre>" + JSON.stringify(obj, null, 4) + "</pre></i>"
+
+}
```
robyoung: This is not a text file :trollface:

migrations/versions/178754d7108a_.py
```diff
@@ -0,0 +1,41 @@
...
+    op.add_column('services', sa.Column('updated_reason', sa.String(), nullable=True))
+
+    services = table('services',
+                     column('updated_by', String),
+                     column('updated_reason', String)
+    )
+
+    op.execute(
+        services.update(). \
+        values({'updated_by': op.inline_literal('import'), 'updated_reason': op.inline_literal('initial import')})
```
robyoung: nice! `inline_literal` is useful

app/main/views.py
```diff
@@ -99,6 +165,43 @@ def get_service(service_id):
     return jsonify(services=jsonify_service(service))
 
 
+@main.route('/services-archive/<int:service_archive_id>', methods=['GET'])
```
robyoung: `services-archive` doesn't seem like the right name for the resource. We chose `/services` because the resource we're dealing with is `service` and we decided to go with plurals for our resource endpoints. Using the same train of thought for this; the individual resource we're talking about is a service archive so the resource endpoint should be `service-archives`.

app/main/views.py
```diff
@@ -99,6 +165,43 @@ def get_service(service_id):
...
+    try:
+        int(service_archive_id)
+    except ValueError:
+        abort(400, "Invalid service id supplied")
+
+    service = ServiceArchive.query.filter(
+        ServiceArchive.id == service_archive_id
+    ).first_or_404()
+
+    return jsonify(services=jsonify_service(service))
```
robyoung: I think the key here should be `services-archives` to match the resource name. You wouldn't be able to pass this in as a keyword argument though. Would have to be
```python
return jsonify(**{"services-archive": jsonify_service(service)})
```

app/main/views.py
```diff
@@ -99,6 +165,43 @@ def get_service(service_id):
     return jsonify(services=jsonify_service(service))
 
 
+@main.route('/services-archive/<int:service_archive_id>', methods=['GET'])
+def get_archived_service(service_archive_id):
```
robyoung: In fact, this seems like a much better resource name 'archived service'.

app/explorer/__init__.py
```diff
@@ -0,0 +1,5 @@
+from flask import Blueprint
+
+explorer = Blueprint('explorer', __name__)
```
robyoung: This feels like it should be a separate pull request.

app/main/views.py
```diff
@@ -49,35 +49,101 @@ def list_services():
...
     now = datetime.now()
     service = Service.query.filter(Service.service_id == service_id).first()
-    http_status = 204
-    if service is None:
-        http_status = 201
-        service = Service(service_id=service_id)
-        service.created_at = now
+
+    if service is not None:
+        abort(409, "Service already exists")
```
robyoung: I'm uncomfortable with this. It seems like we're adopting quite a non-standard approach to REST. If we're deviating from the norm we should have good (and documented) reasons for it.

minglis:
nice comments all, will pick them up.

app/main/views.py
minglis: granted. was in my mind deprecating this. on retrospect you're right. will kill this.

app/explorer/__init__.py
minglis: i stuck it in so that i could do the demo.

minglis: I'll leave it in this pull request. Doesn't feel worthwhile to pull it out.

app/main/views.py
minglis: I think I've missed a trick - maybe the table should be archive. But I've not thought it all through yet - how do we archive supplier changes? Do we need to in the same way? Is it auditing?

Like the naming though - I'll take you at this and update.

app/main/views.py
minglis: Below we talk about archived-services, which would make the table ArchivedServices, but would the endpoint be POST /archived-services?

I guess so. Maybe this would be cleaner.

app/main/views.py
quis: Have you considered something like:
- `/service/<service_id>/at/<change_id>` or `/service/<service_id>/archive/<change_id>`
- and `/service/<service_id>/archives?

Don't know if that fits with the API/REST conventions but feels cleaner to me.

tests/app/helpers.py
```diff
@@ -89,7 +98,7 @@ class JSONUpdateTestMixin(object):
 
     def test_non_json_causes_failure(self):
         response = self.client.open(
-            self.endpoint,
+            'services/1000',
```
allait: What's the reason for this? `JSONUpdateTestMixin` is supposed to be a mixin class, it shouldn't have a hardcoded URL in it.

tests/app/test_services.py
```diff
@@ -132,17 +132,521 @@ def first_by_rel(rel, links):
...
+                    'updated_by': 'joeblogs',
+                    'update_reason': 'whateves'},
+                 'services': {
+                     'serviceName': 'new service name'}}),
+            content_type='application/json')
+
+        assert_equal(response.status_code, 404)
+
+    def test_can_not_update_without_updater_details(self):
+        with self.app.app_context():
```
allait: Can we move this common "create a service" part to the class `setUp`?

tests/test_json_validation.py
```diff
@@ -20,6 +21,22 @@ def test_all_schemas_are_valid():
...
+    valid_updater = {'updated_by': 'this', 'update_reason': 'hi'}
+
+    assert_equal(validates_against_schema(
+        UPDATER_SCHEMA, invalid_updater_no_reason), False)
+    assert_equal(validates_against_schema(
+        UPDATER_SCHEMA, invalid_updater_no_username), False)
+    assert_equal(validates_against_schema(
+        UPDATER_SCHEMA, invalid_updater_no_fields), False)
+    assert_equal(validates_against_schema(
+        UPDATER_SCHEMA, valid_updater), True)
```
allait: It would be nice to stick to using test generators (instead of having a series of asserts) for things like this. That way each check will pass/fail independently.

app/main/views.py
allait: I think we should keep the `{"services": ...}` key in response since we're using it everywhere right now and the JSON response is the same as for normal services

minglis:
@quis pk meant primary key, i.e. most of our APIs use the service id as the reference, for this as there are many rows per service id, I used the primary key - trying to make the distinction between primary key and service id.

scripts/import.py
minglis: Nope - debugging. Good spot. Gone.

.gitignore
minglis: Done.

app/main/views.py
minglis: Done, didn't realize it did that. Nice.

app/main/views.py
minglis: Yep, done.

app/main/views.py
minglis: Looked at this - split it up so there are two methods (not completly happy and think there is a JSONParse template solution - but I'm pushed for time prior to holidays).

The existing method does basic checks, and the new method checks for keys.

app/static/explorer.js
robyoung: Damn! It is now. Now my troll face just looks silly

migrations/versions/178754d7108a_.py
minglis: didn't know what that was called thanks!

app/main/views.py
minglis: Happy to commit to the use  of POST. This is something for the manual - which I haven't written this up in. Hope to do so tomorrow.

app/main/views.py
minglis: Yeah - didn't rename this, it was the import method from the first draft of the API. Maybe it's insert? Not sure. Will leave it for now.

app/main/views.py
minglis: As we addressing a domain object - the service archive - that forms the root part of the URL, then it's unique id for a canonical row, or a query string for filtering.

We're not hitting the services table, hence not using that in the URL.

Happy to chat more, will not change it for now though.

tests/app/helpers.py
minglis: It didn't work with the 409 coming from the PUT methods. I'm going to revert the 409 change so this should go back.

app/main/views.py
minglis: OK, it's all renamed now - Table is ArchivedService, and all URLs are archived-services. Feels a lot nicer.

app/main/views.py
minglis: Right so updated back to as it was, the notion of PUT as a create but not allowing updating was in retrospect a mistake. Was viewing the POST as only method for updating post creation.

We need to think about the difference - and defining it - between PUT and POST.

tests/app/test_services.py
minglis: Done

tests/test_json_validation.py
minglis: Done.

PR #49, opened 2015-03-13
Preload jsonschema validators

robyoung:
Tests were taking 13s to run, most of the time being spent in
jsonschema. This commit pre-loads the jsonschema validators at startup
rather than just loading the actual JSON at startup. This has two
main benefits. It means that the app will fail to start if the JSON
schemas are invalid and it improves performance considerably. After this
commit the tests take a little over 6s to run and the bottleneck is now the database.

### DIFF
--- a/app/validation.py
+++ b/app/validation.py
@@ -2,26 +2,42 @@ import json
 
 from flask import abort
 from jsonschema import validate, ValidationError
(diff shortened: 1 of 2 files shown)


TheDoubleK:
:+1:  yep much faster here.

PR #50, opened 2015-03-20
Drop additional fields in service update requests

allait:
Service update should ignore `links` and `supplierName` fields since they're present in the GET response. This is what `PUT /services/...` does right now.

### DIFF
--- a/tests/app/test_services.py
+++ b/tests/app/test_services.py
@@ -466,6 +466,25 @@ class TestPostService(BaseApplicationTest):
                 self.service_id).get_data()
             assert_equal(len(json.loads(archived_state)['services']), 5)
 
+    def test_writing_full_service_back(self):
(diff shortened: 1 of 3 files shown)


robyoung:
Looks good, tests pass. I'll wait for travis to catch up and then :+1:

tests/app/test_services.py
```diff
@@ -466,6 +466,25 @@ def test_updated_service_should_be_archived_on_each_update(self):
                 self.service_id).get_data()
             assert_equal(len(json.loads(archived_state)['services']), 5)
 
+    def test_writing_full_servie_back(self):
```
TheDoubleK: Typo here:  servie -> service

allait: Well spotted, thanks. Fixed.

PR #51, opened 2015-03-23
Add version to the /_status endpoint

robyoung:
This is pulled from a file in the root of the application which is placed by the release tool.

This is for review only at this stage. I thought it would be a good way of proving the deployment pipeline in the showcase.


allait:
LGTM :+1:

PR #55, opened 2015-03-26
Mi update services database for data migration

minglis:
First part of delivering: https://www.pivotaltracker.com/story/show/91085102

in order to add the G5 services there needed to be a couple of alterations to the services database, this pull request delivers the core updates to enable G5 and G4 to be imported. It does not however change the API interfaces to utilize these new fields other than to support their population on import.

(1) New status column added.

the status column is to manage the various Boolean flags present on the G5/4 schema. These services had notions of:

    - expired - relates to the overall framework - expired services are hidden from all users

    - enabled - relates to admin control over visibility - disabled services hidden from suppliers and buyers but visible to admins

    - status - this was a string field that allowed suppliers to publish and unpublish services - unpublished are visible to admins and suppliers but not to buyers.

These have all been rolled up into a new *status* column that has a check constraint to allow one of three values, described below,

    - published: visible to buyers, suppliers and admins
    - enabled: visible to  suppliers and admins 
    - disabled: visible only to admins

there is no logic implemented in the API to utilize these yet. This is just for information at this point. The notion of expired will be in another pull request.

To implement this there is a change in the PUT endpoint to set all services to enabled. This will need to be expanded upon when we start to use this field in anger so that we don't have unexpected side effects.

There is a migration script to set all existing services to enabled, and to then make this field not null.

(2) Change type of service_id
Service Ids for G6 were integers, for G5 and G4 they are strings, so the DB column type has been changed to string. this impacted quite a lot of code

- import script now converts ints to strings prior to PUTing the requests
- endpoints that utilized the Flask path parameter type to handle validation now have to validate services_ids explicitly
    
    @main.route('/services/<int:service_id>', methods=['PUT'])

becomes

    @main.route('/services/<string:service_id>', methods=['PUT'])

so we need to do: 
    
    if not is_valid_service_id(service_id):
        abort(400, "Invalid service id supplied")

This impacted a lot of tests, both in terms of expectations and setups.

- new validation method to do this - valid char group is
   
    [A-z0-9-] Alphanumeric and dashes, G5/4 is of form 5-g5-0379-325


app/main/views.py
```diff
@@ -182,7 +197,7 @@ def get_service(service_id):
     return jsonify(services=jsonify_service(service))
 
 
-@main.route('/archived-services/<int:archived_service_id>', methods=['GET'])
+@main.route('/archived-services/<string:archived_service_id>', methods=['GET'])
```
allait: This should still be integer, `archived_service_id` is table primary key, not `service_id`. And we're probably missing a test somewhere since this went unnoticed.

migrations/versions/3d3fb04d1a98_.py
```diff
@@ -0,0 +1,36 @@
+"""empty message
```
allait: We're creating more and more migrations, so I think it's worth running `python application.py db migrate` with `-m MESSAGE` in the future. It will add a `_<message>` to the migration file name, which makes it easier to tell them apart.

I think we can rename these files safely too by adding something after `_` without breaking migration versions.

app/validation.py
```diff
@@ -3,7 +3,10 @@
 from flask import abort
 from jsonschema import validate, ValidationError
 from jsonschema.validators import validator_for
+import re
 
+minimum_service_id_length = 10
```
allait: PEP-8 recommends to use uppercase for global constant names

app/models.py
```diff
@@ -30,6 +32,8 @@ class Service(db.Model):
                                nullable=False)
     data = db.Column(JSON)
 
+    status = db.Column(db.String, index=False, unique=False, nullable=False)
```
allait: Shouldn't status be indexed? It seems that we're going to filter enabled/published services on almost every request.

allait: Also, maybe it's worth making an integer status code column instead of the string and then adding an enum of status names in the model?

robyoung: To your second point. I would expect the database to handle optimisation of low cardinality columns for me. Caveat: I have done no research to verify that Postgres actually does this.

allait: It's not so much the query speed that I'm worried about with the string columns, but rather the validation/flexibility implications. E.g. what if we decide to rename "published" to something else.

robyoung: That is a fair point. To the question of whether status should be indexed. According to some random on the internet low cardinality columns shouldn't be indexed at all.
http://image.slidesharecdn.com/fivestepsperform2009-101006154147-phpapp01/95/5-steps-to-postgresql-performance-38-728.jpg?cb=1286398019

robyoung: edit: that 'some random on the internet' is a core contributor

app/main/views.py
minglis: Oversight, fixed and added test.

migrations/versions/3d3fb04d1a98_.py
minglis: Probably a seperate story to "tidy up migrations scripts". I'll put it in as a chore into pivotal.

https://www.pivotaltracker.com/story/show/91376444

app/validation.py
minglis: Done.

app/models.py
minglis: Had a look at some work around enums, seems a little tricky to get the full benefits: see http://techspot.zzzeek.org/2011/01/14/the-enum-recipe/

Going to miss it out for now, but one to come back too.

tests/app/test_services.py
```diff
@@ -558,7 +559,7 @@ def test_update_a_service(self):
             with self.app.app_context():
```
TheDoubleK: The method `test_update_a_service` is indented so I don't think it will be run.

PR #56, opened 2015-03-26
Mi add framework table

minglis:
This is an pull request completes: https://www.pivotaltracker.com/story/show/91085102

The existing services in G5 and G4 have a notion of expired, this is to allow all services to be switched off when a framework itself expires. For this pull request we have created a table to capture the frameworks independent of the service.

There is a new *frameworks* table with 3 columns:

    - ID - primary key
    - Name - arbitrary descriptive text
    - expired - Boolean to determine if framework is expired.

This pull request only implements the existance of G-Cloud 6, and assumes all uses of the PUT API are creating an instance of G-Cloud 6. When we import GCloud 4 or 5, or DSS or other frameworks this will need to be expanded.

Not all tests have been extended to create a framework and due to the limitations of PUT mentioned above, all create a G-Cloud 6 framework.


app/models.py
```diff
@@ -30,8 +41,14 @@ class Service(db.Model):
                                nullable=False)
     data = db.Column(JSON)
 
+    framework_id = db.Column(db.BigInteger,
+                             db.ForeignKey('frameworks.id'),
+                             index=True, unique=False, nullable=False)
+
     supplier = db.relationship(Supplier, lazy='joined', innerjoin=True)
 
+    framework = db.relationship(Frameworks, lazy='joined', innerjoin=True)
```
allait: Do we have a separate supplier record for each framework? Or can one supplier have services on more than one framework?

app/models.py
```diff
@@ -2,6 +2,17 @@
...
 
 
+class Frameworks(db.Model):
+    __tablename__ = 'frameworks'
+
+    id = db.Column(db.Integer, primary_key=True)
+
+    name = db.Column(db.String(255), nullable=False)
+
+    expired = db.Column(db.Boolean, index=False, unique=False,
```
allait: I wonder if this should be a `expired_at` datetime instead? That way we can just set start and end dates on frameworks in advance, without having to flip the switch manually.

app/models.py
robyoung: That is a really nice way of making corrections to comments. I'm nicking that pattern!

app/models.py
minglis: Thought about dates, then I thought that when things expire is a movable issue, and keeping it up to date is an issue, so maybe a button and a boolean would be easier - we launch when we're ready etc rather than automatically. that might be one for the future.

PR #58, opened 2015-03-30
Pc add supplier route

pcraig3:
Hey everyone, added in a `/supplier/{supplier_id}` endpoint to the API

This also meant:
* Adding a few unit tests
* Updating the explorer files to return suppliers
* And, finally, renaming the `app/main/views.py` file to `app/main/views/services.py`, in order to avoid a cryptic error (thanks @allait!)

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -3,11 +3,11 @@ from flask import jsonify, abort, request
 from flask import url_for as base_url_for
 from sqlalchemy.exc import IntegrityError, DatabaseError
 
-from . import main
-from .. import db
-from ..models import ArchivedService, Service, Supplier, Framework
+from .. import main
+from ... import db
+from ...models import ArchivedService, Service, Supplier, Framework
 import traceback
-from ..validation import validate_json_or_400, \
+from ...validation import validate_json_or_400, \
     validate_updater_json_or_400, is_valid_service_id
 
 API_FETCH_PAGE_SIZE = 100
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -15,7 +15,8 @@ def get_supplier(supplier_id):
 
 def row2dict(row):
     """
-    Creates a dictionary object from the {column_name}:{value} pairs of an sqlAlchemy query object
+    Creates a dictionary object from the {column_name}:{value} pairs
+    of an sqlAlchemy query object
     @see http://stackoverflow.com/questions/1958219/
             convert-sqlalchemy-row-object-to-python-dict#comment12704946_1960546
     """
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -3,6 +3,7 @@ from flask import jsonify, abort, request
 from .. import main
 from ...models import Supplier
 
+# TODO: This should probably not be here
 API_FETCH_PAGE_SIZE = 2
 
 
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -1,0 +1,29 @@
+from flask import jsonify
+
+from .. import main
+from ...models import Supplier
+
+
+@main.route('/suppliers/<int:supplier_id>', methods=['GET'])
+def get_supplier(supplier_id):
+    supplier = Supplier.query.filter(
+        Supplier.supplier_id == supplier_id
+    ).first_or_404()
+
+    return jsonify(suppliers=row2dict(supplier))
+
+"""
+Creates a dictionary from sqlAlchemy query object's {column_name}:{value}
+@see http://stackoverflow.com/questions/1958219/convert-sqlalchemy-row-object-to-python-dict#comment12704946_1960546
+"""
+row2dict = lambda row: dict((col, getattr(row, col)) for col in row.__table__.columns.keys())
+
+
+def jsonify_supplier(supplier):
+    data = {
+        'id': supplier.id,
+        'supplier_id': supplier.supplier_id,
+        'name': supplier.name
+    }
+
+    return data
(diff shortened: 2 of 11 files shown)


app/main/views/suppliers.py
```diff
@@ -0,0 +1,34 @@
...
+    Creates a dictionary object from the {column_name}:{value} pairs
+    of an sqlAlchemy query object
+    @see http://stackoverflow.com/questions/1958219/
+            convert-sqlalchemy-row-object-to-python-dict#comment12704946_1960546
+    """
+    return dict((col, getattr(row, col))
+                for col in row.__table__.columns.keys())
+
+
+def jsonify_supplier(supplier):
```
TheDoubleK: We only need one of these functions - I think jsonify_supplier is clearer, if less general purpose.

app/main/views/suppliers.py
```diff
@@ -0,0 +1,23 @@
...
+    supplier = Supplier.query.filter(
+        Supplier.supplier_id == supplier_id
+    ).first_or_404()
+
+    return jsonify(suppliers=jsonify_supplier(supplier))
+
+
+def jsonify_supplier(supplier):
+    data = {
+        'id': supplier.id,
```
allait: Do we want to expose the table primary key as `'id'`? I think `service['id'] == service.service_id` and there's no `service['service_id']`

app/main/views/services.py
```diff
@@ -3,10 +3,10 @@
 from flask import url_for as base_url_for
 from sqlalchemy.exc import IntegrityError, DatabaseError
 
-from . import main
-from .. import db
-from ..models import ArchivedService, Service, Supplier
-from ..validation import validate_json_or_400, validate_updater_json_or_400
+from app.main import main
```
allait: I'd keep the relative imports here since that's what we're using in `views/suppliers.py` and almost everywhere else

app/main/views/suppliers.py
pcraig3: Yeah, I dunno.

screen shot 2015-03-31 at 10 12 54

Here's what my supplier table (with example data that I've added) looks like now.  It's possible to have a divergent `id` and the `supplier_id` if you create them manually, but if they're always meant to be identical than we probably don't need both.

allait: They're not identical, but I think the `id` is only used internally. e.g. when you request a supplier you send the `/suppliers/<supplier_id>`, and that's what the services are linked to.

TheDoubleK: Good spot, yes we shouldn't expose the database id at all.  I think we want `'id': supplier.supplier_id` and no `supplier_id` key in the JSON.

TheDoubleK: Also, for future reference (as not needed here yet), JSON keys should be camelCased rather than underscore_separated, c.f. `supplierId': service.supplier.supplier_id,`

pcraig3:
Just pushed up a new endpoint that returns lists of suppliers based on a matching prefix.
A GET request to `/supplier?prefix={prefix}` returns a list of suppliers whose `name`s start with the prefix. It's a case-insensitive SQL LIKE comparison, so that's the business logic.

Returned suppliers formatted like so.

screen shot 2015-03-31 at 17 02 55

A few questions:
- Is calling by a prefix of indefinite length okay?  Do we want to restrict this to something like (for example) just one letter? What about if supplier names start with numbers?
- What about the response? Do we want to add a `count` variable or is it fine as it is now?

robyoung:
We should add hypermedia links in these responses:
- A `self` link to each supplier like we do for services [[here](https://github.com/alphagov/digitalmarketplace-api/blob/pc_add_supplier_route/app/main/views/services.py#L201)]
- A link to the suppliers list from the root like we do for `services.list` [[here](https://github.com/alphagov/digitalmarketplace-api/blob/pc_add_supplier_route/app/main/views/services.py#L16)]

TheDoubleK:
Digits should definitely be allowed; a count seems unnecessary.

I'm not completely sure what the envisaged use of this endpoint is, so it's hard to say what we should restrict the length to.  If it's only for an A-Z (and 0-9) listing of suppliers then I'd say it should only allow a single character, and reject anything longer with a "400 Bad Request". At the very least It should probably reject anything really long before hitting the database with a query (but "really long" could be anything from 10 to 1000 chars!). 

The response should probably be paginated (we have over 2000 suppliers, and this should grow quite a lot in the near future as we take on new frameworks come online, so there could easily be 200+ suppliers per initial character).  Have a look how this is done for the services listing [here](https://github.com/alphagov/digitalmarketplace-api/blob/pc_add_supplier_route/app/main/views/services.py#L49) - services are paginated at 100 per page, but I'd say a smaller page size (20?) would be better for suppliers.

allait:
LGTM :+1:

pcraig3:
#### Final Response looks like this 

screen shot 2015-04-07 at 10 39 50

***
And just for posterity's sake, here's how the various issues raised were addressed.

##### 1. A `self` link to each supplier like we do for services

Yup, done.  :thumbsup:

##### 2. A link to the suppliers list from the root like we do for `services.list`

:-1: Didn't get included because we don't actually have a `suppliers.list` route at present.  The only way to get a list is to use the `/suppliers?prefix={whatever}` route, which never returns all suppliers.

##### 3. If it's only for an A-Z (and 0-9) listing of suppliers then I'd say it should only allow a single character, and reject anything longer with a "400 Bad Request".

:no_good: Didn't put any logic in about prefix strings because no definitive answers on what that logic should look like.

##### 4. The response should probably be paginated

:+1: Yup, good.  Response is paginated: returns 100 suppliers at a time.   (Incidentally, the variable deciding how many things are returned [is kind of bad](https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/services.py#L13).)

PR #62, opened 2015-04-07
Pc refactor services routes

pcraig3:
Just refactoring `app/main/views/services.py` to be more in line with how `suppliers.py` looks.

This meant, for example, removing methods like `paginate_links` that were near the bottom of `services.py` and putting them in the (Archived)Supplier models themselves.  
Also added `.serialize()` methods to the models in lieu of the prior `jsonify_service`.

Functionality is unchanged -- all the tests still run OK, and the logic is the same -- just the code is a bit cleaner.

### DIFF
--- a/app/main/utils.py
+++ b/app/main/utils.py
@@ -1,0 +1,50 @@
+from flask import url_for as base_url_for
+from flask import abort, request
+
+
+def link(rel, href):
+    if href is not None:
+        return {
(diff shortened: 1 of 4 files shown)


robyoung:
I don't understand why `url_for`, `link` or `pagination_links` are methods on the database model. It feels like they could be shared by just having them as functions imported into each views module.

PR #63, opened 2015-04-08
Only run migrations on one instance

robyoung:
The AWS rolling update [1] stuff does not work as expected and as such
migrations can run at the same time on more than one instance. This
makes beanstalk only run migrations on one instance.

[1] http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/using-features.rollingupdates.html

### DIFF
--- a/.ebextensions/dbmigrations.config
+++ b/.ebextensions/dbmigrations.config
@@ -1,3 +1,4 @@
 container_commands:
   upgrade:
     command: "python application.py db upgrade"
+    leader_only: true


TheDoubleK:
Looks good - have you tried this out at all? Or we believe the documentation in this case?

robyoung:
Good call. Closing while I test it thoroughly

robyoung:
I have tested this and confirmed that it does behave as we expect.

PR #64, opened 2015-04-08
Link to suppliers list

robyoung:
The requirement for services to be listed without providing a prefix
parameter was not explicity given in the story however, it is useful for
things like exporting all services or linking from the root path.

### DIFF
--- a/tests/app/views/test_suppliers.py
+++ b/tests/app/views/test_suppliers.py
@@ -32,9 +32,9 @@ class TestGetSupplier(BaseApplicationTest):
         assert_equal(u"Supplier 585274", data['suppliers']['name'])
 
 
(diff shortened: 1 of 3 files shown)


tests/app/views/test_suppliers.py
```diff
@@ -51,12 +51,16 @@ def setup(self):
             )
 
     def test_query_string_missing(self):
-        response = self.client.get('/suppliers')
-        assert_equal(400, response.status_code)
+        try:
+            response = self.client.get('/suppliers')
+            assert_equal(200, response.status_code)
+        except Exception as e:
```
allait: Should this be here?

robyoung: Nope, fixed.

PR #65, opened 2015-04-08
Mi adding put users endpoint

minglis:
Added Users functionality for pivotal stories: 
https://www.pivotaltracker.com/story/show/91413284
https://www.pivotaltracker.com/story/show/91413552

(1) New Users table, with corresponding create endpoint.
- This is a POST endpoint as not aware of existing IDs.
- unique id is email address
- currently has no roles or supplier relationship
- import script

(2) Auth endpoint
- As we migrated the passwords I did the Auth API to check they work in python
- This is a straight port of existing hashing code from the grail application - bcrypt. see 
https://pypi.python.org/pypi/bcrypt/1.0.1 and https://github.com/spring-projects/spring-security/tree/master/crypto/src/main/java/org/springframework/security/crypto/bcrypt
- POST endpoint to take plaintext username/password and check they validate


robyoung:
We need to add `bcrypt` to the requirements file. Also, have you seen [flask-bcrypt](https://github.com/maxcountryman/flask-bcrypt)? Even if we don't use it directly (it uses a much less popular bcrypt package) we should probably take some of what it does.
- Allow iterations to be configured on the app [[here](https://github.com/maxcountryman/flask-bcrypt/blob/master/flask_bcrypt.py#L132)]
- Do safe string comparisons [[here](https://github.com/maxcountryman/flask-bcrypt/blob/master/flask_bcrypt.py#L180)]
- Py 2 / 3 compatibility (maybe)

app/main/views/services.py
```diff
@@ -4,6 +4,7 @@
 from sqlalchemy.exc import IntegrityError, DatabaseError
 
 from .. import main
+from app.main import helpers
```
robyoung: Should this be a relative import? All the others are.

app/main/views/users.py
```diff
@@ -0,0 +1,81 @@
+from datetime import datetime
+
+from flask import jsonify, request, abort
+from app.main import main
+from app.models import User
+from app import db, encryption
+from sqlalchemy.exc import IntegrityError
+from app.main import helpers
+from app.validation import validate_user_json_or_400, \
+    validate_user_auth_json_or_400
```
robyoung: I think we should decide whether we want to be doing relative or absolute imports and stick to it across modules.

app/main/views/users.py
```diff
@@ -0,0 +1,81 @@
...
+    return jsonify(users=user.serialize())
+
+
+@main.route('/users', methods=['POST'])
+def create_user():
+    json_payload = get_json_from_request('users')
+    validate_user_json_or_400(json_payload)
+
+    user = User.query.filter(
+        User.email_address == 'joeblogs@email.com').first()
```
robyoung: Oops, I'm guessing this should come out of the payload.

app/main/views/users.py
```diff
@@ -0,0 +1,81 @@
...
+
+    if user:
+        abort(409, "User already exists")
+
+    if 'hashpw' in json_payload and not json_payload['hashpw']:
+        password = json_payload['password']
```
robyoung: Should this be `json_payload['hashpw']`?

app/main/views/users.py
```diff
@@ -0,0 +1,81 @@
...
+        password_changed_at=now
+    )
+
+    db.session.add(user)
+
+    try:
+        db.session.commit()
+    except IntegrityError as ex:
+        db.session.rollback()
+        abort(400, ex.message)
```
robyoung: Do we need to do this commit/rollback explicitly ourselves? Seems like the kind of thing that the Flask integration should be able to handle for us.

json_schemas/auth_users.json
```diff
@@ -0,0 +1,21 @@
+{
+  "id": "API Updater Schema",
+  "$schema": "http://json-schema.org/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "email_address": {
+      "type": "string",
+      "pattern": "^[^@^\\s]+@[^@^\\.^\\s]+(\\.[^@^\\.^\\s]+)+$"
```
robyoung: JSONSchema has some built in formats of which email is one. It's not highlighted very well.
http://spacetelescope.github.io/understanding-json-schema/reference/string.html#format
And the Python library does support format checking
http://python-jsonschema.readthedocs.org/en/latest/validate/#validating-formats

scripts/encrypt.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
+""" Test encryption
```
robyoung: Is this script just for testing the bcrypt interface? If so, do we need it committed?

app/main/views/users.py
```diff
@@ -0,0 +1,81 @@
...
+@main.route('/users/auth', methods=['POST'])
+def auth_user():
+    json_payload = get_json_from_request('auth_users')
+
+    validate_user_auth_json_or_400(json_payload)
+
+    user = User.query.filter(
+        User.email_address == json_payload['email_address'].lower()).first()
+
+    if user is None:
```
allait: Do we need different responses for wrong email and wrong password? How about a 403 for any authorization failures?

app/templates/explorer.html
```diff
@@ -24,8 +24,17 @@
                 <li><a href="#" id="import-link" data-toggle="collapse" data-target="#import-block">Import</a></li>
                 <li><a href="#" id="update-link" data-toggle="collapse" data-target="#update-block">Update</a></li>
                 <li><a href="#" id="fetch-link" data-toggle="collapse" data-target="#fetch-block">Fetch</a></li>
-                <li><a href="#" id="fetch-archive-link" data-toggle="collapse" data-target="#fetch-archive-block">Fetch Archive</a></li>
-                <li><a href="#" id="fetch-supplier-link" data-toggle="collapse" data-target="#fetch-supplier-block">Fetch Supplier</a></li>
+                <<<<<<< HEAD
```
allait: Unresolved merge conflict

app/models.py
```diff
@@ -38,15 +38,46 @@ class Framework(db.Model):
...
+    def serialize(self):
+        return {
+            'id': self.id,
+            'email_address': self.email_address,
+            'name': self.name,
+            'active': self.active,
+            'locked': self.locked,
+            'created_at': self.created_at,
+            'updated_at': self.updated_at,
+            'password_changed_at': self.password_changed_at,
```
allait: Should `password_changed_at` be a part of the response? It feels sensitive, and I don't think there's any use for it outside the API.

requirements.txt
```diff
@@ -1,10 +1,22 @@
...
 Flask-Bootstrap==3.3.0.1
 Flask-Migrate==1.3.1
-Flask-Script==2.0.5
 Flask-SQLAlchemy==2.0
-psycopg2==2.5.4
+Flask-Script==2.0.5
+Jinja2==2.7.3
+Mako==1.0.1
+MarkupSafe==0.23
+SQLAlchemy==0.9.8
```
allait: Why are we pinning indirect dependencies?

minglis: My bad, messed up the requirements file, fixed now.

app/main/views/users.py
allait: From the test it looks like `hashpw` is a flag that controls whether the password should be hashed/is already hashed. So `hashpw=False` means password is already a hash (?). I think we need a more descriptive field name, unless there's a reason to keep this one

app/models.py
minglis: Was returning it so users of the API can report this to clients if needed - ie admin app might want to tell a user who can't login when they last changed the password as part of user help.

json_schemas/users.json
```diff
@@ -0,0 +1,30 @@
...
+    "hashpw": {
+      "type": "boolean"
+    }
+  },
+  "required": [
+    "name",
+    "email_address",
+    "password"
+  ]
+}
```
allait: Missing newline

json_schemas/auth_users.json
```diff
@@ -0,0 +1,21 @@
...
+      "type": "string",
+      "maxLength": 255,
+      "minLength": 10
+    }
+  },
+  "required": [
+    "email_address",
+    "password"
+  ]
+}
```
allait: Missing newline

app/main/views/users.py
minglis: 2 types of response - user dosn't exist and user gave incorrect password, returned so users of API can perform auditing/reporting etc of issues as per their own use case.

app/main/views/users.py
minglis: in truth pycharm generates my imports for me, using absolute allows tab completion. Useful?

app/main/views/services.py
minglis: See note below.

app/main/views/users.py
allait: I think the general advise is to have the same response for both cases, e.g. "incorrect username and password". This probably doesn't apply here directly, since it's an API and we have a GET user endpoint anyway, but having a single error response would make sure that the frontend apps don't distinguish it either.

allait: Also, it would be nice to check that the format for "pre-hashed" passwords is recognised by bcrypt

app/main/views/users.py
allait: Also, it would be nice to check that the format for "pre-hashed" passwords is recognised by bcrypt

app/templates/explorer.html
minglis: Fixed

app/main/views/users.py
minglis: yep agreed, this isn't user facing, provides information for client apps to do the right thing, but we know what happened.

json_schemas/auth_users.json
minglis: Not sure how to implement this, changes the way validation works as it all shares a validate method which would take differeing params, and didn't work when I tried it. Bit short of time right now so will park this into a new story to implement formats on validations.

app/main/views/users.py
minglis: This hashpw field exists to get us over the hump of the initial import - the export from the exitsing DM is pre-hashed, once the import is complete this notion of "don't hash this password" will need to be removed. Default is hash, and the field is used only by the import script.

app/main/views/users.py
minglis: Adding a story to change these wherever they may be. Note we need someone of reporting/handling DB errors with regards to how we report his to users. It's not ideal right now.

PR #66, opened 2015-04-09
Adding roles to users table

minglis:
Completes:
https://www.pivotaltracker.com/story/show/92111500

Note this is a branch from the currently open Users branch: https://github.com/alphagov/digitalmarketplace-api/pull/65

This will need merged first. This story simply adds another column to the DB and the validation rules plus import to ensure it's populated.


migrations/versions/407e74de5553_adding_user_table.py
```diff
@@ -8,14 +8,13 @@
 
 # revision identifiers, used by Alembic.
 revision = '407e74de5553'
-down_revision = '3a5eba38e4e8'
+down_revision = '12f9dcc2e36c'
```
TheDoubleK: This looks pretty evil to me.  This rolls back a *long* way and misses a whole bunch of migrations that happened in-between. I think the migration sequence should be linear and not potentially branch like this.

tests/app/test_users.py
```diff
@@ -181,6 +190,20 @@ def test_return_400_for_invalid_user_json(self):
                 'users': {
                     'email_address': 'joeblogs@email.com',
                     'password': '',
+                    'role': 'buyer',
+                    'name': 'joe bloggs'}}),
+            content_type='application/json')
+
+        assert_equal(response.status_code, 400)
```
TheDoubleK: I think it's worth checking for a specific error message as well as the response code - there are a lot of ways to get a 400 out of the system, and we should be confident that the request has failed for the right reason.

app/models.py
```diff
@@ -33,12 +33,14 @@ class User(db.Model):
                            nullable=False)
     password_changed_at = db.Column(db.DateTime, index=False, unique=False,
                                     nullable=False)
+    role = db.Column(db.String, index=False, unique=False, nullable=False)
```
robyoung: Should this be defined as an [sqlalchemy.types.Enum](http://docs.sqlalchemy.org/en/latest/core/type_basics.html#sqlalchemy.types.Enum) given that we are applying the constraint in the migration?

migrations/versions/407e74de5553_adding_user_table.py
robyoung: Agreed, the whole first commit looks dodgy, it adds the `Users` model again. Could this be down to a rebase?

scripts/import_users.py
```diff
@@ -48,61 +47,71 @@ def __init__(self, endpoint, access_token, cert=None):
...
-        mapper = pool.imap
+    print("Filename: {}".format(filename))
 
     putter = UserPutter(endpoint, access_token, cert)
 
     counter = 0
     start_time = datetime.now()
-    for file_path, response in mapper(putter, list_files(listing_dir)):
+
+    data_file = open(filename)
```
robyoung: This should really be done with the `with` statement.
```python
with open(filename) as data_file:
    do_stuff(data_file)
```

tests/app/test_users.py
```diff
@@ -181,6 +190,20 @@ def test_return_400_for_invalid_user_json(self):
...
+
+    def test_return_400_for_invalid_user_role(self):
```
robyoung: We only validate that 'buyer' is valid and 'invalid' is not valid. We never test that 'supplier' and 'admin' are valid roles.

migrations/versions/407e74de5553_adding_user_table.py
minglis: These alembic versions get really messed up with git branches. It's really hard to know the order of application as the DB only has the most recent - which depending on which code you've branched from is not the best measure - not the application history. Maybe we need to manage these outside of the application?

minglis: Kev and I have done some reordering to mend the history here - but it's still a bit dodgy and we need an answer so orderings are obvious.

tests/app/test_users.py
minglis: Stuck an extra clause to check response. At this point the responses are very uninformative - simply JSON not valid. We should think more broadly about our responses so that they inform of the nature of JSON errors.

app/models.py
minglis: These lead to ints in the database. Would prefer strings. There is a way of doing it with python classes and postgres types, but it's quite complex so I haven't taken it on yet.

tests/app/test_users.py
minglis: Added 2 more test cases

app/models.py
robyoung: Good call

PR #68, opened 2015-04-09
Tidy imports and SQLAlchemy commit

robyoung:
- Use relative imports
- Remove manual SQLAlchemy rollback and error

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -1,13 +1,11 @@
 from datetime import datetime
-import traceback
 
 from flask import jsonify, abort, request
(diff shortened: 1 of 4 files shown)


app/main/views/services.py
```diff
@@ -201,11 +193,7 @@ def import_service(service_id):
 
     db.session.add(service)
 
-    try:
-        db.session.commit()
-    except IntegrityError:
-        db.session.rollback()
-        abort(400, "Unknown supplier ID provided")
```
allait: We should have a test to check that when a service is added with an unknown supplier_id the response code is 400

PR #69, opened 2015-04-09
Move pagination page size to config object

robyoung:
I'm not sure if this is the right thing to do. I'm not completely against it being set in the module but it had a TODO so I took a stab at addressing it. Thoughts?

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 
-from flask import jsonify, abort, request
+from flask import jsonify, abort, request, current_app
 from sqlalchemy.exc import IntegrityError
 
 from .. import main, helpers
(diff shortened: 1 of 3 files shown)


app/main/views/services.py
```diff
@@ -55,8 +51,11 @@ def list_services():
 
         services = services.filter(Service.supplier_id == supplier_id)
 
-    services = services.paginate(page=page, per_page=API_FETCH_PAGE_SIZE,
-                                 error_out=False)
+    services = services.paginate(
+        page=page,
+        per_page=current_app.config['DM_SUPPLIERS_PAGE_SIZE'],
```
allait: Should be `DM_API_SUPPLIERS_PAGE_SIZE`

app/main/views/suppliers.py
```diff
@@ -27,7 +24,7 @@ def list_suppliers():
 
     suppliers = suppliers.paginate(
         page=page,
-        per_page=API_FETCH_PAGE_SIZE,
+        per_page=current_app.config['DM_SUPPLIERS_PAGE_SIZE'],
```
allait: Should be `DM_API_SUPPLIERS_PAGE_SIZE`

TheDoubleK:
Yep, we definitely want separate pagination sizes for services and suppliers I reckon.

TheDoubleK:
I'm not against having the values set as code - I don't think it's something that needs to be particularly configurable.  Maybe include default values in the code that can be overridden by config if necessary?

PR #71, opened 2015-04-10
G4-G5 service import

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/90588402

We need to port the G4 and G5 service listings out of the Grails app into the new database.

This PR adds new JSON schemas for G4 and G5 services.  These are general purpose rather than one per lot as with G6, because we are converting fields to their "G6 compatible-ish" form for these services and there is little variation between lots.

The PUT endpoint now detects the framework that a service belongs to based on which schema it validates against and sets the 'Framework' in the database accordingly.

The exported G4 and G5 services have a 'status' field (published, enabled or disabled), which is set as the status in the database and then removed from the JSON payload before the service is saved. (An earlier migration has already set all G6 services to be 'published').

The GET endpoints have been modified to only ever return 'published' services (a later story will need to make 'enabled' services visible to their suppliers).

I've also updated some tests to check for specific error messages - there's a lot of ways to get a 400 response, and checking the error message gives us more confidence that the tests are passing for the right reason.

There are two new scripts: `validate-g4-g5.py` can probably be deleted very soon, but I've left it here for now as it might be useful for testing some of the functionality in the PR. `import-g4-g5.py` copies the pattern from the existing import script but processes a listings dump file rather than folders of individual files.

The latest listing dump file can be found here: http://37.26.90.26:8080/job/DUMP%20DM%20tables%20as%20JSON/ws/listings_dump.json


app/main/views/services.py
```diff
@@ -8,9 +8,11 @@
 from app.main import helpers
 from ... import db
 from ...models import ArchivedService, Service, Supplier, Framework
-from ...validation import validate_json_or_400, \
-    validate_updater_json_or_400, is_valid_service_id
-from ..utils import url_for, pagination_links, drop_foreign_fields, link
+import traceback
```
robyoung: This is an unused import

app/validation.py
```diff
@@ -68,22 +80,25 @@ def validate_user_auth_json_or_400(submitted_json):
...
-    elif validates_against_schema(G6_SAAS_VALIDATOR, submitted_json):
-        return 'G6-SaaS'
-    elif validates_against_schema(G6_PAAS_VALIDATOR, submitted_json):
-        return 'G6-PaaS'
-    elif validates_against_schema(G6_IAAS_VALIDATOR, submitted_json):
-        return 'G6-IaaS'
+    return framework
+
+
+def detect_framework(submitted_json):
```
robyoung: I find this a confusing name. While it does detect the framework this function is mostly used for validating service JSON.

app/main/views/services.py
TheDoubleK: Yes - leftover from the merge hell!  I removed it at least three times during rebasing, but it's somehow got back in. Will sort it.

PR #74, opened 2015-04-10
Refactor validation

robyoung:
Load all schemas at startup and make them available in a dictionary.
This removes a lot of duplication and makes it easier to add things to
the validator initialization such as a FormatChecker.

Rename JSON schema files to follow a consistent pattern.


app/validation.py
```diff
@@ -1,81 +1,47 @@
...
-    AUTH_USERS_VALIDATOR = validator_for(AUTH_USERS_SCHEMA)
-    AUTH_USERS_VALIDATOR.check_schema(AUTH_USERS_SCHEMA)
-    AUTH_USERS_VALIDATOR = AUTH_USERS_VALIDATOR(AUTH_USERS_SCHEMA)
+JSON_SCHEMAS_PATH = './json_schemas'
+FORMAT_CHECKER = FormatChecker()
+
+
+def load_schemas(schemas_path):
+    loaded_schemas = {}
+    for schema_file in os.listdir(schemas_path):
```
allait: Talked with @robyoung and decided to replace the autoloading with an explicit list of schema names. That way we'll  know that the required schema files exist on start.

PR #77, opened 2015-04-20
Pc suppliers fo real

pcraig3:
#### Defines real-live Suppliers and their concomitant Contact Information.

* Models and JSON validation schemas defined for Suppliers and ContactInformation 
* PUT Route (`/suppliers/{id}`) creates a Supplier + (potentially multiple) ContactInformation(s) in the database

##### Notes: 

* Settled on [PUT]ting with the `id` in the route instead of [POST]ing Suppliers because we already know their eventual location.  Discussed it with @TheDoubleK, and also [read about it on StackOverflow](http://stackoverflow.com/a/630475).

* Removed the unique constraints for the `duns_number` and `e_sourcing_id` in the Supplier Model.  @TheDoubleK suggested that it shouldn't be the responsibility of our app to enforce correct supplementary Id values.  If a Supplier gives us incorrect information, we should just accept it.

* `null` values are not being returned in GET JSON response.

* Similarly, Supplier `duns_number`s and `e_sourcing_id`s are not returned in the GET response.

***

Finally, I also cleaned up the regex in the `is_valid_service_id` method in `app/validation.py`  
I think it makes the logic more understandable.


app/main/views/suppliers.py
```diff
@@ -50,3 +54,89 @@ def get_supplier(supplier_id):
     ).first_or_404()
 
     return jsonify(suppliers=supplier.serialize())
+
+
+# Route to insert new Suppliers, not update existing ones
+@main.route('/suppliers/<int:supplier_id>', methods=['PUT'])
```
TheDoubleK: This endpoint will have to overwrite any existing record with the given ID, otherwise we won't be able to import data to the prod databases without deleting everything first.

robyoung:
Could some of the commits be squashed?

pcraig3:
Squashed a bunch of commits together all by myself with no help from @allait!

app/main/views/suppliers.py
```diff
@@ -50,3 +54,97 @@ def get_supplier(supplier_id):
...
+
+    for contact_information_data in contact_informations_data:
+        json_has_required_keys(
+            contact_information_data,
+            ['contactName', 'email', 'postcode']
+        )
+
+    validate_supplier_json_or_400(supplier_data)
+
+    # Check that `supplier_id` matches the JSON-supplied `id`
```
TheDoubleK: This should probably use the `json_has_matching_id` method in utils.

migrations/versions/427bdc69b94f_.py
```diff
@@ -0,0 +1,38 @@
+"""add 'description' to suppliers & make 'email' and 'postcode' required fields
```
TheDoubleK: It would be good to have the description of the migrations as part of the filenames - I think it can just be added after the underscore, as with the later scripts in the folder here: https://github.com/alphagov/digitalmarketplace-api/tree/master/migrations/versions

scripts/import_suppliers.py
```diff
@@ -0,0 +1,322 @@
...
+            return []
+
+        return [value for value in
+                map(unicode.strip, string.split(',')) if len(value) is not 0]
+
+    @staticmethod
+    def un_nest_key_value_pairs(obj, key_of_nested_obj):
+        if key_of_nested_obj not in obj.keys():
+            # abort(400, "No '{0}' key found".format(key_of_nested_obj))
+            # TODO: FAIL LOUDLY
```
TheDoubleK: Are the TODOs in this script going to be done, or are they unnecessary now?  If the former, then they should be done now as this is a one-off script and there will be no point doing them later.   If the latter then please delete the comments.

migrations/versions/550715127385_.py
```diff
@@ -0,0 +1,50 @@
...
+    op.add_column(u'suppliers', sa.Column('duns_number', sa.String(), nullable=True))
+    op.add_column(u'suppliers', sa.Column('esourcing_id', sa.String(), nullable=True))
+    op.create_unique_constraint(None, 'suppliers', ['duns_number'])
+    op.create_unique_constraint(None, 'suppliers', ['esourcing_id'])
+    ### end Alembic commands ###
+
+
+def downgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint(None, 'suppliers', type_='unique')
```
TheDoubleK: This line in the downgrade() method fails (even on an empty database).
The error message is: `AttributeError: 'NoneType' object has no attribute 'lower'`
It seems that this is because the constraints do not have a name but need one in order to roll back (don't ask why!).  There are two options here: 
1. (Better, probably correct, but more work) Add a name as a parameter to the constraints in the model definition and then regenerate these Alembic migration scripts (with the `-m` flag to include a description in the generated filename)
2. Hack these migration scripts to provide a name for each constraint generated/dropped by swapping `None` for a String describing the constraint.

Actually, as these two constraints are dropped by the next migration, it might be worth re-generating the migration scripts so that the constraints are never created in the first place.

TheDoubleK: Yep, looking at the scripts again I think it would definitely be best to wrap the three migrations here into a single migration file.

tests/app/views/test_suppliers.py
```diff
@@ -118,9 +106,124 @@ def test_query_string_prefix_page_out_of_range(self):
...
+        response = self.put_import_supplier(payload)
+        assert_equal(response.status_code, 400)
+
+    def test_when_supplier_has_extra_keys(self):
+        payload = self.load_example_listing("Supplier")
+
+        payload.update({'newKey': 1})
+
+        response = self.put_import_supplier(payload)
+        assert_equal(response.status_code, 400)
```
TheDoubleK: There are lots of ways to get a 400 response from the API.  If each of these different errors returns a specific error message at the point when the 400 is triggered then the tests should check that the error message is the one expected.  Otherwise the tests could continue to pass (i.e. the API returns 400) but for a different reason than the one they are supposed to be testing.

pcraig3:
Incorporated changes suggested by @TheDoubleK:

* Import supplier route using `json_has_matching_id` method
* Migration files have been merged into one which has been (re)named more meaningfully
* TODOs in `import_suppliers.py` dealt with
* 400 responses check for the right error messages.

scripts/import_suppliers.py
```diff
@@ -0,0 +1,298 @@
...
+            print("Skipping {}: not a valid JSON file".format(filename))
+
+    for supplier in json_from_file['suppliers']:
+        supplier_id, response = putter.import_supplier(supplier)
+
+        if response is None:
+            print("ERROR: Supplier (id={}) not imported".format(supplier_id),
+                  file=sys.stderr)
+        elif response.status_code / 100 != 2:
+            print("ERROR: {} on Supplier (id={})".format(response.status_code, supplier_id),
```
TheDoubleK: Line too long! If you can make Travis go green we should be able to get this merged today.

migrations/versions/550715127385_add_contact_information_table_and_extend_suppliers.py
```diff
@@ -0,0 +1,48 @@
+"""add contact_information table and extend suppliers
```
TheDoubleK: :+1: This migration script is much better than the previous version.

tests/app/views/test_suppliers.py
```diff
@@ -118,9 +106,139 @@ def test_query_string_prefix_page_out_of_range(self):
...
+    def test_when_supplier_contact_information_has_missing_keys(self):
+        payload = self.load_example_listing("Supplier")
+
+        payload['contactInformation'][0].pop('email')
+        payload['contactInformation'][0].pop('postcode')
+        payload['contactInformation'][0].pop('contactName')
+
+        response = self.put_import_supplier(payload)
+        assert_equal(response.status_code, 400)
+        assert_in('Invalid JSON must have '
```
TheDoubleK: :+1:

PR #79, opened 2015-04-20
Use DM_ENVIRONMENT to select config environment

allait:
Rename environments to match AWS deployment stages


robyoung:
:+1: :+1: :grin:

PR #81, opened 2015-04-21
Index on service creation

minglis:
This pull request is to complete stories:

https://www.pivotaltracker.com/story/show/92217318

https://www.pivotaltracker.com/story/show/92217404

There is a new SearchApiClient object. This is registered as a Flask extension (http://flask.pocoo.org/docs/0.10/extensiondev/).

It is tested by using requests_mock (http://requests-mock.readthedocs.org/en/latest/api/requests_mock.html) So that actual HTTP calls are mocked out. And the users of the SearchApiClient then use a mocked version of the Client to test those interactions.

This Client has a single method - index - that will index a G-Cloud service based on a mapping defined in that class. This is then POSTed to the search-api which will index that service. The same method is called on both PUT and POST so that service creation and update are both indexed.

The SearchApi  can be disabled in the config with the ES_ENABLED property. the config also needs to have the DM_SEARCH_API_URL and DM_SEARCH_API_AUTH_TOKEN properties set.


app/main/views/services.py
```diff
@@ -200,18 +207,14 @@ def import_service(service_id):
 
     db.session.add(service)
 
-    try:
-        db.session.commit()
-    except IntegrityError:
-        db.session.rollback()
-        abort(400, "Unknown supplier ID provided")
+    db.session.commit()
+    search_api_client.index(service_id, service.data, supplier.name)
```
robyoung: Is this right? Don't we still want to return a 400 if the supplier ID is unknown?

minglis: Removed as was moving to auto-commit/rollback. Put it back as Rob is right in that no supplier is a special case.

tests/app/views/test_services.py
```diff
@@ -490,6 +493,197 @@ def test_should_400_if_mismatched_service_id(self):
...
+                    '/services/1234567890123456',
+                    data=json.dumps(
+                        {
+                            'update_details': {
+                                'updated_by': 'joeblogs',
+                                'update_reason': 'whateves'},
+                            'services': payload}
+                    ),
+                    content_type='application/json')
+            except IntegrityError:
```
allait: Is this testing for a 500? IntegrityError should probably be handled by the app

tests/app/flask_search_api_client/test_search_api_client.py
```diff
@@ -0,0 +1,127 @@
...
+from flask import json
+from app import create_app
+
+
+class TestSearchApiClient():
+
+    search_api_client = None
+
+    def __init__(self):
+        os.environ['ES_ENABLED'] = 'True'
```
allait: How does this work? It's setting an environment value, but it doesn't look like it's accessed anywhere. Is this relying on the flask env to config? Can we set/patch the config directly instead?

minglis: Property is set on config as part of the reading of the environment that sets tokens etc. Setting on environ is how it gets set in running apps - proves round trip. Has no effect on the code under test.

Updated to use config as cleaner for the test. Should maybe be picked up by test that explicity verifies the setting of envionrment vars, as otherwise this relationship between ENV and code is not tested.

tests/app/views/test_services.py
minglis: It's testing a DB commit failure does not subsequently lead to an indexing of that change in ES - so no calls to the API. As the POST method does not handle integrity exceptions - it's a 500 - I need to catch the exception to make the assertion on my mock.

Talked with Rob, and I'm going to add integrity checking to the POST method, as it does on the PUT method.

tests/app/views/test_services.py
```diff
@@ -490,6 +493,191 @@ def test_should_400_if_mismatched_service_id(self):
...
+        super(TestShouldCallSearchApiOnPutToReplaceService, self).setup()
+        now = datetime.now()
+        payload = self.load_example_listing("G6-IaaS")
+        with self.app.app_context():
+            db.session.add(
+                Framework(id=1, expired=False, name="G-Cloud 6")
+            )
+            db.session.add(
+                Supplier(supplier_id=1, name=u"Supplier 1")
+            )
```
pcraig3: Seeing as how my pull request has recently been merged, it's worth pointing out that Suppliers will not be returned from the database if they haven't got any Contact Information ([as it's modelled using a Inner Join](https://github.com/alphagov/digitalmarketplace-api/blob/master/app/models.py#L118)).  So all of the tests that expect a Supplier to be returned will fail.  

I added Contact Information to [unit tests that create Suppliers](https://github.com/alphagov/digitalmarketplace-api/blob/master/tests/app/views/test_services.py#L170), but these newer ones don't have them.  

(I can add them in for you if you like.)

requirements_for_test.txt
```diff
@@ -1,3 +1,4 @@
 -r requirements.txt
 nose==1.3.6
 pep8==1.5.7
+requests-mock==0.6.0
```
TheDoubleK: This file also needs to add: 
`mock==1.0.1`
for the tests to pass.

PR #82, opened 2015-04-21
Add application logs to CloudWatch

robyoung:
- Adds the application name to the log entries
- Adds a request ID to the log entries
- Uses ISO 8601 as the date time format

## Request IDs
This adds request IDs to logs and to responses. The request ID is (in order)
- the request ID header (default: DM-Request-ID)
- the downstream request ID header (default: X-Amz-Cf-Id, the CloudFront request ID) 
- a new ID generated with uuid4


app/__init__.py
```diff
@@ -1,13 +1,12 @@
-import re
-import os
+from __future__ import absolute_import
```
allait: Why do we need this?

robyoung: Good spot, we don't. This was a hangover from when some of the logging was being developed in here.

allait:
Have you looked at https://docs.python.org/2/howto/logging-cookbook.html#filters-contextual? Since logging filters can modify the log record they seem like a better fit than the `CustomFormatter`.

About the request id: it seems that it's useful not only for logging, but for downstream clients (to preserve ID between apps) and maybe other stuff. So keeping it on `g` feels like coupling a lot of things together.
Two possible alternatives that I can think about are: using custom Request class with an added `id` attribute, or an WSGI middleware that would set the `DM-Request-ID` header for all requests if it's not set and then just using the `get_request_id` function (without the `uuid` part) everywhere else.

app/logging.py
```diff
@@ -0,0 +1,78 @@
+from __future__ import absolute_import
```
allait: Another absolute_import, do we need this one?

tests/test_logging.py
```diff
@@ -0,0 +1,107 @@
+from __future__ import absolute_import
```
allait: And another one

app/logging.py
robyoung: This one we do need because we're making use of the standard library logging package that is the same name as ours.

robyoung:
To your first point, I don't see that we get a huge amount of benefit really. Do you see a specific problem with using a custom formatter?

To your second point, that may be a useful thing to do in the future. The solution we have here works well for what what we need it to do right now. I'd be inclined to get this logging in and then look at refactoring as the need arises.

robyoung:
Depends on https://github.com/alphagov/digitalmarketplace-utils/pull/1

robyoung:
This should be good to go now.

.ebextensions/cwl-logs-application.config
```diff
@@ -0,0 +1,40 @@
...
+                log_stream_name = `{"Fn::Join": ["-", ["application", {"Fn::FindInMap":["CWLogs", "ApplicationLogs", "ApplicationName"]}]]}`
+                datetime_format = `{"Fn::FindInMap":["CWLogs", "ApplicationLogs", "TimestampFormat"]}`
+                [application-logs-combined]
+                file = `{"Fn::FindInMap":["CWLogs", "ApplicationLogs", "LogFile"]}`
+                log_group_name = `{"Fn::FindInMap":["CWLogs", "ApplicationLogs", "LogGroupName"]}`
+                log_stream_name = application-combined
+                datetime_format = `{"Fn::FindInMap":["CWLogs", "ApplicationLogs", "TimestampFormat"]}`
+              mode  : "000400"
+              owner : root
+              group : root
```
TheDoubleK: Is 'root' the right owner and group for this?  Or would we be better with a 'logs' user/group?

app/__init__.py
```diff
@@ -22,12 +19,10 @@ def create_app(config_name):
     application.wsgi_app = ProxyFix(application.wsgi_app)
     application.config.from_object(config[config_name])
 
-    for name in config_attrs(config[config_name]):
-        if name in os.environ:
-            application.config[name] = convert_to_boolean(os.environ[name])
```
TheDoubleK: @minglis just copied this pattern you've deleted into the supplier app.  We should probably standardise how config is set up a bit across all the apps.

TheDoubleK: Or is this deleted to move out into the new utils package?

.ebextensions/cwl-logs-application.config
robyoung: This is copied from the example config files AWS provides. The mode means it is only readable by root.

app/__init__.py
robyoung: To immortalise the conversation we just had in real life. The config logic is one of the things we should move into the utils package.

PR #83, opened 2015-04-21
Adding supplier to user table

minglis:
Completes story:

https://www.pivotaltracker.com/story/show/92112152

Adds the foreign key to supplier onto the user table. 

- When users have a role of 'supplier' they MUST have a supplier ID set in the users table. This is enforced by a constraint on the DB.
- Users of the other 2 roles, buyer and admin don't need a supplier id. This is not enforced in the database.

Updates:

- Import script to set the supplier ID from origin JSON

- users.py to read the supplierId and set on the model

- Users model to handle relationship

Also this Pull request does a fundamental change to how the tests run. We now no longer use the model to set up the DB between tests, now we run the migration scripts. This should be better as it means our local tests run against a DB as created on production. But it has a couple of drawbacks and needs some more work:

- Time takes - 1 1/2 minutes on my local mac.
- Should refactor tests to run DB setup before all and tear down after all, not after/before each test.
- This will mean all tests now need to tear down the data not the db - on a per test basis.
- noisy - the INFO logging on alembic means we see the migrations running a lot.
- Our downgrade scripts don't work, consequently we have to drop the model objects and then manually kill the versions table, which is a bit ugly. Fixing the downgrades means we can do:

    upgrade(head)

    downgrade(base)

At start/end of a test run.

Happy to discuss


migrations/versions/10_adding_supplier_to_user_table.py
```diff
@@ -0,0 +1,36 @@
+"""empty message
+
+Revision ID: 10_adding_supplier_to_user_table
```
TheDoubleK: I'm tempted to get https://github.com/alphagov/digitalmarketplace-api/pull/77 merged before this one (updating the suppliers table for supplier import) - it also contains a migration so the down-revision here would need to change.

json_schemas/users.json
```diff
@@ -22,6 +22,11 @@
       "type": "string",
       "pattern": "^(buyer|supplier|admin)$"
     },
+    "supplier_id": {
+      "type": "string",
```
TheDoubleK: supplier_id is an integer in the database, so probably makes sense to have a numeric field in the JSON?

tests/app/helpers.py
```diff
@@ -32,12 +35,24 @@ class BaseApplicationTest(object):
         "scs": "G6-SCS.json"
     }
 
+    ALEMBIC_CONFIG = \
```
TheDoubleK: Nice.

app/__init__.py
```diff
@@ -18,6 +19,9 @@ def create_app(config_name):
     application = Flask(__name__)
     application.wsgi_app = ProxyFix(application.wsgi_app)
     application.config.from_object(config[config_name])
+    migrate = Migrate(application, db)
+
+    manager = Manager(db, MigrateCommand)
```
robyoung: Do we want to do this on all apps or just when testing? If we only need it during testing could we add this in the setup?

json_schemas/users.json
minglis: was a bit worried given our travails with the service_id. so we want to commit to integer IDs?

json_schemas/users.json
```diff
@@ -22,6 +22,11 @@
       "type": "string",
       "pattern": "^(buyer|supplier|admin)$"
     },
+    "supplier_id": {
```
robyoung: This should be camelCase.

json_schemas/users.json
TheDoubleK: I know what you mean, but if DB is numeric already then maybe we just stick with that? Service IDs changed to retrofit G4 and G5 services, but I can't immediately think of a scenario where we'd want to change supplier ID to string in future?  We will need to come up with some way of generating supplier IDs before long though - at the moment they are just the old database IDs.

TheDoubleK: And I think for now the JSON should match the database - either way is fine, but a mix of number and string seems definitely wrong.

migrations/versions/10_adding_supplier_to_user_table.py
```diff
@@ -0,0 +1,36 @@
...
+
+def upgrade():
+    op.add_column('users', sa.Column('supplier_id', sa.BigInteger(), nullable=True))
+    op.create_index(op.f('ix_users_supplier_id'), 'users', ['supplier_id'], unique=False)
+    op.create_foreign_key(None, 'users', 'suppliers', ['supplier_id'], ['supplier_id'])
+
+    op.create_check_constraint(
+        "ck_user_supplier_has_supplier_id",
+        "users",
+        "((role='buyer') or (role='admin') or (role = 'supplier'  and supplier_id is not null))"
```
robyoung: We can model these constraints in SQLAlchemy http://docs.sqlalchemy.org/en/latest/core/constraints.html#check-constraint

app/__init__.py
minglis: Moved this. works fine.

json_schemas/users.json
minglis: Actually should align with supplier imports etc, so made it an integer

migrations/versions/10_adding_supplier_to_user_table.py
robyoung: As discussed in real life: We're going to add the check constraint to the SQLAlchemy model so that it's more obvious to people developing against that model. With an eye to replacing this field with the [enum recipe](http://techspot.zzzeek.org/2011/01/14/the-enum-recipe/) in the future

json_schemas/users.json
TheDoubleK: There are a couple of underscore/camelCase anomalies I'm sorting out in the branch I'm working on at the moment.  In the interest of getting this merged I'll wrap this one up with those changes later.

migrations/versions/10_adding_supplier_to_user_table.py
```diff
@@ -0,0 +1,35 @@
+"""empty message
+
+Revision ID: 10_adding_supplier_to_user_table
+Revises: 550715127385
+Create Date: 2015-04-21 07:45:29.958447
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '10_adding_supplier_to_user_table'
```
TheDoubleK: This will actually be revision 20 rather than 10, so maybe the name of this file/Revision ID should reflect this.  Would be good to include this revision in the VERSION_HISTORY file too: https://github.com/alphagov/digitalmarketplace-api/blob/master/migrations/VERSION-HISTORY.txt

tests/test_validation.py
```diff
@@ -120,7 +120,22 @@ def test_user_creation_validates():
...
+        ({'email_address': 'this@that.com',
+          'role': 'buyer',
+          'name': exactly_255,
+          'supplier_id': 123,
+          'password': exactly_255}, True, "valid supplier id"),
+        ({'email_address': 'this@that.com',
+          'role': 'buyer',
+          'name': exactly_255,
+          'supplier_id': '',
+          'password': exactly_255}, False, "invalid supplier id (to short)"),
```
TheDoubleK: These last two don't really make sense - supplier ID is numeric, so whether they are supplied as a long or short string should make no difference: if it's a string of any length then it's invalid.

migrations/versions/10_adding_supplier_to_user_table.py
minglis: Very annoyingly adding the constraint to the model generates an empty migration, and consequently EVERY run of the migrate command generates a new empty file, as it has no way of knowing it's done it all ready, as the file is empty.

PR #84, opened 2015-04-22
Extract logging

robyoung:
Depends on #82


app/__init__.py
```diff
@@ -19,12 +18,11 @@ def create_app(config_name):
     application.wsgi_app = ProxyFix(application.wsgi_app)
     application.config.from_object(config[config_name])
 
-    for name in config_attrs(config[config_name]):
-        if name in os.environ:
-            application.config[name] = convert_to_boolean(os.environ[name])
-
     config[config_name].init_app(application)
 
+    if not application.debug:
```
allait: I still think we want the same logs in debug going to stderr. Otherwise there's no easy way to test that logging is working ok and that we get the messages we want.

robyoung:
Closing until I get travis working.

app/__init__.py
robyoung: Seems fair, although that would mean disabling Flask's own debug logging. I'll have a play with it.

robyoung: I have implemented this in #82 and the utils package.

PR #86, opened 2015-04-23
Delete ContactInformation when overwriting Suppliers

pcraig3:
Overwriting an existing Supplier would remove the _reference_ between a Supplier and its prior Contact Information, but not delete the Contact Information outright.

Small update deletes Contact Information if a Supplier is found.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -97,13 +97,15 @@ def import_supplier(supplier_id):
     ).first()
 
     if supplier is None:
-            supplier = Supplier(supplier_id=supplier_data['id'])
+        supplier = Supplier(supplier_id=supplier_data['id'])
+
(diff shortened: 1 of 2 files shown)


app/main/views/suppliers.py
```diff
@@ -97,13 +97,15 @@ def import_supplier(supplier_id):
...
+    # if a supplier was found, remove all contact information
+    else:
+        for contact in supplier.contact_information:
+            db.session.delete(contact)
 
     supplier.name = supplier_data.get('name', None)
     supplier.description = supplier_data.get('description', None)
-
-    # TODO: this erases all prior contact information :/
-    supplier.contact_information = []
```
pcraig3: Turns out this just _disassociated_ all prior Contact Information.

PR #90, opened 2015-04-26
Added status query param to fetch services

minglis:
Added query parameter to allow us to ask for combinations of statues from the fetch api. Takes multiple values:

status=1&status=2

would do an "in" query for services with status 1 or status 2

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -37,7 +37,13 @@ def list_services():
 
     supplier_id = request.args.get('supplier_id')
 
-    services = Service.query.filter(Service.status == 'published')
+    services = Service.query
+
+    if request.args.get('status'):
+        services = Service.query.filter(
+            Service.status.in_(request.values.getlist('status'))
+        )
+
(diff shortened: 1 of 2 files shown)


app/main/views/services.py
```diff
@@ -37,7 +37,13 @@ def list_services():
 
     supplier_id = request.args.get('supplier_id')
 
-    services = Service.query.filter(Service.status == 'published')
+    services = Service.query
```
TheDoubleK: Are we sure we want the default behaviour to be to return everything?  An alternative would be to return only published  by default, but take parameters to include expired/unpublished? Actually, I don't think there is any logic to exclude services belonging to expired Frameworks at the moment - should we get that in now too?

minglis: We need a story for the framework stuff.

I figures APIs return everything on the GET by Id and the GET list queries, with limits on the list endpoint if needed. We're the client and should configure accordingly.

In the future we should give our API tokens roles and deal with it there.

PR #91, opened 2015-04-27
Keep version history up-to-date

TheDoubleK:
As a side not on this, the latest migration (10_adding_supplier_to_user_table) will not run if there is already any user data in the database, because adding the constraint fails (users with 'supplier' role but no supplier_id).

This is OK for us at the moment because there is no user data in the prod databases, but we might want to get into the habit of adding constraints in two steps, making sure the database is in a consistent state before adding the constraint.


robyoung:
There is data in the preview database though right? Do we have to drop and reload the preview database for this to work?

TheDoubleK:
I don't think there is User data in preview?  But I could be wrong.  The PR has already been merged, so if everything is green in preview beanstalk then I guess it went ok :/

TheDoubleK:
I'm pretty sure we haven't run either the `import_suppliers`, `import_users` or `import-g4-g5` into any API in Amazon yet.  So "breaking" changes to the database up until we do that import should all be OK.

PR #94, opened 2015-04-27
Add update endpoint for service.status and relevant tests.

pcraig3:
Adding a `[POST]` endpoint to update the status of Services.
Services can be:
- `published`
- `enabled`
- `disabled` 

With the constants defined within the route.
Successfully changing the status of a Service returns its updated JSON as a response.

##### Note: 

Figured we didn't need to archive services every time a status is updated.  Really, we're only changing meta-data.  
Sound off if you don't agree.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -20,12 +20,12 @@ def index():
         {
             "rel": "services.list",
             "href": url_for('.list_services', _external=True),
-        },
+            },
         {
             "rel": "suppliers.list",
             "href": url_for('.list_suppliers', _external=True),
-        },
-    ]), 200
+            },
+        ]), 200
 
 
 @main.route('/services', methods=['GET'])
@@ -254,3 +254,38 @@ def get_archived_service(archived_service_id):
     ).first_or_404()
 
     return jsonify(services=service.serialize())
+
+
+@main.route(
+    '/services/<string:service_id>/status/<string:status>',
+    methods=['POST']
+)
+def update_service_status(service_id, status):
+
+    # Statuses are defined in the Supplier model
+    valid_statuses = [
+        "published",
+        "enabled",
+        "disabled"
+    ]
+
+    is_valid_service_id_or_400(service_id)
+
+    service = Service.query.filter(
+        Service.service_id == service_id
+    ).first_or_404()
+
+    if status not in valid_statuses:
+        abort(400, "'{0}' is not a valid status.".format(status))
+
+    service.status = status
+    db.session.add(service)
+
+    try:
+        db.session.commit()
+        search_api_client.index(service_id, service.data,
+                                service.supplier.name)
+        return jsonify(services=service.serialize()), 200
+    except IntegrityError as e:
+        db.session.rollback()
+        abort(400, e.orig)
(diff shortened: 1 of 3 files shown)


app/main/views/services.py
```diff
@@ -248,3 +248,38 @@ def get_archived_service(archived_service_id):
     ).first_or_404()
 
     return jsonify(services=service.serialize())
+
+
+@main.route(
+    '/services/<string:service_id>/status/<string:status>',
```
robyoung: I don't think we need to use `string:`, it should be the default.

app/main/views/services.py
```diff
@@ -248,3 +248,38 @@ def get_archived_service(archived_service_id):
...
+    ]
+
+    is_valid_service_id_or_400(service_id)
+
+    service = Service.query.filter(
+        Service.service_id == service_id
+    ).first_or_404()
+
+    if status not in valid_statuses:
+        abort(400, "'{0}' is not a valid status.".format(status))
```
robyoung: Is it worth saying what valid statuses are here?

TheDoubleK:
I think we do need to keep an audit trail of who changed the status and when - so when suppliers come to us and say "my service isn't showing up" we can say "so-and-so at CCS disabled it for non-compliance", or they say "that service shouldn't be live" and we can say "so-and-so in your company set it to published last Friday afternoon".

If archiving the service each time with an associated update-reason is the only way to do it for now then we should do it.

app/main/views/services.py
TheDoubleK: Doesn't hurt to be explicit though?

robyoung: Other than it means we have to wrap the line so it's not as easy to scan down the file for routes, no. It's purely a visual thing.

app/main/views/services.py
```diff
@@ -248,3 +248,38 @@ def get_archived_service(archived_service_id):
...
+    db.session.add(service)
+
+    try:
+        db.session.commit()
+        search_api_client.index(service_id, service.data,
+                                service.supplier.name)
+        return jsonify(services=service.serialize()), 200
+    except IntegrityError as e:
+        db.session.rollback()
+        abort(400, e.orig)
```
allait: What does the message look like? I think we usually return our own error description instead of the exception one, but we can keep the exception text if it's not too database-related.

app/main/views/services.py
pcraig3: I think leave it explicitly declared; it's philosophically self-evident what variable we're expecting and it's the way we've declared everything else.

app/main/views/services.py
pcraig3: Leaving it as it because it's what we've done with the [update route](https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/services.py#L166).

pcraig3:
* Updating the status of a service now requires a JSON payload with an `update_details` key but no `services` key (because we're only changing one field).  Past services are archived, like we do in the update route.
* Also, to appease @robyoung, there's now a nice message for people who try to update services with a bad status name.
```json
{
  "error": "'peppermint' is not a valid status. Valid statuses are 'published', 'enabled', and 'disabled'"
}
```

PR #95, opened 2015-04-27
Silence alembic [INFO] messages from tests (and bootstrapping).

pcraig3:
Removing the `handlers = console` line from the `migrations/alembic.ini` file means we no longer have alembic.migration [INFO] messages in our unit tests. 
It also means we miss out on them when bootstrapping the app, just so we're all aware.


robyoung:
YES!! :boom:

PR #96, opened 2015-04-27
Update `_status` route with db connection info

pcraig3:
Return alembic version if connection with db established, otherwise show an error.

### DIFF
--- a/app/status/utils.py
+++ b/app/status/utils.py
@@ -1,4 +1,5 @@
 import os
+from .. import db
 
 
 def get_version_label():
@@ -9,3 +10,10 @@ def get_version_label():
             return f.read().strip()
     except IOError:
         return None
+
+
+def get_db_version():
+
+    return db.engine.execute(
+        "SELECT version_num FROM alembic_version"
+    ).scalar()
--- a/app/status/views.py
+++ b/app/status/views.py
@@ -9,9 +9,16 @@ from . import utils
 def status_no_db():
 
     try:
-        return jsonify(status="ok", app_version=utils.get_version_label(),
-                       db_version=utils.get_db_version())
+        return jsonify(
+            status="ok",
+            version=utils.get_version_label(),
+            db_version=utils.get_db_version()
+        )
 
     except SQLAlchemyError:
-        current_app.logger.exception('Cannot connect to database.')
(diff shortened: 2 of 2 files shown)


app/status/utils.py
```diff
@@ -9,3 +10,10 @@ def get_version_label():
             return f.read().strip()
     except IOError:
         return None
+
+
+def get_db_version():
+
+    return db.engine.execute(
+        "SELECT version_num FROM alembic_version"
+    ).scalar()
```
TheDoubleK: :+1: - I learned something here... hadn't seen .scalar before.

app/status/views.py
```diff
@@ -6,4 +7,11 @@
...
 def status_no_db():
-    return jsonify(status="ok", version=utils.get_version_label())
+
+    try:
+        return jsonify(status="ok", app_version=utils.get_version_label(),
+                       db_version=utils.get_db_version())
+
+    except SQLAlchemyError:
+        current_app.logger.exception('Cannot connect to database.')
+        return jsonify(status="error", message="Database is down"), 500
```
TheDoubleK: I think we could still have the app version number included in the status on database error.

pcraig3: I was sort of thinking the error messages should just be as simplistic as possible.

[Here](https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/42/files), I'm returning a similarly pared-down message for the admin-frontend app. 

Is there a conceivable advantage to having version numbers in all our error messages?
And then, in the case we include one here, do we want the admin-frontend app to know about the API app_version number if it gets an error?

TheDoubleK: I think the `_status` endpoints on all apps should give any information that could be useful in debugging issues.  If we have the code version to hand then we should include it.  e.g. What if we already know the DB is down (because we're trying to fix it) - it would still be useful to be able to easily check which version of the code is deployed for any particular app.

TheDoubleK: And in future we can add to what the status endpoint returns.  If there are multiple errors (e.g. DB is down and can't connect to search either) - then `_status` should report everything, not just the fist/last failure to be detected.  It won't necessarily be as detailed as this: https://www.digitalmarketplace.service.gov.uk/status - but the more the better I think (within reason).

PR #97, opened 2015-04-28
User update endpoint

TheDoubleK:
Adds endpoint to update user details.

Also makes all JSON key camelCase - there will be a corresponding update in the supplier frontend coming soon that will make it use all camelCased JSON keys for users.


app/main/views/users.py
```diff
@@ -36,6 +37,15 @@ def get_user_by_id(user_id):
     return jsonify(users=user.serialize())
 
 
+@main.route('/users', methods=['GET'])
+def get_user_by_email():
+    email_address = request.args.get('email')
```
robyoung: This effectively a required field. I think it would be useful to give a useful 400 error message if it's not provided.

TheDoubleK: Done - although I've gone with 404 rather than 400 due to arguments here (tl;dr: 400 is for syntax errors, and there's nothing wrong with the syntax in this case): http://stackoverflow.com/questions/9454811/which-http-status-code-to-use-for-required-parameters-not-provided

I'm not completely sold on 404 though so happy to go with 400 if you prefer.

PR #99, opened 2015-04-29
Ordering services listing

minglis:
Order list responses by:

- Framework
- Lot
- Service Name

- note this gives the following message under testing:
/Users/martyninglis/Envs/dm-api/lib/python2.7/site-packages/sqlalchemy/sql/base.py:530: SAWarning: Column 'CAST(services.data ->> :data_1 AS VARCHAR)' on table <sqlalchemy.sql.selectable.Select at 0x10de7c390; Select object> being replaced by <sqlalchemy.sql.elements.ColumnClause at 0x10de8dfd0; %(4528261328 anon)s>, which has the same key.  Consider use_labels for select() statements.
  (key, getattr(existing, 'table', None), value))

- Which is an advisory around syntax. To remove this would require using SQLAlchemy select not Flask::SQLAlchemy query for our queries. Whoever picks this up we should chat about implications.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -6,6 +6,7 @@ from .. import main
 from ... import db
 from ... import search_api_client
 from ...models import ArchivedService, Service, Supplier, Framework
+from sqlalchemy import asc
 from sqlalchemy.exc import IntegrityError
 from ...validation import detect_framework_or_400, \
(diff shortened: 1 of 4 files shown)


minglis:
Closing for now as logging issues are not fixed.

PR #100, opened 2015-04-29
Do not activate venv if already in a venv

robyoung:


TheDoubleK:
Nice.  Probably should do this in the other apps too.

PR #105, opened 2015-05-02
Only index current frameworks

minglis:
- When PUTing a G4 service (or any expired framework) don't index it.

- Note POSTing G4 fails currently. We shouldn't be updating these so should be OK.


minglis:
CI Issues. Closing this and raising a new pull request to replace it.

PR #108, opened 2015-05-05
Added notion of 'other' to suppliers prefix

minglis:
Allows retrieval of suppliers starting with non-alphanumeric characters

- When getting suppliers the query string can contain the key 'prefix' this is used to retrieve suppliers with names commencing with the letter provided - ie prefix=a returns all the 'A' suppliers

- To support the companies with non-alphanumeric names the prefix param can contain the special value 'other' which will retrieve all the numerics and so on.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -1,5 +1,6 @@
 from flask import jsonify, abort, request, current_app
 from sqlalchemy.exc import IntegrityError
+from sqlalchemy import not_
 
 from .. import main
 from ... import db
@@ -11,7 +12,6 @@ from ...utils import pagination_links, drop_foreign_fields, \
 
 @main.route('/suppliers', methods=['GET'])
 def list_suppliers():
-
     try:
         page = int(request.args.get('page', 1))
     except ValueError:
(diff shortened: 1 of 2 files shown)


app/main/views/suppliers.py
```diff
@@ -22,9 +22,13 @@ def list_suppliers():
     suppliers = Supplier.query.order_by(Supplier.name)
 
     if prefix:
-        # case insensitive LIKE comparison for matching supplier names
-        suppliers = suppliers.filter(
-            Supplier.name.ilike(prefix + '%'))
+        if prefix == 'other':
+            suppliers = suppliers.filter(
+                Supplier.name.op('~')('^[^A-z]+'))
```
robyoung: `[A-z]` includes other ascii characters, this should be `[A-Za-z]`.
Also, we are only interested in the first character so the `+` could be removed.

TheDoubleK: It looks as though just [A-Z] works case-insensitively with Postgres regexes, but yes it's probably better to be explicit about lowercase:
```
gdmdev=# select name from supplier where name ~* '^[^A-Z]';
            name
-----------------------------
 4sl Group
 100 Percent IT Ltd
 5NINES Data Centres Limited
 3gamma Limited
 4C Strategies Ltd
 1Tech Limited
 365 ITMS Ltd
 3M United Kingdom
 4Projects
 2T Security Ltd
 23 Technology Ltd
 6 POINT 6
 247 Time Ltd
 365 Response
 2nd Quadrant Limited
 3Cubed Limited
 1st Milestone Ltd
 1Spatial Group Ltd
 3R Telecom Limited
 8x8 Solutions
 4Secure
 3Zero Solutions Limited
(22 rows)
```

minglis: Updated this - I did not know the A-z thing with other chars. Fixed.

PR #111, opened 2015-05-05
Adding clients to supplier response + couple of tests.

pcraig3:
Where `Supplier['clients']` is either a list of client names or an empty list.  
This is in contrast to the way that some of our other keys are not in the response if they have `null` values.

Also added a few tests because I didn't think our unit tests are taking long enough to run.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -113,7 +113,8 @@ class Supplier(db.Model):
             # 'dunsNumber': self.duns_number,
             # 'eSourcingId': self.esourcing_id,
             'contactInformation': contact_information_list,
-            'links': links
+            'links': links,
+            'clients': self.clients
         }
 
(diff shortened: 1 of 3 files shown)


minglis:
Paul and I talked yesterday about the empty array Vs missing key for those suppliers with no clients. What do others think about this model or using an empty list?

TheDoubleK:
I don't mind an empty list - it feels much more descriptive than a null for some reason.  As we had in a previous discussion though - clients shouldn't be dependent on particular keys existing; it's the values of keys that do exist they should be worrying about.

TheDoubleK:
:+1: All looks good to me - happy to merge if we agree to return keep empty client arrays. I think either way should be fine though.

PR #114, opened 2015-05-06
Ordering services listing

allait:


TheDoubleK:
Is there a story attached to this PR at all? Is it for the A-Z pages?

app/main/views/services.py
```diff
@@ -41,6 +43,10 @@ def list_services():
 
     services = Service.query.filter(
         Service.framework.has(Framework.expired == false())
+    ).order_by(
+        asc(Service.framework_id),
```
TheDoubleK: The framework_id is an internal database ID - not guaranteed to be in any particular order, it will depend how the migrations ran and what mood Postgres was in at the time.  (I know that in fact it will probably give us the order we want, but I don't think we should rely on it.)
Below is how my Frameworks table looks, but I'm sure I looked in the preview (or maybe prod) DB and the IDs were completely different - I think one framework had ID 60 or something.
```
digitalmarketplace=# select * from frameworks;
 id |   name    | expired
----+-----------+---------
  1 | G-Cloud 6 | f
  2 | G-Cloud 4 | t
  3 | G-Cloud 5 | f
(3 rows)
```

app/main/views/services.py
```diff
@@ -6,13 +6,15 @@
...
+from sqlalchemy.sql.expression import false
 from ...validation import detect_framework_or_400, \
     validate_updater_json_or_400, is_valid_service_id_or_400
 from ...utils import url_for, pagination_links, drop_foreign_fields, link, \
     json_has_matching_id, get_json_from_request, json_has_required_keys, \
     display_list
+from sqlalchemy.types import String
 
 
 @main.route('/')
```
TheDoubleK: Why do we want ordering on the "fetch all" endpoint?  JSON isn't supposed to be ordered...

app/main/views/services.py
```diff
@@ -41,6 +43,10 @@ def list_services():
 
     services = Service.query.filter(
         Service.framework.has(Framework.expired == false())
+    ).order_by(
```
TheDoubleK: Actually, I think the point does make sense - why do we want ordering on the "fetch all services" endpoint by default? JSON isn't supposed to be ordered...

If the story is the supplier page then perhaps ordering should only be applied to queries with a supplier - i.e. in the `if supplier_id is not None:` part of the method.

allait: I'm not sure what you mean by "JSON isn't supposed to be ordered". We return a list of services, lists are inherently ordered.

In case of postgres, not having ordered_by results in the "whatever is fastest" order and it's not really clear what are the guarantees it provides: i.e. does it stay the same between page requests, or is it possible to get duplicate services if order changes between page requests.

TheDoubleK: I'm pretty sure the way pagination is implemented by Flask we are guaranteed not to get duplicates on the next page - I think pagination links in the API responses contain a token that references a specific query.

app/main/views/services.py
allait: Migrations are ordered and even though we do have occasional issues applying them, I don't think we can end up with different order on different environments, especially since we deploy the same version to all the stages. So I'm not sure what's the concern here.

ID order is more arbitrary than name order would be, but unless we have another preference, I don't think it's a problem.

app/main/views/services.py
allait: I'm not sure if that's true (at least not due to anything Pagination does). The next page link in the API response only has `?page=n` and the `.paginate` method doesn't add an `order_by`.

PR #117, opened 2015-05-08
Index or Delete services in search_api if status changes.

pcraig3:
- Re-index the service when it's state has been *changed* to `published` from any other state
- Delete the service from elasticsearch when its state has been changed (to either `enabled` or `disabled`) from `published`

If we `publish` a Service that was previously published (for example), we will still return a successful response, but we won't bother with updating elasticsearch. 
Similarly, we don't delete anything from elasticsearch if a service goes from 'disabled' to 'enabled' or vice-versa.

***

Also moved some of the `search_api_client.index()` calls out of try/except loops because I think it makes it more readable.

### DIFF
--- a/tests/app/views/test_services.py
+++ b/tests/app/views/test_services.py
@@ -5,6 +5,7 @@ from nose.tools import assert_equal, assert_in, assert_not_equal, \
     assert_almost_equal, assert_false, assert_is_not_none
 
 from app.models import Service, Supplier, ContactInformation, Framework
+import mock
 from mock import Mock
 from app import db, search_api_client
 from ..helpers import BaseApplicationTest, JSONUpdateTestMixin, \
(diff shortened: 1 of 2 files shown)


tests/app/views/test_services.py
```diff
@@ -986,6 +986,185 @@ def test_should_not_index_on_service_on_expired_frameworks(self):
...
+            content_type='application/json'
+        )
+
+    def test_should_index_on_service_status_changed_to_published(self):
+
+        old_status = 'enabled'
+        new_status = 'published'
+
+        search_api_client.index = Mock(return_value=True)
+        search_api_client.delete = Mock(return_value=True)
```
robyoung: I think this would be safer using `mock.patch`. Currently this is leaving the search API client's index and delete methods mocked after the test is done.

robyoung:
The solution looks good to me. I think the tests could be simplified a bit.

tests/app/views/test_services.py
pcraig3: Here, I was modelling it on [these tests](https://github.com/alphagov/digitalmarketplace-api/blob/master/tests/app/views/test_services.py#L922) in the `TestShouldCallSearchApiOnPost` class.  
Should I have two patch declarations before every test then?

allait: I think you can patch them in `setUp`/`tearDown` and overwrite the `return_value` in the individual tests if needed.

tests/app/views/test_services.py
```diff
@@ -5,6 +5,7 @@
     assert_almost_equal, assert_false, assert_is_not_none
 
 from app.models import Service, Supplier, ContactInformation, Framework
+import mock
 from mock import Mock
```
pcraig3: Gonna to remove importing `Mock` in a separate branch.

PR #118, opened 2015-05-10
Move test db migration to package setup

robyoung:
This speeds the tests up a lot (~45 seconds down to ~16).

Move the running of all migrations and dropping of all tables to a package
setup which means that it's only run once for all tests in that package.
The contents of tables are dropped after each test.

One bad thing in this pull request is that because we bootstrap the
frameworks table in migrations we have to do that bootstrapping manually
in the tests as well.

### DIFF
--- a/tests/app/helpers.py
+++ b/tests/app/helpers.py
@@ -1,6 +1,4 @@
 from __future__ import absolute_import
-from flask.ext.migrate import Migrate, MigrateCommand
-from flask.ext.script import Manager
 
 import os
 import json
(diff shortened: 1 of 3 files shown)


tests/app/helpers.py
```diff
@@ -71,9 +51,16 @@ def setup_authorization(self):
     def do_not_provide_access_token(self):
         self.app.wsgi_app = self.app.wsgi_app.app
 
-    def apply_migrations(self):
+    def setup_frameworks(self):
         with self.app.app_context():
-            upgrade(self.config, 'head')
+            if Framework.query.count() == 0:
```
minglis: Is it possible to exclude reference tables from the delete data phase? We could keep some record of bootstrapped tables and ignore them.

I can see some of these tables becoming more complex in our future.

tests/app/helpers.py
```diff
@@ -157,8 +144,8 @@ def teardown_authorization(self):
     def teardown_database(self):
         with self.app.app_context():
             db.session.remove()
-            db.drop_all()
-            db.engine.execute("drop table alembic_version")
+            for table in reversed(db.metadata.sorted_tables):
+                db.engine.execute(table.delete())
```
TheDoubleK: As discussed in real life - if we don't delete from Frameworks then we won't need the `setup_frameworks` method at all.

PR #119, opened 2015-05-10
Refactor json schemas

robyoung:
## Change JSON schema ids to titles
The spec states that ids must be valid URIs [1].
The title field [2] has no such restriction and seems more appropriate.

[1] http://json-schema.org/latest/json-schema-core.html#anchor27
[2] http://json-schema.org/latest/json-schema-validation.html#anchor98

## 	Refactor G6 IaaS priceInterval regex
Rather than having two empty alternation options doesn't compile in some
languages.

## Only test schemas with .json extension
Before this change tests fail if you have a schema open in vim (because
of the vim swap file).

### DIFF
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -36,7 +36,7 @@ def test_for_valid_service_id():
 def test_all_schemas_are_valid():
     for file_name in os.listdir('json_schemas'):
         file_path = 'json_schemas/%s' % file_name
-        if os.path.isfile(file_path):
+        if os.path.isfile(file_path) and file_path.endswith(".json"):
             yield check_schema_file, file_path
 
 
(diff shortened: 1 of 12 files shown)


tests/test_validation.py
```diff
@@ -36,7 +36,7 @@ def test_for_valid_service_id():
 def test_all_schemas_are_valid():
     for file_name in os.listdir('json_schemas'):
         file_path = 'json_schemas/%s' % file_name
-        if os.path.isfile(file_path):
+        if os.path.isfile(file_path) and file_path.endswith(".json"):
```
TheDoubleK: I'd be tempted to say that there probably shouldn't ever be any non-json files in the `json_schemas` directory, so throwing an error here (rather than skipping them) might be the right thing.

PR #120, opened 2015-05-11
Swap `Mock`s for `mock.patch`es

pcraig3:
Replace all of the `Mock()` constructors in our unit tests with `@mock.patch()` decorators because this way we're not introducing persistent Mock objects unintentionally.

Note that merging this change before my [Index or Delete services](https://github.com/alphagov/digitalmarketplace-api/pull/117) one probably bring about a merge conflict because this one removes an import statement.

### DIFF
--- a/tests/app/views/test_services.py
+++ b/tests/app/views/test_services.py
@@ -6,8 +6,7 @@ from nose.tools import assert_equal, assert_in, assert_not_equal, \
 
 from app.models import Service, Supplier, ContactInformation, Framework
 import mock
-from mock import Mock
-from app import db, search_api_client
+from app import db
 from ..helpers import BaseApplicationTest, JSONUpdateTestMixin, \
     TEST_SUPPLIERS_COUNT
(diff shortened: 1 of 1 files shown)


tests/app/views/test_services.py
```diff
@@ -863,12 +868,13 @@ def test_should_not_index_on_service_put_if_db_exception(self):
                 content_type='application/json')
 
             assert_equal(search_api_client.index.called, False)
-            db.session.commit = Mock(side_effect=c)
             db.session.rollback()
```
pcraig3: Can we do without this?  The tests pass whether it's there or not.
Since we delete everything in the database after each test, we probably don't need the rollback, right?

robyoung: I think we do but give it a try. We have only mocked out the commit so the deletes / updates are still going through. If we do not either roll back or commit I think that subsequent calls may break.

pcraig3:
:+1:

PR #121, opened 2015-05-11
Add import-from-api script

allait:
Allows copying services from one API endpoint to another: e.g. to
copy data between environments.


robyoung:
In order to work this depends on https://github.com/alphagov/digitalmarketplace-api/pull/122. Code wise it looks fine though and it doesn't affect the app so I'm going to merge.

PR #123, opened 2015-05-12
Verify certain fields do not appear in JSON

robyoung:


tests/app/views/test_services.py
```diff
@@ -735,6 +734,33 @@ def test_should_404_without_status_parameter(self):
...
         assert_equal(response.status_code, 404)
 
+    def test_json_postgres_field_should_not_include_column_fields(self):
+        non_json_fields = [
+            'supplierName', 'links', 'frameworkName', 'status', 'id']
+        with self.app.app_context():
+            response = self.client.get('/services/{}'.format(self.service_id))
+            data = json.loads(response.get_data())
+
+            response = self.client.post(
```
allait: I think we need the same for `.put` too

robyoung: Fair point

PR #124, opened 2015-05-13
Make import-from-api.py Python 3 compatible

robyoung:
I was getting SSL errors running this in Python 2 due to TLS1.0 support. These changes make it possible to run the script with Python 3.

### DIFF
--- a/scripts/import-from-api.py
+++ b/scripts/import-from-api.py
@@ -13,10 +13,10 @@ Example:
     ./import.py --serial http://localhost:5000 myToken http://source-api token
 """
 from __future__ import print_function
+from six.moves import map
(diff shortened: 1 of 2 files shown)


scripts/import-from-api.py
```diff
@@ -13,10 +13,10 @@
     ./import.py --serial http://localhost:5000 myToken http://source-api token
 """
 from __future__ import print_function
+from six.moves import map
```
allait: `six` doesn't seem to be in `requirements.txt`

robyoung: Good spot, it comes in with `requests-mock` but we should be explicit.

PR #127, opened 2015-05-18
[#94850520] Fix bug updating g5 services

robyoung:
https://www.pivotaltracker.com/story/show/94850520

## Change update_service to validate whole service
The change validates the entire serialized service object (minus a small number of foreign fields) against the JSON schema rather than just what is stored in the `data` field. This makes the JSON schema a better reflection of what the HTTP API accepts and produces.

## Make import_service more like update_service
Move code around so that it flows more similarly to update_service.

Dissallow updates via PUT. All our updates currently go through POST and the import_service method already makes assumptions about it always creating new services (it always returns a 201 status code and always updates the created_at field).

Remove some redundant tests that are either in the wrong place (testing post in the put test) or no longer necesarry (testing the search API is called on update through the PUT endpoint).

Ensure supplierId does not make it into the data field, it is in the supplier_id field of the model.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -122,7 +122,6 @@ def list_archived_services_by_service_id():
 def update_service(service_id):
     """
         Update a service. Looks service up in DB, and updates the JSON listing.
-        Uses existing JSON Parse routines for validation
     """
 
     is_valid_service_id_or_400(service_id)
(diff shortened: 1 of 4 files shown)


robyoung:
Closing until I've verified the result with @andrewchong

robyoung:
I have verified the fix with @andrewchong

app/main/views/services.py
```diff
@@ -135,33 +134,21 @@ def update_service(service_id):
...
 
-    data = drop_foreign_fields(data, ['id'])
-    now = datetime.now()
-    service.data = data
-    service.updated_at = now
-    service.updated_by = update_json['updated_by']
-    service.updated_reason = update_json['update_reason']
+    data = service.serialize()
+    data = drop_foreign_fields(data,
+                               ['supplierName', 'links', 'frameworkName'])
```
TheDoubleK: Don't we need update_service to drop `status` too?  Otherwise it might end up in the JSON data field.  (And I'm won round to your idea that status updates should eventually be done through this same update endpoint, but that's not for this story).

robyoung: It get's popped at [here](https://github.com/alphagov/digitalmarketplace-api/pull/127/files#diff-32bf67afda65ad26faef3bf3efd522b1R222). Granted, it's a bit confusing.

TheDoubleK: Ah OK, sorry I hadn't followed it right through.

TheDoubleK: But that's only in the import method, not on update?  Are we just trusting that if it was removed when first imported then it shouldn't be there in updates?

robyoung: In the update it happens [here](https://github.com/alphagov/digitalmarketplace-api/pull/127/files#diff-413e9aa8961126d20e2d90befa13e0f1R231). I wanted to move the `from_json` work into the model so that it's closer to it's inverse (ie. `serialize`). It's not done for `import_service` as this was a bit trickier and it's documented as not being a model to follow.

TheDoubleK:
OK, I'm convinced that 'status' doesn't end up in the JSON data, but this change would now mean that the service status can be updated in two places.  Is that what we want?  Previously only the /services/.../status/... endpoint updated a service but now the regular update endpoint also updates status.
e.g: `
curl -X POST -H "Authorization: Bearer myToken" -H "Content-Type: application/json" -d '{"update_details": {"updated_by":"kevkeenoy","update_reason":"for a try"},
  "services": {"status":"disabled"}
}' http://localhost:5000/services/5-G2-0304-313`

Is this what we want?
If it is then I'll merge it, but good to make it clear that this change makes something possible that wasn't before.

TheDoubleK:
Also, using this endpoint to update the status of a service won't do the necessary work with the search API that is done in `/services/<string:service_id>/status/<string:status>`

robyoung:
Very good point! Thanks.

PR #128, opened 2015-05-19
Add index_services script to sync Elasticsearch with data API

allait:
Script takes URLs and auth tokens for search and data APIs and
for each service in /services either adds or removes the it from
the search api index based on the service "published" status.

Story: https://www.pivotaltracker.com/story/show/92217466

### DIFF
--- a/scripts/index_services.py
+++ b/scripts/index_services.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python
+"""Read services from the API endpoint and write to search-api for indexing.
+
+Usage:
+    index_services.py <search_endpoint> <search_access_token> <api_endpoint>
+              <api_access_token> [options]
+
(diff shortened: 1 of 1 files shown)


scripts/index_services.py
```diff
@@ -0,0 +1,107 @@
...
+        try:
+            services_page = data_client.find_services(page=page)
+        except apiclient.APIError as e:
+            print('API request failed: {}'.format(e.message), file=sys.stderr)
+            return
+
+        for service in services_page['services']:
+            yield service
+
+        if list(filter(lambda l: l['rel'] == 'next', services_page['links'])):
```
robyoung: This will break when [this story](https://www.pivotaltracker.com/story/show/94576802) is done. I've added a link to this pull request to that story.

PR #130, opened 2015-05-21
Convert JSON schema patterns to enums

robyoung:
Where possible convert JSON schema string regular expression patterns into enums.

With the development of the SSP we will need to take a closer look at how we describe what's needed for a given framework. Currently some of this logic is duplicated between these schemas and the SSP content repo. Making 'choice' regular expressions enums is intended to make it easier to reuse this validation logic elsewhere, where providing a choice is required.

I've added the script used to automate this change for reference. It replaces the patterns directly in the raw JSON string to avoid problems with the JSON being reordered.


json_schemas/services-g6-saas.json
```diff
@@ -89,12 +86,10 @@
...
     },
     "priceUnit": {
-      "type": "string",
-      "pattern": "^(Unit|Person|Licence|User|Device|Instance|Server|Virtual machine|Transaction|Megabyte|Gigabyte|Terabyte)$"
+      "enum": ["Unit", "Person", "Licence", "User", "Device", "Instance", "Server", "Virtual machine", "Transaction", "Megabyte", "Gigabyte", "Terabyte"]
     },
     "priceInterval": {
-      "type": "string",
-      "pattern": "^(||Second|Minute|Hour|Day|Week|Month|Quarter|6 months|Year)$"
+      "enum": ["", "Second", "Minute", "Hour", "Day", "Week", "Month", "Quarter", "6 months", "Year"]
```
TheDoubleK: Have you tested that the empty string here works as expected?  Looks like it should do, but would be good to be sure.

allait:
@robyoung do we need to keep the script? It'd be nice to document the three "pattern" formats it recognises, just so it's not accidentally used for things it can't rewrite properly (e.g. patterns like `^(a|b)-c$`)

allait:
There are more patterns that weren't converted: e.g. in https://github.com/alphagov/digitalmarketplace-api/blob/convert-schema-regex-to-enum/json_schemas/services-g5.json: priceInterval, serivceTypes, networksConnected

json_schemas/services-g6-saas.json
robyoung: Yes it does, I've got a test that I can show you but I'm know sure it's worth committing.

PR #131, opened 2015-05-21
Ignore search api errors

allait:
Changes to the **index_services** script:
If any of the search or data API requests fail the script exits
with non-zero status code. Search API request failures don't
stop further requests, but the failure is stored for the exit
code.

Index add and delete request failures are logged as warnings but
don't affect the API response. All inconsistencies due to failed
requests are expected to be fixed by nightly re-indexing script.

Pivotal: https://www.pivotaltracker.com/story/show/94409704

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -6,15 +6,18 @@ from .. import main
 from ... import db
 from ... import search_api_client
 from ...models import ArchivedService, Service, Supplier, Framework
+
 from sqlalchemy import asc
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.sql.expression import false
 from ...validation import detect_framework_or_400, \
     validate_updater_json_or_400, is_valid_service_id_or_400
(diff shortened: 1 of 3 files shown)


app/main/views/services.py
```diff
@@ -233,11 +241,16 @@ def import_service(service_id):
...
 
     if not framework.expired:
+        try:
             search_api_client.index(
                 service_id,
                 service.data,
                 supplier.name,
                 framework.name)
+        except apiclient.HTTPError as e:
+            current_app.logger.warning(
```
TheDoubleK: Don't know if we should log these as error rather than warning? Something is definitely wrong if the index updates are failing.

allait: Index failures should definitely be out of the ordinary, but since we're not returning an error status code for the response I don't think we should log it as ERROR either. I might be wrong about this, but I'd keep errors for "request failed" events. So I'd expect the search-api to log an ERROR if there's something wrong.

PR #132, opened 2015-05-22
[#94576802] Update Json api format

TheDoubleK:
See this story in Pivotal: https://www.pivotaltracker.com/story/show/94576802 - the main aim is to make API response formats consistent across the Data and Search APIs.

There are FOUR related pull-requests to complete this story. The other three are:
https://github.com/alphagov/digitalmarketplace-search-api/pull/42
https://github.com/alphagov/digitalmarketplace-utils/pull/39
https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/60

Changes in the data API are to the 'links' element - instead of 'rel':'x' and 'href':'y' the 'rel' value is now the key for the link e.g. 'x':'y'.  This is already the way the search API formats the links element.

See these gists for examples of the new API responses:
[Example services data API response](https://gist.github.com/TheDoubleK/0d7cd9363440e700a50d)
[Example suppliers data API response](https://gist.github.com/TheDoubleK/4fde0400e052f20ed2c0)
[Example services search API response](https://gist.github.com/TheDoubleK/8e82782502e611faabcc)

### DIFF
--- a/scripts/import-from-api.py
+++ b/scripts/import-from-api.py
@@ -34,14 +34,9 @@ def request_services(api_url, api_access_token, page=1):
                 'Authorization': 'Bearer {}'.format(api_access_token),
             }
         ).json()
-
         for service in services_page['services']:
             yield service
-
-        page_url = list(
-            filter(lambda l: l == 'next', services_page['links']))
-        if page_url:
-            page_url = page_url[0]['href']
+        page_url = services_page['links'].get('next')
 
 
 def print_progress(counter, start_time):
--- a/scripts/import-from-api.py
+++ b/scripts/import-from-api.py
@@ -39,7 +39,7 @@ def request_services(api_url, api_access_token, page=1):
             yield service
 
         page_url = list(
-            filter(lambda l: l['rel'] == 'next', services_page['links']))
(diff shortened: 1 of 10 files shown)


scripts/import-from-api.py
```diff
@@ -39,7 +39,7 @@ def request_services(api_url, api_access_token, page=1):
             yield service
 
         page_url = list(
-            filter(lambda l: l['rel'] == 'next', services_page['links']))
+            filter(lambda l: l == 'next', services_page['links']))
```
allait: `page_url = services_page['links'].get('next')`? The filter was an unfortunate consequence of 'links' being a list of dictionaries.

scripts/import-from-api.py
```diff
@@ -39,7 +39,7 @@ def request_services(api_url, api_access_token, page=1):
...
         if page_url:
             page_url = page_url[0]['href']
```
allait: I don't think we need this, since there's no `'href'` anymore

TheDoubleK: Good spot, missed this one (this script was the last thing I did on Friday, got sloppy!)

PR #133, opened 2015-05-28
Import from API scripts with data cleansing

allait:
Pivotal story: https://www.pivotaltracker.com/story/show/94589786

`import_services_from_api` script rewritten with `dmutils.apiclient`

### import_suppliers_from_api script with data cleaning
Reads suppliers from one API endpoint and writes to another one.
Recursively replaces all fields named 'email' and 'phoneNumber'
with placeholder values in:
* main supplier fields
* fields containing nested dictionaries
* fields containing lists of dictionaries

### An option to create users when importing suppliers
Calling import_suppliers_from_api with --user=PASS options will
create user accounts for each supplier, using PASS as password
and `supplier-<supplierId>@user.dmdev` as email.


scripts/import_suppliers_from_api.py
```diff
@@ -0,0 +1,166 @@
...
+class SupplierUpdater(object):
+    def __init__(self, endpoint, access_token, user_password=None):
+        self.endpoint = endpoint
+        self.access_token = access_token
+        self.user_password = user_password
+
+    def __call__(self, supplier):
+        client = apiclient.DataAPIClient(self.endpoint, self.access_token)
+        try:
+            print(self.clean_data(supplier, supplier['id']))
```
robyoung: Does this need to be here? It makes the output very noisy.

robyoung:
The two new importers look good to me. It would be good to rationalise the naming of the import scripts. The more recent scripts are quite obvious but the earlier ones are not very explicit.

PR #134, opened 2015-06-01
[WIP] Edit suppliers info

allait:
Pivotal story: https://www.pivotaltracker.com/story/show/95432212

## WIP
This adds supplier and contact-information edit endpoints and changes history/audit events using SQLAlchemy-Continuum. Main questions at this point are:
* Do we want to use the SQLAlchemy-Continuum/AuditPlugin versioning and are we happy with the database schemas they create?
* Do we want a separated edit endpoint for the contact information or should it be part of the suppliers edit?

I've run the performance tests against the current service update with archive and the throwaway `__versioned__` services, and the Continuum version is ~10% slower (28 RPS vs 25 RPS on my laptop) with search-api disabled.

### About model versioning
Uses SQLAlchemy-Continuum to create archived versions of Supplier and
ContactInformation objects on data changes. The versions are created
automatically on INSERT, UPDATE and DELETE queries and are stored in
`<model>_version` table.

Each object can reference previous version using the `.versions`
attribute. Each version contains a `.changeset`, which is a list of
fields that were changed when it was created. Versions can also be
used to replace the current object using `.restore`.

The inital version is created when the object is added for the first
time.

When an object is modified a version record is also created for any
related objects (e.g. modifying supplier will create a new version
for related ContactInformation objects). This shouldn't be an issue
since only direct (not backref) references are traversed, so modifying
supplier won't create archives for all related services.

## About audit events
SQLAlchemy-Continuum ActivityPlugin allows creating Activity events
that are connected to version transactions.

Each event can store:

* a reference to the changed object (both current version and the
  version that was active during the transaction are stored)
* an additional target object when related things are changed
* an action verb (e.g. 'update', 'create', ...)
* a JSON 'data' for additional information


app/main/views/suppliers.py
```diff
@@ -143,3 +141,84 @@ def import_supplier(supplier_id):
...
+        db.session.commit()
+    except IntegrityError as e:
+        db.session.rollback()
+        abort(400, "Database Error: {0}".format(e))
+
+    return jsonify(suppliers=supplier.serialize())
+
+
+@main.route(
+    '/suppliers/<int:supplier_id>/contact-information/<int:contact_id>',
```
TheDoubleK: Given out model I think having a separate endpoint for updating contact information makes sense.  Given that a supplier can have multiple contact details we will also need "add" and "delete" contact details endpoints.  Is it worth also doing these now along with this update endpoint? Though it might be a while before they get used... I'm not sure if we're planning on rolling out contact-details-per-service for the G-Cloud 7 SSP? If we are then we should probably make the endpoints now.  If not then maybe leave it for the time being.

app/main/views/suppliers.py
```diff
@@ -143,3 +141,84 @@ def import_supplier(supplier_id):
...
+    if contact is None:
+        abort(404, "contact_id '%d' not found" % contact_id)
+
+    json_has_required_keys(request_data, ['contactInformation'])
+
+    contact_data = contact.serialize()
+    contact_data.update(request_data['contactInformation'])
+    contact_data = drop_foreign_fields(
+        contact_data,
+        ['links']
```
TheDoubleK: I don't think there are any `links` in contact information responses?

app/models.py
```diff
@@ -109,8 +131,8 @@ def serialize(self):
             'id': self.supplier_id,
             'name': self.name,
             'description': self.description,
-            # 'dunsNumber': self.duns_number,
-            # 'eSourcingId': self.esourcing_id,
+            'dunsNumber': self.duns_number,
```
TheDoubleK: I don't think we want these numbers coming out of the API at all.

json_schemas/suppliers.json
```diff
@@ -38,7 +38,6 @@
   },
   "required": [
     "id",
-    "name",
-    "contactInformation"
```
TheDoubleK: Hmm... this is tricky.  We do want to enforce that new suppliers must provide some contact information.  But I can see that for updates we don't want to have to bundle it in every time.

app/models.py
allait: Not returning them complicates things: we accept them on PUT and they're part of the supplier schema, so it's reasonable to accept them on POST. But if they're not returned from the API then there's no indication if they've been stored or not and the data update can't use the `.serialize` since it will clear out the hidden fields. We should either drop the fields completely or treat them as normal fields, this "you can write, but can't read" seems to make little sense for the private API.

TheDoubleK: Apparently we would have to pay money to serve the Duns numbers (although while the API is private this probably isn't an issue).  We should talk with Cath & @minglis about whether these will ever be used for anything.  If not then maybe we should drop them altogether.

app/main/views/suppliers.py
allait: It's not easier to add these at this point than it would be later, and after talking with Ralph and Katie it looks like it's not clear whether we're going to have multiple contacts for suppliers at all. I feel like having a separate contact-information endpoint is a reasonable compromise at this point: it allows us to work with multiple contacts in the future if we need to, but creating a full CRUD flow for contact-details is probably out of scope right now.

app/main/views/suppliers.py
allait: Good call. I'm probably going to add `self` links for contact information objects in the supplier response, since otherwise the `/contact-infomation/` URLs aren't mentioned anywhere in the API responses.

tests/app/views/test_suppliers.py
```diff
@@ -321,3 +322,337 @@ def test_when_supplier_contact_information_email_invalid(self):
...
+        }})
+
+        assert_equal(response.status_code, 200)
+
+        with self.app.app_context():
+            contact = ContactInformation.query.filter(
+                ContactInformation.id == self.contact_id
+            ).first()
+
+        assert_equal(contact.contact_name, "New Value")
```
TheDoubleK: It might be nice for the test to have different values for each of the updated fields - so that we are sure that the right fields are getting the right updated values.

tests/app/views/test_suppliers.py
```diff
@@ -321,3 +322,337 @@ def test_when_supplier_contact_information_email_invalid(self):
...
+        response = self.update_request({'contactInformation': {
+            "contactName": "New Value",
+            "phoneNumber": "New Value",
+            "email": "new-value@example.com",
+            "website": "example.com",
+            "address1": "New Value",
+            "address2": "New Value",
+            "city": "New Value",
+            "country": "New Value",
+            "postcode": "New Value",
```
TheDoubleK: The trailing comma here isn't valid JSON, even if Python is happy with it.

allait: This is not JSON, it's a python dictionary.

TheDoubleK: I know, but the trailing commas disconcert me.

PR #135, opened 2015-06-04
Draft services

minglis:
Create draft services API.

Endpoints:

### Create draft:

    PUT /services/{:service_id}/draft

- Creates a new draft from an existing service. It is not possible to create a draft without an existing service in this pull request.

- Requires the same "updater" JSON used in the services endpoints:

    {
         'update_details': {
            'updated_by': 'Person',
            'update_reason': 'Reason'
        }
    }

- Returns the draft

- Drafts are serialized in the same manner as a normal service

### Edit Draft

    POST /services/{:service_id}/draft

- Edits a draft. Follows same rules as Editing a service - EXCEPT it will not validate the service after an update. This so we can create services, that will be incomplete during their construction/

- Requires updater JSON

- Returns a draft serialized like a service

### Fetch Draft

    GET /services/{:service_id}/draft

Fetches a draft

- Returns a draft serialized like a service

### Delete Draft

Deletes a draft from the drafts table - has no effect on the service / archived services.

    DELETE /services/{:service_id}/draft

- Requires updater JSON

- Returns a 200, with a "done" message

### Publish Draft

    POST /services/{:service_id}/draft/publish

Promotes a draft service to live. 

- This will update the service table with the changes made to the draft

- Creates an Archive from the previous service

- Deletes the draft

- Requires updater JSON

- Returns the new service, serialized as in the services API.

### Caveats

- the updater information IS NOT persited onto the drafts / services. This will be replaced by the AUDIT stories that are coming. the code is there to require, validate and access this information. But it is unused until the AUDIT story comes online.

- It is possible to make a draft service in an invalid state using the DRAFTs functionality. It DOES NOT validate services until publication. Consequently this step may fail with validation errors.


minglis:
Sorry, also, I refactored code out of the services API routes so that it can be shared with the drafts API. I didn't change any of the actual lines, more of an encapsulation into methods in the services_utils.

app/main/views/drafts.py
```diff
@@ -0,0 +1,154 @@
...
+        db.session.rollback()
+        abort(400, "Database Error: {0}".format(e))
+
+    return jsonify(message="done"), 200
+
+
+@main.route('/services/<string:service_id>/draft/publish', methods=['POST'])
+def publish_draft_service(service_id):
+    """
+    Delete a draft service
```
TheDoubleK: This should be "Publish" - copy & paste fail :)

app/main/views/drafts.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+@main.route('/services/<string:service_id>/draft', methods=['PUT'])
+def create_draft_service(service_id):
+    """
+    Create a draft service from an existing service
+    :param service_id:
+    :return:
+    """
+    is_valid_service_id_or_400(service_id)
```
TheDoubleK: Should this method also fail if a draft already exists with the supplied service ID?

app/main/views/drafts.py
```diff
@@ -0,0 +1,154 @@
...
+    :param service_id:
+    :return:
+    """
+
+    is_valid_service_id_or_400(service_id)
+
+    updater_json = validate_and_return_updater_request()
+    update_json = validate_and_return_service_request(service_id)
+
+    service = DraftService.query.filter(
```
TheDoubleK: Might be nice to call this `draft` rather than `service`, as in the PUT endpoint.

app/main/views/drafts.py
```diff
@@ -0,0 +1,154 @@
...
+        Service.service_id == service_id
+    ).first_or_404()
+
+    draft = DraftService.from_service(service)
+
+    db.session.add(draft)
+
+    try:
+        db.session.commit()
+        return jsonify(services=draft.serialize()), 201
```
TheDoubleK: The return could come after the `try...except`.  Not that it really matters, but it would be the same pattern as the POST version then.

app/main/views/drafts.py
```diff
@@ -0,0 +1,154 @@
...
+def fetch_draft_service(service_id):
+    """
+    Return a draft service
+    :param service_id:
+    :return:
+    """
+
+    is_valid_service_id_or_400(service_id)
+
+    service = DraftService.query.filter(
```
TheDoubleK: `draft` here too?

app/main/views/services.py
```diff
@@ -181,31 +157,26 @@ def import_service(service_id):
...
-    now = datetime.now()
+    service = Service.query.filter(
+        Service.service_id == service_id
+    ).first()
 
-    json_payload = get_json_from_request()
-    json_has_required_keys(json_payload,
-                           ['services', 'update_details'])
+    if service is not None:
+        abort(400, "Cannot update service by PUT")
```
TheDoubleK: Should the message rather be that you can't PUT to an ID that already exists? (I know this is just what was there already)

app/models.py
```diff
@@ -302,6 +302,88 @@ def serialize(self):
...
+
+    framework = db.relationship(Framework, lazy='joined', innerjoin=True)
+
+    @staticmethod
+    def from_service(service):
+        return DraftService(
+            framework_id=service.framework_id,
+            service_id=service.service_id,
+            supplier_id=service.supplier_id,
+            created_at=datetime.now(),
```
TheDoubleK: The created_at and updated_at won't necessarily be millisecond-equal here - it's nicer to create a single datetime and use it for both fields to get proper equality.  Again, it probably doesn't matter that much...

app/models.py
```diff
@@ -302,6 +302,88 @@ def serialize(self):
...
+            updated_at=datetime.now(),
+            updated_by=service.updated_by,
```
TheDoubleK: Can we get the currently logged in user for the `updated_by` of the draft?

app/service_utils.py
```diff
@@ -0,0 +1,58 @@
...
+    service.update_from_json(
+        service_payload,
+        updated_by=updater_payload['updated_by'],
+        updated_reason=updater_payload['update_reason'])
+
+    data = service.serialize()
+
+    data = drop_foreign_fields(
+        data,
+        ['service_id', 'supplierName', 'links', 'frameworkName'])
```
TheDoubleK: Do we also need to drop `status` here?

app/service_utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+    detect_framework_or_400(data)
+    return service
+
+
+def index_service(service):
+    if not service.framework.expired:
```
TheDoubleK: Do we want to move the check for `"status": "published"` into this method too?

app/service_utils.py
```diff
@@ -0,0 +1,58 @@
...
+                'Failed to add {} to search index: {}'.format(
+                    service.service_id, e.message))
+
+
+def delete_service_from_index(service):
+    try:
+        search_api_client.delete(service.service_id)
+    except apiclient.HTTPError as e:
+        current_app.logger.warning(
+            'Failed to add {} to search index: {}'.format(
```
TheDoubleK: The message here should be `failed to remove xxx from search index`.

migrations/versions/37ac005a48c3_add_draft_services.py
```diff
@@ -0,0 +1,46 @@
+"""empty message
+
+Revision ID: 37ac005a48c3
+Revises: 20_adding_json_index_to_services
+Create Date: 2015-06-02 18:32:08.399213
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '37ac005a48c3_add_draft_services'
```
TheDoubleK: We were trying to make the revision identifiers nice and sequential and human-readable...

tests/app/views/test_drafts.py
```diff
@@ -0,0 +1,336 @@
...
+                'update_details': {
+                    'updated_by': 'joeblogs',
+                    'update_reason': 'whateves'
+                }
+            }),
+            content_type='application/json')
+
+        assert_equal(res.status_code, 400)
+
+    def test_should_not_be_able_to_a_draft_if_service_doesnt_exist(self):
```
TheDoubleK: Should not be able to do what to a draft service? Go over the line length limit?

TheDoubleK:
Lots of minor comments, but generally looking good to go.

app/service_utils.py
```diff
@@ -0,0 +1,58 @@
+from flask import current_app
+from .utils import get_json_from_request, \
+    json_has_matching_id, json_has_required_keys, drop_foreign_fields
+from .validation import validate_updater_json_or_400, detect_framework_or_400
+from . import search_api_client, apiclient
+
+
+def validate_and_return_updater_request():
```
allait: `update_details` should probably be the same for service and supplier edits, so it might be worth adding this to `validation` instead of `service_utils`

app/models.py
minglis: Deliberately not really done this as the audit stories that Alexey has coming will replace this functionality. This story contains the capture of the user, but it doesn't save it anywhere. the audit events will have that.

app/service_utils.py
minglis: Did this.

migrations/versions/37ac005a48c3_add_draft_services.py
minglis: Yep - forgot about this. Done.

tests/app/views/test_drafts.py
minglis: publish. nothing wrong with long lines

app/service_utils.py
minglis: It's in here as it's request based - so it checks the JSON and then does validate, so validate aspects are in validation and request aspects are in here.

May be need to rename "service_utils" It's not massively focused on services.

tests/app/views/test_drafts.py
minglis: Actually now it's too long 😜

minglis:
Changes up.

PR #136, opened 2015-06-04
Edit suppliers info with audit events

allait:
Pivotal story: https://www.pivotaltracker.com/story/show/95432212

### Add AuditEvent model

Audit fields:
* `type` - type of the event
* `user` - user login or email address. Not a reference to the user
  account, since the admin access is not controled by a User account
  right now
* `data` - a general JSON field for any information about the event
* `object` - a generic reference and can point to objects of any type
  to trace the history of events related to a given model object

`object` uses primary keys for reference since there's builtin support
for them and the natural keys (supplier_id and service_id) have
different types (int and string). This means that if the object is
replaced completely the AuditEvents references will not point to the
replacement.

### Add audit events to supplier and contact information updates

Store the related supplier object (for both direct supplier and
related contact information updates) and the incoming request data
as "request" field of event `data` field.

**NOTE**: ~~Doesn't store update_details since they're not passed in to the API. We should talk about what the `user` and `reason` fields for AuditEvents should look like, since they're not really informative the way they're used right now (`updated_by` is not linked to a User and `updated_reason` is usually set to a static value where the request is made).~~ Updated author is required in the `updated_by` request field and is saved to the `AuditEvent.user`.

### Set empty list as default value for supplier.clients

Since the clients field is not required by the suppliers schema it
can be left out when importing a supplier, which makes it default
to `null`. Setting a default value for the model field makes sure
that clients is always a list.

### Add links to ContactInformation.serialize

Adds a link to update contact information details. url_for points
to the update endpoint, since there's no read endpoint for contacts
yet, but the URL should stay the same.

Nested contactInformation links are dropped in supplier PUT to pass
the JSON schema validation.

### Store update author when editing supplier details 
Adds required 'updated_by' field to the edit request.
Field value is saved to the audit event user field and as part of
the request data.

There's no `update_reason` for the supplier update since it's being
deprecated in other edit endpoints once they're switched to audit
events. `update_reason` was always set to a static value and since
there are no plans to add a reason field to frontend apps it's likely
to stay that way if it's kept.

### Replace uses of .now with .utcnow when saving update times 
.now is local time, so we should always use .utcnow when storing
timestamps in the database.


allait:
Completely forgot about timestamps for audit events. Leaving this open for travis/review for now, but **shouldn't be merged yet**.

app/main/views/services.py
```diff
@@ -181,7 +181,7 @@ def import_service(service_id):
     """
     is_valid_service_id_or_400(service_id)
 
-    now = datetime.now()
+    now = datetime.utcnow()
```
TheDoubleK: There are a couple more of these in the tests, it might be nice to have them all using `utcnow`, although obviously doesn't make a lot of difference in tests.

app/models.py
```diff
@@ -50,10 +51,28 @@ class ContactInformation(db.Model):
...
+        self.city = data.get("city")
+        self.country = data.get("country")
+        self.postcode = data.get("postcode")
+
+        return self
+
     def serialize(self):
-        # Should there be links for the associated service(s) / supplier?
+        links = link(
+            "self", url_for(".update_contact_information",
```
TheDoubleK: Will we add a GET endpoint at some point so that this link works?

json_schemas/suppliers.json
```diff
@@ -33,12 +33,12 @@
...
       },
-      "minItems": 0
+      "minItems": 0,
+      "maxItems": 10
     }
   },
   "required": [
     "id",
-    "name",
-    "contactInformation"
```
TheDoubleK: We shouldn't allow suppliers to be created without any contact information - if we have to remove it from the schema to enable supplier updates then we could add a check in the import_supplier method that there is at least one contact info?

TheDoubleK:
I like this much better than the plugin version.  No auto-merge, but looking good.

app/models.py
allait: The link works now, but only for POST. I'm not sure if we need a GET endpoint for individual contact details, but we can always add it later.

json_schemas/suppliers.json
allait: Looks like there are checks for this in the PUT endpoint already. And the test for missing contact information seem to pass.

PR #137, opened 2015-06-08
Incorporate Feature Flags.

pcraig3:
The upgraded `dmutils` sets up FeatureFlags (among others) in its `init_app` function, and includes the `get_version_label` and `get_flags` methods.

The `init_app` function also sets up `logging`, `config`, and `proxy_fix`, so we don't have to import them into our api anymore.


app/__init__.py
```diff
@@ -1,27 +1,28 @@
 from flask import Flask
 from flask.ext.bootstrap import Bootstrap
 from flask.ext.sqlalchemy import SQLAlchemy
-from dmutils import logging, config, apiclient, proxy_fix
```
TheDoubleK: Removing proxy_fix makes me nervous. Are you sure we no longer need it?

app/status/views.py
```diff
@@ -11,14 +12,16 @@ def status_no_db():
     try:
         return jsonify(
             status="ok",
-            version=utils.get_version_label(),
-            db_version=utils.get_db_version()
+            version=get_version_label(),
+            db_version=utils.get_db_version(),
```
TheDoubleK: If we're removing the `utils.` from other lines then we could also add `get_db_version to the imports and drop it for this line?

requirements.txt
```diff
@@ -6,7 +6,8 @@ Flask-Script==2.0.5
 Flask-SQLAlchemy==2.0
 psycopg2==2.5.4
 jsonschema==2.3.0
-git+https://github.com/alphagov/digitalmarketplace-utils.git@0.14.0#egg=digitalmarketplace-utils==0.14.0
+git+https://github.com/pcraig3/Flask-FeatureFlags.git@master#egg=Flask-FeatureFlags==0.6-dev
+git+https://github.com/alphagov/digitalmarketplace-utils.git@0.17.0#egg=digitalmarketplace-utils==0.17.0
```
TheDoubleK: Utils is up to 0.18.0 now - keep up :)

TheDoubleK: I think it's probably worth pushing a new version tag for 0.18.0 and updating these pull-requests to use it.

app/status/views.py
pcraig3: `get_version_label()` is used by all of the apps using exactly the same code. 
`get_db_version()` is only used by this one, so I figured it didn't make sense to bring it into utils.
Does it?

requirements.txt
pcraig3: Sure, can do.

app/status/views.py
TheDoubleK: No that's fine.  I was confused between utils and dmutils, thought they were the same thing... not having a good day for following what's going on, sorry!

PR #138, opened 2015-06-08
Updates to DATA API:

minglis:
- Add updated at field to JSON response for a service. Format is as input - YYYY-MM-DDTHH:MM:SS
- Note timezone is curious here - we use utcnow() to create dates but this doesn't set timezone - so the test asserts the above, even though allowed input formats include the timezone
- In terms of format strings:
	- %Y-%m-%dT%H:%M:%S%Z (prefered JSON Schema)
	- %Y-%m-%dT%H:%M:%S (whats returned from API)

Also removed page size / pagination from API calls to get services by supplier. Supplier API calls now return ALL services.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -211,6 +211,7 @@ class Service(db.Model):
             'supplierId': self.supplier.supplier_id,
             'supplierName': self.supplier.name,
             'frameworkName': self.framework.name,
+            'updatedAt': self.updated_at.strftime("%Y-%m-%dT%H:%M:%S%Z"),
             'status': self.status
         })
 
--- a/app/models.py
+++ b/app/models.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from sqlalchemy.dialects.postgresql import JSON
 
 from . import db
+from . import formats
 from .utils import link, url_for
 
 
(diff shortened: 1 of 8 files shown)


minglis:
Also - added an endpoint to get all drafts by supplier;

As in the other supplier endpoints - this is UNPAGINATED.

app/models.py
```diff
@@ -211,6 +211,7 @@ def serialize(self):
             'supplierId': self.supplier.supplier_id,
             'supplierName': self.supplier.name,
             'frameworkName': self.framework.name,
+            'updatedAt': self.updated_at.strftime("%Y-%m-%dT%H:%M:%S%Z"),
```
robyoung: `updatedAt` in the `User` model doesn't explicitly format the timestamp. We should be consistent. Also, there is [`datetime.isoformat`](https://docs.python.org/2/library/datetime.html#datetime.datetime.isoformat)

robyoung:
Oops, sorry, didn't mean to close then.

app/models.py
minglis: Pulled the format out into a file, used on User dates now too.

PR #140, opened 2015-06-11
Wire in audit events

minglis:
Delivers: https://www.pivotaltracker.com/story/show/96550602
And a bit of: https://www.pivotaltracker.com/story/show/95513156

Audit events are now added to the following routes:

User:
- create
- update

Drafts
- create
- update
- delete
- publish

Services
- create
- update
- change status

In addition to the ones the suppliers route already had.

these events are at this stage all the "happy path" states, that is the action has been completed. there are no current "bad event" audits, such as validation errors, failed logins and so on.

AuditEvents are in an Enum in the utils repo:
https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/audit.py

Caveat:
- there is an "object" field on an audit event. When creating a new thing (user say) that has yet to be persisted you can't stick this in the AuditEvent, in the way you can with updates and so on. So these are None. happy to chat about what we do here. I think it's cause it hasn't got an ID yet.


minglis:
Doh, need to release the utils.

app/models.py
```diff
@@ -431,6 +412,35 @@ class AuditEvent(db.Model):
         object_type, object_id
     )
 
+    def __init__(self, audit_type, user, data, db_object):
+        if not AuditTypes.is_valid_audit_type(audit_type):
+            raise ValueError("{} is not a valid audit type".format(audit_type))
+        self.type = audit_type
```
allait: Can we add `.value` here instead of on every call? I assume we're only ever going to create events using the `AuditTypes`

app/main/views/services.py
```diff
@@ -194,11 +208,17 @@ def import_service(service_id):
...
     service.status = service_data.pop('status', 'published')
-    service.updated_by = updater_json['updated_by']
-    service.updated_reason = updater_json['update_reason']
     service.data = service_data
 
+    audit = AuditEvent(
+        audit_type=AuditTypes.import_service.value,
+        user=updater_json['updated_by'],
+        data=service_json,
+        db_object=None
```
allait: I think if you do:
```
db.session.add(service)
db.session.flush()
db.session.add(audit)
db.session.commit()
```
then you could use `db_object=service`

app/models.py
```diff
@@ -431,6 +412,35 @@ class AuditEvent(db.Model):
...
+        self.user = user
+
+    def serialize(self):
+        """
+        :return: dictionary representation of an audit event
+        """
+
+        data = dict()
+
+        data.update({
```
allait: `data = `?

app/models.py
minglis: Nice, tidier, also ensures model has only right types passed in. Done.

app/main/views/services.py
minglis: Done this, note that 
1) Audit events have to be made after flush

2) flush will throw the integrity check exceptions that normally happen on commit, so the whole session.add, session.flush, make audit event, session.add(audit) are now in the same try/catch as commit.

app/main/views/services.py
```diff
@@ -194,11 +206,19 @@ def import_service(service_id):
     service.updated_at = now
     service.created_at = now
     service.status = service_data.pop('status', 'published')
-    service.updated_by = updater_json['updated_by']
-    service.updated_reason = updater_json['update_reason']
     service.data = service_data
 
     db.session.add(service)
+    db.session.flush()
```
allait: I guess this should be inside the `try` block too like the user one

migrations/versions/50_remove_updated_details.py
```diff
@@ -0,0 +1,36 @@
...
+
+Revision ID: 4236b7b9bf0f
+Revises: 30_add_audit_events
+Create Date: 2015-06-10 08:04:00.175499
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '50_remove_updated_details'
+down_revision = '30_add_audit_events'
```
allait: This revision doesn't exist, should be `50_add_audit_events` instead

json_schemas/services-update.json
```diff
@@ -6,13 +6,9 @@
   "properties": {
     "updated_by": {
       "type": "string"
-    },
-    "update_reason": {
```
allait: `dmutils.apiclient` is still sending `update_reason`, so we should either keep it as non-required field or set `additionalProperties: true` for now

PR #145, opened 2015-06-16
Adds old and new archived service links to audit events

allait:
Part of the diff services story: [#95431862](https://www.pivotaltracker.com/story/show/95431862)

Changes "service_update" audit events data to:
```json
 "data": {
    "new_archived_service": "http://localhost:5000/archived-services/12403",
    "old_archived_service": "http://localhost:5000/archived-services/12402",
    "service_id": "5769999037759488"
}
```

With links to the related archived services API client can request the archives and create a diff of the change.

### Add models.ServiceTableMixin for common service columns
Removes duplication between Service, ArchivedService and DraftService
models by creating a common base class.

Also updates ArchivedService and DraftsService .serialize method and
API responses using them with new fields that were added to Service
.serialize.

### Add a migration to archive current services
Since the new update endpoint creates an archive for the updated
service version and the existing code created version for services
before updates there are no corresponding archived_services rows
for current versions of services. So we copy all the services rows
to archived_services with a migration.

### Create archived service immediately after update
Instead of archiving the previous service version during update we
archive the new one right away. This way there's always an archive
version to reference from the audit event.

This also means that service import now creates a service archive.

### Use 'update' as a common key for request data in audit event
Only store the request key that contains the data update instead
of all request data. Uses the same AuditEvent.data['update'] key
for all object types.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -187,7 +187,7 @@ def import_service(service_id):
 
     service_data = drop_foreign_fields(
         service_json,
-        ['supplierName', 'links', 'frameworkName']
+        ['supplierName', 'links', 'frameworkName', 'updatedAt']
     )
 
     framework = detect_framework_or_400(service_data)
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -4,18 +4,22 @@ from dmutils.audit import AuditTypes
 from flask import jsonify, abort, request, current_app
 
 from .. import main
-from ... import db
-from ...models import ArchivedService, Service, Supplier, Framework, AuditEvent
+from ...models import ArchivedService, Service, Supplier, Framework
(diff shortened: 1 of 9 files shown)


app/main/views/services.py
```diff
@@ -286,39 +258,23 @@ def update_service_status(service_id, status):
...
-            "old_status": prior_status,
-        },
-        db_object=None
-    )
+    update_json = validate_and_return_updater_request()
 
-    db.session.add(service)
-    db.session.add(audit)
-    db.session.add(service_to_archive)
+    prior_status, service.status = service.status, status
```
pcraig3: :open_mouth:

quis: That is well Pythonic.

PR #146, opened 2015-06-17
[#96551096] SCS validation in API

TheDoubleK:
This partially fulfils: https://www.pivotaltracker.com/story/show/96551096

This adds a new validation method that returns a list of all validation errors when validating against a JSON schema, with the option to enforce 'required' fields or not.

It isn't yet wired in to any endpoint, but in future validation messages on editing/saving pages of questions about a service should use this method to get the list of fields with errors (if any), to display in the frontend app.

This also pulls in the libraries required for the `uri` and `date-time` JSON schema formats to be validated properly (previously any string was accepted).


app/validation.py
```diff
@@ -36,13 +37,22 @@ def load_schemas(schemas_path, schema_names):
             validator = validator_for(schema)
             validator.check_schema(schema)
             loaded_schemas[schema_name] = schema
-    return loaded_schemas
 
-_SCHEMAS = load_schemas(JSON_SCHEMAS_PATH, SCHEMA_NAMES)
+            optional_schema = {}
```
allait: Another way to do this:
```python
optional_schema = schema.copy()
optional_schema.pop('required', None)
```

app/validation.py
```diff
@@ -115,6 +125,22 @@ def validates_against_schema(validator_name, submitted_json):
...
+def get_validation_errors(validator_name, json_data, enforce_required=True):
+    error_map = {}
+    try:
+        validator = get_validator(validator_name, enforce_required)
+        errors = sorted(validator.iter_errors(json_data), key=lambda e: e.path)
+        for index, error in enumerate(errors):
+            if error.path:
+                key = error.path.pop()
+            else:
+                key = "err_{}".format(index)
```
allait: How are we going to tie the errors to individual fields for display? Is it possible to do something like this instead:
```json
{
  "errors": [
     "Request isn't even JSON"
  ],
  "field_errors": {
    "serviceName": "Bad name",
    "pricingDocumentURL": "Not a URL"
}
```

If we can get more than one error per field from the jsonschema, then I'd change that to `"field": ["error 1", "error 2"], "other_field": ["single error"]`.

Alternatively, instead of splitting "errors" and "field_errors" we could have a special key for form errors (eg "_form").

TheDoubleK: > How are we going to tie the errors to individual fields for display?
The keys here (from `error.path`) are the individual field names that are in error.  The frontend apps will be responsible for mapping these keys to the message to be displayed to the user.  The plan is that each field will only have a single error message, so we don't really care if there are multiple errors on the same field - the same error message will be displayed regardless. 

It's only "form errors" that will get the 'err_x' key - these are effectively a special key for form errors.

allait: What's the reason behind the `err_%d` format? It seems like it'd be harder to use than a JSON list of errors.

TheDoubleK:
This is fixed up now - with a new method to pass in a list of required fields, and schema-level errors in a list under the `_form` key.  (Not yet a JSON list as it will be up to the endpoints that use this method to convert to JSON for the response).

app/validation.py
```diff
@@ -41,8 +42,12 @@ def load_schemas(schemas_path, schema_names):
 _SCHEMAS = load_schemas(JSON_SCHEMAS_PATH, SCHEMA_NAMES)
 
 
-def get_validator(schema_name):
-    schema = _SCHEMAS[schema_name]
+def get_validator(schema_name, enforce_required=True, required_fields=[]):
```
allait: It's a good idea to avoid using mutable objects as default argument values since they're only created once during function object creation and are shared between all calls. The usual workaround is:
```python
def get_validator(..., required_fields=None):
    if required_fields is None:
        required_fields = []
```

TheDoubleK: Ah didn't know that, good spot.

TheDoubleK:
Fixed and rebased.  Don't think we need the workaround in the `get_validation_errors` method, as it works OK passing the `None` through.

app/validation.py
```diff
@@ -115,6 +122,28 @@ def validates_against_schema(validator_name, submitted_json):
...
+        for error in errors:
+            if error.path:
+                key = error.path.pop()
+                error_map[key] = error.message
+            else:
+                form_errors.append(error.message)
+        if form_errors:
+            error_map['_form'] = form_errors
+        return error_map
+    except:
```
allait: The `try..except` block doesn't seem to do anything

app/validation.py
```diff
@@ -41,8 +42,14 @@ def load_schemas(schemas_path, schema_names):
...
 
-def get_validator(schema_name):
-    schema = _SCHEMAS[schema_name]
+def get_validator(schema_name, enforce_required=True, required_fields=None):
+    if required_fields is None:
+        required_fields = []
+    if enforce_required:
+        schema = _SCHEMAS[schema_name]
+    else:
+        schema = _SCHEMAS[schema_name].copy()
```
robyoung: This only does a shallow copy, given that the schemas are nested dictionaries I think this should use [`copy.deepcopy`](https://docs.python.org/3.4/library/copy.html#copy.deepcopy)

allait: Nope

robyoung: Thanks. To add a bit more detail, the only change to the schema is at the top level so a deep copy isn't needed.

TheDoubleK: Bu the copy will need to include any deeper nested elements for doing validation against, so maybe we do need it?

allait: The only reason we need `.copy` here is to avoid modifying the original schema. The copied object still has all the nested elements, but they're shared with the original. `.deepcopy` would create copies of the nested structures, but since we're not actually modifying any of them there's no need for us to do this.

TheDoubleK:
Removed the try block and rebased.

PR #148, opened 2015-06-17
[#96549780] Add G-Cloud 7 and Framework list

robyoung:
This builds on #142 to actually add new G-Cloud 7 framework with a 'pending' status.

This change removes the now redundant 'expired' flag as it is covered by the 'status' field.

This change also fixes a bug in the list services view where the Framework status filter was not being applied when filtering by status. Some of the querying logic has been moved into the Service model so that it can be tested in isolation and shared between multiple places. This logic has been added to the Service rather than the mixin because the logic is only needed on the Service right now.

### DIFF
--- a/tests/app/helpers.py
+++ b/tests/app/helpers.py
@@ -71,40 +71,46 @@ class BaseApplicationTest(object):
                 )
             db.session.commit()
 
-    def setup_dummy_services_including_unpublished(self, n):
+    def setup_dummy_framework(self, id=123, name='expired', framework='gcloud',
+                              status='expired'):
+        db.session.add(Framework(
+            id=id,
+            name=name,
+            framework=framework,
+            status=status))
+
+        return id
+
+    def setup_dummy_service(self, service_id, supplier_id=1, data=None,
+                            status='published', framework_id=1):
         now = datetime.utcnow()
+        db.session.add(Service(service_id=service_id,
+                               supplier_id=supplier_id,
+                               status=status,
+                               data=data or {
+                                   'serviceName': 'Service {}'.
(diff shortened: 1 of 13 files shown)


tests/app/helpers.py
```diff
@@ -71,40 +71,46 @@ def setup_dummy_suppliers(self, n):
                 )
             db.session.commit()
 
-    def setup_dummy_services_including_unpublished(self, n):
+    def setup_dummy_framework(self, id=123, name='expired', framework='gcloud',
```
allait: We probably can get away without this since frameworks are created by migrations and we have the G4 expired one.

TheDoubleK:
Won't we need the API to also return services in 'pending' frameworks sometimes?  e.g. suppliers who need to view their submitted G7 services. It should be up to the front-end rather than the API to decide whether to display a service, perhaps?  Or failing that we will need a new endpoint to retrieve services in 'pending' frameworks.

allait:
@TheDoubleK we can add a /services query parameter for other framework statuses. @robyoung do we need it for G-Cloud 7 now or will the G7 services list be based on drafts?

robyoung:
@TheDoubleK yes, good point. I think that will come out from making the G7 dashboard. I'm not entirely sure how we want to retrieve them yet.

TheDoubleK:
I think this is open for discussion - it's a question of whether we stick with drafts-only until the framework goes live (and add a "submitted" flag to the draft service JSON), or whether the service becomes a "real" service with a shiny new ID when it is complete and the supplier clicks "submit", even if it's not live yet.

PR #149, opened 2015-06-17
Adding ack audit events

minglis:
Adds endpoints to handle fetch and acknowledge of audit events:

for story:
https://www.pivotaltracker.com/story/show/95513156

- GET /audit-events fetches audit events
- uses query params: 
- page (for pagination)
- audit-type (based on AuditEvent enum)
- acknowledged (filter on acknowledged or otherwise)
- audit-date (filter by day of audit event)

- POST /audit-event/{:id}/acknowledged updates an audit event to acknowledged state
- uses updated JSON block for "who" informatiopn

Also updated the "update-service" audit event to capture supplier id and name, for display elsewhere.


app/main/views/audits.py
```diff
@@ -39,3 +67,34 @@ def list_audits():
...
+    try:
+        audit_id = int(audit_id)
+    except ValueError:
+        abort(400, "Invalid audit id argument")
+
+    updater_json = validate_and_return_updater_request()
+
+    audit_event = AuditEvent.query.filter(
+        AuditEvent.id == audit_id
+    ).first()
```
robyoung: This can be replaced with `Audit.query.get(audit_id)`.

app/models.py
```diff
@@ -364,11 +364,24 @@ class AuditEvent(db.Model):
...
+        db.Boolean,
+        index=True,
+        unique=False,
+        nullable=False)
+
+    acknowledged_by = db.Column(db.String)
+    acknowledged_at = db.Column(
+        db.DateTime,
+        nullable=True,
+        default=datetime.utcnow)
```
robyoung: Providing a default here will set the `acknowledged_at` time when the record is created.

app/models.py
```diff
@@ -364,11 +364,24 @@ class AuditEvent(db.Model):
         object_type, object_id
     )
 
+    acknowledged = db.Column(
+        db.Boolean,
+        index=True,
+        unique=False,
+        nullable=False)
+
+    acknowledged_by = db.Column(db.String)
```
robyoung: `nullable=True` is the default but in other places we're being explicit.

app/models.py
```diff
@@ -364,11 +364,24 @@ class AuditEvent(db.Model):
...
+        db.DateTime,
+        nullable=True,
+        default=datetime.utcnow)
+
     def __init__(self, audit_type, user, data, db_object):
         self.type = audit_type.value
         self.data = data
         self.object = db_object
         self.user = user
+        self.acknowledged = False
```
robyoung: Why can this not be default on the column?

app/models.py
```diff
@@ -385,6 +399,14 @@ def serialize(self):
                 "self", url_for(".list_audits"))
         }
 
+        if self.acknowledged:
+            data.update({
+                'acknowledged_at':
+                    self.acknowledged_at.strftime("%Y-%m-%dT%H:%M:%S%Z"),
```
robyoung: These should use the format string from dmutils. See: #143

app/validation.py
```diff
@@ -176,3 +178,17 @@ def is_valid_service_id_or_400(service_id):
...
+        datetime.strptime(date, default_format)
+        return True
+    except ValueError:
+        return False
+
+
+def is_valid_acknowledged_state(acknowledged):
+    if acknowledged in ['all', 'acknowledged', 'not-acknowledged']:
+        return True
+    return False
```
robyoung: Really nitpicky, but this could just be `return acknowledged in ['all', 'acknowledged', 'not-acknowledged']`

migrations/versions/60_archive_current_services.py
```diff
@@ -8,7 +8,7 @@
 
 # revision identifiers, used by Alembic.
 revision = '60_archive_current_services'
-down_revision = '50_remove_updated_details'
+down_revision = '60_add_acknowledged_column'
```
robyoung: Updating the down revision of previous migrations sounds like a bad idea. Surely this will fail if the previous migration has already gone out?

migrations/versions/70_add_framework_and_status.py
```diff
@@ -8,7 +8,7 @@
 
 # revision identifiers, used by Alembic.
 revision = '70_add_framework_status'
-down_revision = '60_archive_current_services'
+down_revision = '70_acknowledged_not_null'
```
robyoung: Again, this feels very wrong.

app/validation.py
```diff
@@ -176,3 +178,17 @@ def is_valid_service_id_or_400(service_id):
         return True
     else:
         abort(400, "Invalid service ID supplied: %s" % service_id)
+
+
+def is_valid_date(date, default_format='%Y-%m-%d'):
```
pcraig3: Would it be worth setting this as DATE_FORMAT in [dmutils.formats](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/formats.py)?

<sub>I've [used the same format](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/status.py#L31) for our FeatureFlags, so it might be a good idea.</sub>

app/models.py
minglis: yep good spot.

app/main/views/audits.py
```diff
@@ -39,3 +67,32 @@ def list_audits():
             request.args
         )
     )
+
+
+@main.route('/audit-events/<string:audit_id>/acknowledge', methods=['POST'])
```
allait: `<int:audit_id>`

app/main/views/audits.py
```diff
@@ -26,6 +40,20 @@ def list_audits():
         else:
             abort(400, "Invalid audit type")
 
+    acknowledged = request.args.get('acknowledged', None)
+    if acknowledged:
+        if is_valid_acknowledged_state(acknowledged):
+            if acknowledged == 'acknowledged':
```
allait: Is it worth using `dmutils.config.convert_to_boolean` here? `&acknowledged=acknowledged` doesn't look very nice

app/main/views/services.py
```diff
@@ -190,8 +195,14 @@ def import_service(service_id):
     service.status = service_data.pop('status', 'published')
     service.data = service_data
 
+    audit_data = service_data.copy()
```
allait: Why do we need the full service data in the audit event?

app/main/views/services.py
```diff
@@ -135,8 +135,13 @@ def update_service(service_id):
         service, update, update_details
     )
 
+    audit_data = {
+        'supplierName': service.supplier.name,
+        'supplierId': service.supplier.supplier_id
```
allait: `commit_and_archive_service` adds service id key `service_id`. We should probably change either that one or `supplierId` to have the same format.

app/main/views/audits.py
```diff
@@ -39,3 +67,32 @@ def list_audits():
...
+
+    updater_json = validate_and_return_updater_request()
+
+    audit_event = AuditEvent.query.get(audit_id)
+
+    if audit_event is None:
+        abort(404, "No audit event with this id")
+
+    audit_event.acknowledged = True
+    audit_event.acknowledged_at = datetime.now()
```
allait: `.utcnow()`

app/main/views/services.py
minglis: Think this correct, the other needs to change. I'll change it now. May have knock on effects.

app/main/views/services.py
minglis: Yeah, seeing as we have the archived links now esp.

app/main/views/audits.py
minglis: Gives a 404 not a 400. Was thinking that the bad request response was preferable. Thoughts?

app/main/views/audits.py
minglis: changing to true / false / all

allait: Do we need "all" as a value? For other filters missing value means "all".

app/main/views/audits.py
allait: I don't really mind either status. We do use `<int:...>` for suppliers and users though, and it looks slightly easier than the `try: int(...) ...` check.

app/models.py
```diff
@@ -426,20 +438,29 @@ def serialize(self):
...
-                    self.data.get('new_archived_service_id')
+                "newArchivedService": ArchivedService.link_object(
+                    self.data.get('newArchivedServiceId')
                 )
             })
         }
 
+        if self.acknowledged:
+            data.update({
+                'acknowledged_at':
```
allait: `acknowledgedAt`?

migrations/versions/60_add_acknowledged_column.py
```diff
@@ -0,0 +1,41 @@
...
+
+Revision ID: 2d2309ce61f
+Revises: 50_remove_updated_details
+Create Date: 2015-06-17 11:40:51.427138
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '60_add_acknowledged_column'
+down_revision = '100_add_open_status'
```
robyoung: The down revision is `100_add_open_status` but this revision is `60_add_acknowledged_column`. Also, it doesn't match what's in the comment above.

minglis: I had to replay my migrations on top of the existing ones to sort the multiple heads issues. Thinking the numbers are fairly redundant?

migrations/versions/110_add_acknowledged_column.py
```diff
@@ -0,0 +1,41 @@
+"""empty message
+
+Revision ID: 2d2309ce61f
```
TheDoubleK: The revisions/revises in the comments in these migration scripts are all wrong.  Would be nice if they match up to the revision/down_revision numbers, for clarity/sanity.

PR #150, opened 2015-06-18
Add supplier service counts

allait:
[#97275278](https://www.pivotaltracker.com/story/show/97275278)

Adds 'service_counts' key to the supplier response with service
counts for each framework:

```json
{
  ...
  "service_counts": {
    "G-Cloud 5": 10,
    "G-Cloud 6": 5,
  }
}
```

Only published services for "live" frameworks are counted.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -2,11 +2,11 @@ from datetime import datetime
 from flask_sqlalchemy import BaseQuery
 
 from sqlalchemy import asc
+from sqlalchemy import func
 from sqlalchemy.dialects.postgresql import JSON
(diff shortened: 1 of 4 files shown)


app/models.py
```diff
@@ -138,10 +138,21 @@ class Supplier(db.Model):
 
     clients = db.Column(JSON, default=list)
 
+    def get_service_counts(self):
```
minglis: Do we need a query here or can we use existing "fetch all" and just group in code? 

Also is this query properly indexed? how does it perform? Counts in postgres can be slow.

PR #151, opened 2015-06-19
Update DATETIME and use `utcnow()`

pcraig3:
A few small (but important!) changes -- shouldn't cause any problems.

* Update to new dmutils with updated `DATETIME_FORMAT`
* Replace date format strings in tests or model definitions.
* Change all uses of `.now()` to `.utcnow()`


TheDoubleK:
:+1:  - will merge once Travis catches up.

pcraig3:
Updated!  Should be good.

PR #153, opened 2015-06-22
[#97275564] DB migration: service IDs are not required for draft services

TheDoubleK:
Newly created drafts for service submissions will not have a related service ID
- service IDs will only be generated when a draft service is published, so should
be able to be null in the draft services table.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -29,6 +29,10 @@ def create_draft_service(service_id):
         Service.service_id == service_id
     ).first_or_404()
 
+    if DraftService.query.filter(
+        DraftService.service_id == service_id
+    ).first():
+        abort(400, "Draft already exists for service {}".format(service_id))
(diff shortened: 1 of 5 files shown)


app/main/views/drafts.py
```diff
@@ -29,6 +29,10 @@ def create_draft_service(service_id):
         Service.service_id == service_id
     ).first_or_404()
 
+    if DraftService.query.filter(
+        DraftService.service_id == service_id
+    ).first():
+        abort(400, "Draft already exists for service {}".format(service_id))
```
robyoung: I find that quite hard to read. Would this be easier to read?
```
draft_service = DraftService.query.filter(
    DraftService.service_id == service_id
).first()
if draft_service:
```

TheDoubleK: It's marginal ;)

But can change it no worries.

PR #155, opened 2015-06-23
Add selection questions

robyoung:
This PR adds a slug field to the `Framework` model so that we can use meaningful framework identifiers in URLs.

It also adds a new `SelectionQuestions` model that joins `Supplier` and `Framework` and allows answers to the selection questionnaire to be stored. There is a new supplier route that allows these questions to be set and retrieved for a framework.

`/suppliers/<supplier-id>/selection-questions/<framework-slug>`

Note that selection questions cannot be submitted for a framework that does not have an `open` status.


robyoung:
Closing while I add the auditing; see: https://github.com/alphagov/digitalmarketplace-utils/pull/83

PR #156, opened 2015-06-23
[#97275564] Create new G7 drafts

TheDoubleK:
NOTE: This PR is rebased to include the database migrations in this separate pull-request, which should be reviewed and merged first: https://github.com/alphagov/digitalmarketplace-api/pull/153

Together they fulfil: https://www.pivotaltracker.com/story/show/97275564

This creates a new endpoint to create a new draft service (previously drafts could only be made by copying an existing service).

It also refactors the existing draft endpoints to use draft_id instead of service_id.


robyoung:
Do we need a route for listing draft services? ie. What lives at `/draft-services`?

TheDoubleK:
There is a get on that route - it requires at least `supplier_id` - to get the drafts for a supplier.  It can optionally be filtered by `framework` and `service_id` (to get drafts of a particular existing service).

I don't think there is a use case for retrieving all drafts in the system, unfiltered by supplier.

TheDoubleK:
Rebased to fix merge conflicts.

app/main/views/drafts.py
```diff
@@ -101,8 +112,10 @@ def edit_draft_service(service_id):
 @main.route('/draft-services', methods=['GET'])
 def list_drafts():
```
pcraig3: I know it's not technically part of your pull request, but I'm wondering why this method isn't called `list_draft_services`.
All of the other method names have the word "service" in them somewhere.

app/main/views/drafts.py
```diff
@@ -36,16 +39,19 @@ def create_draft_service(service_id):
         abort(400, "Draft already exists for service {}".format(service_id))
 
     draft = DraftService.from_service(service)
+
+    db.session.add(draft)
+    db.session.flush()
```
allait: Flush can trigger an `IntegrityError`, so it might be worth wrapping it with `try ... except`

app/main/views/drafts.py
```diff
@@ -36,16 +39,19 @@ def create_draft_service(service_id):
...
+
     audit = AuditEvent(
         audit_type=AuditTypes.create_draft_service,
         user=updater_json['updated_by'],
         data={
+            "draft_id": draft.id,
```
allait: Audit events in `service_utils` were changed to use the `serviceId` as key, we should probably change this one to match the new key convention

app/main/views/drafts.py
```diff
@@ -180,37 +204,40 @@ def delete_draft_service(service_id):
...
-    archived_service = ArchivedService.from_service(service)
-    new_service = update_and_validate_service(
-        service,
-        draft.data,
-        updater_json)
+    else:
+        new_service = Service.create_from_draft(draft, "enabled")
+        validate_service(new_service)
 
     audit = AuditEvent(
```
allait: It'd be nice to use `service_utils.commit_and_archive_service` here since it has the common audit event format for the service updates that allows displaying diff of the change and does the right save order when archiving the service: update first, then create the archive copy of the new data.

app/main/views/drafts.py
```diff
@@ -180,37 +204,40 @@ def delete_draft_service(service_id):
...
     draft = DraftService.query.filter(
-        DraftService.service_id == service_id
+        DraftService.id == draft_id
     ).first_or_404()
 
-    service = Service.query.filter(
-        Service.service_id == draft.service_id
-    ).first_or_404()
+    if draft.service_id:
+        service = Service.query.filter(
```
allait: This is somewhat similar to what the service update endpoint does, is it possible to extract it/parts of it to `service_utils`?

app/main/views/drafts.py
```diff
@@ -229,3 +256,58 @@ def publish_draft_service(service_id):
...
+    draft_json = validate_and_return_draft_request()
+
+    # TODO: Get framework id by matching the slug once in Framework table.
+    framework_id = Framework.query.filter(
+        Framework.name == 'G-Cloud 7'
+    ).first_or_404().id
+
+    # TODO: Reject any requests on a Framework that is not currently 'open'
+
+    supplier_id = draft_json['supplierId']
```
allait: I couldn't find any checks for `supplierId` and `lot` in `validate_and_return_draft_request`, so this could end up being a `KeyError`

app/main/views/drafts.py
```diff
@@ -229,3 +256,58 @@ def publish_draft_service(service_id):
...
+    lot = draft_json['lot']
+    errs = get_draft_validation_errors(draft_json, lot, slug=framework_slug)
+    if errs:
+        return jsonify(errors=errs), 400
+
+    draft_json = drop_foreign_fields(draft_json, ['supplierId'])
+    now = datetime.utcnow()
```
allait: I think we can just set this as default value for `created_at` and `updated_at` (like https://github.com/alphagov/digitalmarketplace-api/blob/master/app/models.py#L407), that way the correct time will be set when the object is saved for the first time.

app/models.py
```diff
@@ -390,12 +404,13 @@ def from_service(service):
 
     def serialize(self):
         data = super(DraftService, self).serialize()
+        data['id'] = self.id
         data['service_id'] = self.service_id
```
allait: `serviceId`

app/models.py
```diff
@@ -471,3 +486,8 @@ def filter_null_value_fields(obj):
     return dict(
         filter(lambda x: x[1] is not None, obj.items())
     )
+
+
+def generate_new_service_id():
+    # TODO: Decide what we want G7 service IDs to look like and implement
```
allait: Is this comment still valid or have we chosen to use the `uuid4`? The method is being used by the model.

app/main/views/drafts.py
TheDoubleK: I think making it use `service_utils.commit_and_archive_service` should be enough in terms of extracting common parts.

app/main/views/drafts.py
TheDoubleK: The best place to do this looks like it would be in the ServiceTableMixin.  The only issue being that there is the chance that the initial updatedAt and createdAt might not be microsecond-equivalent.  Do we care about that?

app/models.py
TheDoubleK: The comment is still valid - the uuid4 is here as a placeholder until we resolve the issue.  @minglis was going to add a story to Pivotal about this.

TheDoubleK:
Updated and comments replied to.

app/main/views/drafts.py
allait: I don't see us using them in any queries with direct comparison, so I think it should be ok. I think we can even make SQLAlchemy to set the `updatedAt` value for as on each write (http://stackoverflow.com/a/13978770), but that's probably more than we need for now.

app/models.py
allait: Do we expect to call the draft publishing endpoint before we fix the service id generation? If not,  we could make the method `raise NotImplementedError()`. Do uuid4 pass our validation rules for service ids?

I see the reason for doing this since we'll reset our database once we go live anyway, but it still feels slightly dangerous to roll out code that will knowingly create incorrect data.

TheDoubleK: I think the publish flow is still very much a work in progress with a fair amount of iteration to be done still (e.g. I think the publish endpoint will need to be aware of framework status and do different things with a draft depending on the status of it's framework).  I think the benefits of having end-to-end functionality outweigh the problems with having a temporary patch in here, so I'd argue to leave the method in.

The point about validation is a good one though - UUID won't match the current rules, and the fact that the tests pass show that we're missing at least one test on this - I'll fix it up.

app/draft_utils.py
```diff
@@ -0,0 +1,35 @@
...
+        json_has_matching_id(json_payload['services'], draft_id)
+    return json_payload['services']
+
+
+def get_draft_validation_errors(draft_json, lot,
+                                framework_id=0, slug=None, required=None):
+    if not slug and not framework_id:
+        raise Exception('Validation requires either framework_id or slug')
+    if not slug:
+        # TODO: Get framework slug from Framework table once it exists.
```
robyoung: #157 has been merged now so this could be uncommented.

TheDoubleK: Let's get this merged to unblock things and I'll do the TODOs soon.

PR #157, opened 2015-06-23
Add the ability to answer selection questions

robyoung:
This PR adds a slug field to the Framework model so that we can use meaningful framework identifiers in URLs.

It also adds a new SelectionAnswers model that joins Supplier and Framework and allows answers to the selection questionnaire to be stored. There is a new supplier route that allows these questions to be set and retrieved for a framework.

/suppliers/<supplier-id>/selection-answers/<framework-slug>

Note that answers to the selection questions cannot be submitted for a framework that does not have a status set to 'open'.

The response to `/suppliers/<supplier-id>/selection-answers/<framework-slug>` is
```json
{
  "selectionAnswers": {
    "supplierId": 1234,
    "frameworkSlug": "g-cloud-7",
    "questionAnswers": {
      "question": "answer"
    }
  }
}
```

The minimum required request is:
```json
{
  "updated_by": "user",
  "selectionAnswers": {
    "questionAnswers": {
      "question": "answer"
    }
  }
}
```


app/main/views/suppliers.py
```diff
@@ -245,3 +246,63 @@ def update_contact_information(supplier_id, contact_id):
         abort(400, "Database Error: {0}".format(e))
 
     return jsonify(contactInformation=contact.serialize())
+
+
+@main.route('/suppliers/<supplier_id>/selection-answers/<framework_slug>',
+            methods=['GET'])
+def get_selection_questions(supplier_id, framework_slug):
+    application = SelectionAnswers.query.find_by_supplier_and_framework(
```
allait: This could only return 1 or 0 items, should we make it a model method instead of a query one, since there's no other meaningful filtering to be done on the results?

robyoung: Good call

robyoung: Discussed in meatspace and decided that a static method is the lesser evil.

PR #158, opened 2015-06-24
[#97502388] Validation errors for service updates

TheDoubleK:
NOTE: This PR is branched from the following (to preserve schema filename changes), which should be reviewed and merged before this one (only the final two commits on this branch are new to this PR): https://github.com/alphagov/digitalmarketplace-api/pull/156

 This is the API end of the following story: https://www.pivotaltracker.com/story/show/97502388

In order to display validation errors on the supplier frontend the API needs to return per-field validation errors. The validation rules added to the schemas in this PR are on the supplier-editable fields only.

I have tested the import of G4 and G5 services against this branch with the latest dump data and the import runs fine with the additional validation rules in the G5 schema here.


TheDoubleK:
Finally fixed the word limit regex (the previous version hung), though it's pretty ugly:
`^\s?(?:\S+\s+){0,9}\S+\s?$`
Here's how it works...
* The first and last `\s?` allow leading and trailing whitespace to validate
* The non-capturing group `(?:\S+\s+){0,9}` matches zero to nine words
* The final `\S+` matches one word - the minimum required for a valid response

PR #159, opened 2015-06-24
[#97502388] Validation errors for service updates

TheDoubleK:
~~NOTE: This PR is branched from the following (to preserve schema filename changes), which should be reviewed and merged before this one (only the final two three commits on this branch are new to this PR): https://github.com/alphagov/digitalmarketplace-api/pull/156~~

 This is the API end of the following story: https://www.pivotaltracker.com/story/show/97502388

In order to display validation errors on the supplier frontend the API needs to return per-field validation errors. The validation rules added to the schemas in this PR are on the supplier-editable fields only.

I have tested the import of G4 and G5 services against this branch with the latest dump data and the import runs fine with the additional validation rules in the G5 schema here.

### DIFF
--- a/app/service_utils.py
+++ b/app/service_utils.py
@@ -3,8 +3,7 @@ from sqlalchemy.exc import IntegrityError
 
 from .utils import get_json_from_request, \
     json_has_matching_id, json_has_required_keys, drop_foreign_fields
-from .validation import validate_updater_json_or_400, detect_framework_or_400, \
-    get_validation_errors
+from .validation import validate_updater_json_or_400, get_validation_errors
 from . import search_api_client, apiclient
 from . import db
 from .models import ArchivedService, AuditEvent
@@ -62,7 +61,6 @@ def validate_service(service):
     errs = get_validation_errors(validator_name, data, enforce_required=True)
     if errs:
         abort(400, errs)
-    # detect_framework_or_400(data)
     return
 
 
--- a/app/service_utils.py
+++ b/app/service_utils.py
@@ -3,7 +3,8 @@ from sqlalchemy.exc import IntegrityError
 
 from .utils import get_json_from_request, \
     json_has_matching_id, json_has_required_keys, drop_foreign_fields
-from .validation import validate_updater_json_or_400, detect_framework_or_400
+from .validation import validate_updater_json_or_400, detect_framework_or_400, \
+    get_validation_errors
 from . import search_api_client, apiclient
 from . import db
 from .models import ArchivedService, AuditEvent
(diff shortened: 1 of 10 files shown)


TheDoubleK:
Finally fixed the word limit regex (the previous version hung), though it's pretty ugly:
`^\s?(?:\S+\s+){0,9}\S+\s?$`
Here's how it works...
* The first and last `\s?` allow leading and trailing whitespace to validate
* The non-capturing group `(?:\S+\s+){0,9}` matches zero to nine words
* The final `\S+` matches one word - the minimum required for a valid response

robyoung:
Is that regular expression supposed to only allow zero or one leading or trailing whitespace characters?

TheDoubleK:
Ah yep, I guess it should have a plus too.

TheDoubleK:
Aargh, no, would need to be a star, but I'll have to test tomorrow if that will break and hang again...

TheDoubleK:
@robyoung, what do you think to dropping the whitespace from the regex and having the validation method trim all values before validating? It would make the regexes much nicer. Can you think of a case where we would need to preserve surrounding whitespace?

TheDoubleK:
Fixed the word count regexes and added a test for them.

TheDoubleK:
Rebased. Previous Travis failure was do to a failure to download some requirements, so hopefully should pass OK this time.

tests/app/views/test_services.py
```diff
@@ -599,8 +600,8 @@ def test_invalid_field_value_not_accepted_on_update(self):
                 content_type='application/json')
 
             assert_equal(response.status_code, 400)
-            assert_in('JSON was not a valid format',
-                      json.loads(response.get_data())['error'])
+            assert_in("'per Truth' is not one of",
```
quis: @TheDoubleK Excellent. cc @katieminatie

app/service_utils.py
```diff
@@ -30,6 +31,24 @@ def update_and_validate_service(service, service_payload):
 
 
 def validate_service(service):
+
+    # TODO: Get framework slug from Framework table once it exists.
```
robyoung: This is merged now, can we uncomment this yet?

app/service_utils.py
```diff
@@ -40,7 +59,10 @@ def validate_service(service):
     data = drop_foreign_fields(
         data,
         ['service_id', 'supplierName', 'links', 'frameworkName', 'updatedAt'])
-    detect_framework_or_400(data)
+    errs = get_validation_errors(validator_name, data, enforce_required=True)
+    if errs:
+        abort(400, errs)
+    # detect_framework_or_400(data)
```
robyoung: Can this be removed?

app/service_utils.py
TheDoubleK: There's a few places this can be done now the Frameworks are in.  I don't think it needs to be in this PR, which I am sick of the sight of.

app/service_utils.py
TheDoubleK: Yes.

PR #160, opened 2015-06-24
Api updates for supplier a to z

minglis:
Adds the ability to get suppliers on a specific framework

- fetch suppliers now has an additional param "framework"

- if supplied the query is restricted to only suppliers that have a published service on that framework

- additionally framework must be active

- Framework validates as a string

- Uses enum to check is a valid value before using the query

- Query is a subselect

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -26,7 +26,7 @@ def list_suppliers():
     suppliers = Supplier.query.order_by(Supplier.name)
 
     if prefix:
-        if prefix == 'other':
+        if prefix == '123':
             suppliers = suppliers.filter(
                 Supplier.name.op('~')('^[^A-Za-z]'))
         else:
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -70,7 +70,6 @@ def list_suppliers():
         abort(400, 'invalid framework')
 
 
-
 @main.route('/suppliers/<int:supplier_id>', methods=['GET'])
 def get_supplier(supplier_id):
     supplier = Supplier.query.filter(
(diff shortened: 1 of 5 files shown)


app/main/views/suppliers.py
```diff
@@ -23,10 +23,28 @@ def list_suppliers():
 
     prefix = request.args.get('prefix', '')
 
-    suppliers = Supplier.query.order_by(Supplier.name)
+    framework = request.args.get('framework')
+
+    if framework:
+        is_valid_string_or_400(framework)
```
TheDoubleK: Is the idea that we call this method for all URL query parameters as a basic check before proceeding?  Or is framework a special case for some reason?  If it's only our apps (our API client) calling the API do we really need this check here? Shouldn't we be checking user-supplied params at the frontend routes rather than in the API?

app/main/views/suppliers.py
```diff
@@ -23,10 +23,28 @@ def list_suppliers():
...
+
+        active_services = Service.query.join(
+            Service.framework
+        ).filter(
+            Framework.status == 'live',
```
TheDoubleK: If we're querying by framework why restrict it only to 'live' frameworks at the API?

app/validation.py
```diff
@@ -226,3 +222,20 @@ def is_valid_date(date, default_format=DATE_FORMAT):
...
     return acknowledged in ['all', 'true', 'false']
+
+
+def is_valid_string_or_400(string):
+    if not is_valid_string(string):
+        abort(400, "invalid value {}".format(string))
+
+
+def is_valid_string(string, minlength=1, maxlength=255):
+    regex_match_valid_service_id = r"^[A-z0-9-]{%s,%s}$" % (
```
TheDoubleK: Should the name of this regex change if we're using it to check more than just service IDs now?

app/main/views/suppliers.py
minglis: Thinking that the API might be public at some point.

PR #161, opened 2015-06-29
Change import to respect original dates

minglis:
Services no longer have dates in the JSON data column.

- createdAt can be set in submitted JSON to PUT endpoint
	- this is not  a required field
	- if present it is used as created_at data on the model
	- if not present we use utcnow instead
	- not persisted in the JSON data column

- Removed from the JSON Schemas for G6
	- lastUpdated
	- lastUpdatedByEmail
	- lastCompleted
	- lastCompletedByEmail

- Replaced with createdAt (See above)
- lastUpdated now set as utcnow() on the model in the last_updated column
- lastUpdatedByEmail - not used
- lastCompletedByEmail should be set as the updatedBy field in the importer - this will be used in the creation audit event

### DIFF
--- a/scripts/import_services_from_api.py
+++ b/scripts/import_services_from_api.py
@@ -111,9 +111,9 @@ def do_index(api_url, api_access_token, source_api_url,
             status = status and result
             print_progress(counter, start_time)
 
+    print_progress(counter, start_time)
     return status
 
-    print_progress(counter, start_time)
 
 if __name__ == "__main__":
     arguments = docopt(__doc__)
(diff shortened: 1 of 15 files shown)


scripts/import_services_from_api.py
```diff
@@ -41,8 +41,25 @@ def request_services(api_url, api_access_token, page=1):
...
+            if isinstance(links, list):
+                found = False
+                for link in links:
+                    if 'rel' in link:
+                        if link['rel'] == 'next':
+                            page += 1
+                            found = True
+
+                if not found:
+                    return
```
robyoung: Could this be rewritten as the following?
```if len([link for link in links if link['rel'] == 'next']) != 0:
  page += 1
```

PR #163, opened 2015-07-01
Generate new service ids

robyoung:
When 'publishing' a draft service we have to generate a service ID. This commit handles collisions by attempting to add the service in a nested transaction and then trying again, with a new generated service ID.

See:
http://docs.sqlalchemy.org/en/rel_1_0/orm/session_transaction.html#using-savepoint

After 5 attempts the IntegrityError is just raised and will result in a 500.

New service IDs for G7 are now generated as a random 16 digit number between 7e15 and 8e15. Collisions are handled and retries made.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -12,7 +12,7 @@ from ...validation import is_valid_service_id_or_400
 from ...models import Service, DraftService, Supplier, AuditEvent, Framework
 from ...service_utils import validate_and_return_updater_request, \
     update_and_validate_service, index_service, validate_service, \
-    commit_and_archive_service
+    commit_and_archive_service, create_service_from_draft
 from ...draft_utils import validate_and_return_draft_request, \
     get_draft_validation_errors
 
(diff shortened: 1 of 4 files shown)


TheDoubleK:
Whilst 5 collisions in a row is almost vanishingly unlikely, I'd be tempted to set the tries higher - even a few 100s of attempts would probably clock in under a second and cause no issues in the system. (These figures are, of course, pure speculation, but 5 seems low.)

TheDoubleK:
And while it is theoretically possible that it could loop infinitely trying to hit an unused ID, unless we expect to get up to tens of billions of G7 services I'd be happy letting it retry without a count at all.

robyoung:
That is a really good point. I'm not comfortable with an unbounded loop but maybe this loop is constrained on the wrong thing. What we want to ensure is that finding a new ID doesn't take too long, so the limiting clause could be based on time.

allait:
After a conversation with @robyoung we've decided to rollback the time-constrained loop in favour of a set number of retries. It seems that while neither limit should be normally reached, limit on retries is more predictable and should be easier to debug and measure with log metrics in case something misbehaves in the id generation.

PR #164, opened 2015-07-02
Add login tracking

robyoung:
[[#97201260]](https://www.pivotaltracker.com/story/show/97201260)
[[#98000176]](https://www.pivotaltracker.com/story/show/98000176)

This tracks the last time a user logged in and the number of failed login attempts they have had in a row. After the user has failed to log in 5 times in a row the account will be locked, preventing the user from logging in any more. It can only be unlocked by an API call.

This changes the behaviour of the `User.locked` field. Previously setting it to true had no effect, now it will prevent any further logins.


app/main/views/users.py
```diff
@@ -19,14 +19,25 @@ def auth_user():
...
+        user.failed_login_count = 0
+        db.session.add(user)
+        db.session.commit()
+
         return jsonify(users=user.serialize()), 200
     else:
+        login_limit = current_app.config['DM_FAILED_LOGIN_LIMIT']
+        user.failed_login_count += 1
+        if user.failed_login_count >= login_limit:
+            user.locked = True
```
allait: I don't think there's a need to set the locked flag when we can check the counter directly.
Is there any other use for the `user.locked` flag? And how is it different from `user.active`?

Can we check if they're set for any users in the user DB dump and decide whether we need them at all?
E.g. if the `locked` is not set for any users, I'd replace it with a property that'd check the `failed_login_count` value.
If it is set, then we should probably find out why and avoid overwriting the flag value so that we don't unlock a user later thinking they were locked for the failed attempts when the real reason might've been different.

robyoung: I think the `locked` and `active` flags were carried over from the Grails app, @minglis may know more about this.

TheDoubleK: We only export active and unlocked users from the Grails app, so no need to preserve either of these if we don't need them. Users are only locked for failed logins, no other reason.

allait:
### Replace User.locked column with a property
User.locked now checks the failed login counter and returns True if
it's above the set limit. The serialized user response still has the
"locked" field (with value derived from the login attempt counter),
so this doesn't affect the frontend apps.

migrations/versions/180_remove_user_locked_flag.py
```diff
@@ -0,0 +1,26 @@
...
+revision = '180_remove_user_locked_flag'
+down_revision = '170_add_login_tracking'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('users', 'locked')
```
TheDoubleK: Do we need the `active` column for anything at all?  If not then we could get rid of that at the same time maybe.

app/main/views/users.py
```diff
@@ -19,14 +19,22 @@ def auth_user():
...
-        User.email_address == json_payload['emailAddress'].lower()).first()
+    user = User.get_by_email_address(json_payload['emailAddress'].lower())
 
     if user is None:
         return jsonify(authorization=False), 404
-    elif encryption.checkpw(json_payload['password'], user.password):
+    elif encryption.authenticate_user(json_payload['password'], user):
+        user.logged_in_at = datetime.utcnow()
+        user.failed_login_count = 0
+        db.session.add(user)
```
TheDoubleK: Should we have try-catches around these database operations?  Or should we just let it fail if there is an error updating the user?

migrations/versions/170_add_login_tracking.py
```diff
@@ -0,0 +1,34 @@
...
+# revision identifiers, used by Alembic.
+revision = '170_add_login_tracking'
+down_revision = '160_add_framework_slug'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('users', sa.Column('failed_login_count', sa.Integer(), nullable=True))
```
TheDoubleK: `failed_login_count` is not nullable in the model, so probably shouldn't be nullable in the database either.

migrations/versions/180_remove_user_locked_flag.py
```diff
@@ -0,0 +1,26 @@
...
+    ### end Alembic commands ###
+
+
+def downgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('users', sa.Column('locked', sa.BOOLEAN(), autoincrement=False, nullable=False))
```
TheDoubleK: This downgrade fails - `column "locked" contains null values` - need to create column as nullable, populate then change to not null.

PR #170, opened 2015-07-09
Add a route to list users (optionally by `supplier_id`)

pcraig3:
### The thing I'm actively working on

As part of the [user dashboard](https://www.pivotaltracker.com/story/show/98628954) story on Pivotal, we need an API route that can return users associated with a certain supplier.

This pull request creates a `list_users` route to get back all users, optionally filterable by a `supplier_id` or an `email_address`. If an email address is specified that doesn't exist, a 404 is returned.
<del>It also changes the existing url path for the `get_user_by_email` route because it was in conflict with the `list_users` route.</del> 
The `get_user_by_email` route has been removed entirely.

For example:

```python

# /users?supplier_id=123456
# /users?email_address=liz@royal.gov.uk
@main.route('/users', methods=['GET'])
def list_users():

```

*** 
#### Housekeeping

* I changed the pep8 line number thing to 120 like we've done in a few of the other apps
* Put the page-getting logic into the `utils.py` file and replaced the page-number-getting try/except block everywhere I found it being used

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -49,18 +49,32 @@ def get_user_by_id(user_id):
 def list_users():
     page = get_valid_page_or_1()
 
+    # email_address is a primary key
+    email_address = request.args.get('email_address')
+    if email_address:
+        user = User.query.filter(
+            User.email_address == email_address.lower()
+        ).first()
+
+        if not user:
+            abort(404, "No user with email_address '{}'".format(email_address))
+
+        return jsonify(
+            users=[user.serialize()],
+            links={}
+        )
+
     supplier_id = request.args.get('supplier_id')
 
     if supplier_id is not None:
         try:
             supplier_id = int(supplier_id)
         except ValueError:
-            abort(400, "Invalid supplier_id: %s" % supplier_id)
+            abort(400, "Invalid supplier_id: {}".format(supplier_id))
 
         supplier = Supplier.query.filter(Supplier.supplier_id == supplier_id).all()
         if not supplier:
-            abort(404, "supplier_id '%d' not found" % supplier_id)
-
+            abort(404, "supplier_id '{}' not found".format(supplier_id))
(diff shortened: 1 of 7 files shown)


app/main/views/users.py
```diff
@@ -111,7 +148,7 @@ def create_user():
         db.session.rollback()
         abort(400, "Invalid supplier id")
 
-    return jsonify(users=user.serialize()), 200
+    return jsonify(users=user.serialize()), 201
```
quis: Why `201`? Nothing has been [created](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#201) here has it?

pcraig3: Yeah, man.

quis: My bad.

image

allait:
`/users/email-address` sounds like a REST endpoint to GET user email addresses.
I think we could merge them together (the way filters for services work right now) and either return a single user if the email argument was set, or always return a list and get the first item for email in `dmutils`.

(I'm not sure where the email address lookup is used right now, but I think it should only be used by the reset password flow.)

pcraig3:
Martyn and I talked about that as well, and -- you're right -- the email address lookup is only actually used in the supplier app, [and even then it goes through the Api Client](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/apiclient/data.py#L116).

The `get_user` method in dmutils is making the assumption (quite rightly) that a user can be gotten by id or by email because those are primary keys.

Would it be fine to just extend the `list_users` method but then finagle dmutils so that it a list of users can only be returned if a `supplier_id` is provided?

allait:
Yeah,  sounds good, and should be easy to do in dmutils by making the `supplier_id` argument required.

allait:
Since this returns a list of users for `email_address` now I think we need to update utils to support both single item and a list response and roll the new version of supplier app before this.

PR #172, opened 2015-07-09
Add copy g-cloud-7 draft API endpoint

allait:
[#97536644](https://www.pivotaltracker.com/story/show/97536644)

POST /draft-services/<draft_id>/copy creates a copy of the draft
service. List of copied fields is the same as when creating a draft
from existing service, except the service_id is not copied from the
original draft.

Required for the "Make a copy" supplier-frontend button. Instead of
an additional API endpoint it'd be possible to add a new apiclient
method that would GET the original draft from the API and then POST
it to the create draft URL. But in order to do this apiclient would
have to filter all non-data fields from the draft: service id, links,
timestamps etc. API endpoint avoids this by only copying the relevant
model fields from one object to another.

**Note:** similar to the "make a draft from service" endpoint there's no validation that the supplier that triggered the request is the owner of the original service or draft.

I've left it out for now. It's better to do this the same way for all similar drafts endpoints so it should be a separate change/story.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+sudo: false
 python:
   - "2.7"
   - "3.4"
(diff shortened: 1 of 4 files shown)


allait:
### Order drafts by id instead of serviceName
id-based order doesn't change if service names are changed.
New drafts always appear at the end of the list.

robyoung:
:+1: lgtm

PR #174, opened 2015-07-14
Make Apache logs more specific

robyoung:
This makes the Apache log line format more specific by adding the type
(apache), application (api) and the DM-Request-ID header value. It also
updates the metric filters to only match apache log lines for this
application.

This was motivated by the realisation that metric filters are applied to a
whole log group rather than individual log streams.

### DIFF
--- a/.ebextensions/cwl-logs-apache-access.config
+++ b/.ebextensions/cwl-logs-apache-access.config
@@ -33,10 +33,10 @@ Mappings:
       LogGroupName: {"Fn::GetOptionSetting": {"OptionName": "LogGroupName"}}
       ApplicationName: {"Fn::GetOptionSetting": {"OptionName": "ApplicationName"}}
     FilterPatterns:
-      Http4xxMetricFilter: "[..., status=4*, size, referer, agent]"  
-      HttpNon4xxMetricFilter: "[..., status!=4*, size, referer, agent]"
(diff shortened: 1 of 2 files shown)


.ebextensions/cwl-logs-apache-access.config
```diff
@@ -33,10 +33,10 @@ Mappings:
       LogGroupName: {"Fn::GetOptionSetting": {"OptionName": "LogGroupName"}}
       ApplicationName: {"Fn::GetOptionSetting": {"OptionName": "ApplicationName"}}
     FilterPatterns:
-      Http4xxMetricFilter: "[..., status=4*, size, referer, agent]"  
-      HttpNon4xxMetricFilter: "[..., status!=4*, size, referer, agent]"
-      Http5xxMetricFilter: "[..., status=5*, size, referer, agent]"  
-      HttpNon5xxMetricFilter: "[..., status!=5*, size, referer, agent]"  
+      Http4xxMetricFilter: "[type=apache, app=api, ..., status=4*, size, referer, agent]"
```
allait: Should the `type=apache` match the `apache-access` from LogFormat, or is it referencing something else?

robyoung: Good call, thanks. I changed that after writing the metric filters.

PR #181, opened 2015-07-17
Set Postgres client encoding to UTF8 in config.py.

RichardKnop:
We found out there was an issue with import scripts
not working properly, we had to set client encoding
in the Postgres connection string in the config file.

### DIFF
--- a/config.py
+++ b/config.py
@@ -27,7 +27,7 @@ class Config:
     DM_API_SUPPLIERS_PAGE_SIZE = 100
     SQLALCHEMY_COMMIT_ON_TEARDOWN = False
     SQLALCHEMY_RECORD_QUERIES = True
-    SQLALCHEMY_DATABASE_URI = 'postgresql://localhost/digitalmarketplace'
+    SQLALCHEMY_DATABASE_URI = 'postgresql://localhost/digitalmarketplace?client_encoding=utf8'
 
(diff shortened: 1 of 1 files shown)


allait:
Looking through the postgres docs `client_encoding` will default to the database encoding unless set in the `postgresql.conf`. Database encoding is set by `initdb` initially and is dervied from locale. So either setting any UTF8 locale when creating the database cluster or specifying the encoding on either `createdb` or `initdb` works.

I don't think this should be set in the app itself, since it's a database configuration and the app should use whatever the database prefers, as it does right now.

RichardKnop:
@allait Is that true if you have a remote database such as RDS?

Our RDS has all settings (locale etc) set to UTF-8 and yet we were getting Unicode errors when running the import.

RichardKnop:
```
ubuntu@ip-10-128-11-136:~$ psql -U postgres -l
                                                     List of databases
                Name                 |      Owner       | Encoding |   Collate   |    Ctype    |     Access privileges      
-------------------------------------+------------------+----------+-------------+-------------+----------------------------
 db_590140734167fd33e6a9             | db_5901407_group | UTF8     | en_US.UTF-8 | en_US.UTF-8 | 
 digitalmarketplace_api_dbc6b6960938 | digitalmar_group | UTF8     | en_US.UTF-8 | en_US.UTF-8 | 
```

PR #185, opened 2015-07-22
Add json logs

robyoung:
## Add JSON formatted apache logs

You can create much richer CloudWatch metric filters against JSON logs.

## Use JSON log for 400 - 403 metric filter

The JSON log allows us to group conditions together.


allait:
:+1:

PR #187, opened 2015-07-23
Fix JSON logs configuration

robyoung:
For some reason CloudFormation does not like the call to Fn::Join in the mappings section at the top. It throws an error saying that all mapping values must be strings or lists. Which is strange because the result of the Fn::Join is a string.

This change just does the Fn::Join where it's needed rather than at the top. The duplication is minimal. Ultimately it may be useful to do this in dm-aws.


allait:
I think we can update this to use https://github.com/alphagov/digitalmarketplace-aws/pull/124 right away

allait:
:+1:

PR #188, opened 2015-07-23
Small bugfix in user import script.

pcraig3:
In a past commit (8903c03e4b2e47a5776f6ff370ffa9a3a81959ab), I changed the status code when creating a user from 201 from 200.
Unfortunately, I never updated the import user script.
...until now.


quis:
:+1:

PR #189, opened 2015-07-23
Update JSON schemas and PUT route with `updatedAt`

pcraig3:
Migrating data from the preview API through my local API was failing because the schemas didn't accept the `updatedAt` key.
This change adds `updatedAt` to the schemas as well as includes some logic to preserve updated times when they are included.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -179,11 +179,8 @@ def import_service(service_id):
     service.framework_id = framework.id
     service.status = service_data.pop('status', 'published')
     now = datetime.utcnow()
-    if 'createdAt' in service_data:
-        service.created_at = service_data['createdAt']
-    else:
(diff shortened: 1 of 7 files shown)


app/main/views/services.py
```diff
@@ -179,11 +179,8 @@ def import_service(service_id):
     service.framework_id = framework.id
     service.status = service_data.pop('status', 'published')
     now = datetime.utcnow()
-    if 'createdAt' in service_data:
-        service.created_at = service_data['createdAt']
-    else:
-        service.created_at = now
-    service.updated_at = now
+    service.created_at = service_data.get('createdAt', now)
+    service.updated_at = service_data.get('updatedAt', now)
```
robyoung: Nice

robyoung:
:+1:

PR #190, opened 2015-07-24
Unlock and deactivate users

minglis:
3 changes to the API

1) Now the authenticate method respects whether the user is active and won't allow logins for inactive users.

2) We can now update the locked status of users. Send { "locked": false } as update JSON and the failed login count is reset

3) The update users endpoint now returns the User, not a success message.

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -22,7 +22,7 @@ def auth_user():
 
     if user is None:
         return jsonify(authorization=False), 404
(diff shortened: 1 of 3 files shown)


pcraig3:
**Couple more (small) changes to the API**

1. Added optional filter to `list_users` method whereby we can choose to return
either active or inactive users.
2. Also threw in a default ordering for returning users because tests weren't
consistently passing.

allait:
:+1:

PR #191, opened 2015-07-28
Remove `active` filter when listing users.

pcraig3:
Decided it would be better to handle filtering active users from inactive users in the front-end apps.
Removing `active` filter from the `list_user` route and associated tests.


allait:
I'm not sure about this: it seems like it's the same amount of code to do this in the api or the frontend apps, but the API should be easier to test and can be used by multiple apps if necessary.

robyoung:
@minglis @pcraig3 and myself had a discussion. Short version; we don't actually need this very much, let's get rid of it and pull it up when we do.

PR #192, opened 2015-07-29
Changed prefix for *any* servcies by supplier query

minglis:
- This used to require '123' to perform the any non-alphanumeric supplier name search
- Changed to 'other' to be more clear
- Matched changes in buyer app
- This needs to go first


quis:
:+1:

PR #193, opened 2015-07-31
Minor bug fixes and 'not_a_number' validation

robyoung:
## Correctly handle invalid JSON

Before this change invalid JSON would cause Werkzeug to raise an error which would not be formatted in our standard way. The body of this error would not be JSON.

## Fix user validation and improve error messaging

The role is required as it's used on the very next line. This change also adds in detailed error messaging like for service.

## Fix UTC bug in audit tests

This bug only manifests itself between 12 and 1am BST. Today BST is the next day according to UTC.

## Add validation message for 'not_a_number'

This makes it easier to parse errors where a field is not a number.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -1,6 +1,7 @@
 from flask import url_for as base_url_for
 from flask import abort, request
 from six import string_types
+from werkzeug.exceptions import BadRequest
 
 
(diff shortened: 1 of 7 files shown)


allait:
:+1:

PR #194, opened 2015-07-31
Added script to update services

minglis:
Added a cheeky script to update services

To be used when doing security updates.

### DIFF
--- a/scripts/update_service.py
+++ b/scripts/update_service.py
@@ -16,10 +16,10 @@ from docopt import docopt
 
 
 def update(base_url, access_token, filename, service_id):
-    print (base_url)
-    print (access_token)
-    print (filename)
-    print (service_id)
+    print(base_url)
+    print(access_token)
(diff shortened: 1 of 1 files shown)


scripts/update_service.py
```diff
@@ -0,0 +1,60 @@
...
+    print(filename)
+    print(service_id)
+
+    endpoint = "{}/services/{}".format(base_url, service_id)
+
+    with open(filename) as data_file:
+        try:
+            json_from_file = json.load(data_file)
+        except ValueError:
+            print("Skipping {}: not a valid JSON file".format(filename))
```
robyoung: I don't think we want to catch the error and continue here.

TheDoubleK:
"cheeky"... I just hope to goodness that anyone using this script will apply on staging first and diligently check the service listing pages for the results of each change before running against prod.

robyoung:
:+1:

PR #195, opened 2015-08-03
Add pricing validation

robyoung:
Add a global validation rule for max price being more than min price.

JSONSchema does not support validations across more than one field. This introduces the idea of a global validation. It is applied to all service submissions but only fires if both fields are present.

This change also includes a bit of refactoring.

Refactor detect_framework; I keep finding the multi-line condition hard to read. I'm not sure if this makes it better but I find it easier to read.

Use slug for G-Cloud 4 & 5 validator names.


tests/test_validation.py
```diff
@@ -348,6 +348,37 @@ def test_string_too_long_causes_validation_error():
     assert "under_character_limit" in errs['serviceName']
 
 
+def test_price_not_money_format_validation_error():
+    cases = [
+        "",  # not provided
+        "foo",  # not numeric
+        "12.1",  # too few decimal places
+        "12.001",  # too many decimal places
```
quis: Lots of decimal places is OK if your price is less than £1, eg £0.005 per gigabyte is a reasonable figure for some cloud storage.

Refer to: https://github.com/alphagov/supplier-submission-portal/commit/6a8c5003a434cccfdc5dfa4937d6dfefae3ebc10

json_schemas/services-g-cloud-7-scs.json
```diff
@@ -81,14 +81,12 @@
...
+      "type":"string",
+      "pattern": "^\\d+(?:\\.\\d{2})?$"
     },
     "priceMax":{
-      "type":[
-        "number",
-        "null"
-      ]
+      "type": "string",
+      "pattern": "^\\d+(?:\\.\\d{2})?$"
```
TheDoubleK: I think prices were allowed up to four decimal places for G6.

Are we sure having numbers for G6 but strings for G7 will not cause any problems?

TheDoubleK:
I think we should be able to get rid of `detect_framework` altogether some time very soon, given that we shouldn't ever be doing any imports of G4, G5 or G6 services ever again (or any service imports at all). The method was scaffolding to get the migration done that is looking more and more horrible as we add more schemas.  Now we've migrated we should ditch it altogether.

And if we do need import functionality we should POST to a specific `Framework` (now that we have them in the model), so there is no need to detect it any more.

app/validation.py
```diff
@@ -263,9 +273,10 @@ def _translate_json_schema_error(message):
             # A string that is too long
             return 'under_character_limit'
     if 'does not match' in message:
-        return 'under_{}_words'.format(_get_word_count(message))
-    if "is not of type 'number'" in message:
-        return 'not_a_number'
+        if key in ['priceMin', 'priceMax']:
+            return 'not_money_format'
```
TheDoubleK: The prices are not the only numbers that get validated - there is also the 'serviceAvailability` percentage, which (as far as I know) we are keeping as a number.  So we should keep the `not_a_number` branch in here I think, as well as the new string money-format check for prices.

robyoung:
Regarding the `detect_framework` stuff. Yes, I totally agree, now that we've done the initial migration and we're comfortable that the new system is stable we should do a sweep of the whole API removing endpoint and code that's no longer needed.

allait:
:+1:

quis:
:+1: To the fix for the [thing I commented on](https://github.com/alphagov/digitalmarketplace-api/pull/195#discussion_r36180861).

TheDoubleK:
No I wasn't suggesting it should be part of this PR, sorry - it was just a comment as you mentioned it (and refactored it) here, and I have been thinking for a while we should get rid of it.

Gets a :+1:  from me once we are merging things again.

PR #196, opened 2015-08-03
Add framework slug to service JSON

allait:
## Add framework slug to service JSON
Framework slug is used in new document upload paths as a document
directory.

## Don't save updated/createdAt to service JSON on update
Removes updatedAt and createdAt fields from the update
JSON so that they're not saved as part of service data.

### DIFF
--- a/app/service_utils.py
+++ b/app/service_utils.py
@@ -53,7 +53,7 @@ def validate_service(service):
 
     data = drop_foreign_fields(
         data,
-        ['service_id', 'supplierName', 'links', 'frameworkName', 'updatedAt'])
+        ['service_id', 'supplierName', 'links', 'frameworkSlug', 'frameworkName', 'updatedAt'])
     errs = get_validation_errors(validator_name, data, enforce_required=True)
     if errs:
         abort(400, errs)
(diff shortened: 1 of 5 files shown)


app/service_utils.py
```diff
@@ -53,7 +53,7 @@ def validate_service(service):
 
     data = drop_foreign_fields(
         data,
-        ['service_id', 'supplierName', 'links', 'frameworkName', 'updatedAt'])
+        ['service_id', 'supplierName', 'links', 'frameworkSlug', 'frameworkName', 'updatedAt'])
```
robyoung: Should `createdAt` be in this list as well?

allait: `updatedAt` and `createdAt` are both in the schema and are removed from stored service data by the `.update_from_json`, so it's more likely that there's no need to drop `updatedAt` here. But it's hard to trace exactly what's going on, instead I've changed the tests to check that `updatedAt` and `createdAt` are not saved to the JSON data and fixed the code so the tests pass. I think we need to rethink how we deal with fields later.

robyoung:
:+1:

PR #197, opened 2015-08-03
Only allow generating service ids for G-Cloud 7

allait:
The 7 followed by 15 digits service ID pattern should be used only
for G-Cloud 7 for now.


robyoung:
:+1:

PR #198, opened 2015-08-03
Accept page_questions top-level key in draft update requests

allait:
`page_questions` should not be a part of the service data. This is
the API part of the change that also requres an update to the utils
and supplier frontend that send the `page_questions` to the API.


robyoung:
:+1:

PR #199, opened 2015-08-04
[#100216314] Make delete draft take user ID as a parameter

TheDoubleK:
HTTP DELETEs can't take a JSON payload, so the previous version of the DELETE endpoint wouldn't work.

Fortunately nothing was using it.

This changes the endpoint to take `user_id` as a parameter to allow auditing of who deleted the draft.

There is a corresponding pull-request to update the API client here: https://github.com/alphagov/digitalmarketplace-utils/pull/119

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -9,7 +9,7 @@ from .. import main
 from ... import db
 from ...utils import drop_foreign_fields, json_has_required_keys
 from ...validation import is_valid_service_id_or_400
-from ...models import Service, DraftService, Supplier, AuditEvent, Framework
+from ...models import AuditEvent, DraftService, Framework, Service, Supplier, User
 from ...service_utils import validate_and_return_updater_request, \
     update_and_validate_service, index_service, \
     commit_and_archive_service, create_service_from_draft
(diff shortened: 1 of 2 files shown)


allait:
I'm not sure about "can't take a JSON payload". They're not strictly by the spec, the same way GET request bodies are not mentioned in the spec, but the requests themselves work. It could only be a problem if the requests are going through external proxies, which is not the case for the API.

`update_details` is request metadata, it's not really part of the URL. I think the correct solution would be to send it as additional headers, but delete body didn't seem to cause any issues.

TheDoubleK:
The issue I had that prompted this change was trying to call the delete endpoint from the functional tests using the Ruby `rest_client`, which does not allow any payload to be sent with a delete request. I read around a bit and it seems like many clients do not allow any entity/body/payload to be sent with a delete.

allait:
@TheDoubleK the `rest_client` readme mentions that it's only the `.delete` helper that doesn't accept the body, this:
```
RestClient::Request.execute(method: :delete, url: 'http://example.com/resource',
                            payload: 'foo', headers: {myheader: 'bar'})
```
should still work

PR #201, opened 2015-08-05
Added companies house id to supplier model

minglis:
Adds companies house ID to the model

- can be null
- no required
- not added to any code at this point, to be done in subsequent pull requests
- need to add this as a precursor to other works


robyoung:
:+1:

PR #203, opened 2015-08-06
Create brand new supplier

minglis:
Allows the creation of a brand new supplier

- POST /suppliers
- Does not take any IDs
- Uses a new sequence to populate the supplier id
- Keyed higher than existing suppliers

This PULL request DOES NOT deal with the resultant code duplication. This is tech debt and we need to deal with the PUT/Import endpoints.


app/main/views/suppliers.py
```diff
@@ -147,9 +148,74 @@ def import_supplier(supplier_id):
...
+
+        contact_information.contact_name = contact_information_data.get('contactName')
+        contact_information.phone_number = contact_information_data.get('phoneNumber')
+        contact_information.email = contact_information_data.get('email')
+        contact_information.website = contact_information_data.get('website')
+        contact_information.address1 = contact_information_data.get('address1')
+        contact_information.address2 = contact_information_data.get('address2')
+        contact_information.city = contact_information_data.get('city')
+        contact_information.country = contact_information_data.get('country')
+        contact_information.postcode = contact_information_data.get('postcode')
```
robyoung: Could this be moved to a `from_json` style method on `ContactInformation`?

app/main/views/suppliers.py
```diff
@@ -147,9 +148,74 @@ def import_supplier(supplier_id):
...
+    supplier_data = supplier_data['suppliers']
+    supplier_data = drop_foreign_fields(supplier_data, ['links'])
+
+    json_has_required_keys(
+        supplier_data,
+        ['name', 'contactInformation']
+    )
+
+    contact_informations_data = [
+        drop_foreign_fields(contact_data, ['links'])
```
robyoung: Is links valid in an embedded `ContactInformation` object?

app/main/views/suppliers.py
```diff
@@ -147,9 +148,74 @@ def import_supplier(supplier_id):
...
+
+    supplier_data = drop_foreign_fields(
+        supplier_data,
+        ['contactInformation']
+    )
+
+    supplier = Supplier()
+    supplier.update_from_json(supplier_data)
+
+    for contact_information_data in contact_informations_data:
```
allait: Can we split this to a separate "create contact information" endpoint? This is how updates work right now, so it should make it easier to reuse the frontend code/forms.

minglis: This has quite a different page flow to the edit pages, no chance of reuse.

Was also thinking that making a supplier in the first instance should require all of the details, which includes min contact details. One API call to make one valid supplier.

Updates though are not the same thing.

Could be done as two, depends if we think that suppliers should or must have a contact.

app/main/views/suppliers.py
minglis: Done

app/main/views/suppliers.py
minglis: Not sure. This is as was.

allait:
:+1:

PR #205, opened 2015-08-10
Renamed the companies house id migrations file

minglis:
- file created as companies_house_id and shipped
- renamed on a different branch leading to issues
- renamed back here and a new file added rename column


allait:
:+1:

PR #206, opened 2015-08-12
DUNS Number DB Updates

minglis:
- Duns numbers do not have a unique index as we have some legacy nulls in the database

- But we don't want to have any duplicates

- POST to create a supplier now checks the duns_number is not in the DB
- DUN number is indexed
- New duns_number request param on list suppliers - so we can check up front in create user and fail fast.


app/main/views/suppliers.py
```diff
@@ -185,6 +195,12 @@ def create_supplier():
         ['contactInformation']
     )
 
+    # duns number does not have a unique constraint due to nulls
```
allait: I'm not sure if we need to do this, I think NULLs are exempt for the unique constraints. From the postgres docs: "even in the presence of a unique constraint it is possible to store duplicate rows that contain a null value in at least one of the constrained columns." http://www.postgresql.org/docs/9.5/interactive/ddl-constraints.html

minglis: First DB I've used that allows this. Nice (I think)

- Added the unqiue constraint and update the POST method to not load up to check.

Kept the other stuff to allow fail fast on the create supplier flow. We can chat to Cath tomorrow to see if she wants that or whether failing on create would be OK.

robyoung:
:+1:

PR #207, opened 2015-08-12
Complete draft service

allait:
### Add /complete draft endpoint
Validates the full draft JSON (skipping the service ID error, since
drafts don't have an ID) and sets the status to 'submitted'.

### Remove priceString from G7 JSON schemas
"priceString" was replaced by separate max/minPrice fields, so it
shouldn't be a part of the G7 schema.

It's not removed from G6 to keep existing services valid.

### Order draft services by ID only
Ordering services by lot means that copied and new services appear
in the middle of the list. Instead, we order services by `id` and
let the frontend to reverse the order or apply a different sort
method.

### Return last audit event related to a draft in GET draft response
Finds the last create/update/complete event related to the given
draft and returnes it together with the draft service data.

Audit event data is used for "Last edited" block on the draft page.

### Append ' copy' to the copied draft name
Draft copies are named "<original name> copy" to avoid accidental
duplicate names in submitted services.

Since the name change doesn't go through service validation it's
possible that the new draft name is longer than the serviceName
limit of 100 characters and won't pass validation if the draft
is completed.

To avoid this there's an additional check that adds " copy" only
if the new name will stay below the 100 character limit. Since the
limit can only be increased without breaking existing services this
check should be enough to make sure that the new service name passes
the validation length rule in the future.

### Always set copied draft status to 'not-submitted'
Since there's no way to change draft status from 'submitted' to
'not-submitted' new drafts should always start in the 'not-submitted'
state, even if copied from the 'submitted' draft.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -15,7 +15,7 @@ from ...service_utils import validate_and_return_updater_request, \
     commit_and_archive_service, create_service_from_draft
 from ...draft_utils import (
     validate_and_return_draft_request, get_draft_validation_errors,
-    get_request_page_questions
+    get_request_page_questions, validate_draft
 )
 
 
@@ -82,7 +82,6 @@ def edit_draft_service(draft_id):
 
     draft.update_from_json(update_json)
     errs = get_draft_validation_errors(draft.data,
-                                       draft.data['lot'],
                                        framework_id=draft.framework_id,
                                        required=page_questions)
     if errs:
@@ -264,8 +263,7 @@ def create_new_draft_service(framework_slug):
         abort(400, "'{}' is not open for submissions".format(framework_slug))
 
     supplier_id = draft_json['supplierId']
-    lot = draft_json['lot']
-    errs = get_draft_validation_errors(draft_json, lot, slug=framework_slug)
+    errs = get_draft_validation_errors(draft_json, slug=framework_slug)
     if errs:
         return jsonify(errors=errs), 400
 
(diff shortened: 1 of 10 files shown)


app/main/views/drafts.py
```diff
@@ -298,6 +306,41 @@ def create_new_draft_service(framework_slug):
...
+    draft.status = 'submitted'
+    try:
+        db.session.add(draft)
+        db.session.flush()
+
+        audit = AuditEvent(
+            audit_type=AuditTypes.complete_draft_service,
+            user=updater_json['updated_by'],
+            data={
+                "draftId": draft.id
```
robyoung: Why do we need `draftId` in the payload as well as being linked with `db_object`?

app/models.py
```diff
@@ -464,11 +464,15 @@ def from_service(service):
         )
 
     def copy(self):
+        data = self.data.copy()
+        name = data.get('serviceName', '')
+        if len(name) <= 95:
+            data['serviceName'] = u"{} copy".format(name)
```
robyoung: I think 'Copy of "other service name"' would be clearer.

tests/app/views/test_drafts.py
```diff
@@ -887,3 +910,93 @@ def test_should_not_create_draft_with_invalid_data(self):
...
+        draft = json.loads(draft.get_data())['services']
+
+        res = self.client.post(
+            '/draft-services/%s/complete' % draft['id'],
+            data=json.dumps({'update_details': {'updated_by': 'joeblogs'}}),
+            content_type='application/json')
+
+        assert_equal(res.status_code, 400)
+        errors = json.loads(res.get_data())['error']
+        assert_in('serviceSummary', errors)
```
robyoung: Should we have a test that confirms it fails if no updater details are provided?

app/models.py
allait: That was my first idea too, but multiple "Copy of Copy of" look (somewhat) worse than " copy copy", and the appended text is what both Windows and OS X do.

app/main/views/drafts.py
allait: I'm not really sure, I think this started when we thought that `draftId` might be `service_id`, but I kept it here because all other draft events write `draftId`.

tests/app/views/test_drafts.py
allait: Added tests

robyoung:
:+1:

PR #208, opened 2015-08-12
Add audit events

robyoung:
## Retrieve audit events by object reference
Allow audit events to be retrieved by object reference identified by the object type and the object ID. The object type is as it appears in the URL and the JSON payloads as this is on the external interface. If one
of object-type or object-id is provided both must be provided.

## Add audit events via the API
Allow audit events to be added via the API. This is so that the frontends can create audit events explicitly.

This is required for capturing the suppliers registering interest in a framework ([#100934356](https://www.pivotaltracker.com/story/show/100934356)).

## Depends on

- [ ] https://github.com/alphagov/digitalmarketplace-utils/pull/134

### DIFF
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -6,11 +6,22 @@ from sqlalchemy.exc import IntegrityError
 from sqlalchemy.sql.expression import true, false
 from ...utils import pagination_links, get_valid_page_or_1
 from .. import main
-from ... import db
+from ... import db, models
 from dmutils.audit import AuditTypes
 from dmutils.config import convert_to_boolean
 from ...validation import is_valid_date, is_valid_acknowledged_state
 from ...service_utils import validate_and_return_updater_request
+
+
+AUDIT_OBJECT_TYPES = {
+    "suppliers": models.Supplier,
+    "services": models.Service,
+}
+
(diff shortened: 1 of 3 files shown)


app/main/views/audits.py
```diff
@@ -67,6 +95,47 @@ def list_audits():
...
+            abort(400, "invalid object type supplied")
+        if 'objectId' not in audit_event_data:
+            abort(400, "object type cannot be provided without an object ID")
+        model = AUDIT_OBJECT_TYPES[audit_event_data['objectType']]
+        id_field = AUDIT_OBJECT_ID_FIELDS[audit_event_data['objectType']]
+        db_objects = model.query.filter(
+            id_field == audit_event_data['objectId']
+        ).all()
+        if len(db_objects) != 1:
+            abort(400, "referenced object does not exist")
```
allait: This sounds like a 404

allait:
Other than tests and the status comment everything else looks good :+1:

app/main/views/audits.py
robyoung: We had a chat about this and it boils down to a difference of opinion on what an audit event is (is it a message attached to a database object or is it a message that may or may not have a database object attached to it?). I'm going to merge this for now and add a note for us to figure this out.

PR #209, opened 2015-08-13
PaaS does not require serviceTypes

robyoung:
The fact that this doesn't break any tests is a worry. Currently we only explicitly test SCS services as that was the first that got implemented.


quis:
Did a casual diff of the schemas for G6 and G7 PaaS services, this seems to be the only material difference.

:+1:

PR #211, opened 2015-08-17
Add an endpoint for stats about a framework

quis:
This pull request:
- gives us count for some things that we think we might want to count, based on [#100215966](https://www.pivotaltracker.com/story/show/100215966)

It doesn't:
- deal with snapshotting these counts over time, which will probably happen in a separate app

--

Sample output:

``` json
{
  "interested_suppliers": [
    {
      "count": 2,
      "has_made_declaration": false
    },
    {
      "count": 12,
      "has_made_declaration": true
    }
  ],
  "services": [
    {
      "count": 1,
      "declaration_made": true,
      "lot": "PaaS",
      "status": "not-submitted"
    },
    {
      "count": 1,
      "declaration_made": false,
      "lot": "IaaS",
      "status": "not-submitted"
    },
    {
      "count": 1,
      "declaration_made": false,
      "lot": "SCS",
      "status": "not-submitted"
    },
    {
      "count": 1,
      "declaration_made": false,
      "lot": "SaaS",
      "status": "not-submitted"
    },
    {
      "count": 1,
      "declaration_made": false,
      "lot": "PaaS",
      "status": "not-submitted"
    },
    {
      "count": 1,
      "declaration_made": true,
      "lot": "SCS",
      "status": "submitted"
    },
    {
      "count": 2,
      "declaration_made": true,
      "lot": "SCS",
      "status": "not-submitted"
    }
  ],
  "supplier_users": [
    {
      "count": 2381,
      "recent_login": null
    },
    {
      "count": 3,
      "recent_login": true
    },
    {
      "count": 2,
      "recent_login": false
    }
  ]
}
```

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -1,6 +1,6 @@
 from flask import jsonify
-from sqlalchemy.types import String
-from sqlalchemy import func
+from sqlalchemy.types import String, Boolean
+from sqlalchemy import func, or_
 import datetime
 
 from .. import main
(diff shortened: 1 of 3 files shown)


app/main/views/frameworks.py
```diff
@@ -11,3 +12,18 @@ def list_frameworks():
     return jsonify(
         frameworks=[f.serialize() for f in frameworks]
     )
+
+
+@main.route('/frameworks/g-cloud-7/stats', methods=['GET'])
+def get_framework_stats():
+
+    seven_days_ago = datetime.datetime.now() + datetime.timedelta(-7)
```
allait: `.utcnow`, it doesn't make much of a difference in this case, but makes it easier to check that we're using UTC everywhere

app/main/views/frameworks.py
```diff
@@ -11,3 +12,18 @@ def list_frameworks():
...
+
+    return str({
+        'drafts': DraftService.query.filter(DraftService.status == "not-submitted").count(),
```
allait: Drafts should probably be filtered by framework. We only have G7 drafts right now, but this might change if service edits go live before the end of SSP.

app/main/views/frameworks.py
```diff
@@ -11,3 +12,18 @@ def list_frameworks():
...
+        'complete': DraftService.query.filter(DraftService.status == "submitted").count(),
```
robyoung: These could be combined into a group by query (see [the docs](http://docs.sqlalchemy.org/en/rel_1_0/orm/tutorial.html#counting))
```python
db.session.query(DraftService.status, func.count(DraftService.status)).group_by(DraftService.status)
```
DraftService.query.

quis:
_via @cathrooney:_

We also need to count:
- eligible suppliers (who have at least one complete service and have made the declaration)
- eligible services (which are complete and whose supplier has made the declaration)

app/main/views/frameworks.py
quis: Addressed in 04f8e26

app/main/views/frameworks.py
quis: Addressed in d9abb2f

app/main/views/frameworks.py
quis: Made it so the framework is not hard-coded: 9eb73b3

app/main/views/frameworks.py
```diff
@@ -11,3 +14,42 @@ def list_frameworks():
...
+            DraftService.framework_id == framework.id
+        )),
+        'services_by_lot': dict(db.session.query(
+            lot_column, func.count(lot_column)
+        ).group_by(
+            lot_column
+        ).filter(
+            DraftService.framework_id == framework.id
+        ).all()),
+        'users': User.query.count(),
```
allait: Should we count only accounts with the supplier role here?

app/main/views/frameworks.py
```diff
@@ -11,3 +14,42 @@ def list_frameworks():
...
+    lot_column = DraftService.data['lot'].cast(String).label('lot')
+
+    return str({
+        'services_by_status': dict(db.session.query(
+            DraftService.status, func.count(DraftService.status)
+        ).group_by(
+            DraftService.status
+        ).filter(
+            DraftService.framework_id == framework.id
+        )),
```
allait: It's possible to get service counts for each lot and status combination with a single query and then sum it in the frontend if we want. Since lots have different question counts I think "complete" split will be different from total counts.

app/main/views/frameworks.py
quis: Sounds sensible for the future—bc0cf65

app/main/views/frameworks.py
quis: @allait and I addressed this in ac98809

allait:
#### Count suppliers with/without complete drafts in framework stats

Splits supplier count based on declaration and complete drafts
status.

Uses a subquery to get at most 1 complete draft for each supplier.

robyoung:
These are some pretty complicated queries. I'm not sure I can be confident they're doing what we think they're doing without some tests.

allait:
@robyoung added tests for the stats response

quis:
Intention of the subquery thing and the tests look good to me. If @robyoung is happy then this should be merged.

robyoung:
:+1:

allait:
This will need an update once https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/206 is merged to check the "declaration is complete" flag

quis:
Going to merge now while I'm still young and update it based on @TheDoubleK's work later.

PR #212, opened 2015-08-18
Re-add unique constraint for DUNS number

TheDoubleK:
The duplicate DUNS numbers in preview, staging and prod databases have now been fixed so there are no existing duplicates, so it will now be OK to apply this database upgrade.


allait:
:+1:

PR #213, opened 2015-08-19
Allow audit event with no user

robyoung:
Audit events are allowed to be created with no associated user. This
makes the API create audit event endpoint also allow this. There is a
requirement in the frontend to create an audit event for sending an
invite email on creation of a supplier, before the user has logged in.


allait:
:+1:

PR #215, opened 2015-08-20
admin-ccs user role

allait:
#### Check that user supplier_id is set (only) with the supplier role
Since the previous commit replaced check constraint with an enum,
there's no database check that supplier_id is set for users with
'supplier' role.

Instead, we add a check on user create/update that supplier_id is
always set only for 'supplier' users.

#### Add admin-ccs user role
[#101410132](https://www.pivotaltracker.com/story/show/101410132)

New role will be used for read-only access to the MI page on the
admin app.

#### Add a script to change user password
Takes user email, environment name, API token and new user password
and sends a user password update request. Can optionally unlock users
after changing the password.

Password is read using getpass, so there's no input echoing.


scripts/chpass.py
```diff
@@ -0,0 +1,66 @@
...
+    data_client = apiclient.DataAPIClient(api_endpoint, api_token)
+
+    try:
+        user = data_client.get_user(email_address=user_email)['users']
+        data_client.update_user_password(user['id'], new_password)
+        if unlock:
+            data_client.update_user(user['id'], locked=False)
+            logger.info("User unlocked")
+    except apiclient.APIError:
+        sys.exit(1)
```
robyoung: Would be good to give some indication of what's gone wrong here.

robyoung:
:+1:

scripts/chpass.py
allait: Logging is configured for the script run, so it prints `apiclient` warnings with status codes and error messages

robyoung: Boom! :+1:

PR #217, opened 2015-08-20
Make 'create_supplier' call match new name

tombye:
This method was renamed to be 'import_supplier' when the new create supplier route was added.

See https://github.com/alphagov/digitalmarketplace-utils/commit/e0e19d29e74d6d7ff3f71d8679414a2b4484cd69 for details.


TheDoubleK:
Seen in action :+1:

PR #218, opened 2015-08-21
Add decorator to set PostgreSQL isolation level

robyoung:
## Do not merge!

This is just for discussion.

Add a Flask decorator to allow the PostgreSQL transaction isolation level to be set per request.

On some update requests we can run into concurrency issues because we read a row with a big blob of JSON from the database make some changes and then write it back. If another process was also doing this, even if it was in another part of the JSON there would be a race condition and the last writer would win. PostgreSQL has a transaction isolation level [1] called SERIALIZABLE that will cause a failure if any of the rows that a transaction trys to work with are changed while it is running. We don't want to set this isolation level on all transactions because it's more costly and often we don't care. This change allows us to set the isolation level on a single request transaction.

[1] http://www.postgresql.org/docs/9.1/static/transaction-iso.html

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -47,7 +47,8 @@ def isolation_level(level):
     """
     def decorator(view):
         def view_wrapper(*args, **kwargs):
-            db.session.connection(execution_options={'isolation_level': level})
+            if flask_featureflags.is_active('TRANSACTION_ISOLATION'):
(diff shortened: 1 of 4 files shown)


allait:
Let's merge this. There's not much we can do without running it in an actual environment, and with a feature flag it shouldn't cause any issues in prod :+1:

PR #219, opened 2015-08-24
[#101503356] Filter fields when copying a draft

robyoung:
https://www.pivotaltracker.com/story/show/101503356

Certain fields should not be copied over into the new draft.


allait:
:+1:

PR #220, opened 2015-08-25
[#100722784] Make apiType optional

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/100722784

The minimum length requirement on apiType meant that it was giving a `required` validation error even though `apiType` is not in the list of required fields.

 The regular expression also needed to be modified to match an empty string, or that also caused a validation error when the page was submitted with nothing in the apiType field..


robyoung:
:+1:

PR #221, opened 2015-08-25
[#100513836] Percentages should be less than 100

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/100513836

I've re-used the existing `not_a_number` validations key because it gives the correct validation message from the SSP content.

### DIFF
--- a/json_schemas/services-g-cloud-7-iaas.json
+++ b/json_schemas/services-g-cloud-7-iaas.json
@@ -444,7 +444,10 @@
       "type": "object",
       "properties": {
         "value": {
-          "type": "number"
+          "type": "number",
+          "minimum": 0,
+          "maximum": 100,
+          "exclusiveMaximum": true
         },
         "assurance": {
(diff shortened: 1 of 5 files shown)


json_schemas/services-g-cloud-7-iaas.json
```diff
@@ -444,7 +444,10 @@
       "type": "object",
       "properties": {
         "value": {
-          "type": "number"
+          "type": "number",
+          "minimum": 0,
+          "maximum": 100,
+          "exclusiveMaximum": true
```
allait: Why do we not allow 100%? Are there any existing services that would fail this validation?

TheDoubleK: 100% uptime is a lie :) 
We didn't allow 100% for G6 - so no existing services would fail to validate against this (and these are G7 schemas anyway and won't be applied to existing services).

quis:
:+1:

PR #222, opened 2015-08-26
Snapshot framework stats

allait:
#### Add "frameworks" to the list of audit object types
Framework stats snapshot events are associated with a framework
object, so this adds "frameworks" to the list of object types that
can be used with audit events API endpoints.

Framework slug is used as object_id since similar to service_id it's
used to reference the framework object everywhere outside the API.

#### Add snapshot_framework_stats.py script
Requests the framework stats for the given environment and stores
the response in a new audit event associated with the given
framework.

#### Add audit_events type, created_at and object indexes
Adds three separate indexes:

* Index on audit_events.type, since it's the most useful filter for
  listing audit events
* Index on audit_events.created_at for timestamp filters
* Multicolumn index on (object_type, object_id) for filters based on
  referenced object.

#### Take advantage of new audit_events index in framework stats
When joining suppliers with audit events there's no need to have
a object_type filter, since framework interest events are always
related to a supplier object. But adding additional object_type
filter allows the DB to take advantage of the new multicolumn
(object_type, object_id) index.

### DIFF
--- a/scripts/snapshot_framework_stats.py
+++ b/scripts/snapshot_framework_stats.py
@@ -14,6 +14,7 @@ import logging
 logger = logging.getLogger('script')
 logging.basicConfig(level=logging.INFO)
 
+import backoff
 from docopt import docopt
 
 from dmutils import apiclient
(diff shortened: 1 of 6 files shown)


scripts/snapshot_framework_stats.py
```diff
@@ -0,0 +1,62 @@
...
+    try:
+        stats = data_client.get_framework_stats(framework_slug)
+        data_client.create_audit_event(
+            AuditTypes.snapshot_framework_stats,
+            data=stats,
+            object_type='frameworks',
+            object_id=framework_slug
+        )
+    except apiclient.APIError:
+        sys.exit(1)
```
quis: What's the chance of an API call actually failing? eg if we make 1000 calls over the course of the process, what percentage can we expect to error (if any)?

Is it worth retrying x times before giving up? Would be a shame to miss a data point because of a drop in connectivity or something.

allait:
@quis that's a good point. I've added a retry decorator anyway. It should do 5 attempts with increasing delays (from 2 to 10 seconds). With hourly snapshots we should have more than enough data points even if some of the runs fail.

robyoung:
:+1:

PR #225, opened 2015-08-27
Add JSON application logging to CloudWatch logs

robyoung:
This adds JSON application logs to CloudWatch logs. See: https://github.com/alphagov/digitalmarketplace-utils/pull/151


minglis:
assuming this is supplementary and does not replace the existing log (which it looks like it is) it's a 👍 from me.

PR #226, opened 2015-08-27
Use new supplier declaration status field in framework stats

allait:
Counts suppliers by declaration status:
* `null` if declaration wasn't started (first page wasn't saved)
* `"started"` if declaration was started but not complete
* `"complete"` if declaration is complete


quis:
:+1:

PR #227, opened 2015-08-28
Fix application.py python 3 syntax error

allait:
list_routes used print statement, which wasn't picked up by tests
since they never import application.py


pcraig3:
:+1:

PR #228, opened 2015-08-28
Add `assurance_required` key to validation errors

TheDoubleK:
Required as part of this story/bugfix: https://www.pivotaltracker.com/story/show/102149718

All the different types of assurance array have "Service provider assertion" as the first option, so this should work OK.


quis:
Bit hacky but should work fine for now. :+1:

TheDoubleK:
No more hacky than the other mappings from JSON schema error messages to our error keys...

PR #229, opened 2015-09-01
Update import users to import admins

minglis:
Allow import users scripts to create admin and admin-ccs users


allait:
I think that's an old script we've used to migrate users from the old DB. Are we planning to use it for something else?

TheDoubleK:
Yes I think this script is part of the long list of "things we should kill now the migration is done", along with the import API endpoints.

robyoung:
Closing as the general consensus seems to be that this script should not be used.

PR #232, opened 2015-09-01
Add microsecond Apache request time

robyoung:
We currently only have request time from Nginx, however requests to APIs
do not go through Nginx. This change adds the request time to the Apache
access logs. The time is recorded in microseconds, unfortunately Apache
cannot log the request time in seconds with the sub-second component
like Nginx so I've opted for the microsecond option.

https://httpd.apache.org/docs/2.2/mod/mod_log_config.html#formats


allait:
:+1:

PR #233, opened 2015-09-02
Remove terminationCost from G7 SaaS

allait:
terminationCost is not present in G-Cloud 6 SaaS schema and g6
content applies the question only for SCS, PaaS and IaaS.

Listing it in the G7 SaaS schema makes it impossible to complete a
G7 SaaS service as the form doesn't show the question at all, but
full service validation in the API fails.


TheDoubleK:
:+1:

PR #234, opened 2015-09-02
Add framework interest to user email list

robyoung:
Add an option to allow the user email list to be filtered by whether they've registered interest with a framework.

### DIFF
--- a/scripts/generate-user-email-list.py
+++ b/scripts/generate-user-email-list.py
@@ -1,27 +1,61 @@
 """
 
 Usage:
-    scripts/generate-user-email-list.py <data_api_url> <data_api_token>
+    scripts/generate-user-email-list.py <data_api_url> <data_api_token> [--framework=<slug>]
 """
 import csv
 import sys
+import multiprocessing
+from multiprocessing.pool import ThreadPool
 
 from docopt import docopt
 from dmutils.apiclient import DataAPIClient
+from dmutils.audit import AuditTypes
 
 
-def generate_user_email_list(data_api_url, data_api_token):
+def is_registered_with_framework(client, framework_id):
+    def is_of_framework(audit):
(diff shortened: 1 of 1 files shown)


scripts/generate-user-email-list.py
```diff
@@ -1,27 +1,61 @@
...
 from docopt import docopt
 from dmutils.apiclient import DataAPIClient
+from dmutils.audit import AuditTypes
 
 
-def generate_user_email_list(data_api_url, data_api_token):
-    client = DataAPIClient(data_api_url, data_api_token)
+def is_registered_with_framework(client, framework_id):
+    def is_of_framework(audit):
+        return audit['data'].get('frameworkSlug') == framework_id
```
allait: Oh, good catch, I think we're missing this check in the stats. But it shouldn't affect numbers right now.

scripts/generate-user-email-list.py
```diff
@@ -1,27 +1,61 @@
...
+            audit_type=AuditTypes.register_framework_interest,
+            object_type='suppliers',
+            object_id=user['supplier']['supplierId'])
+        return any(map(is_of_framework, audits['auditEvents'])), user
+    return is_registered
 
-    writer = csv.writer(sys.stdout, delimiter=',', quotechar='"')
 
+def filter_is_registered_with_framework(client, framework_id, users):
+    pool = ThreadPool(100)
```
allait: A 100 threads might be quite a bit of load to the production API, especially since we haven't really tested the audit-events performance. Is loading all `register_framework_interest` events first and then filtering suppliers using the in-memory collection more complicated?

robyoung: Good call about loading them all up front. I forgot that you could just get them all for a given type.

PR #235, opened 2015-09-03
Fix complete counts

quis:
Problem: we were seeing 6 **suppliers** with 'complete' services who hadn't made the declaration, but only 2 **services** which were 'complete' from suppliers who hadn't made the declaration.

This was because we weren't properly differentiating between 'complete' and 'started' delcarations when counting suppliers.

This commit fixes the counts to be correct.


allait:
:+1:

PR #236, opened 2015-09-03
Switch remaining metric filters to JSON

robyoung:
Adding the request time to the end of the Apache logs changes what the metric filter pattern actually does. JSON metric filters are much more robust to this kind of change.


TheDoubleK:
:+1:

PR #237, opened 2015-09-03
Allow custom page size for audit events

quis:
In about a day's time we will have over 100 statistic snapshots saved as audit events. By the end of G-Cloud 7 we will have about 1000 (35 days * 24 snapshots per day = 840).

Because the graphs need all the available snapshots we could either:

1. paginate through the snapshots using [the iter method in utils](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/apiclient/base.py#L21-L43)
2. set the page size to 840 or greater

This commit allows for the possibility of doing 2., which should be less load on the API.


allait:
:+1:

PR #238, opened 2015-09-03
[102665190] Require both value and assurance for assurance questions

TheDoubleK:
Assurance questions haven't required both fields up until now, but they should.

This will be useful for fixing this bug: https://www.pivotaltracker.com/story/show/102562916

And also fixes this bug: https://www.pivotaltracker.com/story/show/102665190

### DIFF
--- a/app/validation.py
+++ b/app/validation.py
@@ -281,6 +281,8 @@ def _translate_json_schema_error(key, message):
         else:
             # A string that is too long
             return 'under_character_limit'
+    if "is not one of [u'Unit', u'Person'" in message:
+        return 'no_unit_specified'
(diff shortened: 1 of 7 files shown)


quis:
How can we be sure this isn't going to make existing draft services explode?

TheDoubleK:
There is currently no way to submit just an assurance or just a value, so no current drafts will have that.  That possibility is only introduced in: https://github.com/alphagov/digitalmarketplace-utils/pull/158.

Also, I have tested it locally with existing draft services and all are fine. (Anecdotal, I know!)

quis:
:+1:

PR #240, opened 2015-09-07
[#102567802] Updates to CCS admin roles

TheDoubleK:
Requirted as part of this story: https://www.pivotaltracker.com/story/show/102567802

* Create two new CCS roles - `admin-ccs-category` and `admin-ccs-sourcing`
* Deprecate the existing `admin-ccs` role

Postgres doesn't allow the removal of values from an enum, so the `admin-ccs` value is still in the list, but should not be used.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -242,7 +242,8 @@ class User(db.Model):
         'buyer',
         'supplier',
         'admin',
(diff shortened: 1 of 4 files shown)


quis:
:+1:

PR #241, opened 2015-09-08
Improve supplier list query performance

robyoung:
While investigating the Apache access logs for the API the `/suppliers` route was identified as one of the slowest routes when filtering by framework. The median response time is ~600ms and the 90th percentile
is ~700ms.

The distinct subquery joining to the services table was not being not being optimised by the query planner very well. The query needs to join into the services table to restrict the suppliers to only those that have live services in the requested framework. However, it only needs to check that each supplier has at least one published service. The query planner can optimise this but seems to confused with the multiple layers of nested queries that SQLAlchemy creates. This change flattens the query out making it easier for the optimiser to do it's stuff. The order by is extended with the supplier ID because PostgreSQL requires the distinct keys to be a prefix of the order keys.

Testing locally shows that this change improves performance by about 10x from ~500ms to ~50ms

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -29,20 +29,15 @@ def list_suppliers():
     if framework:
         is_valid_string_or_400(framework)
 
-        active_services = Service.query.join(
-            Service.framework
+        suppliers = Supplier.query.join(
+            Service.supplier, Service.framework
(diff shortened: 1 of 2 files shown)


minglis:
I think this is OK. 👍

PR #242, opened 2015-09-08
Script to add new users from a JSON file of user objects

TheDoubleK:
Helper script for when we need to add new admin users (or any other sort of user). Uses the API client and avoids having to roll-our-own CURLs.

### DIFF
--- a/scripts/add-users.py
+++ b/scripts/add-users.py
@@ -1,0 +1,37 @@
+#!/usr/bin/env python
+"""
+Add a series of users from a file of JSON objects, one per line.
+
+The JSON user object lines can have the following fields:
+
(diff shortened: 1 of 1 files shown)


quis:
Why not parse the entire file as JSON and have an array of users? eg:
```
[
  {
    "name": …
  },
  {
    "name": …
  }
]
```

TheDoubleK:
Um... because I just unthinkingly copied the way the `update-suppliers` script works?

robyoung:
I think it's ok either way. Both have their benefits. Line oriented files are convenient for command line tools sometimes because they work well with other tools like `sort`, `grep` and `awk`.
:+1:

PR #246, opened 2015-09-17
Remove required `postcode` when importing a supplier

pcraig3:
Last change to `/json_schemas/contact-information.json` removed the postcode as a required part of contact information, but this wasn't reflected in the `PUT` import method.
As a consequence, `/scripts/import_suppliers_from_api.py` was failing to import suppliers without postcodes.
Also updated the implicated test.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -13,6 +13,7 @@ from ...validation import (
 )
 from ...utils import pagination_links, drop_foreign_fields, get_json_from_request, \
     json_has_required_keys, json_has_matching_id, get_valid_page_or_1
+from ...supplier_utils import validate_and_return_supplier_request
 from dmutils.audit import AuditTypes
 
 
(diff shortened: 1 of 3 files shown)


TheDoubleK:
We have suppliers without postcodes? Were we supposed to capture postcode into the contact details from the supplier declaration as they made that?

TheDoubleK:
The change here is fine. :+1: 
I just wonder if we've missed a trick with the declarations.

pcraig3:
I think the reasoning behind no longer requiring postcodes was just to loosen up the requirements, to make it easier for suppliers to sign up.  Commit can be found [here](https://github.com/alphagov/digitalmarketplace-api/commit/f363bd398e65b49654cc82e281052ee54bba4e31) at any rate.

//

Actually, this problem came about because we have a `json_has_required_keys` method which validates JSON data as well as a schema which validates JSON data. 
At the time I was using it, I remember that the error messaging on our JSON schemas was pretty crap, so getting a message back from `json_has_required_keys` was a huge advantage.

Playing around a bit, you do get a better idea of the problem from `json_has_required_keys`, which will tell you about *all* the potentially missing keys, while the schema error message only tells you about one missing key at a time.

```javascript
// Trying to import a supplier missing 'id' and 'name' keys

// error message returned by `json_has_required_keys`
{
  "error": "Invalid JSON must have '['id', 'name', 'contactInformation']' key(s)"
}

// error message returned by the json schema
{
  "error": "JSON was not a valid format. 'id' is a required property"
}
```

Any thoughts on slowly removing `json_has_required_keys` methods so as not to bifurcate validation logic vs. keeping it in because the error messaging is better?

TheDoubleK:
There was a good reason for that method. I'm not 100% sure what it was - I think it was because updates have two parts to the payload - i.e. updater details (who did it and why - although we've dropped the 'why' now) and the update itself (e.g. service, user, whatever), with no single schema that validates both parts are there.  So `has_required_keys` checks that all separate required parts of the payload exist before validating any of them against a JSON schema.

Checking for keys that are also checked in the JSON schemas is feature-creep I think.

If you can face trying to weed that out then good luck, but I think there are some cases where the method is useful and not a duplication of stuff in the schemas.

pcraig3:
Found the original commit where [`json_has_required_keys`](https://github.com/alphagov/digitalmarketplace-api/commit/2aa1e60abe94236d284d13907fcf4b3fa46b0db4) was added, and -- as @TheDoubleK hypothesised -- it's only being used to check for the existence of keys that the schema doesn't accept. 

So I decided to remove `json_has_required_keys` in `suppliers.py` unless:

- It was checking a root object key (ie, `suppliers`)
- It was checking an key external to the schema (ie, `update_details`)

TheDoubleK:
:+1:

PR #247, opened 2015-09-18
Fix example of how to run script, and make it executable

TheDoubleK:
Single quotes aren't allowed in JSON, so if you copy & pasted the example line to edit values the script wouldn't run properly.


quis:
:+1:

PR #249, opened 2015-09-18
Sourcing exports

minglis:
2 new scripts for CCS Sourcing exports

1) generate_declaration_export_for_ccs_sourcing.py
Used to generate a CSV file containing the supplier declaration data
- Gets every supplier from the supplier API
- For each supplier looks for a declaration for G7
- Checks that the declaration is complete by looking for the first compulsary field from page 4 - this is a proxy for "completed". All pages must be valid to be saved and all pages must be complete to get to page 4, so this *should* be sufficient in lieu of an actual status.
- Parses the declaration data into human readable form for presentation in the CSV
  - Booleans -> yes | no
  - Strings -> encoded to UTF-8, presented content as is
  - Required checkbox lists -> These are valid if all boxes are ticked. So presented as "passed" or "failed" depending on how long the list of values is compared to expected length.
  - Optional checkbox lists -> More informational, data presented as values joined by comma's, encoded to UTF-8.
- Uses Digital marketplace supplier ID, duns number and name as meta data.

2) generate_lots_export_for_ccs_sourcing.py
Generates the data around G7 services for each supplier.
- Get every supplier from supplier API
- For each supplier gets all the drafts
- For each draft filter the G7 ones, which have service names into a sublist
  - note that the service name check filters out "hanging" copies. Services that have been copied but have had no action taken exist in the DB but are not visible to suppliers and therefore are removed.
- Group this by lot, then by status.
- CSV has same meta columsn as (1) above
- Columns are status per lot, so submitted | not-submitted for each lot.

### DIFF
--- a/scripts/generate_declaration_export_for_ccs_sourcing.py
+++ b/scripts/generate_declaration_export_for_ccs_sourcing.py
@@ -79,7 +79,7 @@ expected_totals = \
     }
 
 # map fields to processing method
-fields = \
+question_mapping = \
     {
         "PR1": process_boolean,
         "PR2": process_boolean,
@@ -149,8 +149,8 @@ def process_supplier_declaration(supplier_declaration, questions):
     """
     answers = []
     for question in questions:
-        if question in fields:
-            answers.append(fields[question](supplier_declaration, question))
+        if question in question_mapping:
+            answers.append(question_mapping[question](supplier_declaration, question))
     return answers
 
 # get the list of questions, in order, from the manifest
--- a/scripts/generate_declaration_export_for_ccs_sourcing.py
+++ b/scripts/generate_declaration_export_for_ccs_sourcing.py
@@ -1,7 +1,7 @@
 """
 
 Usage:
-    scripts/generate_declaration_export_for_ccs_sourcing.py <data_api_url> <data_api_token> <path_to_declaration_manifest> # noqa
+    scripts/generate_declaration_export_for_ccs_sourcing.py <data_api_url> <data_api_token> <path_to_manifest>
 
 Example:
     ./generate_declaration_export_for_ccs_sourcing.py http://api myToken /my/path/to/file/manifest.yml
--- a/scripts/generate_declaration_export_for_ccs_sourcing.py
+++ b/scripts/generate_declaration_export_for_ccs_sourcing.py
@@ -1,7 +1,7 @@
 """
 
 Usage:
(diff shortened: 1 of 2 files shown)


TheDoubleK:
To check if the declaration is complete you can check the `status` key in the saved JSON data - possible values are `started` or `complete`.

scripts/generate_declaration_export_for_ccs_sourcing.py
```diff
@@ -0,0 +1,271 @@
+"""
+
+Usage:
+    scripts/generate_declaration_export_for_ccs_sourcing.py <data_api_url> <data_api_token> <path_to_declaration_manifest> # noqa
```
TheDoubleK: Amazingly, adding `#noqa` at the end of this line prevents the script from running.

``
$ python scripts/generate_declaration_export_for_ccs_sourcing.py http://localhost:5000 myToken ../digitalmarketplace-supplier-frontend/app/content/frameworks/g-cloud-7/manifests/declaration.yml
Usage:
    scripts/generate_declaration_export_for_ccs_sourcing.py <data_api_url> <data_api_token> <path_to_declaration_manifest> # noqa
``

allait: Yeah, it's within a string literal, so the `# noqa` would be saved as part of the module docstring – which is parsed by docopt to define script options and parameters, so I think it expects the script call to end with `# noqa`. I don't think there's a way to make pep8 ignore the error in a line of a string literal.

TheDoubleK:
I'm not sure if the behaviour of the second script is exactly what CCS want (it may be - I'm just checking the requirements)? It only counts draft services that have all required fields filled in - i.e. services that could have been marked as complete but that weren't.  Do they also want to know if there were drafts that were started but not all required fields completed?

cathrooney:
I'm happy with just the drafts which could have been marked as complete.  This is informational only for CCS

minglis:
Checking @TheDoubleK comment looks like it does all drafts, not just finished but not submitted ones. Checked locally. @TheDoubleK doing a second pair of eyes check.

TheDoubleK:
Checked mine again and it definitely only seems to be doing the complete-but-not-submitted ones:
screen shot 2015-09-29 at 10 40 16

Results from the counting script:
screen shot 2015-09-29 at 10 39 59

TheDoubleK:
Actually, it could also be counting not-complete-but-not-submitted too, as there are one each of those.  Let me do a bit more digging.

PR #250, opened 2015-09-24
[#103779820] Auditing user updates properly

TheDoubleK:
Part of this bug fix: https://www.pivotaltracker.com/story/show/103779820

The update user endpoint should expect updater details and save them in audit events.

Previously the auditing of user updates only recorded the user who had been updated and not the user who made the change.  We need both to have a proper record of changes made.

Depends on:
- [x] https://github.com/alphagov/digitalmarketplace-utils/pull/166/files


quis:
:+1: when the version is ready

PR #251, opened 2015-09-29
[#103991306] Add declaration status to email list

robyoung:
See: https://www.pivotaltracker.com/story/show/103991306

This change adds an option to the generate_user_email_list script to
add the declaration status to the CSV output. This moves the logic into
a package in the main app package so that it can be more easily included
into an IPython session for manual testing / debugging. In theory this
makes it easier to write tests for this code although I haven't done for
this script. :/

### DIFF
--- a/app/scripts/generate_user_email_list.py
+++ b/app/scripts/generate_user_email_list.py
@@ -1,0 +1,96 @@
+import csv
+import sys
+import multiprocessing
+from multiprocessing.pool import ThreadPool
+
+from dmutils.apiclient import DataAPIClient
+from dmutils.apiclient.errors import HTTPError
+from dmutils.audit import AuditTypes
+
+
+def selection_status(client, framework_id):
+    def something(user):
+        try:
+            answer = client.get_selection_answers(user['supplier']['supplierId'], framework_id)
+            status = answer['selectionAnswers']['questionAnswers']['status']
+        except HTTPError as e:
+            if e.status_code == 404:
+                status = 'unstarted'
+            else:
+                status = 'error-{}'.format(e.status_code)
+        except KeyError:
+            status = 'error-key-error'
+        return (status, user)
+
+    return something
+
+
+def add_selection_status(client, framework_id, users):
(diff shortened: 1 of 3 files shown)


scripts/generate_user_email_list.py
```diff
@@ -1,67 +1,25 @@
 """
 
 Usage:
-    scripts/generate-user-email-list.py <data_api_url> <data_api_token> [--framework=<slug>]
+    scripts/generate-user-email-list.py <data_api_url> <data_api_token> [--framework=<slug>] [--status]
```
TheDoubleK: Could we be more specific about what `status` is here - users also have the `active/deactivated` & `locked` statuses, but this is specifically the status of their declaration.

An example call in the comment, or list of possible values that `status` can take would be useful for future reference I think.

app/scripts/generate_user_email_list.py
```diff
@@ -0,0 +1,96 @@
...
+import multiprocessing
+from multiprocessing.pool import ThreadPool
+
+from dmutils.apiclient import DataAPIClient
+from dmutils.apiclient.errors import HTTPError
+from dmutils.audit import AuditTypes
+
+
+def selection_status(client, framework_id):
+    def something(user):
```
TheDoubleK: `something` seems pretty vague as an identifier...

TheDoubleK: There is a method somewhere in supplier frontend that encapsulates this same logic (see https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/main/helpers/frameworks.py#L186).  Might be more hassle than it's worth trying to share it for use here though I guess.

TheDoubleK:
This looks good.  The `something` doesn't really need a name I guess as it's obvious from `selection_status` what it does.

If we want the code ever to be tested then the tests should probably go in now, because it'll never get done otherwise once it's merged.

robyoung:
Closing as implemented over here https://github.com/alphagov/digitalmarketplace-scripts

PR #254, opened 2015-09-30
Change prices to strings in G5&6 schemas

robyoung:
Change the minPrice and maxPrice fields to be strings (as they are in
G7) in the G5 and G6 schemas.

This is going to mean that G5 and G6 services read from the data API
will not be valid if sent straight back. However, this shouldn't be a
problem because the only way they can be edited is through the admin
app which sends them back as strings.

In terms of the effect this will have on other apps. The supplier and
frontend apps will be using the same price rendering code which will
allow for strings or numbers for the prices. The buyer frontend
currently uses the `priceString` field which isn't set on G7 services so
we need to address this anyway.


quis:
:+1:

allait:
I'm a bit confused by the "the only way they can be edited is through the admin app which sends them back as strings" part. If you edit an unrelated service field in the admin app then changed fields are sent to the API, combined with the original service and the resulting service data is validated – and I'd expect it to be rejected since the price fields no longer pass the validation. Am I missing something?

TheDoubleK:
Good point @allait - you're right I think (not tested though).

robyoung:
Ah, rubbish, I was thinking of the draft behaviour where only the changed section is uploaded. Closing until I have a solution.

allait:
:+1:

PR #255, opened 2015-10-01
[#104625852] Don't audit updated passwords

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/104625852

Currently the data stored with audit events involving a password update includes the un-hashed plain-text password.  We need to not do that.


robyoung:
:+1:

PR #257, opened 2015-10-02
Use response header for requestId

robyoung:
This means that we get a request ID even if one isn't passed in.

### DIFF
--- a/.ebextensions/apache_log.config
+++ b/.ebextensions/apache_log.config
@@ -4,8 +4,8 @@ files:
     owner: root
     group: root
     content: |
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Looks like magic... how does `o` give us an ID? And is that same ID then somehow propagated to the rest of the log entries for that request? (If not then I don't see that a random ID would be much use to us?) 

I don't doubt you - just curious.

robyoung:
The `o` uses the HTTP header in the response (outbound header) rather than `i` that uses the one in the request (inbound header).
https://httpd.apache.org/docs/2.2/mod/mod_log_config.html#formats

allait:
:+1:

PR #258, opened 2015-10-05
Allow the API to be run with gunicorn

robyoung:
Sometimes it is useful to run more threads for the data API. For example, when loading in large amounts of data doing it through a single thread takes a long time.

This runs it through [gunicorn](http://gunicorn.org/) with 4 workers when `gunicorn` is provided as an argument. Gunicorn should not be the default because it is not used anywhere else.


TheDoubleK:
Tried it out and it works as far as I can tell :+1:

PR #259, opened 2015-10-06
Rework logic to get declaration statuses.

pcraig3:
Pull request addresses two things.

1. Script was failing on the parameter name: it was looking for `path_to_declaration_manifest` not `path_to_manifest`
2. Script uses same logic as in the generate email script to add an extra `State of Declaration` column to the resulting .csv file.
 - Only suppliers whose declarations are `started` or `completed` will be returned

Probably better to put this into [digitalmarketplace-scripts](https://github.com/alphagov/digitalmarketplace-scripts) if we're planning to use this very often.

<del>Oh, and p.s., this breaks in Python3.</del>

### DIFF
--- a/scripts/generate_declaration_export_for_ccs_sourcing.py
+++ b/scripts/generate_declaration_export_for_ccs_sourcing.py
@@ -69,7 +69,7 @@ def process_boolean(declaration, field):
 
 
 def process_string(declaration, field):
-    return get_or_else_empty_string(declaration, field).encode("utf-8")
+    return get_or_else_empty_string(declaration, field)
 
 
 def process_checkbox_counts(declaration, field):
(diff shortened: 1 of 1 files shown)


robyoung:
> Oh, and p.s., this breaks in Python3
Then use Python3 and make it break in Python2 :trollface:

scripts/generate_declaration_export_for_ccs_sourcing.py
```diff
@@ -201,21 +201,21 @@ def headers(questions):
...
-    """
-    If declartion has a value for any of the keys on the 4th page we treat it as completed
-    All pages must be valid to save so any key set implies data saved on last page
-    To get to last page required all previous pages completed
-    SQ2-2a is the first compulsory question on the 4th page
-    :param declaration:
-    :return:
-    """
-    return "SQ2-2a" in declaration['selectionAnswers']['questionAnswers']
+# def supplier_has_completed_declaration(declaration):
```
TheDoubleK: We can probably delete rather than comment this out?

TheDoubleK:
Other than deleting the comment, looks :+1:  to me.

robyoung:
This should, at some point, move to https://github.com/alphagov/digitalmarketplace-scripts

PR #262, opened 2015-10-07
Rename selection answers to declaration

robyoung:
The language we have now settled on for this is the supplier declaration however it is still referred to as the selection questions in the data API. This change brings the API in line with the language used elsewhere.

Rather than renaming the table and routes to supplier declaration I have use the name `supplier_frameworks`. There are other pieces of information such as whether a supplier has registered interest in a framework that we currently query from the audit table. This was a temporary hack for expediency, now they can be hung off this table.

## ~~Causes downtime~~

~~This change is backwards compatible with respect to the apps that depend on the endpoint. However, because it changes the name of the table it will cause requests against `/suppliers/{}/selection-answers/{}` to fail for the duration of the release.~~

## :sparkles: this has been tested locally with the functional tests :sparkles: 

SSP functional tests with the framework open as well.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -247,8 +247,67 @@ def update_contact_information(supplier_id, contact_id):
     return jsonify(contactInformation=contact.serialize())
 
 
-@main.route('/suppliers/<supplier_id>/selection-answers/<framework_slug>',
-            methods=['GET'])
+@main.route('/suppliers/<supplier_id>/declarations/<framework_slug>', methods=['GET'])
+def get_a_declaration(supplier_id, framework_slug):
+    declaration = SelectionAnswers.find_by_supplier_and_framework(
+        supplier_id, framework_slug
+    )
+    if declaration is None:
+        abort(404)
+
+    return jsonify(declarations=declaration.serialize())
+
+
+@main.route('/suppliers/<supplier_id>/declarations/<framework_slug>', methods=['PUT'])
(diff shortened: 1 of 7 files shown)


allait:
Since the submissions are closed selection answers is read-only right now. Should we consider doing a zero-downtime migration by creating the new table, copying the data over, switching and then deleting the existing one?

I think we can even keep most steps within the same migration (apart from table deletion), so when deploy starts first instance will move the data over while the remaining ones keep using the old tables until they're updated with the new version.

This should be safer in terms of not losing data if something goes wrong. And as far as I understand `selection-answers` are called by the supplier frontend dashboard to show the submission status.

robyoung:
@allait yes, great point. I've updated the migration (also realised I'd messed up the order).

app/main/views/suppliers.py
```diff
@@ -247,20 +247,76 @@ def update_contact_information(supplier_id, contact_id):
...
+        status_code = 201
+
+    request_data = get_json_from_request()
+    json_has_required_keys(request_data, ['declaration', 'updated_by'])
+
+    supplier_framework.declaration = request_data['declaration']
+    db.session.add(supplier_framework)
+    db.session.add(
+        AuditEvent(
+            audit_type=AuditTypes.answer_selection_questions,
```
allait: We should probably rename this at some point too

allait:
:+1:

PR #263, opened 2015-10-07
Add Framework slug to list response

robyoung:
This is needed for listing frameworks as the slug is used in subsequent requests.


TheDoubleK:
:+1:

PR #265, opened 2015-10-08
Set null prices to empty strings

robyoung:
The new G5 / G6 schema does not allow null prices, rather it expects
empty strings.


TheDoubleK:
:+1:

PR #266, opened 2015-10-13
Add framework filter to /services list endpoint

allait:
#### Add framework filter to /services list endpoint
Allows requesting services for given framework slugs only. Will be
used for creating separate g-cloud-6 & g-cloud-7 and dos indexes.

Not setting the framework filter keeps the current behaviour of
returning services for all live frameworks.

#### Use comma separated list for /service?status= queries
Similar to Search API: repeated `&key=value` params are treated as
AND filters, whereas `&key=value1,value2` is a filter matching any
of the listed values (OR filter).

`?status` service filters aren't supported by dmutils apiclient, so
they're unlikely to be used in any existing code.

#### Add last page link to pagination results
Last link index comes from pagination results and doesn't require
any additional queries.

#### Remove index_services scripts
Moved to https://github.com/alphagov/digitalmarketplace-scripts
(alphagov/digitalmarketplace-scripts#1)


allait:
There's a small issue with how we reference frameworks in API endpoints. e.g. `/suppliers` accepts a `framework` parameter, but uses it to look up the `Framework.framework` value instead of the iteration slug. This works pretty well for the buyer frontend, as it doesn't care about the iterations most of the time (which is probably why API and search-api have to). On the other hand, we do need to reference individual iterations in this case. I wonder whether we should come up with a consistent naming for the framework/iteration slug parameters or maybe we can combine them under a single name (so that `framework`could be `g-cloud` or `g-cloud-6` and the API would handle the different filters on its own)

TheDoubleK:
This PR looks good :+1: 

As for the framework slugs - it would be nice to be consistent in how we use them, but I guess it depends on what the need is at each endpoint - i.e. should it care about the iteration number or not?  Anyway I don't think it should hold up getting this merged.

PR #267, opened 2015-10-14
Stop detecting frameworks (on service import)

pcraig3:
Our import route was using some framework detection logic that we didn't need anymore, as services that are returned by the API have framework information included in their JSON content (set in the [`serialize()`](https://github.com/alphagov/digitalmarketplace-api/blob/master/app/models.py#L359) method for services).

_ 

As we're never creating a new service by directly importing them, we know that the import route only gets used to import services returned by the API.  So we can use the existing `validate_service` method to validate the service based on the correct schema, and now that's what we do.


TheDoubleK:
:+1:

PR #268, opened 2015-10-14
Drop selection_answers table

robyoung:
It has been renamed to supplier_frameworks.


TheDoubleK:
Are we leaving the routes in `suppliers.py` as `/selection-answers/` for the time being?  We'll fix it up later when the DOS declarations are going in?

TheDoubleK:
Version numbering will clash with: https://github.com/alphagov/digitalmarketplace-api/pull/271, but no problem to fix that if this goes in first. 

:+1:

PR #269, opened 2015-10-14
Allow all declarations to be changed

robyoung:
Allow supplier declarations to be changed regardless of what state the framework is in. I have discussed this with Cath and the declaration can change at any point in the process but who changes must be controlled. The who is managed by the frontend applications.


pcraig3:
:+1:

PR #270, opened 2015-10-14
[#103088246] Add skeleton DOS schema and allow DOS drafts to be created

TheDoubleK:
The tests add the `dos` framework to the database to allow the tests to run - once `dos` exists in all environments we should remove the `bootstrap_dos()` part of these tests.


robyoung:
:+1:

PR #272, opened 2015-10-19
Add endpoints for suppliers to register interest in a framework

TheDoubleK:
Interest in a framework will now be stored in the `supplier_frameworks` table rather than in audit events.  This adds the endpoints required to register interest and get the frameworks that a supplier is interested in.

Question: what should the return value from the POST endpoint be?  It only allows adding interest in a new framework, not updating all frameworks that the supplier is interested in.  So I have it return only the new framework that has been registered, e.g. any supplier registering interest in g-cloud-9 would get the response:
`{"frameworkInterest": "g-cloud9"}`

The other possibility would be to return the full updated list of all frameworks that the supplier is now registered in, just like the GET endpoint added here - e.g. something like `{"frameworks"=["g-cloud-5", "g-cloud-7", "g-cloud-9"]}`.  That seems not quite right because this POST endpoint can't be used to change any of those other frameworks.  But comments welcome, I can see either way could make sense.

Or does it not really matter because nothing actually looks at the JSON responses from successful POST requests anyway (except our tests)?

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -1,5 +1,3 @@
-from datetime import datetime
-from dmutils.deprecation import deprecated
 from flask import jsonify, abort, request, current_app
 from sqlalchemy.exc import IntegrityError, DataError
 from .. import main
@@ -8,11 +6,11 @@ from ...models import Supplier, ContactInformation, AuditEvent, Service, Supplie
 from ...validation import (
     validate_supplier_json_or_400,
     validate_contact_information_json_or_400,
-    is_valid_string_or_400,
-    validate_new_supplier_json_or_400
+    is_valid_string_or_400
 )
 from ...utils import pagination_links, drop_foreign_fields, get_json_from_request, \
     json_has_required_keys, json_has_matching_id, get_valid_page_or_1
+from ...service_utils import validate_and_return_updater_request
 from ...supplier_utils import validate_and_return_supplier_request
 from dmutils.audit import AuditTypes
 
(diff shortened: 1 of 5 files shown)


robyoung:
Would this not make more sense under the `/suppliers/<supplier_id>/frameworks/<framework_slug>` routes in the suppliers views?

TheDoubleK:
Do you mean the new route in `views/frameworks.py`?

It seemed better as part of the `/frameworks` route as it's getting everything for a particular framework - the `/supplier` routes are generally about specific suppliers (including a `supplier_id` in the URL). This seemed more framework-specific.

robyoung:
Sorry, I had managed to completely skip over the second set of routes. Looks good to me :+1:

app/main/views/suppliers.py
```diff
@@ -361,3 +359,66 @@ def set_selection_questions(supplier_id, framework_slug):
...
+        SupplierFramework.supplier_id == supplier_id
+    ).all()
+    slugs = []
+    for framework in supplier_frameworks:
+        framework = Framework.query.filter(
+            Framework.id == framework.framework_id
+        ).first()
+        slugs.append(framework.slug)
+
+    return jsonify(frameworks=slugs)
```
robyoung: Why do we only return the slugs?

app/main/views/suppliers.py
```diff
@@ -361,3 +359,66 @@ def set_selection_questions(supplier_id, framework_slug):
         abort(400, "Database Error: {}".format(e))
 
     return jsonify(selectionAnswers=supplier_framework.serialize()), status_code
+
+
+@main.route('/suppliers/<supplier_id>/frameworks/interest', methods=['GET'])
```
robyoung: Is this how this is used on the frontend? Don't we ask about interest for a specific framework?

TheDoubleK: I guess how we use it on the frontend is up for grabs to an extent?  It feels right to list all relevant frameworks to me - e.g. say there is a framework that has just closed and two frameworks that are currently open, and the supplier has interest registered in two of them - in that case a single page might show three different contingent messages, and it would be better for the frontend to be able to make a single API call to find all the relevant frameworks, rather than one for each potential framework.

But I'm prepared to be convinced otherwise by a counter-example.

app/main/views/suppliers.py
TheDoubleK: I'm not sure - yes I guess it would make sense to have the whole framework to save on extra API calls to get the framework details.  I'll change this.

robyoung: In which case, could the route just be `/suppliers/<supplier_id>/frameworks`?

PR #273, opened 2015-10-19
Service Lot model and data-only schemas

allait:
## Add Lot model and service lot relationships
As part of adding new lot-related rules we've decided to pull Lot
into a separate model.

This adds the basic Lot table with a many-to-many relationship with
frameworks: since G-Cloud framework iterations reuse the same lots
but other frameworks might change lots between iterations.

Lots have a one_service_lot flag for separating DOS lots that only
allow one service per supplier.

Only name and slug are stored for each lot, there's no separate field
for capitalized abbreviation that is currently stored in service data
and displayed in some service list pages. Service field will be kept
in existing G-Cloud service responses for now, until content "depends"
rules and other uses are moved to the lot slug checks.

Framework objects are automatically joined to lots, `Framework.get_lot`
method can be used to look up the lot object with a given slug.

#### Add migration to populate service lot_id columns
Uses service.data['lot'] to look up the lot ids and populate the
lot_id column for three service tables.

Archived services might have 'missing' lot from initial data import.
Since 'missing' lots were replaced with 'SCS' we do the same thing
in the migration script for any archived services.

## Add service helper to validate framework, lot and supplier fields
When creating a service or draft checks that incoming JSON has a
valid framework and lot slugs and supplier ID and returns the
referenced objects.

Depending on the endpoint logic this could be run before the
JSON schema validation, so the function checks that required
keys exist and handles the non-integers supplier ID DataError.

Even though the supplier object is rarely used by calling views
and we could continue to rely on the IntegrityError to check for
missing suppliers, adding it to the same function simplifies the
logic a bit by creating a single place responsible for verifying
references. The added Supplier query shouldn't add significant
load.

As a side-effect, this changes the error response message for
non-digit supplierId values (used to be JSON schema error) and
missing supplierIds (used to be database IntegrityError).

## Keep only service.data JSON fields in services JSON schemas
The use and purpose of service JSON schemas have never been properly
defined. As a result the set of fields defined by the schemas is not
the same as the set of fields defined on the service objects within
the application. Currently schemas validate neither the service
responses given by the API (due to missing metadata and fields mixed
in from other models) nor the service.data JSON (since schemas define
id, status and other fields that are stored in separate columns).

This creates the need for additional steps to add or remove relevant
fields before validating against the schemas, which change depending
on the schema involved and the action that is being performed.

There are two possible approaches to avoid this:
1. Make the schemas define the API response. This would mean adding
   all additional fields returned by the API to the schema definitions.
2. Make the schemas define the service.data JSON only and rely on other
   methods to check additional fields when needed.

First approach would generally be more beneficial: it creates a clear
definition of both API responses and possible request payloads. It
would also allow the validation to be performed at an earlier stage
during service/draft creation, since the schema can be applied to the
full request JSON without first removing any non-data fields.

However, current JSON schema specs lack any inheritance or reuse
features that would allow us to avoid defining the common service
fields inside each schema definition. Having to define the fields
each time reintroduces some of the issues we're trying to avoid:
duplication (especially related to possible field values), which
leads to small differences between schemas that could go unnoticed.

At the same time, while the schemas could be used to validate some
basic properties of the additional fields, there's often still a
need to perform additional validation steps within the app, which
often remove the need for schema validation (e.g. validating that
supplier ID, framework or lot slugs exist is better than just
checking the format).

Defining only service.data fields in the JSON schemas means that
schema definitions avoid repeating the same fields, while the
validation and processing of additional fields is handled by the
application code.

## Check lot.one_service_limit when creating DOS drafts
one_service_limit lots only allow one service for each supplier, so
if the draft already exists we return a 400


tests/app/views/test_frameworks.py
```diff
@@ -166,41 +177,40 @@ def test_stats(self):
         self.setup_data('g-cloud-7')
 
         response = self.client.get('/frameworks/g-cloud-7/stats')
+        print response.get_data()
```
robyoung: Brrp

robyoung:
This looks good but we need to make some changes to the buyer and supplier apps to avoid any breakages before this is merged.

app/draft_utils.py
```diff
@@ -17,45 +16,11 @@ def get_request_page_questions():
...
-    if not slug:
-        framework = Framework.query.filter(
-            Framework.id == framework_id
-        ).first()
-        slug = framework.slug
-
+def get_draft_validation_errors(draft_json, framework=None, lot=None, required=None):
     errs = get_validation_errors(
-        "services-{0}-{1}".format(slug, draft_json['lot'].lower()),
+        "services-{0}-{1}".format(framework.slug, lot.slug),
```
TheDoubleK: Seems like this method requires a framework and lot (`.slug` will blow up otherwise), so there's no point in `=None` as defaults in the signature?

app/main/views/drafts.py
```diff
@@ -264,27 +268,32 @@ def create_new_draft_service(framework_slug):
...
+    if lot.one_service_limit:
+        lot_service = DraftService.query.filter(DraftService.supplier == supplier, DraftService.lot == lot).first()
+        if lot_service:
+            abort(400, "'{}' service already exists for supplier '{}'".format(lot.slug, supplier.supplier_id))
 
-    supplier_id = draft_json['supplierId']
-    errs = get_draft_validation_errors(draft_json, slug=framework_slug)
-    if errs:
-        return jsonify(errors=errs), 400
+    if framework.status != 'open':
```
TheDoubleK: I might put this check before the `one_service_limit` check - seems like if the framework isn't open then we shouldn't even start looking into the services and lots for it.

But it doesn't really matter...

app/main/views/services.py
```diff
@@ -162,40 +160,24 @@ def import_service(service_id):
...
+
+    framework, lot, supplier = validate_and_return_related_objects(service_data)
+
+    service = Service(
+        service_id=service_id,
+        supplier_id=supplier.supplier_id,
+        lot_id=lot.id,
+        framework_id=framework.id,
+        status=service_data.pop('status', 'published'),
+        created_at=service_data.pop('createdAt', None),
```
TheDoubleK: Are the timestamps going to be passed in by the frontend apps now? That seems weird to me - the API knows better than the frontend what time the record was created/updated.

allait: This is the import endpoint, so we keep the original service create and update timestamps. But it's a good note since there's no need for `.pop` here anymore, I'll remove it.

TheDoubleK: Ah OK of course.  But shouldn't we have a value other than `None` in case there are no dates - they are `NOT NULL` columns (shouldn't ever happen, I know, so maybe it's right that the import will fail if there are no dates?).

app/models.py
```diff
@@ -345,6 +402,38 @@ def supplier(cls):
...
+        return value
+
+    @validates('status')
+    def validates_status(self, key, value):
+        if value not in self.STATUSES:
+            raise ValidationError("Invalid status value '{}'".format(value))
+
+        return value
+
+    @validates('data')
```
TheDoubleK: I like this `@validates` decorator - nice.  But a validation method feels like an odd place to do all this data manipulation (dropping fields, stripping whitespace).  I'm not sure... it is nice and neat having it here though.

app/main/views/services.py
allait: If they're not passed in then column default values will be applied (setting both to `.utcnow()`). I've just checked again to make sure that if create and update dates are set to `None` the default value gets stored.

migrations/versions/330_add_lot_table_and_relationships.py
```diff
@@ -0,0 +1,97 @@
...
+        sa.Column('name', sa.String(), nullable=False),
+        sa.Column('one_service_limit', sa.Boolean(), nullable=False),
+        sa.PrimaryKeyConstraint('id')
+    )
+
+    op.create_index(op.f('ix_lots_slug'), 'lots', ['slug'], unique=False)
+
+    op.create_table(
+        'framework_lots',
+        sa.Column('framework_id', sa.Integer(), nullable=True),
```
TheDoubleK: I don't think either column in this table should be nullable - you need both IDs to record a relationship, an entry with just one ID is meaningless.

app/models.py
allait: I think the name is a bit misleading – essentially they're just setters for the attribute, so whatever value gets returned is stored. Validation is only achieved by raising exceptions (which is also something you'd do in a setter), so it's different from the common `if x.is_valid()` pattern.

I think it's better to keep the naming convention/methods from SQLAlchemy though (even if we could just use properties here) and keeping all the ".data -> JSON -> .data" within the model definitely feels like a good approach for us.

migrations/versions/330_add_lot_table_and_relationships.py
```diff
@@ -0,0 +1,97 @@
...
+    framework_lots_table = table(
+        'framework_lots',
+        column('framework_id', sa.Integer),
+        column('lot_id', sa.Integer)
+    )
+
+    # Add 4 lots (ids 1-4) to all G-Cloud frameworks (ids 1-4)
+    op.bulk_insert(framework_lots_table, [
+        {'framework_id': framework_id, 'lot_id': lot_id}
+        for framework_id, lot_id in itertools.product(range(1, 5), range(1, 5))
```
TheDoubleK: Ideally this should look up IDs for all `gcloud` frameworks rather than relying on database IDs.

But as we in fact have `gcloud` frameworks with IDs 1-4 on all environments we can probably leave it as is.

migrations/versions/340_extract_service_lot_into_lot_id_column.py
```diff
@@ -0,0 +1,42 @@
...
+                  )
+                )
+        """.format(table=table))
+
+    op.alter_column('archived_services', 'lot_id', existing_type=sa.BIGINT(), nullable=False)
+    op.alter_column('draft_services', 'lot_id', existing_type=sa.BIGINT(), nullable=False)
+    op.alter_column('services', 'lot_id', existing_type=sa.BIGINT(), nullable=False)
+
+
+def downgrade():
```
TheDoubleK: Should downgrade also delete the `lot_id` entries for all g-cloud 4,5,6 and 7 services?

migrations/versions/340_extract_service_lot_into_lot_id_column.py
```diff
@@ -0,0 +1,42 @@
...
+        op.execute("""
+            UPDATE {table} SET lot_id = lots.id
+            FROM framework_lots, lots
+            WHERE
+                framework_lots.framework_id = {table}.framework_id
+                AND framework_lots.lot_id = lots.id
+                AND (
+                    lower({table}.data->>'lot') = lots.slug
+                    OR (
+                        {table}.data->>'lot' = 'missing'
```
TheDoubleK: Ouch.  It's a shame we've never fixed preview and staging.  We could delete the services with missing lot from those environments and then we wouldn't need this in the migration?

allait: This affects the archived services too and then everyone would have to do this locally if they have them in the dev DB. I think having this in a migration should be ok since it just does the same change we've applied to production data and it has the benefit of fixing the preview/staging missing lot services at the same time.

migrations/versions/340_extract_service_lot_into_lot_id_column.py
allait: We could, but it's likely that both migrations will be applied/downgraded at the same time and `330` will remove the whole column on downgrade, so I skipped the update to NULL step here. Happy to add it if you think there might be a need for it.

tests/app/views/test_drafts.py
```diff
@@ -884,8 +897,8 @@ def test_copy_draft(self):
             content_type='application/json')
 
         data = json.loads(res.get_data())
-        assert_equal(res.status_code, 201)
-        assert_equal(data['services']['lot'], 'SCS')
+        assert_equal(res.status_code, 201, res.get_data())
```
TheDoubleK: What does the third parameter to `assert_equal` do here? I can't seem to find any documentation on it.

tests/test_validation.py
```diff
@@ -14,6 +14,12 @@
                                                     '..', 'example_listings'))
 
 
+def drop_api_exported_fields_so_that_api_import_will_validate(data):
```
TheDoubleK: Would it be better to fix the example listing rather than re-introduce this method here?  The example listing should be up-to-date with what the API will handle.

app/main/views/services.py
TheDoubleK: OK great - I looked before and couldn't see any default value in the database.

app/models.py
TheDoubleK: Yep let's keep it here. It's nice and neat.

migrations/versions/340_extract_service_lot_into_lot_id_column.py
TheDoubleK: We only roll forward ;)
And you're right about rolling both back if we ever did want to - leave it as it is.

migrations/versions/340_extract_service_lot_into_lot_id_column.py
TheDoubleK: Ah yes - because we'll never look at the 'lot' field from the JSON again.  I hadn't worked that out.

TheDoubleK:
I'm done reviewing - only minor things - sterling work here!

app/main/views/drafts.py
allait: You're absolutely right, nice catch. Moved!

migrations/versions/330_add_lot_table_and_relationships.py
allait: Set `nullable=False` on both and updated the migration

migrations/versions/330_add_lot_table_and_relationships.py
allait: I went with hardcoded IDs since g-cloud framework records were created with migrations up till now, so the IDs should be consistent across all environments. But you're right, going with a query would've been cleaner.

tests/app/views/test_drafts.py
allait: That's exactly what it does: it'll show the error response instead of the `400 != 201` that we usually get on failures. I changed it while looking into test errors and figured I'd commit it since it is a bit more helpful.

tests/test_validation.py
allait: The example listings contain the full data for service requests, not just fields that get stored to `.data`. I believe changing them would unfortunately break a lot of tests that don't mix all of the request fields into request data, so keeping the helper method here was an easier solution for now.

allait:
@TheDoubleK fixed nullable columns and moved the lot service limit check after the framework.status check.

Updated to merge `get_draft_validation_errors` with `validate_service_data` and remove the additional `framework` and `lot` arguments. New commits:

#### Only show missing keys in required keys validation error
Previous message used to return the full list of required keys when
some of them were missing, which makes it harder to track down the
error when dealing with test fixtures. Changed function only returns
the names of the keys that weren't present.

#### Use relationship attributes instead of foreign keys for assignment
Assigning framework, lot and supplier instances directly to attributes
encapsulates knowledge about foreign key relationships within the models
and makes sure valid service/draft instances always have a reference to
related framework, lot and supplier records.

#### Merge get_draft_validation_errors with validate_service_data
Removes framework and lot parameters from validate_service_data
since service/draft objects should always have references to
lot and framework objects.

Adds enforce_required and required_fields arguments to
validate_service_data so that it can be used both for services and
drafts.

#### Add a test for service.data validation in import endpoint

tests/app/views/test_drafts.py
TheDoubleK: Yep definitely better than the default message.

tests/test_validation.py
TheDoubleK: You know "for now" will mean "forever", but yes fair enough.

TheDoubleK:
I like "Use relationship attributes instead of foreign keys for assignment" - does this just work out of the box without any change to the Draft model? Feels like magic!

TheDoubleK:
With all fingers crossed and holding my breath... :+1:

TheDoubleK:
Let's wait until Monday to merge it, eh?

PR #274, opened 2015-10-26
[#106103948] Pass/fail for supplier frameworks

TheDoubleK:
This is the API part of this story: https://www.pivotaltracker.com/story/show/106103948

Adds two new columns to the `supplier_frameworks` table and an endpoint for updating values in them.

**Note to self: Migration sequence numbers will probably need updating by the time this is ready to merge**

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -341,7 +341,7 @@ def register_interest_in_framework(supplier_id, framework_slug):
     ).first_or_404()
 
     interest_record = SupplierFramework.query.filter(
-        SupplierFramework.supplier_id == supplier_id,
+        SupplierFramework.supplier_id == supplier.supplier_id,
         SupplierFramework.framework_id == framework.id
     ).first()
 
@@ -349,13 +349,13 @@ def register_interest_in_framework(supplier_id, framework_slug):
         return jsonify(frameworkInterest=interest_record.serialize()), 200
     else:
         interest_record = SupplierFramework(
-            supplier_id=supplier_id,
+            supplier_id=supplier.supplier_id,
             framework_id=framework.id
         )
         audit_event = AuditEvent(
             audit_type=AuditTypes.register_framework_interest,
             user=updater_json.get('user'),
-            data={'supplierId': supplier_id, 'frameworkSlug': framework_slug},
+            data={'supplierId': supplier.supplier_id, 'frameworkSlug': framework_slug},
             db_object=supplier
         )
 
@@ -382,16 +382,15 @@ def register_framework_interest(supplier_id, framework_slug):
         Supplier.supplier_id == supplier_id
     ).first_or_404()
 
-    interest_record = SupplierFramework.query.filter(
-        SupplierFramework.supplier_id == supplier_id,
-        SupplierFramework.framework_id == framework.id
-    ).first()
-
     json_payload = get_json_from_request()
     json_payload.pop('update_details')
     if json_payload:
         abort(400, "This PUT endpoint does not take a payload.")
 
+    interest_record = SupplierFramework.query.filter(
+        SupplierFramework.supplier_id == supplier.supplier_id,
+        SupplierFramework.framework_id == framework.id
+    ).first()
     if interest_record:
         return jsonify(frameworkInterest=interest_record.serialize()), 200
 
@@ -399,7 +398,7 @@ def register_framework_interest(supplier_id, framework_slug):
         abort(400, "'{}' framework is not open".format(framework_slug))
 
     interest_record = SupplierFramework(
-        supplier_id=supplier_id,
+        supplier_id=supplier.supplier_id,
         framework_id=framework.id
     )
     audit_event = AuditEvent(
@@ -437,7 +436,7 @@ def update_supplier_framework_details(supplier_id, framework_slug):
     update_json = json_payload["frameworkInterest"]
 
     interest_record = SupplierFramework.query.filter(
-        SupplierFramework.supplier_id == supplier_id,
+        SupplierFramework.supplier_id == supplier.supplier_id,
         SupplierFramework.framework_id == framework.id
     ).first()
 
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -379,6 +379,54 @@ def register_interest_in_framework(supplier_id, framework_slug):
         return jsonify(frameworkInterest=interest_record.serialize()), 201
 
 
+@main.route('/suppliers/<supplier_id>/frameworks/<framework_slug>', methods=['PUT'])
+def register_framework_interest(supplier_id, framework_slug):
+    updater_json = validate_and_return_updater_request()
+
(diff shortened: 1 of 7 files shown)


app/main/views/suppliers.py
```diff
@@ -422,3 +377,103 @@ def register_interest_in_framework(supplier_id, framework_slug):
...
+def register_framework_interest(supplier_id, framework_slug):
+    updater_json = validate_and_return_updater_request()
+
+    framework = Framework.query.filter(
+        Framework.slug == framework_slug
+    ).first_or_404()
+
+    supplier = Supplier.query.filter(
+        Supplier.supplier_id == supplier_id
+    ).first()
```
allait: Why can't we do `.first_or_404` here similar to the framework query above?

TheDoubleK: No reason... for variety? ;)

app/main/views/suppliers.py
```diff
@@ -422,3 +377,103 @@ def register_interest_in_framework(supplier_id, framework_slug):
...
+
+    if supplier is None:
+        abort(404, "supplier_id '{}' not found".format(supplier_id))
+
+    interest_record = SupplierFramework.query.filter(
+        SupplierFramework.supplier_id == supplier_id,
+        SupplierFramework.framework_id == framework.id
+    ).first()
+
+    if interest_record:
```
allait: I think this behaviour might be surprising. There's no indication that the endpoint doesn't take a payload and only saves supplier and framework id, so while it makes sense to return 200 if we know the data is the same, none of the side effects are triggered (i.e. there's no audit event with update details about the request) and there's no indication that any additional data that could've been sent in the request was discarded.

I suggest we either check that request doesn't contain anything besides updater details and keep the 200 or we reject repeated PUT requests (they shouldn't happen anyway once the "Make an application" button is hidden on the supplier frontend)

app/main/views/suppliers.py
```diff
@@ -422,3 +377,103 @@ def register_interest_in_framework(supplier_id, framework_slug):
...
+
+    supplier = Supplier.query.filter(
+        Supplier.supplier_id == supplier_id
+    ).first()
+
+    if supplier is None:
+        abort(404, "supplier_id '{}' not found".format(supplier_id))
+
+    json_payload = get_json_from_request()
+    json_has_required_keys(json_payload, ["update"])
```
allait: I think we should stick with the pattern we use for service updates with update key matching the response key, so this should be `"frameworkInterest"`

app/main/views/suppliers.py
TheDoubleK: Ah I've got it - because I cut & pasted from the (now deprecated) POST method above and didn't want to mess with the pattern.

app/models.py
```diff
@@ -260,6 +260,8 @@ class SupplierFramework(db.Model):
                              db.ForeignKey('frameworks.id'),
                              primary_key=True)
     declaration = db.Column(JSON)
+    on_framework = db.Column(db.Boolean, nullable=True)
```
pcraig3: So this is another 3-way boolean, right?

If a given supplier:

- applies and succeeds -> `on_framework = True`
- applies and fails -> `on_framework = False`
- doesn't apply -> no `SupplierFramework` record

Have I got that right?

app/main/views/suppliers.py
```diff
@@ -422,3 +377,103 @@ def register_interest_in_framework(supplier_id, framework_slug):
...
+        abort(400, "'{}' framework is not open".format(framework_slug))
+
+    interest_record = SupplierFramework(
+        supplier_id=supplier_id,
+        framework_id=framework.id
+    )
+    audit_event = AuditEvent(
+        audit_type=AuditTypes.register_framework_interest,
+        user=updater_json.get('updated_by'),
+        data={'supplierId': supplier.supplier_id, 'frameworkSlug': framework_slug},
```
pcraig3: Not that it's a huge deal, but sometimes you've got `supplier.supplier_id` and other times just `supplier_id`.  
Any reason not to standardise on using/not using the database record?

tests/app/views/test_suppliers.py
```diff
@@ -1261,13 +1157,171 @@ def test_can_get_registered_frameworks_for_a_supplier(self):
...
+
+
+class TestSupplierFrameworkUpdates(BaseApplicationTest):
+    def setup(self):
+        super(TestSupplierFrameworkUpdates, self).setup()
+
+        self.setup_dummy_suppliers(1)
+
+        with self.app.app_context():
+            framework_enum_vals = db.session.execute("SELECT enum_range(NULL::framework_enum);").first()[0]
```
pcraig3: :crying_cat_face:

tests/app/views/test_suppliers.py
```diff
@@ -1261,13 +1157,171 @@ def test_can_get_registered_frameworks_for_a_supplier(self):
...
+            if 'dos' not in str(framework_enum_vals):
+                self.bootstrap_dos()
```
pcraig3: :crying_cat_face:

app/models.py
TheDoubleK: Yep that's right - if no application was made then this column will never get populated (though there could be a completed declaration if they didn't complete any services).

We might or might not want to clear out the "didn't apply" rows in future, but shouldn't be a problem in the meantime.

app/main/views/suppliers.py
TheDoubleK: `supplier.supplier_id ` is when we want the integer from the database record.  The string `supplier_id` comes from the parameter passed in URL and is only used for error messaging (I think?) - seemed to make sense as the error relates to the value passed in the URL.

TheDoubleK: I'll take another look over to see it makes sense though.

pcraig3: Ah, I didn't realise they were different types.  

Although, I'm not really sure that it matters which one you use.  I would have thought that both the `format()` method and inserting into the database will take either value.

TheDoubleK:
Comments addressed:
* Use `first_or_404() where appropriate
* Reject PUTs with a payload
* Use `frameworkInterest` as the JSON key for updates
* Use database supplier ID where appropriate
Also: 
* Fixed up the migrations order after a rebase

app/main/views/suppliers.py
TheDoubleK: It mattered for creating the audit event - using the string type gave the supplier ID quotes in the update JSON, which looked wrong.

app/models.py
robyoung: This isn't very clear, maybe it should be documented in the commit message for https://github.com/alphagov/digitalmarketplace-api/commit/c2227c7512c130c838f18a1a87a19bad45ea85ba

TheDoubleK: Reworded here: https://github.com/alphagov/digitalmarketplace-api/commit/2e14a2bd7b4d3316bc64c12487373f3d3a5ccc65

robyoung:
:+1:

PR #275, opened 2015-10-26
Add new statuses to frameworks

quis:
### coming ⇢ open ⇢ pending ⇢ standstill ⇢ live ⇢ expired

--

- "coming" replaces "pending" as the state before a framework is open for suppliers to make their applications
- "pending" moves in the order to reflect that it is the stage between a supplier applying and CCS telling suppliers if they're successful
- "standstill" is a new state which is in between suppliers being told if they're successful and their services being live on the marketplace

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -1,10 +1,15 @@
 from flask import jsonify, abort
 from sqlalchemy.types import String
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy import func, orm, case
 import datetime
 
+from dmutils.audit import AuditTypes
 from .. import main
-from ...models import db, Framework, DraftService, User, Supplier, SupplierFramework, AuditEvent, Lot
+from ...models import (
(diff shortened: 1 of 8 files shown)


TheDoubleK:
Requires an update to the status enum - see here for something to copy & paste: https://github.com/alphagov/digitalmarketplace-api/blob/master/migrations/versions/100_add_open_status.py

allait:
I'd suggest getting rid of that enum and replace it with model validator similar to https://github.com/alphagov/digitalmarketplace-api/blob/master/app/models.py#L416-L421. That would still require a migration but at least we won't need to do this each time we want to add/remove/rename a status in the future. I don't think the safety provided by enums is worth the migrations issues and `@validators` error message can be much nicer than the database error we get from invalid enum values.

robyoung:
We discussed how we should go about adding or updating a framework now that the validation isn't in the database. It would be good to add that to the README or a script.

quis:
:+1: to the endpoint

robyoung:
:+1: to the rest

PR #277, opened 2015-10-28
Adding a description to hashpw

tommorris:
For a human reader, `hashpw` is ambiguous. `hashpw` being true could mean "the password in this payload has already been hashed" or it could mean "the password in this payload needs to be hashed". Similarly, it being false could mean "the supplied password has not been hashed" or it could mean "don't hash this password, I've already done it".

Adding a brief description in the JSON schema makes it clear what the purpose of it is. Found this because I tried to create a user following the schema and ended up with a test user in the DB with a cleartext password.

### DIFF
--- a/json_schemas/users.json
+++ b/json_schemas/users.json
@@ -25,7 +25,8 @@
       "type": "integer"
     },
     "hashpw": {
-      "type": "boolean"
+      "type": "boolean",
+      "description": "set to False to skip password hashing"
     }
   },
   "required": [


allait:
`hashpw` was used when we imported users from the old database. I don't think there's a need for it now (passwords are one thing we'd never want to move between environment databases), so I think we should remove 'hashpw' altogether at some point. I'd opt for removing it from the schema to begin with, but since that's a bit more work there's no reason not to merge this now, so :+1:

tommorris:
@allait I didn't want to do anything drastic until I understood more about the reasons behind it and the overall architecture.

PR #280, opened 2015-10-30
Added some new tests

tommorris:
Just [being bold](https://en.wikipedia.org/wiki/Wikipedia:BOLD) and adding some new tests and docstrings.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -1,8 +1,9 @@
 from flask import jsonify, abort, request, current_app
+from sqlalchemy import func, orm
 from sqlalchemy.exc import IntegrityError, DataError
 from .. import main
 from ... import db
-from ...models import Supplier, ContactInformation, AuditEvent, Service, SupplierFramework, Framework
+from ...models import Supplier, ContactInformation, AuditEvent, Service, DraftService, SupplierFramework, Framework
 from ...validation import (
     validate_supplier_json_or_400,
     validate_contact_information_json_or_400,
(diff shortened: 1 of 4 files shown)


tests/app/test_utils.py
```diff
@@ -0,0 +1,55 @@
+from app.utils import (display_list,
+                       strip_whitespace_from_data,
+                       json_has_matching_id,
+                       json_has_required_keys,
+                       link)
+
+from nose.tools import assert_equal, raises
```
allait: API was switched to pytest, so we should probably start using pytest tools for new tests

tests/app/test_utils.py
```diff
@@ -0,0 +1,55 @@
...
+
+def test_strip_whitespace_from_data():
+    struct = {"eggs": "  ham "}
+    assert_equal(strip_whitespace_from_data(struct)['eggs'], "ham")
+    struct_with_list = {"eggs": ["  spam  ", "  ham  ", "  eggs "]}
+    after_strip = strip_whitespace_from_data(struct_with_list)
+    for item in after_strip['eggs']:
+        # check to see if there is any whitespace left.
+        # if not, find will return -1
+        assert_equal(item.find(" "), -1)
```
allait: `assert " " not in item`

allait:
I don't think we've ever officially adopted it, but I think most of our commit messages follow the GDS styleguide (https://github.com/alphagov/styleguides/blob/master/git.md), so it might be worth taking a look at it.

app/utils.py
```diff
@@ -5,6 +5,7 @@
 
 
 def link(rel, href):
+    """Generates a link dict from a rel, href pair."""
```
allait: Should be "Generate a link ...", according to https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

tests/app/test_utils.py
```diff
@@ -0,0 +1,55 @@
+from app.utils import (display_list,
```
allait: Again, I don't think we've ever agreed on this, but things that are tested usually are imported after the test tools.

robyoung:
Needs a rebase but :+1: from me.

PR #281, opened 2015-11-02
Add initial JSON schemas for DOS lots …

allait:
### Add initial JSON schemas for DOS lots
Adds schemas for 4 Digital Outcomes and Specialiasts lots. Schemas
are configured with `additionalProperties: true` since there's no
list of service fields yet, but this should be changed once the
properties are added.

### Remove unused /framework/<slug>/status endpoint
GET /framework/<slug> returns the framework status with other
framework info, so a separate endpoint is no longer needed.

It's not used by any of the apps at the moment.

### Add page_questions validation to create draft endpoint
Returns validation errors for page_questions fields to allow
creating a draft object from any section data. This will be
used by the supplier frontend to create drafts after the first
section page for the selected lot is submitted.


robyoung:
:+1:

PR #282, opened 2015-11-02
Add supplier frameworks endpoint

quis:
## Get frameworks that a supplier has interacted from `/suppliers/1/frameworks`

This commit adds an endpoint to list the `supplier_frameworks` table.

For any framework from G-Cloud 7 onwards, this table has an entry when a supplier does stuff like register interest, answer declaration questions, etc

## Add service counts to `/suppliers/1/frameworks` response

For each framework that a supplier has interacted with, this commit adds:
- count of published services
- count of submitted draft services
- count of draft draft services

It doesn’t add this for the individual framework endpoint because we don’t have a use case for this yet.

## Add SupplierFramework entry for all G5/6 suppliers

G-Cloud 5 and 6 suppliers have 'interacted' with a framework, it just wasn’t on our system. So we should retroactively record this interaction.

If we *don’t* do this then it means 3 API queries:
- get all frameworks
- get all frameworks for a supplier
- get frameworks with a supplier’s services
- …and much munging

And things would get even more complicated when G6 and G7 are both live.

_Paired with @allait_

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -1,8 +1,9 @@
 from flask import jsonify, abort, request, current_app
+from sqlalchemy import func, orm
 from sqlalchemy.exc import IntegrityError, DataError
 from .. import main
 from ... import db
-from ...models import Supplier, ContactInformation, AuditEvent, Service, SupplierFramework, Framework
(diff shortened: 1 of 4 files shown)


robyoung:
:+1:

PR #283, opened 2015-11-04
Re-add nose as requirements

carolinegreen:
Nose deprecated, however some old tests are still using this and it needs to be kept as a requirment
in the meantime.


allait:
Hmm, that's interesting. I was under the impression that py.test mocked nose imports since the tests passed on fresh Travis builds without installing nose. But it's possible that nose is always preinstalled on Travis environments.

Starting with a fresh virtualenv locally I get the nose import error, so :+1:

PR #285, opened 2015-11-04
Replace validation message rewrites with validator type checks

allait:
Instead of matching validation messages returned by the jsonschema
we use additional properties set on the ValidationError to replace
the message with the content validator name.

Some of the checks map directly from JSON schema validator to the
content validators, others require additional checks.

"required" and word count are the only ones that rely on substring
matching, since they don't expose any other values that could be
used instead (required for Draft v4 doesn't return the name of the
missing property, only the full list of required properties, so we
have to match the message text).

### DIFF
--- a/app/validation.py
+++ b/app/validation.py
@@ -37,9 +37,6 @@ SCHEMA_NAMES = [
 ]
 FORMAT_CHECKER = FormatChecker()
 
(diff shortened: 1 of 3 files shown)


robyoung:
I like this! :+1:

PR #286, opened 2015-11-04
Return 410 for services on expired frameworks

TheDoubleK:
When the G-Cloud 5 framework expires, we should return 410 ("Gone") rather than 404s for G5 service pages.

See: https://www.pivotaltracker.com/story/show/107382498

This is related to, but not dependent on: https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/185

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -187,9 +187,14 @@ def import_service(service_id):
 def get_service(service_id):
     service = Service.query.filter(
         Service.service_id == service_id
-    ).framework_is_live().first_or_404()
+    ).first_or_404()
 
-    return jsonify(services=service.serialize())
+    if service.framework.status == 'live':
+        return jsonify(services=service.serialize())
+    elif service.framework.status == 'expired':
+        abort(410)
+    else:
+        abort(404)
 
 
 @main.route('/archived-services/<int:archived_service_id>', methods=['GET'])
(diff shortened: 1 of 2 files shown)


quis:
LGTM :+1:

allait:
I don't think we should be doing this here. We allow listing services with any framework and status, there's no reason to remove the API ability to return a service based on the framework status.

RIght now buyer frontend checks the service status and returns 404 if it's not published (https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/main/views.py#L164), I think we need to either add the framework status check there, or move that logic to the dmutils `get_framework` method since it will probably be used by supplier frontend too.

TheDoubleK:
This doesn't remove any capability - this endpoint only ever returned `live` services and 404'd for everything else.  All this change does is change the error code for expired services.

allait:
@TheDoubleK you're right, I've missed the `.framework_is_live` bit. I still don't agree with keeping this in the API – we allow listing expired services, but then don't return individual resources.

TheDoubleK:
Agreed - it doesn't seem right behaviour for the API, but pushing the check to buyer & supplier apps (probably via utils) will be a fair bit more work and doesn't feel like a priority right now... could we add that as a story in the icebox?

allait:
@TheDoubleK yeah, let's do that. You're right that we're not changing behaviour here, so there's no reason not to merge this as is. :+1:

TheDoubleK:
I added a story here: https://www.pivotaltracker.com/story/show/107478318

PR #287, opened 2015-11-06
Filter suppliers on a framework

robyoung:
Add the route `/frameworks/<framework_slug>/suppliers` for listing suppliers associated with a framework. This expands on what is provided through `/frameworks/<framework_slug>/interest` with full supplier framework objects and allows the list to be filtered by whether they have returned their agreement.

This is required for the admin interface for CCS which will allow them to list suppliers that have returned their framework agreements with the most recent first.


app/main/views/frameworks.py
```diff
@@ -136,6 +137,33 @@ def label_columns(labels, query):
...
+            supplier_frameworks = supplier_frameworks.filter(
+                SupplierFramework.agreement_returned == false()
+            )
+
+    return jsonify(supplierFrameworks=[
+        supplier_framework.serialize() for supplier_framework in supplier_frameworks
+    ])
+
+
 @main.route('/frameworks/<string:framework_slug>/interest', methods=['GET'])
```
TheDoubleK: Should we deprecate this `interest` endpoint in favour of the new one? (I'm not saying make all the changes now! But to tidy up at some later date.)

app/main/views/frameworks.py
```diff
@@ -136,6 +137,33 @@ def label_columns(labels, query):
...
+    agreement_returned = request.args.get('agreement_returned')
+
+    supplier_frameworks = SupplierFramework.query.filter(
+        SupplierFramework.framework_id == framework.id
+    )
+
+    if agreement_returned is not None:
+        if convert_to_boolean(agreement_returned):
+            supplier_frameworks = supplier_frameworks.filter(
+                SupplierFramework.agreement_returned == true()
```
TheDoubleK: By having `== true()` here does it mean that both `?agreement_returned=true` and  `?agreement_returned=false` will return the list of suppliers who have returned their agreement, which seems odd.

Should it be `== convert_to_boolean(agreement_returned)` instead to get either suppliers who have or suppliers who haven't returned their agreements?  Or have I missed something?

tests/app/views/test_suppliers.py
```diff
@@ -1,5 +1,8 @@
+from datetime import datetime
 from flask import json
-from nose.tools import assert_equal, assert_in, assert_is_not_none, assert_true
+import pytest
+from freezegun import freeze_time
```
TheDoubleK: I like this `freeze_time` - nice find.

TheDoubleK:
Reviewed - looks good, other than being able to filter on `agreement_returned=false`

app/main/views/frameworks.py
robyoung: Yea, I was going to do that when I rename the response key.

app/main/views/frameworks.py
robyoung: Talked in real world.

TheDoubleK:
:+1: once rebased

PR #288, opened 2015-11-06
Add supplierName to SupplierFramework

robyoung:
Add the supplierName to the serialised version of SupplierFramework.
This is needed particularly for the admin app where we list suppliers
that have uploaded framework agreements.


TheDoubleK:
:+1:

PR #290, opened 2015-11-09
Add frameworkStatus to service response

robyoung:
This is required so that frontends can decide whether or not to show a service. Once this change is merged and the buyer frontend is updated to make use of it then the get service endpoint can be changed to always return services regardless of framework status.


TheDoubleK:
:+1:

PR #292, opened 2015-11-11
G7 launch

robyoung:
## Allow all services to be returned by API
Previously the buyer frontend expected the API to raise a 404 or 410 if the framework status was not valid. It now handles that itself by looking at the frameworkStatus so now the API can return all services it knows about.

## Test search API call on draft publishing
Test that the search API is called correctly during the draft publishing process. Services should only be indexed in the search API if the framework is live.

## Prevent draft deletion on publish submission
This change has been introduced to prevent drafts created through the submission process (submissions) from being deleted when they are published. In implementing this I have tightened up what actions can happen in various different states.

The `not-submitted` and `submitted` statuses are only possible on drafts that were created during the submissions process. They will not have a service ID until they are published. Because only submissions with the 'submitted' status can be published only they can have service IDs. When this kind of draft is published it is not deleted so as to preserve the submission.

The `enabled`, `disabled` and `published` statuses are inherited from the service model. Drafts can only have these statuses if they are created from an existing service. These drafts will always have a service ID. When this kind of draft is published it is deleted.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -227,10 +227,7 @@ def publish_draft_service(draft_id):
             Service.service_id == draft.service_id
         ).first_or_404()
 
-        service_from_draft = update_and_validate_service(
-            service,
-            draft.data)
-
+        service_from_draft = update_and_validate_service(service, draft.data)
     else:
         service_from_draft = create_service_from_draft(draft, "enabled")
 
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -235,7 +235,7 @@ def publish_draft_service(draft_id):
 
         service_from_draft = update_and_validate_service(service, draft.data)
     else:
-        service_from_draft = create_service_from_draft(draft, "enabled")
+        service_from_draft = create_service_from_draft(draft, "published")
 
     commit_and_archive_service(service_from_draft, update_details,
                                AuditTypes.publish_draft_service,
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -36,7 +36,8 @@ def copy_draft_service_from_existing_service(service_id):
     ).first_or_404()
 
     draft_service = DraftService.query.filter(
-        DraftService.service_id == service_id
+        DraftService.service_id == service_id,
+        DraftService.status.notin_(('not-submitted', 'submitted')),
     ).first()
     if draft_service:
         abort(400, "Draft already exists for service {}".format(service_id))
@@ -222,6 +223,11 @@ def publish_draft_service(draft_id):
         DraftService.id == draft_id
     ).first_or_404()
 
+    if draft.status == 'not-submitted':
+        abort(400, "Cannot publish a draft if it is not submitted: {}".format(draft.status))
+    if draft.status == 'submitted' and draft.service_id:
+        abort(400, "Cannot re-publish a submitted service")
+
     if draft.service_id:
         service = Service.query.filter(
             Service.service_id == draft.service_id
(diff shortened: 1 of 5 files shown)


app/main/views/drafts.py
```diff
@@ -222,15 +223,17 @@ def publish_draft_service(draft_id):
...
             Service.service_id == draft.service_id
         ).first_or_404()
 
-        service_from_draft = update_and_validate_service(
-            service,
-            draft.data)
-
+        service_from_draft = update_and_validate_service(service, draft.data)
     else:
         service_from_draft = create_service_from_draft(draft, "enabled")
```
allait: While this helps with not indexing the G7 drafts before the framework is live, that's not quite what we want. New G7 services should be "published" since we won't have an additional step to update all the service statuses, and we need an additional bit of logic that prevents them from being sent to search API when promoted from submitted drafts

robyoung: That is a good call, also, the endpoint clearly says "publish".

app/main/views/drafts.py
```diff
@@ -222,24 +223,30 @@ def publish_draft_service(draft_id):
...
     try:
-        db.session.delete(draft)
+        if draft.status == 'submitted':
+            draft.service_id = service_from_draft.service_id
+            db.session.add(draft)
+        else:
+            db.session.delete(draft)
         db.session.commit()
     except IntegrityError as e:
         db.session.rollback()
```
TheDoubleK: The error messaging in this catch is no longer quite right - it could now also be the update of service id that failed.

app/main/views/services.py
```diff
@@ -189,12 +189,7 @@ def get_service(service_id):
         Service.service_id == service_id
     ).first_or_404()
 
-    if service.framework.status == 'live':
-        return jsonify(services=service.serialize())
-    elif service.framework.status == 'expired':
-        abort(410)
-    else:
-        abort(404)
+    return jsonify(services=service.serialize())
```
TheDoubleK: Frontend apps will need to change to check statuses - are you on that already? (I'm particularly thinking of buyer app service pages, but supplier app might also use this endpoint in the turned-off edit live services flow.)

tests/app/views/test_drafts.py
```diff
@@ -808,33 +886,9 @@ def test_service_id_collisions_should_be_handled(self,
...
-        assert_equal(res.status_code, 200)
-        with assert_raises(IntegrityError):
-            res = self.publish_new_draft_service()
-
-        with self.app.app_context():
-            db.session.rollback()
-            assert_equal(Service.query.count(), 2)
-            assert_equal(DraftService.query.count(), 1)
+        # Count is 3 because we create one in the setup
+        assert self.service_count() == 3
```
TheDoubleK: Is it also worth testing that the assigned service IDs are the two that we expect from the array?

TheDoubleK:
Reviewed - couple of very minor comments but otherwise looks good.

app/main/views/services.py
robyoung: This only applies to the buyer frontend where [the change](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/186) has already been merged. We have to remember that that must be deployed before this one.

TheDoubleK: Cool, sorry - I completely missed that PR go through.

tests/app/views/test_drafts.py
robyoung: I've added a check in for this now.

app/main/views/drafts.py
robyoung: I've updated this.

TheDoubleK:
Nice one :+1:

PR #293, opened 2015-11-13
Refactor error handling

robyoung:
The bug this fixes is that we didn't have a handler for 409. I've
refactored how we do error handling so that it's generic for all HTTP
error codes.

### DIFF
--- a/app/main/errors.py
+++ b/app/main/errors.py
@@ -9,31 +9,17 @@ def validatation_error(e):
     return jsonify(error=e.message), 400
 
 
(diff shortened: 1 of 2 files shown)


app/main/errors.py
```diff
@@ -2,6 +2,7 @@
 
 from . import main
 from ..models import ValidationError
+from werkzeug.exceptions import HTTPException
```
allait: Don't need this

allait:
:+1:

PR #294, opened 2015-11-16
Add data migration for dos entries in lots table

carolinegreen:
Add data migration so can be certain lots are consistant through all the enivironments.

This has been rebased on top of https://github.com/alphagov/digitalmarketplace-api/pull/295 because both migrations need to go in together.

### DIFF
--- a/tests/app/views/test_suppliers.py
+++ b/tests/app/views/test_suppliers.py
@@ -1146,9 +1146,7 @@ class TestRegisterFrameworkInterest(BaseApplicationTest):
         super(TestRegisterFrameworkInterest, self).setup()
 
         with self.app.app_context():
-            framework_enum_vals = db.session.execute("SELECT enum_range(NULL::framework_enum);").first()[0]
-            if 'dos' not in str(framework_enum_vals):
-                self.bootstrap_dos()
+            self.set_framework_status('digital-outcomes-and-specialists', 'open')
 
             db.session.add(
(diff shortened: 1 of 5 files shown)


quis:
This should probably go in after https://github.com/alphagov/digitalmarketplace-api/pull/295, with the lot-framework relationship defined.

quis:
Tested locally and does the job. @carolinegreen @robyoung If you're happy with mine and @TheDoubleK’s earlier commits then this is :+1:.

tests/app/views/test_suppliers.py
```diff
@@ -1257,9 +1255,7 @@ def setup(self):
         self.setup_dummy_suppliers(1)
 
         with self.app.app_context():
-            framework_enum_vals = db.session.execute("SELECT enum_range(NULL::framework_enum);").first()[0]
-            if 'dos' not in str(framework_enum_vals):
-                self.bootstrap_dos()
```
pcraig3: This is great, getting rid of this.

pcraig3:
I think after removing the enum check in those 3 tests, you can also remove the [`bootstrap_dos()`](https://github.com/alphagov/digitalmarketplace-api/blob/b597eee092f2ede31173ffffb5cd8db9e7f3c2b8/tests/app/helpers.py#L184) method entirely now.

TheDoubleK:
Yep we should definitely be able to :fire: :fire: :fire: `bootstrap_dos()` from the tests everywhere.

PR #295, opened 2015-11-16
Add Digital Outcomes and Specialists to frameworks

quis:
Depends on:
- [x] https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/187 being merged **and** deployed through to production

--

This commit checks to see if:
- the framework exists in the enum before trying to add it
- the framework exists in the table before trying to add

This means that it won’t fall over on environments where DOS has already been created (eg local dev machines, preview).

The downgrade will fail if data associated with a framework has been created (eg a supplier has registered interest). We made this trade off because:
- can rollback production if anything goes wrong if we deploy this
- no framework-related data will be created on production until the framework is open—by which time it will be way too late to rollback this far

_Paired with @TheDoubleK_

### DIFF
--- a/migrations/versions/420_dos_is_coming.py
+++ b/migrations/versions/420_dos_is_coming.py
@@ -1,0 +1,34 @@
+"""DOS is coming
+
+Revision ID: 420
+Revises: 410_remove_empty_drafts
+Create Date: 2015-11-16 14:10:35.814066
+
+"""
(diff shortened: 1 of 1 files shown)


quis:
Closing in favour of https://github.com/alphagov/digitalmarketplace-api/pull/294

PR #297, opened 2015-11-23
Add validation errors to get draft response

robyoung:
https://www.pivotaltracker.com/story/show/108748808

Add a flag to the response when getting a single draft service showing whether the entire service is valid or not. This is useful for knowing when to show the 'mark as complete' button on the supplier frontend.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -12,7 +12,8 @@ from ...service_utils import (
     validate_and_return_updater_request,
     update_and_validate_service, index_service,
     commit_and_archive_service, create_service_from_draft,
-    validate_and_return_related_objects, validate_service_data
+    validate_and_return_related_objects, validate_service_data,
+    get_service_validation_errors
 )
 
 from ...draft_utils import (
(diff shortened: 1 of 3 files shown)


app/main/views/drafts.py
```diff
@@ -170,7 +171,8 @@ def fetch_draft_service(draft_id):
 
     return jsonify(
         services=draft.serialize(),
-        auditEvents=last_audit_event.serialize(include_user=True)
+        auditEvents=last_audit_event.serialize(include_user=True),
+        isValid=is_service_data_valid(draft)
```
allait: Can we return the validation errors instead/in addition to this? Frontend can display two separate messages instead of the "Mark as complete" button: about missing answers and about specialists/outcomes, it would be nice if we could eventually derive both from the API response

app/service_utils.py
```diff
@@ -73,6 +77,14 @@ def validate_service_data(service, enforce_required=True, required_fields=None):
     return
 
 
+def is_service_data_valid(service):
+    return len(get_validation_errors(
```
allait: `not get_validation_errors(...)`?

app/main/views/drafts.py
robyoung: Yes, really good point, done.

allait:
:+1:

PR #298, opened 2015-11-23
Don't allow copying a one_service_limit draft

allait:
For lots that can only have one service per supplier there shouldn't
be a way to create additional services, so we return an error if the
copy draft endpoint was called.

Since there's no "Make a copy" button displayed for one-service lots
this requires no changes in the frontend apps.

### Add explicit order_by to Framework.lots
Lots should always be ordered by id since they could be used for
display.


pcraig3:
Not sure what the protocol is about coveralls, but it seems okay to me. :+1:

PR #300, opened 2015-11-25
Add framework.clarification_questions_open flag

allait:
Part of [#107995312](https://www.pivotaltracker.com/story/show/107995312)

Flag tels supplier frontend if clarification questions are open for
the framework. Can be updated at the same time as the framework status.

Migration sets the flag to 'false' for all existing non-open frameworks
and to 'true' for open frameworks.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -46,7 +46,8 @@ class Lot(db.Model):
             'id': self.id,
             'slug': self.slug,
             'name': self.name,
-            'one_service_limit': self.one_service_limit,
+            'one_service_limit': self.one_service_limit,  # TODO deprecated
+            'oneServiceLimit': self.one_service_limit,
         }
 
 
@@ -86,7 +87,7 @@ class Framework(db.Model):
             'slug': self.slug,
             'framework': self.framework,
             'status': self.status,
-            'clarification_questions_open': self.clarification_questions_open,
+            'clarificationQuestionsOpen': self.clarification_questions_open,
             'lots': [lot.serialize() for lot in self.lots],
         }
 
--- a/app/models.py
+++ b/app/models.py
@@ -65,6 +65,7 @@ class Framework(db.Model):
     status = db.Column(db.String(),
                        index=True, nullable=False,
                        default='pending')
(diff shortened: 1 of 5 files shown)


quis:
This is a good idea.

:+1:

app/models.py
```diff
@@ -86,6 +87,7 @@ def serialize(self):
             'slug': self.slug,
             'framework': self.framework,
             'status': self.status,
+            'clarification_questions_open': self.clarification_questions_open,
```
TheDoubleK: Other serializations use camelCase for keys - e.g. `frameworkSlug`, `contactName`, `phoneNumber`, `dunsNumber`, etc.

TheDoubleK: Although we do have `one_service_limit` I think the rest are camelCase. Might be nice to fix that one up to be the same too.

migrations/versions/440_add_framework_clarification_questions_.py
```diff
@@ -0,0 +1,30 @@
...
+down_revision = '430'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('frameworks', sa.Column('clarification_questions_open', sa.Boolean(), nullable=True))
+    op.execute("""
+        UPDATE frameworks SET clarification_questions_open = (CASE status
```
TheDoubleK: I've never seen `CASE` used before. Hardcore.

app/models.py
allait: Good call. I've renamed both, but keeping `one_service_limit` too for now since it's used by the supplier frontend.

migrations/versions/440_add_framework_clarification_questions_.py
tommorris: It has plenty of uses.

robyoung:
Is there a PR for changing whatever uses `one_service_limit`?

allait:
@robyoung I'll add it to the clarification questions PR on the supplier frontend

robyoung:
Needs a rebase but looks good :+1:

PR #301, opened 2015-11-25
[#107188798] Remove keys from draft services if updated values are None

TheDoubleK:
Required as part of this story: https://www.pivotaltracker.com/story/show/107188798

At the moment there is no way to remove keys from draft services.  This adds a step where keys that have their value set to `None` are removed from the draft service.

This is required to allow removal of specialist roles, etc. from DOS draft services.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -86,6 +86,7 @@ def edit_draft_service(draft_id):
     ).first_or_404()
 
     draft.update_from_json(update_json)
+    draft.purge_nulls()
     validate_service_data(draft, enforce_required=False, required_fields=page_questions)
 
     audit = AuditEvent(
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -86,7 +86,6 @@ def edit_draft_service(draft_id):
     ).first_or_404()
 
     draft.update_from_json(update_json)
-    draft.purge_nulls()
     validate_service_data(draft, enforce_required=False, required_fields=page_questions)
 
     audit = AuditEvent(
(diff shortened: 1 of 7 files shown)


app/main/views/drafts.py
```diff
@@ -86,6 +86,7 @@ def edit_draft_service(draft_id):
     ).first_or_404()
 
     draft.update_from_json(update_json)
+    draft.purge_nulls()
```
allait: I think this could be part of `.data` validator. Both require us to disallow nulls as field values for service data, so I think we should make it an explicit rule ("null key are removed from data") rather than a delete workaround ("to delete set the value to null"). Making sure nulls are never written to the database seems better than removing them on first update too.

app/models.py
```diff
@@ -511,6 +511,9 @@ def update_from_json(self, data):
         self.data = current_data
         self.updated_at = datetime.utcnow()
 
+    def purge_nulls(self):
+        self.data = dict((k, v) for k, v in iteritems(self.data) if v is not None)
```
allait: `for k, v in self.data.items()` would've worked here too since we're not too fussed about laziness on python2

tests/app/test_model.py
```diff
@@ -120,3 +120,26 @@ def test_has_statuses_should_accept_multiple_statuses(self):
...
+            'serviceName': 'Service with nulls',
+            'empty': None,
+            'serviceSummary': None,
+            'price': 'Not a lot'
+        }
+        same_service_without_nulls = {
+            'serviceName': 'Service with nulls',
+            'price': 'Not a lot'
+        }
+        with self.app.app_context():
```
allait: I don't think this test needs a persisted service in the database – we could do the same thing with ad-hoc `Service(data=service_with_nulls)` objects, which means there's no need to set up suppliers or any other database records

TheDoubleK:
Yep I think it's better here.

allait:
:+1:

PR #302, opened 2015-11-25
[#107188910] Add Digital Services and Outcomes' Digital Services and Digital Outcomes schemas, make sure they validate correctly, and write some tests for them.

pcraig3:
**This pull request:**

- Adds the new schemas
- Changes how validating works in a few places
- Tests things

[>> Story in Pivotal](https://www.pivotaltracker.com/story/show/107188910)

***

#### Adds the new schemas

Now that we've got what seem to be the final DOS schemas, the API can start validating against the right things :+1: 

#### Changes how validating works in a few places

1.  Process `dependencies` errors
If a key that's part of a group is included without the rest of the group, the schema will fail to validate with a default `dependencies` error message that's not very helpful. The new behaviour is to processing the error the same way we do for `required` keys because the implications are the same.
 Adding `developerLocations` without any other developer-related keys will return 
  ```json
  `{ "error": { "developerPriceMin": "answer_required", "developerPriceMax": "answer_required" } }
  ```

2. Process `anyOf` errors
Similarly to the last thing, if a schema fails to validate because none of the `anyOf` groups are present (ie, we have no specialists or outcomes), a form error message is returned that our frontend will understand.  
Trying to mark as complete a digital-specialists service without any specialists will return
  ```json
  `{ "error": { "_form": "specialist_required" } }
  ```
Note that the `specialist` substring comes from [the `any_of` key in a multiquestion's manifest file](https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/digital-outcomes-and-specialists/questions/services/agileCoach.yml#L9).

3. Liberalise pricing check in `_translate_json_schema_error()`
Previously, we would previously only return a `not_money_format` if the failing key exactly matched `priceMin` or `priceMax`. 
Of course, now pricing keys in digital-specialists all look like `{specialistName}PriceMax` and `{specialistName}PriceMin`.  I've expanded the check to find all of our new pricing fields.

  - @allait pointed out that our manifest files ([for example](https://github.com/alphagov/digitalmarketplace-frameworks/blob/81a317c4d6bcf8f560e1cb1dcf9cd225801c6124/frameworks/digital-outcomes-and-specialists/questions/services/labAddressTown.yml#L14)) contain a message which describes the problem.  This arguably obviates the need for [our janky key detection](https://github.com/alphagov/digitalmarketplace-api/blob/master/app/validation.py#L232) anyway, since we could use a more general key (`incorrect_format_error`, maybe) for both cases and just return the corresponding message.

4. Drop `anyOf` fields when validating against a partial schema.
Our [`get_validator`](https://github.com/alphagov/digitalmarketplace-api/blob/master/app/validation.py#L57) function allows us to override the `required` fields with a passed-in subset of schema fields -- makes it easier to validate smaller bits of the schema at a time.  However, if we don't remove the `anyOf` keys, we're still validating against a bunch of stuff we probably don't care about.

#### More unit tests

Added more unit tests.
Specifically: 
 - :white_check_mark: Can still successfully edit a service if `page_questions` are invalid
 - :x: Cannot edit a digital-specialists draft with a bad price string
 - :x: Cannot edit a digital-specialists draft with a min_price greater than a max_price
 - :x: Cannot edit a digital-specialists draft with missing dependencies
 - :white_check_mark: Can edit a digital-specialists draft if `page_questions` are invalid or required in `anyOf`
 - :white_check_mark: Can mark as complete a valid digital-specialists draft
 - :x: Cannot mark as complete a digital-specialists draft without a specialist

:muscle: :muscle: :muscle: :muscle: :muscle: :muscle: :muscle: :muscle: :muscle: :muscle: :muscle: :muscle: :muscle: :muscle: 
screen shot 2015-11-25 at 19 13 00

### DIFF
--- a/app/validation.py
+++ b/app/validation.py
@@ -237,9 +237,8 @@ def _translate_json_schema_error(key, validator, validator_value, message):
     elif validator == 'pattern':
         # Since error messages are now specified in the manifests, we can (in the future) generalise the returned
         # string and just show the correct message
-        for price_str in ['priceMin', 'priceMax', 'PriceMin', 'PriceMax']:
-            if price_str in key:
-                return 'not_money_format'
+        if key.endswith(('priceMin', 'priceMax', 'PriceMin', 'PriceMax')):
+            return 'not_money_format'
 
         return 'under_{}_words'.format(_get_word_count(validator_value))
 
--- a/app/validation.py
+++ b/app/validation.py
@@ -141,6 +141,9 @@ def get_validation_errors(validator_name, json_data,
         elif error.validator == 'required':
             key = re.search(r'\'(.*)\'', error.message).group(1)
             error_map[key] = 'answer_required'
+        elif error.validator == 'anyOf':
+            if error.validator_value[0].get('title'):
+                form_errors.append('{}_required'.format(error.validator_value[0].get('title')))
         else:
             form_errors.append(error.message)
     if form_errors:
--- a/app/validation.py
+++ b/app/validation.py
@@ -139,7 +139,7 @@ def get_validation_errors(validator_name, json_data,
                 key, error.validator, error.validator_value, error.message
             )
         elif error.validator in ['required', 'dependencies']:
-            regex = r'\'(\w+)\' is a dependency of \'(\w+)\'' if error.validator == 'dependencies' else r'\'(.*)\''
+            regex = r'u?\'(\w+)\' is a dependency of u?\'(\w+)\'' if error.validator == 'dependencies' else r'\'(.*)\''
             key = re.search(regex, error.message).group(1)
             error_map[key] = 'answer_required'
         elif error.validator == 'anyOf':
--- a/app/validation.py
+++ b/app/validation.py
@@ -138,8 +138,9 @@ def get_validation_errors(validator_name, json_data,
             error_map[key] = _translate_json_schema_error(
                 key, error.validator, error.validator_value, error.message
             )
-        elif error.validator == 'required':
-            key = re.search(r'\'(.*)\'', error.message).group(1)
+        elif error.validator in ['required', 'dependencies']:
+            regex = r'\'(\w+)\' is a dependency of \'(\w+)\'' if error.validator == 'dependencies' else r'\'(.*)\''
+            key = re.search(regex, error.message).group(1)
             error_map[key] = 'answer_required'
         elif error.validator == 'anyOf':
             if error.validator_value[0].get('title'):
--- a/app/validation.py
+++ b/app/validation.py
@@ -64,7 +64,8 @@ def get_validator(schema_name, enforce_required=True, required_fields=None):
         schema['required'] = [
             field for field in schema.get('required', [])
             if field in required_fields
-            ]
+        ]
+        schema.pop('anyOf', None)
     return validator_for(schema)(schema, format_checker=FORMAT_CHECKER)
 
 
--- a/app/validation.py
+++ b/app/validation.py
@@ -155,10 +155,26 @@ def get_validation_errors(validator_name, json_data,
 
 
 def min_price_less_than_max_price(error_map, json_data):
-    if 'priceMin' in json_data and json_data.get('priceMax'):
-        if 'priceMin' not in error_map and 'priceMax' not in error_map:
-            if Decimal(json_data['priceMin']) > Decimal(json_data['priceMax']):
-                return {'priceMax': 'max_less_than_min'}
+
+    def return_min_and_max_price_keys(json_data):
+        if 'priceMin' in json_data:
+            return 'priceMin', 'priceMax'
+
+        specialist_price_keys = [k for k in json_data.keys() if k.endswith(('PriceMin', 'PriceMax'))]
+        if specialist_price_keys:
+            return \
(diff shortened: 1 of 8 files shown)


app/validation.py
```diff
@@ -229,10 +235,13 @@ def _translate_json_schema_error(key, validator, validator_value, message):
             return 'answer_required'
 
     elif validator == 'pattern':
-        if key in ['priceMin', 'priceMax']:
-            return 'not_money_format'
-        else:
-            return 'under_{}_words'.format(_get_word_count(validator_value))
+        # Since error messages are now specified in the manifests, we can (in the future) generalise the returned
+        # string and just show the correct message
+        for price_str in ['priceMin', 'priceMax', 'PriceMin', 'PriceMax']:
```
robyoung: This needs to be `endswith` to catch the DOS ones.

robyoung:
Also, `min_price_less_than_max_price` needs to be updated to support DOS prices.

app/validation.py
```diff
@@ -60,11 +61,12 @@ def get_validator(schema_name, enforce_required=True, required_fields=None):
     if enforce_required:
         schema = _SCHEMAS[schema_name]
     else:
-        schema = _SCHEMAS[schema_name].copy()
+        schema = copy.deepcopy(_SCHEMAS[schema_name])
```
allait: I'm guessing this was left from the `anyOf` modification commit. I think shallow `.copy()` should be still ok here.

app/validation.py
allait: That's what I thought at first too, but the next line is the key check, this is just a list of substrings we're checking for

app/validation.py
```diff
@@ -136,9 +138,13 @@ def get_validation_errors(validator_name, json_data,
             error_map[key] = _translate_json_schema_error(
                 key, error.validator, error.validator_value, error.message
             )
-        elif error.validator == 'required':
-            key = re.search(r'\'(.*)\'', error.message).group(1)
+        elif error.validator in ['required', 'dependencies']:
+            regex = r'u?\'(\w+)\' is a dependency of u?\'(\w+)\'' if error.validator == 'dependencies' else r'\'(.*)\''
```
allait: <sub>Obligatory comment on json schema errors being unhelpful</sub>
Do we get an error message for each of the missing keys? E.g. if we have `["agileCoachMinPrice", "agileCoachMaxPrice"]` in the dependency and both are missing – will we get `answer_required` error for both or only for one of them?

pcraig3: **tl:dr** Yes

Basically, you get two errors in both cases. 

Scenario 1:
```
# keys which depend on each other 
`["agileCoachPriceMin", "agileCoachPriceMax", "agileCoachLocations"]`

# submitted json
`{"agileCoachPriceMin": "100"}`

# unmodified error messages
errors = [{'message': "'agileCoachPriceMax' is a dependency of 'agileCoachPriceMin'"}, {'message': "'agileCoachLocations' is a dependency of 'agileCoachPriceMin'"}]
```

Scenario 2: 
```
# keys which depend on each other 
["agileCoachPriceMin", "agileCoachPriceMax", "agileCoachLocations"]

# submitted json
{"agileCoachPriceMin": "100", "agileCoachPriceMax": "101"}

# unmodified error messages
errors = [{'message': "'agileCoachLocations' is a dependency of 'agileCoachPriceMin'"}, {'message': "'agileCoachLocations' is a dependency of 'agileCoachPriceMax'"}]
```

Since we're just assigning errors to a dictionary and using the service key name, a second assignment to an already-existing key happens in some cases, but the practical result is the same. ie, in Scenario 2, `agileCoachLocations` gets overwritten with the same value, which seems fine.

app/validation.py
pcraig3: Is there any reason to replace it other than that we can? 
It still might cause the same kind of problem in the future.

allait: `deepcopy` is slower since it has to do quite a bit more work, and the intent is not as clear when we're only modifying the top-level. But other than that it should work, so we can keep it if you think we're likely to need it in the future.

app/validation.py
```diff
@@ -149,10 +155,26 @@ def get_validation_errors(validator_name, json_data,
...
-    if 'priceMin' in json_data and json_data.get('priceMax'):
-        if 'priceMin' not in error_map and 'priceMax' not in error_map:
-            if Decimal(json_data['priceMin']) > Decimal(json_data['priceMax']):
-                return {'priceMax': 'max_less_than_min'}
+
+    def return_min_and_max_price_keys(json_data):
+        if 'priceMin' in json_data:
+            return 'priceMin', 'priceMax'
+
+        specialist_price_keys = [k for k in json_data.keys() if k.endswith(('PriceMin', 'PriceMax'))]
```
allait: Neat, I didn't know 'endswith' accepts tuples. I think we could use this for the `not_money_format` check too.

robyoung:
:+1:

app/validation.py
pcraig3: That's a good point.

app/validation.py
pcraig3: I think keep it. It doesn't seem to me like a huge tradeoff.

PR #303, opened 2015-11-26
Update stats endpoint for DOS

robyoung:
Registering interest in a framework is now defined by adding a SupplierFramework record. This change removes the join the audit table and replaces it with a join to the supplier frameworks table.


TheDoubleK:
Looks good to me once it's had a rebase :+1:

PR #304, opened 2015-11-27
Validate complete draft on update if already marked as complete

TheDoubleK:
If changes are made to a draft that has already been marked as complete then the complete draft should still be valid after the update.


allait:
Would be nice to have a test for this, since it's behaviour we're going to rely on quite a bit and `enforce_required` and `required_fields` interaction is not obvious.

allait:
:+1:

PR #305, opened 2015-12-01
Update schemas for DOS

carolinegreen:
Update schemas for both DOS specialists and DOS labs.  Joint commit with @carolinegreen and @pcraig3

### DIFF
--- a/json_schemas/services-digital-outcomes-and-specialists-digital-specialists.json
+++ b/json_schemas/services-digital-outcomes-and-specialists-digital-specialists.json
@@ -92,6 +92,14 @@
     },
     {
       "required": [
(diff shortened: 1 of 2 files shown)


pcraig3:
:+1:

carolinegreen:
Not sure you can thumbs up a joint commit you're part of?

TheDoubleK:
:+1: 

With a caveat that we will need to tell all testers that they need to delete their Labs lot and start again, as any saved labs will have a max price in and so won't validate.

pcraig3:
Ooh, good point.  I'll remove them from preview/staging.

TheDoubleK:
Remove what - labs for all users?

PR #306, opened 2015-12-01
Fix `min_price_less_than_max_price`

TheDoubleK:
This method wasn't working as expected.
It will now return all errors where max price is less than min price for price fields for the same thing.

### DIFF
--- a/app/validation.py
+++ b/app/validation.py
@@ -159,9 +159,8 @@ def min_price_less_than_max_price(error_map, json_data):
     for key in json_data.keys():
         if key.lower().endswith('pricemin'):
             prefix = key[:-8]
-            max_price_key = prefix + ('PriceMax' if prefix else 'priceMax')
+            max_price_key = (prefix + 'PriceMax') if prefix else 'priceMax'
             if (
-                max_price_key in json_data and
                 json_data.get(max_price_key, None) and
                 key not in error_map and
                 max_price_key not in error_map
--- a/app/validation.py
+++ b/app/validation.py
@@ -155,27 +155,21 @@ def get_validation_errors(validator_name, json_data,
 
 
 def min_price_less_than_max_price(error_map, json_data):
+    price_errors = {}
+    for key in json_data.keys():
+        if key.lower().endswith('pricemin'):
+            prefix = key[:-8]
+            max_price_key = prefix + ('PriceMax' if prefix else 'priceMax')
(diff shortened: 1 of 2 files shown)


app/validation.py
```diff
@@ -155,27 +155,21 @@ def get_validation_errors(validator_name, json_data,
...
-        if min_price_key in json_data and json_data.get(max_price_key):
-            if min_price_key not in error_map and max_price_key not in error_map:
-                if Decimal(json_data[min_price_key]) > Decimal(json_data[max_price_key]):
-                    return {max_price_key: 'max_less_than_min'}
-    return {}
+    price_errors = {}
+    for key in json_data.keys():
+        if key.lower().endswith('pricemin'):
+            prefix = key[:-8]
+            max_price_key = prefix + ('PriceMax' if prefix else 'priceMax')
```
allait: Isn't `(prefix + 'PriceMax') if prefix else 'priceMax'` what we actually want? The result is the same, but it makes the intent clearer.

app/validation.py
```diff
@@ -155,27 +155,21 @@ def get_validation_errors(validator_name, json_data,
...
+            if (
+                max_price_key in json_data and
+                json_data.get(max_price_key, None) and
```
allait: We don't need `max_price_key in json_data` check since we're doing a `.get` right after

allait:
I think we need a test for this with multiple price pairs, to replicate the problem we had initially

app/validation.py
TheDoubleK: Yep that's nicer.

TheDoubleK:
Tests added, comments addressed, rebased up to date.

tests/test_validation.py
```diff
@@ -444,6 +444,50 @@ def test_max_price_larger_than_min_price():
...
+        'agileCoachPriceMin': '200',
+        'agileCoachPriceMax': '250',
+        'developerPriceMin': '200',
+        'developerPriceMax': '25',
+        'designerPriceMin': '300',
+        'designerPriceMax': '299.99',
+        'windowCleanerPriceMin': '12.50',
+        'windowCleanerPriceMax': '300',
+    }
+    previous_errors = {'designerPriceMax': 'non_curly_quotes'}
```
allait: :joy:

allait:
:+1:

PR #307, opened 2015-12-02
Update dos schemas with maximum price string length

allait:
Limits the max and min price strings to 15 characters before the
decimal point.

Related to alphagov/digitalmarketplace-frameworks#153


TheDoubleK:
Looks good once Travis is happy :+1:

PR #309, opened 2015-12-07
Fix stats

robyoung:
## Make declaration default to an empty dict
When we query declarations for stats we need to be able to look inside them. PostgreSQL complains if you try to query into a null value.

## Make stats handle null declarations
There was a bug fixed in the previous commit that would result in JSON 'null' values for the supplier declaration which would in turn cause the stats queries to fail. As this is a subtle bug and the stats are depended on they need to be pretty robust. This change excludes JSON 'null' values from the query.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -1,7 +1,7 @@
 from flask import jsonify, abort, request
 from sqlalchemy.types import String
 from sqlalchemy.exc import IntegrityError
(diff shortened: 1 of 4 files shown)


TheDoubleK:
"Too many blank lines"

TheDoubleK:
Other than that the code looks good and tests pass locally - :+1: once PEP8 is appeased.

PR #310, opened 2015-12-09
List audit events for a user

robyoung:
Allow audit events for a given user to be listed through the API.
Currently the only way to view these through the API is to list all
audit events and filter them client side.


allait:
:+1:

PR #312, opened 2015-12-18
Create route that returns information about users, suppliers, and their frameworks

pcraig3:
Added a new route to export a list of users for suppliers who have (at least) shown interest in a framework.
User/Supplier data is returned, as well as information about whether a supplier has applied to the framework or returned their framework agreement.
This will be used by administrators who are currently relying on developers generating csv files by running various scripts.

Writing tests for the user exporting route was pretty involved because in order to test the various states, we have to insert quite a bit of information into the database, but they're done now, so that's nice.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/107188322)

***

### Sample output

```json
{
  "users": [
  {
      "application_result": "pass",
      "application_status": "application",
      "declaration_status": "complete",
      "framework_agreement": true,
      "supplier_id": 10000,
      "user_email": "supplier-10000@user.com",
      "user_name": "DM Test Supplier"
    },
    {
      "application_result": "fail",
      "application_status": "no_application",
      "declaration_status": "complete",
      "framework_agreement": false,
      "supplier_id": 10101,
      "user_email": "supplier-10101@user.com",
      "user_name": "1Cyber Group Ltd"
    }
  ]
}
```

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -228,15 +228,9 @@ def export_users_for_framework(framework_slug):
     supplier_frameworks_and_suppliers_and_users = db.session.query(
         SupplierFramework, Supplier, User
     ).join(
-        Framework
+        Supplier, User, Framework
     ).filter(
         Framework.slug == framework_slug
-    ).filter(
-        SupplierFramework.framework_id == Framework.id
-    ).filter(
-        SupplierFramework.supplier_id == Supplier.supplier_id
-    ).filter(
-        User.supplier_id == Supplier.supplier_id
     ).all()
 
     submitted_draft_counts_per_supplier = {}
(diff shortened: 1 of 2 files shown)


app/main/views/users.py
```diff
@@ -213,6 +214,77 @@ def update_user(user_id):
...
+        SupplierFramework, Supplier, User
+    ).join(
+        Framework
+    ).filter(
+        Framework.slug == framework_slug
+    ).filter(
+        SupplierFramework.framework_id == Framework.id
+    ).filter(
+        SupplierFramework.supplier_id == Supplier.supplier_id
+    ).filter(
```
robyoung: These multiple filters become quite verbose to read. These can just be multiple arguments to a single `filter` call (see `AND` in http://docs.sqlalchemy.org/en/latest/orm/tutorial.html#common-filter-operators)

app/main/views/users.py
```diff
@@ -213,6 +214,77 @@ def update_user(user_id):
...
+    if not framework:
+        abort(400, 'invalid framework')
+
+    if framework.status == 'coming':
+        abort(400, 'framework not yet open')
+
+    supplier_frameworks_and_suppliers_and_users = db.session.query(
+        SupplierFramework, Supplier, User
+    ).join(
+        Framework
```
robyoung: Why do we need `Framework` in a separate `join` cal here?

app/main/views/users.py
robyoung: Also, I don't think we need to list all the join conditions, SQLAlchemy should already know them.

pcraig3: For posterity, what we found was that using a `.join()` function gives you an inner join while just listing model names in the `query()` argument returns the cartesian product.

robyoung:
:+1:

PR #314, opened 2015-12-31
Add detail to service status updates

tombye:
We will need to show some information on service status changes for those that have been made unavailable.

Story: https://www.pivotaltracker.com/story/show/110695856

This should let us do page state like this:

status_change_banner

We already add some audit data to requests for drafts:

https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/drafts.py#L163

It was proposed we follow this pattern for services, restricting this only to where needed (services made unavailable).

### DIFF
--- a/tests/app/views/test_services.py
+++ b/tests/app/views/test_services.py
@@ -1754,33 +1754,58 @@ class TestGetService(BaseApplicationTest):
         assert_equal(data['services']['frameworkName'], u'G-Cloud 6')
         assert_equal(data['services']['frameworkStatus'], u'live')
 
-    def test_get_service_returns_last_status_update_audit_if_none(self):
-        response = self.client.get('/services/123-published-456')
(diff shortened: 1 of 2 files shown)


robyoung:
:+1:

tests/app/views/test_services.py
```diff
@@ -1751,3 +1753,70 @@ def test_get_service_returns_framework_info(self):
...
+                    "serviceId": "123-disabled-456",
+                    "old_status": "published",
+                    "oldArchivedServiceId": 1
+                }
+            )
+            db.session.add(audit_event)
+            db.session.commit()
+        response = self.client.get('/services/123-disabled-456')
+        data = json.loads(response.get_data())
+        assert_equal(data['statusUpdateAuditEvent']['type'], 'update_service_status')
```
allait: Maybe we should check for the audit event date here, since it's the part the buyer frontend expects

PR #315, opened 2016-01-04
Add option to have audit events return newest-first

quis:
This makes more sense for the front end. By default it is most useful to see the latest events, and then paginating to get to older ones.

Otherwise you could reload the page for today and might not even notice that new events had been added to end since you last looked (especially if they got paginated onto the next page).

### DIFF
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -1,7 +1,7 @@
 from flask import jsonify, abort, request, current_app
 from datetime import datetime
 from ...models import AuditEvent
-from sqlalchemy import asc, Date, cast
+from sqlalchemy import asc, desc, Date, cast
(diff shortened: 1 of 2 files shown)


allait:
I think this affects the admin app service change history (which should be fine with reversed order) and the stats page (which we might need to update)

quis:
@allait Made it optional because it broke a bunch of the tests, so it shouldn’t affect any existing stuff now.

allait:
@quis options seems like a good way to go, I'm not sure about the name though. How about `latest_first=true` using `convert_to_boolean` for value?

quis:
@allait sounds reasonable, have made the change and squashed into 5cf60ee

allait:
:+1:

PR #316, opened 2016-01-05
Update dmutils and add dmapiclient package dependency

allait:
Fixes imports to use the new dmapiclient package instead of
dmutils.apiclient and dmutils.audit.


quis:
When the tests pass, :ship:

quis:
:+1:

PR #317, opened 2016-01-05
Remove unused and insecure requirements

allait:
* Removes unused `requirements_for_build.txt`
* Removes json-logger from requirements.txt since the package is
  now in PyPi and is installed with dmutils
* Upgrades insecure version of `requests` used by the scripts


TheDoubleK:
:+1:

PR #318, opened 2016-01-06
Only show active users in export

robyoung:
When generating an email list we noticed that it was bigger than the list generated by the original script. This is due to user's active flag not being checked and therefore disabled users being included in the list.

This change also moves the `test_users` module under the same package that all the other views tests live. It's probably easier to review the two commits separately.


TheDoubleK:
:+1:

TheDoubleK:
We should do another sanity check of script vs. admin app lists once this is deployed.

PR #319, opened 2016-01-07
Upgrade Flask-Bcrypt

robyoung:
Internally this upgrades from python-bcrypt 0.3.1 to bcrypt 2.0.0

Hashed passwords are now returned as bytes so we need to decode to get
strings.


allait:
:+1:

PR #321, opened 2016-01-08
Add additional indexes for audit_events table

allait:
### Add additional indexes for audit_events table
We've seen some performance issues leading to request timeouts
on preview mostly related to the audit events table queries.

Slowest queries we've observed were:
* "Find last audit event for object" queries - were using the
  created_at index scan with filters that can take a long time
  for objects that were last updated a long time ago (query times
  up to 60s on preview).
* Count audit events with given type and acknowledged status query.
  Executed by flask-sqlalchemy `.paginate` method to get the total
  count. The query was using a Bitmap Heap Scan to check the
  acknowledged status after a type index scan, which slows down as
  table size increases.

Other common queries triggered by `/audit-events` endpoint are mostly
using the `created_at` index due to pagination limit.

### Move draft and services 'last event' query to model query class
Since we've noticed some performance issues with the 'last event'
queries, moving them to a single location to make sure they're
using the same order and filter conditions that can be covered
by a single database index.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -2,7 +2,7 @@ from dmapiclient.audit import AuditTypes
 from flask import jsonify, abort, request, current_app
 
 from sqlalchemy.exc import IntegrityError
-from sqlalchemy import asc, desc
+from sqlalchemy import asc
 
 from .. import main
 from ... import db, isolation_level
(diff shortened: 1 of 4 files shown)


robyoung:
:+1:

PR #322, opened 2016-01-11
Add framework expired to service status audit events

tombye:
An extension to the functionality added in https://github.com/alphagov/digitalmarketplace-api/pull/314.

Services on expired frameworks should also return an audit event for the status change, just one triggered by the framework status change instead of that for the service.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -3,7 +3,7 @@ from dmapiclient.audit import AuditTypes
 from flask import jsonify, abort, request, current_app
 
 from .. import main
-from ...models import ArchivedService, Service, Supplier, AuditEvent
+from ...models import ArchivedService, Service, Supplier, AuditEvent, Framework
 
 from sqlalchemy import asc
 from ...validation import is_valid_service_id_or_400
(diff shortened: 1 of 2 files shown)


robyoung:
This is starting to feel a bit weird. On an endpoint for the `Service` model we're returning an audit event that is not for that `Service`.

I wondering if something can be done to how it's returned to make it clear that it's not necessarily the status of the service, it's more just the reason why this service is considered gone.

tombye:
I wonder if it's just the naming (which was defined when I thought it was just associated with service status changes)?

What about if `status_update_audit_event` was renamed `service_unavailability_audit_event`?

robyoung:
Yea, I think it is just about the naming, would `service_unavailability_reason` give enough information, or do you think it should specifically say it's an audit event?

tombye:
I quite like the explicitness of saying it's an audit event. Means you know what to expect from its `data` property (which changes depending on it being an event for a framework or service).

tombye:
@robyoung I've given it a try with a slightly different name. See what you think about how it reads now.

robyoung:
Yea, I think anything sounds strange now. I'm good with this :+1:

PR #324, opened 2016-01-13
[#111613066] Fix audit events object-id and object-type filtering

allait:
List audit events endpoint used a join between the audit event
table and the `object-type` model, but didn't check the audit event
`object_type`. This means that eg an event with `object_id=1` could
be joined to Services, Suppliers or Frameworks record depending on
the `object-type` argument, which lead to incorrect audit events
being returned by the endpoint.

Instead of a join query we request the referenced object separately
and then filter the audit events based on that.

### DIFF
--- a/tests/app/views/test_audits.py
+++ b/tests/app/views/test_audits.py
@@ -116,14 +116,13 @@ class TestAuditEvents(BaseApplicationTest):
         assert_equal(len(data['auditEvents']), 1)
         assert_equal(data['auditEvents'][0]['user'], 'rob')
 
-    def test_get_audit_event_for_missing_object_returns_no_events(self):
+    def test_get_audit_event_for_missing_object_returns_404(self):
         self.add_audit_events_with_db_object()
 
         response = self.client.get('/audit-events?object-type=suppliers&object-id=100000')
         data = json.loads(response.get_data())
 
-        assert_equal(response.status_code, 200)
-        assert_equal(len(data['auditEvents']), 0)
+        assert_equal(response.status_code, 404)
 
     def test_should_only_get_audit_event_with_correct_object_type(self):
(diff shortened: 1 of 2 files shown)


tests/app/views/test_audits.py
```diff
@@ -116,6 +116,38 @@ def test_should_get_audit_event_by_object(self):
         assert_equal(len(data['auditEvents']), 1)
         assert_equal(data['auditEvents'][0]['user'], 'rob')
 
+    def test_should_get_audit_event_by_missing_object(self):
```
TheDoubleK: The name of this test seems odd - it's testing that there are no audit events for a missing object but this name sounds like it should get one.

tests/app/views/test_audits.py
```diff
@@ -116,6 +116,38 @@ def test_should_get_audit_event_by_object(self):
...
+
+    def test_should_only_get_audit_event_with_correct_object_type(self):
+        self.add_audit_events_with_db_object()
+
+        with self.app.app_context():
+            supplier = Supplier.query.filter(Supplier.supplier_id == 1).first()
+            event = AuditEvent(
+                audit_type=AuditTypes.supplier_update,
+                db_object=supplier,
+                user=1,
```
TheDoubleK: It isn't obvious from the code how this test works. I think the idea is that add_audit_events adds one Supplier-event, then we add another event with the sane object ID but with Service as the object type, and then check that only the original Supplier-event is returned? I'm not sure that checking the user is 'rob' is a good test for this - is user=1 definitely not 'rob' too? If so it would be nice to have it clearer in the code that the second audit event does not have 'rob' as the user.

allait:
Updated to return 404 when /audit-events referenced object is missing

TheDoubleK:
:+1:

PR #325, opened 2016-01-14
Add migration for G4 expiry

tombye:
We need audit events for each time a framework expired. G4 is the only framework without an event so this migration can be used to add it to an environment.


migrations/versions/460_add_audit_events_for_g4_and_g5_frameworks.py
```diff
@@ -0,0 +1,29 @@
...
+from alembic import op
+
+
+def upgrade():
+    # G-Cloud 4
+    op.execute("""
+        INSERT INTO audit_events
+            ("type", "created_at", "user", "data", "object_type", "object_id", "acknowledged")
+        VALUES 
+            ('framework_update', '2015-01-31T14:00:00', 'migration', '{"update": {"status": "expired"}}', 'Framework', 2, FALSE)
```
TheDoubleK: We could even have the user as `migration 460` to make it completely obvious to anyone looking at the event later on.

TheDoubleK:
Good stuff - :+1:  with or without the user name change.

tombye:
@TheDoubleK I've made that change, can you see any problems with using that string for the user name?

TheDoubleK:
Not that I can think of at all - why does it worry you?

Upgrade and downgrade both tested locally - :+1:

tombye:
Just it having a space in it, combined with my limited experience with Postgres :)

PR #326, opened 2016-01-14
Swap order of test for unavailable audit event

tombye:
If the framework a service is in has expired, the
message about it being unavailable should relate
to the framework expiring.

If the framework is not expired, the availability
of the service should be determined by the status
of the service.


robyoung:
It feels like there should be a test that covers this change.

tombye:
@robyoung good point. I've added one to make sure framework expiration trumps a service status which would make it unavailable for which audit event gets added.

robyoung:
:+1:

PR #327, opened 2016-01-15
Add 'failed' as a possible status for draft services

TheDoubleK:
For lots which include declaration questions those services could be 'failed' because of incorrect answers.


robyoung:
:+1:

PR #329, opened 2016-01-19
Fix framework stats service counting drafts twice for old declarations

allait:
Framework drafts query uses an outerjoin between Drafts and framework
declarations on supplier_id field, which creates duplicate Draft rows
for suppliers who have completed declarations for other frameworks.

Filtering out rows with incorrect SupplierFrameworks.framework_id fixes
the problem.

Modified stats test to create a single set of suppliers and two sets
of framework applications for two different frameworks to check that
data from other frameworks doesn't affect the stats.


TheDoubleK:
Nice work - how did you spot the error? Anyway, :+1:

PR #330, opened 2016-01-21
Bump digitalmarketplace api-client.

carolinegreen:
Bump for update in digitalmarketplace admin frontend using new audi events.


allait:
:+1:

PR #331, opened 2016-01-22
Add a migration to remove old imported buyer accounts

allait:
Users with a "buyer" role were imported from the Grails app and are
a mix of buyer and supplier users. We used to convert them to supplier
accounts during registration, but now that we're planning to open buyer
registration for valid buyers we're going to delete all old accounts and
ask user to re-register if needed instead of trying to validate if the
existing buyer accounts are valid buyers.

**Since the migration completely removes user data, we need to create a manual production snapshot before releasing this**

### DIFF
--- a/migrations/versions/470_remove_old_imported_buyer_accounts.py
+++ b/migrations/versions/470_remove_old_imported_buyer_accounts.py
@@ -1,0 +1,24 @@
+"""Remove old imported buyer accounts
+
+Revision ID: 470
+Revises: 460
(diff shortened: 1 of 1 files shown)


TheDoubleK:
:fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: 
Burn them :+1: 
:fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire:

PR #332, opened 2016-01-25
Prevent buyer user being created with invalid email domain

robyoung:
[#111536536](https://www.pivotaltracker.com/story/show/111536536)

Buyers must be from allowed public sector organisations. This change prevents buyer user accounts being created unless their email address is either on one of the allowed domains or on a subdomain.

Allowed domains are stored in a text file rather than handled with a JSON schema to make it easier to manage the list.


app/validation.py
```diff
@@ -32,6 +32,7 @@
     'services-update',
     'users',
     'users-auth',
+    'users-buyer-email-domain',
```
allait: This is breaking the tests since the schema file doesn't exist

app/main/views/users.py
```diff
@@ -100,6 +104,9 @@ def create_user():
     json_payload = json_payload["users"]
     validate_user_json_or_400(json_payload)
 
+    if not user_email_domain_allowed(json_payload):
```
allait: Should we do this in the model validation instead? Users are unlikely to be created anywhere outside this endpoint, but it still feels like the model attribute validation is a more reliable place for this

robyoung: Yep, good idea.

TheDoubleK:
The list of just over 100 email domains here seems very small compared to, e.g. the list of governmental bodies listed here: https://www.gov.uk/government/organisations (and that's just central gov), or even just the list of arms-length public bodies listed here: https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/372212/Annex_-_Updated_list_of_public_bodies.pdf

It feels like we may be putting a barrier to signing up in the way of a huge number of potential DOS buyers here (e.g. `housing-ombudsman.org.uk` isn't there but probably should be).

robyoung:
This is the list from DS2. I believe the idea is that the copy in the frontend will direct people towards the helpdesk where we can review and then add them to the list. We have another story in the backlog for improving how we manage this list to make that process quicker.

TheDoubleK:
:+1:

PR #333, opened 2016-01-25
Add new endpoint to update the status of a draft

TheDoubleK:
Currently we only have an endpoint to mark a service as complete, but that is the only change to `status` that can be made.

We now need to mark drafts as `failed`, so this commit adds a new endpoint hat can be used to update the status of a draft to any valid status (valid values are as listed in `DraftService.STATUSES`).

Depends on: 
 - [x] https://github.com/alphagov/digitalmarketplace-apiclient/pull/5

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -6,6 +6,7 @@ from sqlalchemy import asc
 
 from .. import main
 from ... import db, isolation_level
+from ...utils import json_only_has_required_keys
 from ...validation import is_valid_service_id_or_400
 from ...models import Service, DraftService, Supplier, AuditEvent, Framework
 from ...service_utils import (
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -9,7 +9,7 @@ SQLAlchemy==1.0.5
 SQLAlchemy-Utils==0.30.5
 
 git+https://github.com/alphagov/digitalmarketplace-utils.git@16.0.0#egg=digitalmarketplace-utils==16.0.0
-git+https://github.com/alphagov/digitalmarketplace-apiclient.git@1.2.0#egg=digitalmarketplace-apiclient==1.2.0
```
robyoung: I think this change is supposed to be on the [scripts PR](https://github.com/alphagov/digitalmarketplace-scripts/pull/28)

tests/app/views/test_drafts.py
```diff
@@ -1493,3 +1493,82 @@ def test_should_not_complete_invalid_dos_draft(self):
...
+            '/draft-services',
+            data=json.dumps(create_draft_json),
+            content_type='application/json')
+
+        self.draft = json.loads(draft.get_data())['services']
+        self.draft_id = self.draft['id']
+
+    def test_update_draft_status(self):
+        res = self.client.post(
+            '/draft-services/%s/update-status' % self.draft_id,
```
pcraig3: Why not use the `.format()` method for these?  Percentage formatting is less good, [apparently](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting).

robyoung:
:+1:

app/main/views/drafts.py
```diff
@@ -347,6 +348,44 @@ def complete_draft_service(draft_id):
     return jsonify(services=draft.serialize()), 200
 
 
+@main.route('/draft-services/<int:draft_id>/update-status', methods=['POST'])
```
allait: This is quite different from the service status update URL, while doing a similar thing. I think we should look at them and decide on a more consistent endpoint URL.

app/main/views/drafts.py
```diff
@@ -347,6 +348,44 @@ def complete_draft_service(draft_id):
...
+def update_draft_service_status(draft_id):
+    updater_json = validate_and_return_updater_request()
+    update_json = validate_and_return_draft_request()
+    json_only_has_required_keys(update_json, ['status'])
+
+    new_status = update_json['status']
+    if not new_status or new_status not in DraftService.STATUSES:
```
allait: I don't think we need to do this since status validator on `ServiceTableMixin` would run the same check

PR #334, opened 2016-01-27
Bugfix: delete keys with None values when saving declaration changes

TheDoubleK:
Updating a declaration with the admin app on a page with optional questions that have not been answered creates keys with `None` values in the saved declaration.

 This causes the template that displays the declarations to break.

 This change purges keys with `None` values before saving.


robyoung:
:+1:

PR #335, opened 2016-01-29
Buyer should be able to write a brief

robyoung:
[#111139752](https://www.pivotaltracker.com/story/show/111139752)

Adds a `Brief` model and routes to create, update, list, get and update the status.

There are various constraints around brief creation and modification.
- There is a composite foreign key to `FrameworkLot` to ensure that a brief isn't created for a lot on a different framework.
- Only 'buyer' users can be associated with a brief
- Status must be 'draft' or 'live' and 'live' briefs cannot be turned into 'draft' briefs.
- Briefs can only be created for lots that require briefs (ie. not for G-Cloud lots or studios)

I have created stub JSON schemas just to allow their integration.

## depends on
- [x] https://github.com/alphagov/digitalmarketplace-apiclient/pull/6


app/models.py
```diff
@@ -750,6 +759,112 @@ def serialize(self, include_user=False):
...
+    def update_from_json(self, data):
+        current_data = dict(self.data.items())
+        current_data.update(data)
+
+        self.data = current_data
+
+    def serialize(self):
+        data = dict(self.data.items())
+
+        data.update({
```
allait: The frontend will need to check that a user belongs to a brief and can view/modify/publish, so we need to return a list of brief users here (maybe only when an individual brief is requested).

app/main/views/briefs.py
```diff
@@ -0,0 +1,162 @@
...
+    ).first_or_404()
+
+    return jsonify(briefs=brief.serialize())
+
+
+@main.route('/briefs', methods=['GET'])
+def list_briefs():
+    page = get_valid_page_or_1()
+
+    briefs = Brief.query.order_by(Brief.id)
```
allait: Pretty sure filtering by brief user is the main query we need right now, so it might be worth adding it right away

app/main/views/briefs.py
```diff
@@ -0,0 +1,162 @@
...
+    brief = Brief.query.filter(
+        Brief.id == brief_id
+    ).first_or_404()
+
+    if brief_json['status'] == brief.status:
+        abort(400, "Brief is already {}".format(brief.status))
+
+    brief.status = brief_json['status']
+
+    validate_brief_data(brief, enforce_required=True)
```
allait: We only need to do this when making the brief `live`, although I suppose it's the only status change we could have right now.

One other thing we could check here is that the framework is still 'live', so that it's not possible to create a draft brief and then publish it once the framework has expired.

app/models.py
```diff
@@ -38,6 +38,10 @@ class Lot(db.Model):
     name = db.Column(db.String, nullable=False)
     one_service_limit = db.Column(db.Boolean, nullable=False, default=False)
 
+    @property
+    def requires_brief(self):
```
allait: The flag controls whether it's possible to create a brief for a given lot, so should it be `allows_briefs` rather than "requires"?

app/models.py
```diff
@@ -750,6 +759,112 @@ def serialize(self, include_user=False):
...
+    framework_id = db.Column(db.Integer, db.ForeignKey('frameworks.id'), nullable=False)
+    _lot_id = db.Column("lot_id", db.Integer, db.ForeignKey('lots.id'), nullable=False)
+
+    data = db.Column(JSON)
+    status = db.Column(db.String, index=False, unique=False,
+                       nullable=False, default='draft')
+
+    created_at = db.Column(db.DateTime, index=True, nullable=False,
+                           default=datetime.utcnow)
+    updated_at = db.Column(db.DateTime, index=True, nullable=False,
```
allait: This is a bit speculative, but it feels like the most important date for briefs is the publishing date – it makes the most sense for ordering and might be useful as a base point for expiration etc. With edit endpoints we have right now, publishing date should be the same as the last updated_at. But if we plan to store clarification questions as part of the brief that will allow some updates for live briefs, so will automatically change the updated_at timestamp.

We could also use audit event to find the publishing date, but it'd be hard to use for ordering.

app/models.py
```diff
@@ -750,6 +759,112 @@ def serialize(self, include_user=False):
...
+            raise ValidationError("The brief user must be a buyer")
+        return user
+
+    @property
+    def lot_id(self):
+        return self._lot_id
+
+    @lot_id.setter
+    def lot_id(self, lot_id):
+        raise ValidationError("Cannot update lot_id directly, use lot relationship")
```
allait: Why are we doing this?

app/models.py
```diff
@@ -750,6 +759,112 @@ def serialize(self, include_user=False):
...
+        data = purge_nulls_from_data(data)
+
+        return data
+
+    @validates('status')
+    def validates_status(self, key, status):
+        if status not in self.STATUSES:
+            raise ValidationError("Invalid brief status '{}'".format(status))
+        if self.status == 'live' and status == 'draft':
+            raise ValidationError("Cannot make a live brief draft")
```
allait: "Cannot change brief status from 'live' to 'draft'?"

app/validation.py
```diff
@@ -15,6 +15,10 @@
 
 JSON_SCHEMAS_PATH = './json_schemas'
 SCHEMA_NAMES = [
+    'briefs-digital-outcomes-and-specialists-digital-outcomes',
+    'briefs-digital-outcomes-and-specialists-digital-specialists',
+    'briefs-digital-outcomes-and-specialists-user-research-studios',
```
allait: studios shouldn't be in the list

app/main/views/briefs.py
```diff
@@ -0,0 +1,162 @@
...
+
+    json_payload = get_json_from_request()
+    json_has_required_keys(json_payload, ['briefs'])
+    brief_json = json_payload['briefs']
+
+    json_has_required_keys(brief_json, ['frameworkSlug', 'lot', 'userId'])
+
+    framework, lot = validate_and_return_lot(brief_json)
+
+    if framework.status != 'live':
```
TheDoubleK: I think this check could be done as part of the `validate_and_return_lot` method?  I can't think of a situation where we will want to do anything with briefs on frameworks in any state other than `live`?

app/main/views/briefs.py
```diff
@@ -0,0 +1,162 @@
...
+    json_has_required_keys(json_payload, ['briefs'])
+    brief_json = json_payload['briefs']
+    json_has_required_keys(brief_json, ['status'])
+
+    brief = Brief.query.filter(
+        Brief.id == brief_id
+    ).first_or_404()
+
+    if brief_json['status'] == brief.status:
+        abort(400, "Brief is already {}".format(brief.status))
```
TheDoubleK: Does this need to be an error?  If the status is already as requested it doesn't feel like an error to me - could just no-op and return a 200?

app/models.py
TheDoubleK: Yep agreed -`allow` is the word we never got to in the Slack conversation about this.

app/models.py
TheDoubleK: Agreed - using audit events for this would be pretty horrible.

json_schemas/briefs-digital-outcomes-and-specialists-user-research-studios.json
```diff
@@ -0,0 +1,16 @@
+{
```
TheDoubleK: We don't need a schema for studios.

migrations/versions/480_add_briefs_table.py
```diff
@@ -0,0 +1,36 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '480'
+down_revision = '470'
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+def upgrade():
```
TheDoubleK: Do we not want `status` as a column? It is for Services and Drafts.

migrations/versions/490_add_users_to_briefs.py
```diff
@@ -0,0 +1,32 @@
...
+# revision identifiers, used by Alembic.
+revision = '490'
+down_revision = '480'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
```
TheDoubleK: A minor annoyance... can we remove the `commands auto generated by Alembic - please adjust!` comments from the migrations?  Removing these comments feels like saying "I've reviewed Alembic's work and am happy with it".

app/main/views/briefs.py
robyoung: `validate_and_return_lot` is also used by the services and drafts views which can be used when the framework is not `live`.

tests/app/test_model.py
```diff
@@ -56,6 +58,137 @@ def test_framework_should_accept_valid_statuses():
...
+            'frameworkSlug': 'test',
+            'frameworkName': 'test',
+            'lot': 'test',
+            'lotName': 'test',
+            'title': 'test',
+        }
+
+        assert brief.data == {'title': 'test'}
+
+    def test_nulls_are_stripped_from_declaration(self):
```
TheDoubleK: `test_nulls_are_stripped_from_brief_data`?

tests/app/test_model.py
```diff
@@ -56,6 +58,137 @@ def test_framework_should_accept_valid_statuses():
...
+        brief = Brief(data={}, framework=self.framework, lot=self.lot)
+        brief.data = {'foo': 'bar', 'bar': None}
+
+        assert brief.data == {'foo': 'bar'}
+
+    def test_whitespace_values_are_stripped_from_declaration(self):
```
TheDoubleK: `test_whitespace_values_are_stripped_from_brief_data`

tests/app/test_model.py
```diff
@@ -71,6 +204,18 @@ def test_framework_is_live_only_returns_live_frameworks(self):
             assert_equal(services.count(), 3)
             assert(all(s.framework.status == 'live' for s in services))
 
+    def test_lot_must_be_associated_to_the_framework(self):
+        with self.app.app_context():
+            self.setup_dummy_suppliers(1)
+            self.setup_dummy_service(
```
TheDoubleK: It's not that clear from the setup here how this is testing for a lot associated to the framework - a comment to explain would be useful.

tests/app/views/test_briefs.py
```diff
@@ -0,0 +1,471 @@
...
+                    'lot': 'digital-specialists',
+                    'title': 'the title',
+                },
+                'update_details': {
+                    'updated_by': 'example'
+                }
+            }),
+            content_type='application/json')
+
+        assert res.status_code == 400
```
TheDoubleK: Might be nice to test the error message here too as there may be several reasons for a 400 and it's good to know the failure is for the right reason.

tests/app/views/test_briefs.py
```diff
@@ -0,0 +1,471 @@
...
+                'briefs': {'status': 'live'},
+                'update_details': {'updated_by': 'example'}
+            }),
+            content_type='application/json')
+        data = json.loads(res.get_data(as_text=True))
+
+        assert res.status_code == 400
+        assert data['error'] == 'Brief is already live'
+
+    def test_cannot_make_a_live_brief_pending(self):
```
TheDoubleK: `cannot_make_a_live_brief_draft`? Or `cannot_return_a_live_brief_to_draft`?

app/main/views/briefs.py
TheDoubleK: Ah OK, fair enough then. From the diff it looked like it was a completely new method.

migrations/versions/480_add_briefs_table.py
robyoung: Yes, but not without squashing quite a lot of the whole commits together as well. I'll have a go at this in a branch and see how it looks.

allait:
Looks great :+1:

migrations/versions/480_add_briefs_table.py
allait: I think it's fine either way. It might be surprising during review, but once it's merged there's no expectation that the initial table creation migration describes the current table schema, so I don't see a problem with leaving this as is.

app/models.py
```diff
@@ -750,6 +759,126 @@ def serialize(self, include_user=False):
...
+        return self._lot_id
+
+    @lot_id.setter
+    def lot_id(self, lot_id):
+        raise ValidationError("Cannot update lot_id directly, use lot relationship")
+
+    @validates('lot')
+    def validates_lot(self, key, lot):
+        if not lot.allows_brief:
+            raise ValidationError("Lot '{}' does not require a brief".format(lot.name))
```
TheDoubleK: does not allow a brief?

tests/app/test_model.py
```diff
@@ -56,6 +58,149 @@ def test_framework_should_accept_valid_statuses():
...
+    def test_brief_lot_must_be_associated_to_the_framework(self):
+        with self.app.app_context():
+            other_framework = Framework.query.filter(Framework.slug == 'g-cloud-7').first()
+
+            brief = Brief(data={}, framework=other_framework, lot=self.lot)
+            db.session.add(brief)
+            with pytest.raises(IntegrityError):
+                db.session.commit()
+
+    def test_brief_lot_must_require_briefs(self):
```
TheDoubleK: `must_allow_briefs`

tests/app/views/test_briefs.py
```diff
@@ -0,0 +1,535 @@
...
+                }
+            }),
+            content_type='application/json')
+        data = json.loads(res.get_data(as_text=True))
+
+        assert res.status_code == 201
+        assert data['briefs']['frameworkSlug'] == 'digital-outcomes-and-specialists'
+        assert data['briefs']['title'] == 'the title'
+
+    def test_create_fails_if_lot_does_not_require_briefs(self):
```
TheDoubleK: `does_not_allow`

robyoung:
Rebased on top of https://github.com/alphagov/digitalmarketplace-api/pull/336

PR #336, opened 2016-01-29
Rename framework gcloud to g-cloud

robyoung:
In the frameworks table change gcloud to g-cloud so that it can be used as the slug for buyer frontend URLs.

This is required for https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/210


TheDoubleK:
Nice - upgrade and downgrades tested locally and all good.
:+1:

PR #337, opened 2016-02-01
Set apache KeepAliveTimeout to greater than ELB idle timeout

allait:
One of the possible explanations for the 504 responses we're seeing
is that ELBs keep sending requests to connections closed by the app
webserver.

Recommendation we've received from AWS support is to set keep-alive
timeout to a value greater than ELB idle connection timeout (which
is 60s by default).

Looking at apache config on Elastic Beanstalk instances it doesn't
look like an explicit value is set anywhere, so Apache is probably
using the default KeepAliveTimeout of 5s.

Setting a value above 60s to see if it fixes the issue. There's a
risk that a high keep-alive value will tie up Apache workers with
idle connections if ELBs are opening more backend connections than
the number of Apache threads.

(http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/ts-elb-error-message.html#ts-elb-errorcodes-http504)

### DIFF
--- a/.ebextensions/apache-keepalive.config
+++ b/.ebextensions/apache-keepalive.config
@@ -1,0 +1,8 @@
+files:
+  "/etc/httpd/conf.d/keepalive.conf":
+    mode: "000644"
+    owner: root
+    group: root
+    content: |
+      KeepAlive On
+      KeepAliveTimeout 61


robyoung:
Do we have a plan for how we will test that this fixes the problem? Have we been able to recreate 504s ourselves?

allait:
@robyoung haven't been able to replicate 504s. I've tried disabling all request sources to staging and sending requests with different intervals, but it doesn't resolve in 504s. I couldn't find any explanation of how ELB decides to open a new backend connection vs reusing an existing one, so it's hard to come up with a pattern of requests that would trigger the issue.

On the other hand, we get a few of them in production every day for API, so a few days without any ELB 5xx should indicate that the problem is fixed.

The only possible problem with high keep-alives is if they lock Apache workers with idle connections, but that's something we can test by throwing a lot of simple requests to preview.

TheDoubleK:
:+1: for a tryout on preview at least.

PR #338, opened 2016-02-01
Turn Off Apache KeepAlive

allait:
Setting a high KeepAliveTimeout value seems to interfere with ELB
health checks and either prevents the deployment with the default
ELB health check intervals or causes ELB_5xx errors during the
deployment.

I'm not sure if this is caused by ELB/Apache interaction or if it's
Apache/mod_wsgi.

Instead, let's try disabling keep-alive altogether. This might force
ELB to close the backend connection after each request instead of
trying to reuse it, but I couldn't find an explanation of ELB
behaviour, so I'm not sure if this will improve the ELB_5xx rate.

### DIFF
--- a/.ebextensions/apache_keepalive.config
+++ b/.ebextensions/apache_keepalive.config
@@ -4,5 +4,4 @@ files:
     owner: root
     group: root
     content: |
-      KeepAlive On
-      KeepAliveTimeout 61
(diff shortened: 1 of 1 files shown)


pcraig3:
:+1: :pray:

PR #339, opened 2016-02-03
Add framework framework

robyoung:
## Add framework.framework to serialized output
For briefs and services (ie. `services`, `draft_services` and `archived_services`) add the frameworkFramework field to the output. This is required for the admin app to be able to generate buyer app service URLs in a framework agnostic way.

The key `frameworkFramework` is not nice but it follows the existing pattern. This pattern should probably be changed to include sub-resources as sub-objects rather than having one big flat key space.

## Add `lotSlug` to serialized Brief and Service
This adds in `lotSlug` with the intention of replacing the `lot` key. This follows the pattern we're using elsewhere and also opens us up to adding sub-resources as objects rather than flattening them into the object.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -509,7 +509,7 @@ class ServiceTableMixin(object):
             'id', 'status',
             'supplierId', 'supplierName',
             'frameworkSlug', 'frameworkFramework', 'frameworkName', 'frameworkStatus',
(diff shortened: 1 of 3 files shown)


TheDoubleK:
Like you say, it's not pretty but :+1:

PR #340, opened 2016-02-04
Introduce Makefile

robyoung:
Originally this PR was about running database migrations from the `run_app.sh` script to help prevent people forgetting to run migrations. This solution was not popular.

A Makefile was discussed as a solution that allows us to have a single command that runs migrations and runs the app while still allowing people to run just the application.


allait:
:+1:

PR #341, opened 2016-02-05
Pre-flight database version check

robyoung:
Check that the database version is up to date before starting the app.
This originally ran the database migrations before running the app but there was a concern that this may not be expected and there there may be use cases where this was not desired. To address these concerns this change makes no changes but alerts the user to a problem and tells them how to fix it. It also provides a way to skip the check if it is not needed.


allait:
:+1:

robyoung:
Closing in favour of https://github.com/alphagov/digitalmarketplace-api/pull/340

PR #342, opened 2016-02-05
Move env defaults to config.py

robyoung:
This way there is one place where config defaults are managed. They can
still be overridden with environment variables.


TheDoubleK:
:+1:

PR #343, opened 2016-02-08
Fix Makefile when not in a virtual environment

robyoung:
We need to provide the full path to executables that should come from the virtual environment in case the user is not already in a virtual environment.


allait:
:+1:

PR #344, opened 2016-02-08
Fix how DM_API_AUTH_TOKENS are loaded

robyoung:
Default development and test values are not provided in config.py Therefore loading them directly from the environment will no longer work.


TheDoubleK:
:+1:

PR #345, opened 2016-02-08
Add default empty DM_API_AUTH_TOKENS

robyoung:
This needs to be present in order for the environment variable to be
used.


allait:
:+1:, as long as we're sure this complains loudly if the environment variable isn't set properly

PR #346, opened 2016-02-09
Make update details payload consistent

allait:
### Move validate_and_return_updater_request to utils
This is a general function to check updated_by for all write requests.

### Add support for both update_details and updated_by JSON fields
Change `validate_and_return_updater_request` to accept either `update_details` or a top-level `updated_by` field.

This is an intermediate step to make all API endpoints use the helper function and make the update payload format consistent across all endpoints.

Once the apiclient and frontend apps have been updated to always use one of the formats we can drop support for the other one.

### Use updater validation helper in framework and supplier views
Makes sure all API endpoints are using the valudate updater details utils function.

### Use JSONUpdateTestMixin to test for missing update details
Remove ad-hoc endpoint tests in favour of JSONUpdateTestMixin test for missing update details.

Clean-up some duplicate tests and set endpoint and method on update view tests. Add support for using object attributes in endpoint, to make it possible to use the mixin with test classes that create the relevant object during test setup.

This updates most of the view tests with the exception of drafts and briefs. Briefs and drafts don't split tests between classes for each endpoint, so setting JSONUpdateTestMixin with one endpoint and method won't have the same benefit.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -3,6 +3,15 @@ from flask import abort, request
 from six import iteritems, string_types
 from werkzeug.exceptions import BadRequest
 
+from .validation import validate_updater_json_or_400
+
+
+def validate_and_return_updater_request():
+    json_payload = get_json_from_request()
+    json_has_required_keys(json_payload, ['update_details'])
+    validate_updater_json_or_400(json_payload['update_details'])
+    return json_payload['update_details']
+
 
 def link(rel, href):
     """Generate a link dict from a rel, href pair."""
--- a/app/utils.py
+++ b/app/utils.py
@@ -8,9 +8,13 @@ from .validation import validate_updater_json_or_400
 
 def validate_and_return_updater_request():
     json_payload = get_json_from_request()
-    json_has_required_keys(json_payload, ['update_details'])
-    validate_updater_json_or_400(json_payload['update_details'])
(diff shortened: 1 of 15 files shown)


app/utils.py
```diff
@@ -3,6 +3,19 @@
...
+
+def validate_and_return_updater_request():
+    json_payload = get_json_from_request()
+
+    if 'update_details' in json_payload:
+        json_payload = json_payload['update_details']
+
+    validate_updater_json_or_400(json_payload)
+
+    return {'updated_by': json_payload['updated_by']}
```
TheDoubleK: Can we not just `return json_payload` here? I don't think `update_details` ever has any other keys now? (`update_reason` got killed a long time ago, didn't it?)

TheDoubleK:
One minor comment, but looks fine as it is - nice to get this consistent, and the Mixin is sweet :+1:

app/utils.py
allait: You're right about `update_details`, but for requests with top-level `updated_by` `json_payload` would have the rest of request fields. It would still work fine, but this way we know we never access any other keys from updater request.

TheDoubleK: Ah of course, I was only thinking of it the other way round.

PR #348, opened 2016-02-12
Fix DOS schema enum values

robyoung:
This is the API side of https://github.com/alphagov/digitalmarketplace-frameworks/pull/192

Existing brief data will need to be deleted or updated.

### DIFF
--- a/json_schemas/briefs-digital-outcomes-and-specialists-digital-outcomes.json
+++ b/json_schemas/briefs-digital-outcomes-and-specialists-digital-outcomes.json
@@ -95,11 +95,11 @@
     },
     "phase": {
       "enum": [
-        "Not started",
-        "Discovery",
-        "Alpha",
-        "Beta",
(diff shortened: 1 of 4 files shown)


TheDoubleK:
Does this mean that in future we have the potential to change the labels but keep the same value so things don't all break (e.g. changes in capitalisation, hyphenation, etc.)?  Living the dream!

TheDoubleK:
Of course, in that case there is the potential for us to be storing values that don't reflect the answer provided... e.g. like if we change a label from "Scotland" to "Wales" we would have 'scotland' stored for people who ticked the "Wales" box :/

TheDoubleK:
Anyway, these metaphysical questions aside, :+1: for the changes here - we need them.

robyoung:
Yes, and yes, it may contain it's own set of pain, but we'll at least be able to decouple the two.

PR #349, opened 2016-02-12
Supplier responds to a brief

allait:
# New brief responses model and endpoints
### Add BriefResponse model
Stores supplier responses to briefs. There's no status or update, publish dates since we're not implementing edits for responses right now.

### Drop foreign fields from BriefResponses data
Only removes briefId and supplierId since they're sent as part of request. Other fields like 'links' are unlikely to appear in requests since there's little reason to reuse a brief response JSON as request body.

### Only allow one response per brief per supplier
This is likely to change in the future since we might allow multiple responses for digital specialists briefs, but for now we're limiting all briefs to 1 response per supplier.

### Add BriefResponse.validate method
Data validation is a common action we have for models that contain a JSON field. So far we've been using helper functions in utils modules, but a method on the model might be slightly more convenient.

### Validate the number of answers in the brief response
Checks that the number of answers matches the number of questions in the brief for both essential and nice to have requirements.

This assumes that requirement answers are either boolean or have a similar uniform structure (eg "Yes"/"No" enum values) and each item is validated by the JSON schema, so the only invalid answer type not caught by schema validation is the wrong number of answers.

### Add list brief responses endpoint
Can be filtered by supplier_id and brief_id.

Returns a paginated list only if not filtered, otherwise returns all responses at once ⁉️ (similar to what we do with supplier services to avoid having to deal with pagination in frontend code).

Returns a 200 ⁉️ with no results for non-integer supplier_id and brief_id. Different endpoints handle this differently (some 400, some will raise a DataError). We don't really expect those requests from the frontend apps, so I'm opting for the simplest way to handle them - which is ignoring the invalid value.

# App-level changes ‼️
### Check item types when removing whitespace for data arrays
Essential/non-essential brief question answers are lists of booleans so whitespace utility function has to check for item types.

### Don't remove falsy values when filtering empty items from data
Whitespace filter removed False values from lists. It should only filter empty strings and nulls, since they should never appear in lists.

### Add 'self' link to pagination_links
Since we have a 'self' link for serialized objects we should have one for pagination results too.

# Testing changes
### Add live and expired framework pytest fixtures 💭
This is a tiny attempt to move away from relying on database migrations to set up framework records for tests. Using migration framework records means we need to use actual (sometimes expired) frameworks to write tests ties us to existing frameworks and require manual rollbacks for any record changes since frameworks/lots tables are not reset between tests.

Using fixtures for frameworks makes it possible to depend explicitly on database state required to run the test. This works better than creating objects in setup/teardown since we can create more complicated objects by using fixture dependencies (eg a service fixture relying on framework and supplier fixtures etc.).

The fixture is still using DOS for now, but we could change it to create a test framework with test lots as long as any additional test objects are created using the fixture dependencies.

### Use hypothesis to create example brief response data 💭
Creates a dictionary strategy to create brief response data example. Sets default settings to run each example only once, so that our test times are pretty much unchanged, but it's always possible to change the value to get all the benefits of generative testing.

(I wasn't really planning on committing this, but it actually caught a bug in the whitespace filter, so I'm keeping it.)


tests/app/views/test_brief_response.py
```diff
@@ -0,0 +1,351 @@
...
+            content_type='application/json'
+        )
+
+        data = json.loads(res.get_data(as_text=True))
+
+        assert res.status_code == 400
+        assert data['error']['essentialRequirements'] == 'answer_required'
+        assert 'niceToHaveRequirements' in data['error']
+
+    def test_cannot_create_brief_without_supplier_id(self, live_framework):
```
pcraig3: Probably should be `test_cannot_create_brief_response_without_supplier_id`.

tests/app/views/test_brief_response.py
```diff
@@ -0,0 +1,351 @@
...
+        res = self.create_brief_response(dict(brief_response_data, **{
+            'briefId': self.brief_id,
+            'supplierId': 0,
+        }))
+
+        assert res.status_code == 400, res.get_data(as_text=True)
+        assert 'Brief response already exists' in res.get_data(as_text=True)
+
+    @given(example_listings.brief_response_data(None, 5).filter(lambda x: len(x['essentialRequirements']) != 5))
+    def test_cannot_respond_to_abrief_with_wrong_number_of_essential_reqs(self, live_framework, brief_response_data):
```
pcraig3: `test_cannot_respond_to_a_brief_with_wrong_number_of_essential_reqs`

pcraig3:
Seems okay to me.  Just noticed a couple of errors in test names was all.

allait:
Updated to fix test names

app/models.py
```diff
@@ -892,6 +892,71 @@ class BriefUser(db.Model):
...
+            'supplierId', 'briefId',
+        ])
+        data = strip_whitespace_from_data(data)
+        data = purge_nulls_from_data(data)
+
+        return data
+
+    def validate(self, enforce_required=True, required_fields=None):
+        errs = get_validation_errors(
+            'brief-responses-{}-{}'.format(self.brief.framework.slug, self.brief.lot.slug),
```
robyoung: Do we need to have separate schemas per lot?

robyoung:
I don't agree with returning 200 for invalid arguments when listing brief responses. I'm guessing you thought this would be contentious what with the interrobang. I think it should be a 400. I think returning a 200 is misleading and can mask bugs in our client code (either in the frontend or in quickly written scripts). It's even more of a problem if we ever want to make a read only version of the API public (which we're already getting use cases for).

app/main/views/brief_responses.py
```diff
@@ -0,0 +1,117 @@
...
+        brief = None
+
+    if brief is None:
+        abort(400, "Invalid brief ID '{}'".format(brief_response_json['briefId']))
+
+    if brief.status != 'live':
+        abort(400, "Brief must be live")
+
+    if brief.framework.status != 'live':
+        abort(400, "Brief framework must be live")
```
robyoung: Could these also be handled by the `validate` method? They are invariants that make creating the `BriefResponse` invalid.

tests/app/views/test_brief_response.py
```diff
@@ -0,0 +1,351 @@
...
+    def list_brief_responses(self, **parameters):
+        return self.client.get('/brief-responses', query_string=parameters)
+
+
+class TestCreateBriefResponse(BaseBriefResponseTest, JSONUpdateTestMixin):
+    endpoint = '/brief-responses'
+    method = 'post'
+
+    @given(example_listings.brief_response_data())
+    def test_create_new_brief_response(self, live_framework, brief_response_data):
```
robyoung: I like the framework pytest fixture and the hypothesis strategy. The only thing that worries me is that we now have quite a few things involved in setting up this test. Debugging it in the future will be that much more of a headache. I'm not sure if it's a big deal, it just worries me a little.

app/models.py
allait: I think so. Just like with briefs, there's a separate set of questions for brief responses, depending on the lot.

app/main/views/brief_responses.py
allait: Similar to briefs – this is a check that only applies when creating a response through the API. Stored brief response attached to an "expired" brief or to a brief on an expired framework is still valid.

allait:
Changed to return 400 when supplier_id or brief_id parameters are set to non-integer values https://github.com/alphagov/digitalmarketplace-api/commit/ac9919854bd05d4f4fa2fc9c076ec5176e3b3749

app/models.py
robyoung: Unless I'm missing something the schemas will always be the same though. We have `essentialRequirements` and `niceToHaveRequirements`.

app/main/views/brief_responses.py
robyoung: Ah yes, good point.

app/main/views/brief_responses.py
```diff
@@ -0,0 +1,117 @@
...
+
+    supplier = validate_and_return_supplier(brief_response_json)
+
+    # Check if brief response already exists from this supplier
+    if BriefResponse.query.filter(BriefResponse.supplier == supplier, BriefResponse.brief == brief).first():
+        abort(400, "Brief response already exists for supplier '{}'".format(supplier.supplier_id))
+
+    brief_response = BriefResponse(
+        data=brief_response_json,
```
TheDoubleK: Should we strip the `briefId` and `supplierId` out of the `brief_response_json` before saving, as we have the linked `brief` and `supplier` objects as columns.  If we do strip them out here though then we'll need to preserve the submitted JSON for the audit event, otherwise it will be impossible to track anything from the auditing.

app/models.py
TheDoubleK: One to check with Ralph/Sarah I think - at the moment I think Rob is right (see, e.g. http://dm-prototype.herokuapp.com/page/respond/template-brief?key=outcomes.0), but not sure if this might change in future.

allait: We have more questions for specialists and user research participants in the frameworks PR https://github.com/alphagov/digitalmarketplace-frameworks/pull/196/files

TheDoubleK: Ah righto!

json_schemas/brief-responses-digital-outcomes-and-specialists-digital-outcomes.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "additionalProperties": false,
+  "properties": {
+    "essentialRequirements": {
+      "items": {
+        "type": "boolean"
+      },
+      "maxItems": 10,
```
TheDoubleK: We should check this number with Ralph and/or Sarah - I noticed this morning that one of the examples in the prototype has 11 essential requirements.  We can add a `number_of_items` key in the content to change the max items that can be added to the list.

But I guess deciding the number isn't essential for this PR.

migrations/versions/570_create_brief_responses_table.py
```diff
@@ -0,0 +1,34 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+
+def upgrade():
+    op.create_table('brief_responses',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('data', postgresql.JSON(), nullable=True),
```
TheDoubleK: I can't think how/why we would have a brief response with NULL data? Seems odd to allow "the response is no response".

allait: We haven't set `nullable=False` on any of JSON fields, I think because NOT NULL is less useful on JSON columns since it doesn't apply to JSON with just `null`. But I suppose there's no reason to skip it. Fixed.

TheDoubleK: Don't we?  Ah well, like you say it probably doesn't make much difference either way.

TheDoubleK:
As far as I can tell everyone's concerns are addressed and this should be :+1:

PR #353, opened 2016-02-22
None requirements answers

allait:
### Don't drop None values when stripping whitespace
Changes utils function to only remove empty strings from lists (no longer removes `None` values :exclamation:). None can be either a valid or invalid value depending on the field, so the validation logic should be encoded in the field schema. If None is valid, the schema should define it as one of the possible item values, otherwise a validation error should be returned.

### Add None values to brief response requirements tests
Makes type validator return `answer_required` error message code for boolean types. :exclamation:
For requirements questions `None` value means that the quesiton wasn't answered, which should trigger a validation error similar to wrong number of answers.

This will replace a generic "there was a problem with the answer" message with "question requires an answer" for other boolean fields if a value triggers a type validator error, but I don't think this happens on requests made by frontend apps right now.

Checks that None values cause an error response when creating brief responses.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -111,7 +111,7 @@ def strip_whitespace_from_data(data):
         if isinstance(value, list):
             # Strip whitespace and remove empty items from lists
             data[key] = list(
-                filter(lambda x: x not in ['', None],
+                filter(lambda x: x != '',
                        map(lambda x: x.strip() if isinstance(x, string_types) else x, value))
             )
         elif isinstance(value, string_types):
--- a/tests/example_listings.py
+++ b/tests/example_listings.py
@@ -1,4 +1,9 @@
-from hypothesis import strategies, settings
+from hypothesis import settings
+from hypothesis.strategies import (
+    fixed_dictionaries, lists,
+    booleans, integers, text, none,
+    composite, sampled_from, one_of,
+)
 
 settings.register_profile("unit", settings(
     database=None,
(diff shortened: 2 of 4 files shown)


app/utils.py
```diff
@@ -111,7 +111,7 @@ def strip_whitespace_from_data(data):
         if isinstance(value, list):
             # Strip whitespace and remove empty items from lists
             data[key] = list(
-                filter(lambda x: x not in ['', None],
+                filter(lambda x: x != '',
```
pcraig3: This seems okay to me.  As it's used for `list`s and `boolean_list`s (and empty list fields are sent back as `''`, right?), it looks like a fairly small change.

pcraig3:
Looks good to me, but I think it's worth getting a second pair of :eyes:

tests/example_listings.py
```diff
@@ -7,31 +12,28 @@
 settings.load_profile("unit")
 
 
+@composite
+def requirements_list(draw, length, answers=False):
+    if answers:
+        elements = booleans() if length is not None else one_of(booleans(), none())
+    else:
+        elements = text(min_size=1, alphabet='abcdefgh', average_size=10)
```
robyoung: Why do we restrict this so much?

allait: We want to exclude spaces since we drop space-only strings when stripping whitespace. And using a smaller alphabet also seems to speed up the initialisation (I'm not sure if it might be explained by something else, but I've seen a +1s difference when running without alphabet and average_size. I suspect it's caused by sampling unicode characters)

robyoung: Does reducing the alphabet not also reduce the effectiveness of the tests. We want the API to not blow up when bad data is passed in.

allait: It does, but we're not touching the values too much either way, it'd be mostly testing the JSON schema validator. I think we're getting most of the benefit with the small alphabet giving us unpredictable values of varying length. We could throw in a few unicode characters there, but I wouldn't expect them to change much (mostly because unicode issues are often caused by how data is sent and received, which is different when running unit tests).

robyoung:
:+1:

PR #354, opened 2016-02-25
Publish clarification questions and answers on briefs

robyoung:
[#112340959](https://www.pivotaltracker.com/story/show/112340959)

Depends on https://github.com/alphagov/digitalmarketplace-apiclient/pull/15

This allows clarification questions and answers to be published on a live brief.

I've done validation on the model rather than with a JSON schema as we are not using a JSON data column.

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -203,3 +203,39 @@ def delete_draft_brief(brief_id):
         abort(400, "Database Error: {0}".format(e))
 
     return jsonify(message="done"), 200
+
+
+@main.route("/briefs/<int:brief_id>/clarification-questions", methods=["POST"])
+def add_clarification_question(brief_id):
+    updater_json = validate_and_return_updater_request()
+
+    json_payload = get_json_from_request()
+    json_has_required_keys(json_payload, ['clarificationQuestion'])
+    question_json = json_payload['clarificationQuestion']
+    json_has_required_keys(question_json, ['question', 'answer'])
+
+    brief = Brief.query.filter(
+        Brief.id == brief_id
+    ).first_or_404()
+
+    question = brief.add_clarification_question(
+        question_json.get('question'),
+        question_json.get('answer'))
+
+    try:
+        db.session.flush()
+    except IntegrityError as e:
+        db.session.rollback()
+        abort(400, e.orig)
+
+    audit = AuditEvent(
+        audit_type=AuditTypes.add_brief_clarification_question,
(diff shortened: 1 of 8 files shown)


app/models.py
```diff
@@ -843,6 +848,16 @@ def validates_status(self, key, status):
             self.published_at = datetime.utcnow()
         return status
 
+    def add_clarification_question(self, question, answer):
+        clarification_question = BriefClarificationQuestion(
+            brief=self,
+            question=question,
+            answer=answer)
+
+        Session.object_session(self).add(clarification_question)
```
pcraig3: What does this do?  Is this how you reference the current database entity?

tests/app/test_model.py
```diff
@@ -207,6 +208,33 @@ def test_cannot_update_lot_by_id(self):
...
                       lot_id=self.framework.get_lot('user-research-studios').id)
 
+    def test_add_brief_clarification_question(self):
+        with self.app.app_context():
+            brief = Brief(data={}, framework=self.framework, lot=self.lot, status="live")
+            db.session.add(brief)
+            db.session.commit()
+
+            clarification = brief.add_clarification_question(
+                "How do expect to deliver this?",
```
pcraig3: How is _who_ going to deliver this?

app/main/views/briefs.py
```diff
@@ -203,3 +203,39 @@ def delete_draft_brief(brief_id):
...
+    json_payload = get_json_from_request()
+    json_has_required_keys(json_payload, ['clarificationQuestion'])
+    question_json = json_payload['clarificationQuestion']
+    json_has_required_keys(question_json, ['question', 'answer'])
+
+    brief = Brief.query.filter(
+        Brief.id == brief_id
+    ).first_or_404()
+
+    question = brief.add_clarification_question(
```
pcraig3: Is it worth checking that the questions/answers exist?
I was able to empty strings for both "question" and "answer", which seems like it's probably not very helpful.

app/models.py
robyoung: It gets the SQLAlchemy session associated with the `Brief` object (ie. the session that was used to retrieve it) and then adds the `BriefClarificationQuestion` object to it.

app/main/views/briefs.py
robyoung: Good call, I've added some validation to the `BriefClarificationQuestion` model.

pcraig3:
<sub>:+1:</sub> :+1: <sup>:+1:</sup>

PR #355, opened 2016-03-03
Fix invalid framework status error message

allait:
ValidationError should be printed with `e.message`, which is done
by the app error handler, so there's no need to handle the exception
in any of the views.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -8,7 +8,7 @@ from dmapiclient.audit import AuditTypes
 from dmutils.config import convert_to_boolean
 from .. import main
 from ...models import (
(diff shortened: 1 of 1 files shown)


app/main/views/frameworks.py
```diff
@@ -57,7 +57,7 @@ def update_framework(framework_slug):
                 data={'update': json_payload['frameworks']})
         )
         db.session.commit()
-    except (IntegrityError, ValidationError) as e:
+    except IntegrityError as e:
         db.session.rollback()
         abort(400, "Validation Error: {}".format(e))
```
TheDoubleK: Should the message here not mention `Validation` any more?

allait: Good point, changed to `Database Error`, since that's what we use for other `IntegrityError` exceptions

TheDoubleK:
:+1:

PR #357, opened 2016-03-07
Brief applications closed at

allait:
### Replace Brief.status column with a property
Replaces the column with a computed property. Setting the status now sets the related date fields instead so that the API clients can still reference `status` field both when reading and updating the brief.

### Add ability to filter briefs by status
Uses the sqlalchemy hybrid_property with a separate function to be used in query expressions.

This means we need to reimplement the logic twice (both for getting the status property for python and for filtering based on status in the database), but the `.status` property behaves like a regular column for the rest of the API code.

http://docs.sqlalchemy.org/en/latest/orm/extensions/hybrid.html#module-sqlalchemy.ext.hybrid

### Add Brief 'closed' status and applications_closed_at property
Adds a closed date calculated from the brief's published date for published briefs.

Similar to status, applications_closed_at is a hybrid property and has a separate implementation when used in SQLAlchemy expressions, which is covered by additional query tests.

Closed date is also checked to return 'closed' status if the brief application period has closed.

### Add Brief.clarification_questions_closed_at property
Sets the clarification questions close date, similar to the close date for brief responses.

### Add Brief.clarificaiton_quesitons_are_closed property
Checking whether the close date has passed seems like it will be the most common operation related to close dates. When checking whether applications are closed the app can check the brief status instead, but for clarification questions the frontend apps would need to parse the date first.

This adds a helper property to Brief objects and serialized data that returns true if clarification close date is in the future and false otherwise.

### DIFF
--- a/migrations/versions/590_remove_brief_status_column.py
+++ b/migrations/versions/590_remove_brief_status_column.py
@@ -1,0 +1,26 @@
+"""Remove brief status column
+
+Revision ID: 590
+Revises: 580
+Create Date: 2016-03-03 14:56:59.218753
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '590'
+down_revision = '580'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.drop_column('briefs', 'status')
+
+
+def downgrade():
+    op.add_column('briefs', sa.Column('status', sa.VARCHAR(), autoincrement=False, nullable=True))
+    op.execute("""
+        UPDATE briefs SET status = (CASE WHEN published_at is not NULL THEN 'live' ELSE 'draft' END)
+    """)
+    op.alter_column('briefs', sa.Column('status', sa.VARCHAR(), nullable=False))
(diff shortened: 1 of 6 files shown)


robyoung:
:+1: That is a really nice way of representing the logic.

migrations/versions/590_remove_brief_status_column.py
```diff
@@ -0,0 +1,22 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.drop_column('briefs', 'status')
+
+
+def downgrade():
+    op.add_column('briefs', sa.Column('status', sa.VARCHAR(), autoincrement=False, nullable=False))
```
TheDoubleK: I don't think this downgrade will work if there are any briefs because there's no default value set.  I think the way to do it would be to add the column as `nullable=True`, run a query to populate the status for existing briefs (depending on published time, etc.), and then set the column to `nullable=False`.

allait: Updated with a proper downgrade. It will only set 'draft' or 'live' status, since we didn't have any closed briefs before the migration.

TheDoubleK:
Thanks.  I know we're unlikely ever to use it in any production environment, and if we need to then it will be a sign that something has really gone horribly wrong.  But still nice to know it will work.
:+1: (until next week, when we decide that times need to be set for MVP or whatever...)

PR #358, opened 2016-03-08
Allow service IDs to be generated for DOS

robyoung:
Remove the restriction on serivce IDs that only allows them to be
generated for g-cloud-7.

This generates a string of integers between 16 characters long.
I've changed from exponential notation to raising to the power because
the exponential notation creates floats and the larger float overflows.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1052,7 +1052,4 @@ def filter_null_value_fields(obj):
 
 
 def generate_new_service_id(framework_slug):
-    if framework_slug == 'g-cloud-7':
-        return str(random.randint(7e15, 8e15-1))
-    else:
-        raise NotImplemented("Can only generate service_id for G-Cloud 7")
+    return str(random.randint(10 ** 14, 10 ** 15 - 1))


pcraig3:
One argument against 16-character ids is that in a normal import Excel will treat them as numbers and set the final digit to zero.  See [here](http://superuser.com/questions/634565/how-to-stop-excel-from-formating-the-16-digit-account-number-in-a-way-to-avoid-m) and [here](http://www.cnet.com/forums/discussions/displaying-long-numbers-greater-than-15-char-in-excel-292538/).
There seems to be no way to format a 16 digit column so that it will just be imported as a string by default.

Ash has brought this up before and so has Nicole.

robyoung:
I've updated to reduce the number of digits to 15.
I have also passed this by @cathrooney and she is ok with it so long as existing IDs don't change. Which they will not.

TheDoubleK:
Our `is_valid_service_id()` method allows any string between 10 and 20 characters long, which seems pretty liberal, but probably fine for now still.

:+1:

PR #359, opened 2016-03-08
[#113271913] Buyer filters suppliers

robyoung:
https://www.pivotaltracker.com/story/show/113271913

This change consists of two related parts.

## Filter services by lot, location and role
This change involves adding `lot_slug`, `location` and `role` parameters to the `/services` route.

## Check if a supplier is eligible for a brief
This adds a new route `/suppliers/<supplier_id>/briefs/<brief_id>` that accepts `HEAD` requests and will respond with a `200` if the supplier is eligible and a `404` if the supplier is not eligible.


app/main/views/suppliers.py
```diff
@@ -430,3 +435,31 @@ def update_supplier_framework_details(supplier_id, framework_slug):
         return jsonify(message="Database Error: {0}".format(e)), 400
 
     return jsonify(frameworkInterest=interest_record.serialize()), 200
+
+
+@main.route("/suppliers/<supplier_id>/briefs/<brief_id>", methods=["HEAD"])
```
allait: How would we end an endpoint to get a full list of suppliers for a brief, which we might need for eg notifications? The URL also looks a bit strange to me, as briefs don't really belong to suppliers the way it seems to suggest. What do you thing about `/briefs/<brief_id>/suppliers?supplier_id=<id>`? We could then extend it to return the full list of eligible suppliers when called without `supplier_id`.

app/main/views/suppliers.py
```diff
@@ -430,3 +435,31 @@ def update_supplier_framework_details(supplier_id, framework_slug):
...
+        statuses=["published"],
+        lot_slug=brief.lot.slug,
+        location=brief.data["location"],
+        role=brief.data["specialistRole"] if brief.lot.slug == "digital-specialists" else None
+    )
+
+    services = services.filter(Supplier.supplier_id == supplier.supplier_id)
+
+    if len(services.all()):
+        return "", 200
```
allait: Is the apiclient going to be ok with an empty response? I think our default request methods try to parse the response as JSON?

app/main/views/services.py
```diff
@@ -42,16 +42,25 @@ def list_services():
...
-        ])
+        statuses = [status.strip() for status in request.args['status'].split(',')]
+    else:
+        statuses = None
+
+    try:
+        services = filter_services(
+            framework_slugs=frameworks,
+            statuses=statuses,
+            lot_slug=request.args.get('lot_slug', None),
```
allait: We use `framework` when passing framework slugs as query arguments, should we just use `lot` for lots?

app/main/views/suppliers.py
```diff
@@ -430,3 +435,28 @@ def update_supplier_framework_details(supplier_id, framework_slug):
...
+
+    services = filter_services(
+        framework_slugs=[brief.framework.slug],
+        statuses=["published"],
+        lot_slug=brief.lot.slug,
+        location=brief.data["location"],
+        role=brief.data["specialistRole"] if brief.lot.slug == "digital-specialists" else None
+    )
+
+    services = services.filter(Supplier.supplier_id == supplier.supplier_id)
```
allait: Should be `.filter(Services.supplier_id == supplier.supplier_id)`. `Supplier.supplier_id` is not joined to this table, so the check is ignored (or always True)

TheDoubleK: Good spot!

app/main/views/services.py
robyoung: It bugs me that internally we try to only use the noun on it's own for an instance of the model or a JSON blob describing it and therefore the `_slug` suffix becomes a good signal to what is supposed to be in the variable. It is a lot harder to change the `framework` query argument though.

allait:
:+1:

PR #360, opened 2016-03-09
Avoid using cast on audit date

robyoung:
[#115324811](https://www.pivotaltracker.com/story/show/115324811)

Casting the audit created_at field to a date prevents the index on created_at being used. As the audit table grows this index becomes much more significant for the service status change page which only filters on date and audit type.

See: http://stackoverflow.com/a/18485137/15720


TheDoubleK:
Looks good - I think the existing tests cover this OK (well, there is one test that gets an audit event by date and I think that's all we need still).

:+1:

PR #362, opened 2016-03-10
Remove `update_details` from everywhere

TheDoubleK:
The API client never sends `update_details` any more, so the API should just expect `updated_by`.

Thought this was a one-line change (in `suppliers.py`), but ended up sorting out the tests too. The only change everywhere here is to remove `update_details` in favour of plain old `updated_by`.


allait:
I was holding off on this thinking some of the scripts might still be using `update_details`, but quick search suggests it's only in the API. So I think we might be able to remove `update_details` support from utils altogether https://github.com/alphagov/digitalmarketplace-api/blob/master/app/utils.py#L12.

Thanks for doing this :+1:

PR #363, opened 2016-03-11
[#114474769] Add endpoint to check if an email address has a valid buyer domain

TheDoubleK:
This is required to address this bug with buyer user sign-up:
https://www.pivotaltracker.com/story/show/114474769


robyoung:
:+1:

PR #364, opened 2016-03-14
Don't allow updating or deleting closed briefs

allait:
### Change setup_dummy_briefs helper to support 'closed' status
Sets the published_at field directly instead of setting the status property.

### Don't allow updating or deleting closed briefs
Changes brief status checks to look for non-'draft' briefs instead of checking for 'live' status only. This way, 'live' and 'closed' briefs aren't allowed to be updated or removed and the check will cover any similar statuses added in the future.

Fixes the missing `test_` in test name for live brief delete.

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -74,8 +74,8 @@ def update_brief(brief_id):
         Brief.id == brief_id
     ).first_or_404()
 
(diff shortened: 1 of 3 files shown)


TheDoubleK:
:+1:

PR #365, opened 2016-03-15
Only allow creating a brief response for eligible suppliers

allait:
### Only allow creating a brief response for eligible suppliers
Checks if the supplier is elgible to respond to a brief and returns
400 otherwise.

- [x] Requires #359

### DIFF
--- a/tests/example_listings.py
+++ b/tests/example_listings.py
@@ -34,6 +34,7 @@ def brief_response_data(essential_count=5, nice_to_have_count=5):
 
 def brief_data(essential_count=5, nice_to_have_count=5):
     return fixed_dictionaries({
+        'location': text(alphabet='abcdefghijkl'),
         'essentialRequirements': requirements_list(essential_count),
         'niceToHaveRequirements': requirements_list(nice_to_have_count),
     })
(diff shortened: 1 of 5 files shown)


tests/example_listings.py
```diff
@@ -34,6 +34,7 @@ def brief_response_data(essential_count=5, nice_to_have_count=5):
 
 def brief_data(essential_count=5, nice_to_have_count=5):
     return fixed_dictionaries({
+        'location': text(alphabet='abcdefghijkl'),
```
TheDoubleK: Why use a random text here?  We have fixed sets of locations we could use?

TheDoubleK:
Left a question out of interest, but looks good :+1:

tests/example_listings.py
allait: We caught a unicode issue with the initial generator (https://github.com/alphagov/digitalmarketplace-api/pull/365/commits/857f8ff7a2b4c415fe369ca64c156a7c97b5a418) but then had to restrict the alphabet because it triggered some escaping error in SQLAlchemy with some of the strings.

We could've used a set of actual location choices, but since this doesn't go through schema validation there's no real reason to tying it to content values right now.

PR #366, opened 2016-03-16
[#114207705] Allow filtering briefs by multiple statuses, lot and framework

TheDoubleK:
Required for: https://www.pivotaltracker.com/story/show/114207705

We need both `live` and `closed` briefs in a single list for the catalogue of briefs in the frontend.

This mimics the `has_statuses` query that we already have for selecting services with multiple different statuses.

Also now allows filtering briefs by framework and lot slugs.


allait:
:+1:

PR #367, opened 2016-03-16
Add briefs object type support to audit-events endpoints

allait:
Adds 'briefs' as possible value for audit event `object-type`, both for creating and filtering audit events.

This allows us to create audit events related to brief objects from other apps, eg creating events for brief clarification questions.


TheDoubleK:
:+1:

PR #368, opened 2016-03-18
Do not paginate briefs fetch if user_id is supplied

TheDoubleK:
For a single user we should return all briefs - this is useful, e.g. for the buyer dashboard in the frontend.


robyoung:
:+1: for the moment, I guess we revisit pagination on the dashboard if it becomes a problem

PR #369, opened 2016-03-23
Abort draft list if invalid framework is provided

robyoung:


app/main/views/drafts.py
```diff
@@ -139,6 +139,8 @@ def list_draft_services():
         framework = Framework.query.filter(
             Framework.slug == framework_slug
         ).first()
```
allait: `.first_or_404` would've worked just fine, I think we use it quite a lot in other places

allait:
:+1:

app/main/views/drafts.py
robyoung: The reason I didn't use `first_or_404` here was that the error message didn't give any indication as to what had gone wrong. Given that multiple things could be causing the 404 I think a specific error message is more useful.

robyoung: It's a bit of a pain that you can't provide an override error message to `first_or_404`.

PR #370, opened 2016-03-24
Update brief JSON schema with new specialist values

allait:
New values should match the service location/day rate specialist keys.


robyoung:
:+1:

PR #371, opened 2016-03-30
Add endpoint to add a framework

robyoung:
The error message for invalid clarification questions flag is less than ideal. This is because if handled with a validtor method we would raise an error with a field name that does not match that given in the request and when handled with a DataError we can't easily be sure that the error is caused by invalid clarification questions flag.


allait:
:+1:

PR #372, opened 2016-04-04
Update JSON schemas from latest content

TheDoubleK:
I still think we also need to make `niceToHaveRequirements` required, but that can be part of the wider overhaul of content that will have to come at some point soon.

This adds the email address required for brief responses, so we know who to tell buyers to contact.

**UPDATE** - the email format validation in the second commit is auto-generated from the content in this pull-request: https://github.com/alphagov/digitalmarketplace-frameworks/pull/215


robyoung:
:+1:

robyoung:
:+1:

PR #373, opened 2016-04-05
Criteria weighting validation

allait:
### Add criteria weighting 100% total validation
Checks the criteria weighting sum if all criteria fields are set.
This relies on all three fields being required.

If the fields don't add up to a 100 an error is added for each field
that doesn't have any other validation errors.

### DIFF
--- a/app/brief_utils.py
+++ b/app/brief_utils.py
@@ -13,6 +13,14 @@ def validate_brief_data(brief, enforce_required=True, required_fields=None):
         required_fields=required_fields
     )
 
+    criteria_weighting_keys = ['technicalWeighting', 'culturalWeighting', 'priceWeighting']
+    # Only check total if all weightings are set
+    if all(key in brief.data for key in criteria_weighting_keys):
+        criteria_weightings = sum(brief.data[key] for key in criteria_weighting_keys)
+        if criteria_weightings != 100:
+            for key in set(criteria_weighting_keys) - set(errs):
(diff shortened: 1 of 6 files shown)


app/brief_utils.py
```diff
@@ -13,6 +13,14 @@ def validate_brief_data(brief, enforce_required=True, required_fields=None):
         required_fields=required_fields
     )
 
+    criteria_weighting_keys = ['technicalWeighting', 'culturalWeighting', 'priceWeighting']
+    # Only check total if all weightings are set
+    if all(key in brief.data for key in criteria_weighting_keys):
+        criteria_weightings = sum(brief.data[key] for key in criteria_weighting_keys)
+        if criteria_weightings != 100:
+            for key in set(criteria_weighting_keys) - set(errs):
+                errs[key] = 'total_should_be_100'
```
robyoung: Why is this done here rather than in `validation` [like pricing](https://github.com/alphagov/digitalmarketplace-api/blob/master/app/validation.py#L159)?

allait: Because it only applies to briefs. Pricing is a bit more general now, it has a separate question type and is not tied to the field keys directly, while this is a more ad-hoc solution right now.

robyoung: Fair enough. That does raise the question of why, if this is closer to brief responses, is it not changed to be implemented more like brief responses? ie. In `Brief.validate`

robyoung:
:+1:

PR #375, opened 2016-04-11
Return correct error message for non-integer values

pcraig3:
Passing in non-integer values to fields that expected them (for example, [some of our 'weighting' keys](https://github.com/alphagov/digitalmarketplace-api/blob/master/json_schemas/briefs-digital-outcomes-and-specialists-digital-specialists.json#L20) used to return the wrong error message.

Also changed the logic around validating the sum of all of our weightings (they have to add up to 100) such that we won't try and sum the three values unless they've individually been validated successfully.


allait:
👍

PR #376, opened 2016-04-13
Update JSON schemas with question and answer session details question

allait:
Adds the new question to the schemas. Since the question is optional the API can be released separately from the frontend apps.


pcraig3:
I'm already using them, so :+1: after a rebase

PR #377, opened 2016-04-13
Change brief application and questions closing times to 23:59

allait:
Setting the closing date and time to 23:59 + 14 days means that we can show the date part of the timestamp separately since it's the last day applications/questions are open as opposed to the first day they're closed for 00:00 dates. Making the change in the API allows frontend apps to display the date without any modifications.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -867,16 +867,16 @@ class Brief(db.Model):
         if self.published_at is None:
             return None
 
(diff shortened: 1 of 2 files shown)


TheDoubleK:
This is fine for the time being :+1:

However, in future I think we will end up counting the clarification period back from the closing date, rather than forward from the published date. But probably leave it as it is for now and change that when we need to.

PR #378, opened 2016-04-15
Limit brief response specialist day rate

allait:
### Update brief responses JSON schemas
Sets dayRate pricing pattern and updates email field property after replacing 'email' type with string limit format in the content repo.

### Add 'dayRate' to the list of pricing field name suffixes
If dayRate is not recognised as a pricing field the validation will raise an exception when trying to get word count from the pricing pattern.

There's no other easy way to mark pricing fields right now and renaming dayRate to contain 'price*' doesn't seem appropriate.

Checking the question names in frameworks repo, dayRate doesn't appear anywhere else at the moment.

### Return 'invalid_format' error for JSON schema 'format' validator
Allows content questions that specify string format to set error message (eg setting a message for invalid email strings).

### Check that brief response day rate is less than the service maximum
For digital-speficalists responses dayRate should be less than or equal to the maximum price for the given specialist set in the supplier's DOS service.

### :bug: Strip trailing whitespace from brief clarification questions
Values with leading or trailing whitespace don't match the word-count regular expression resulting in a "Too many words" error message being displayed by the frontend apps.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -930,8 +930,8 @@ class Brief(db.Model):
     def add_clarification_question(self, question, answer):
         clarification_question = BriefClarificationQuestion(
             brief=self,
-            question=question,
-            answer=answer)
+            question=question.strip(),
+            answer=answer.strip())
         clarification_question.validate()
 
         Session.object_session(self).add(clarification_question)
--- a/app/models.py
+++ b/app/models.py
@@ -1019,7 +1019,7 @@ class BriefResponse(db.Model):
 
         return data
 
-    def validate(self, enforce_required=True, required_fields=None):
+    def validate(self, enforce_required=True, required_fields=None, max_day_rate=None):
         errs = get_validation_errors(
             'brief-responses-{}-{}'.format(self.brief.framework.slug, self.brief.lot.slug),
             self.data,
@@ -1038,6 +1038,10 @@ class BriefResponse(db.Model):
             len(self.data.get('niceToHaveRequirements', [])) != len(self.brief.data.get('niceToHaveRequirements', []))
         ):
             errs['niceToHaveRequirements'] = 'answer_required'
+
+        if max_day_rate and 'dayRate' not in errs:
+            if float(self.data['dayRate']) > float(max_day_rate):
+                errs['dayRate'] = 'max_less_than_min'
 
         if errs:
             raise ValidationError(errs)
--- a/tests/app/views/test_brief_response.py
+++ b/tests/app/views/test_brief_response.py
@@ -29,10 +29,27 @@ class BaseBriefResponseTest(BaseApplicationTest):
                 supplier_id=0,
             )
 
-            db.session.add_all([service, brief])
+            specialist_brief = Brief(
+                data=example_listings.brief_data().example(),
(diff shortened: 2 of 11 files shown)


TheDoubleK:
There is a `maxDayRate` to come in briefs: https://github.com/alphagov/digitalmarketplace-frameworks/pull/225/files#diff-c59b4f1ba35968bdbf9a2427fee3811bR47

But looking at this it will be fine.  But I do wonder if maybe I should rename in that new content to something like `maxPossibleDayRate` to try to disambiguate a bit.

allait:
@TheDoubleK can we merge that with `budgetRange`? It's a textbox question, and right now the name sounds like it'd be an actual price

TheDoubleK:
The story note specifically says `The following questions need to be kept separate: Maximum day rate (specialists) [...] Budget range (outcomes)` (https://www.pivotaltracker.com/n/projects/997836/stories/116711753)

app/models.py
```diff
@@ -1039,6 +1039,10 @@ def validate(self, enforce_required=True, required_fields=None):
         ):
             errs['niceToHaveRequirements'] = 'answer_required'
 
+        if max_day_rate and 'dayRate' not in errs:
+            if float(self.data['dayRate']) > float(max_day_rate):
```
TheDoubleK: Should we catch cast errors here?  Or we should fail because there's a real problem if dayRate isn't a number?

tests/app/views/test_brief_response.py
```diff
@@ -29,10 +29,27 @@ def setup(self):
                 supplier_id=0,
             )
 
-            db.session.add_all([service, brief])
+            specialist_brief = Brief(
+                data=example_listings.brief_data().example(),
+                status='live', framework_id=5, lot=Lot.query.get(6)
```
TheDoubleK: Relying on existing framework and lot IDs in tests feels wrong to me, but we do it all over the place and fixing it isn't part of this PR... another one to chalk up for "after DOS live" though.

app/models.py
allait: We should fail, since we don't expect any cast errors here – there are no errors in the field after schema validation, and that checks for values matching the float format (only digits, one dot etc).

tests/app/views/test_brief_response.py
allait: These are set up by migrations, which is why they exist in tests. I think we shouldn't be using actual frameworks in tests at all, but it's definitely a separate discussion.

app/models.py
```diff
@@ -1039,6 +1039,10 @@ def validate(self, enforce_required=True, required_fields=None):
...
+                errs['dayRate'] = 'max_less_than_min'
```
TheDoubleK: This key name seems confusing here.
Is there a benefit in re-using this key rather than having something new and more accurately descriptive? (e.g. `over_max_value`)

allait: The name is confusing, but that's something we need to fix either way. I like reusing it here since it fits quite well and it helps keep the list of validation message keys smaller.

TheDoubleK:
OK cool, I still think that key is more confusing in this case than how it's been used elsewhere, but :+1:

PR #384, opened 2016-04-22
Only strip brief supplier question and answer if they're strings

allait:
Content loader replaces empty strings with `None` values when sending the request to the API, which makes the `.strip()` call fail.

Instead, we move the logic to model field validators and only call strip if the field value is of string type.

With an upcoming change to the buyer frontend the fields will be stripped before being sent to the API, so we could potentially get rid of the `.strip` call completely (similar to how we don't strip whitespace from other direct database fields), but it feels like the API should be responsible for this.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -4,6 +4,8 @@ import re
 
 from flask import current_app
 from flask_sqlalchemy import BaseQuery
+from six import string_types
+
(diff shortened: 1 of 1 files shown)


pcraig3:
👍

PR #387, opened 2016-04-25
Update essential and nice-to-have maxItems for brief responses

allait:
The limit of nice-to-have and essential requirements for briefs has been raised to 20, but the same changes hasn't been applied to the brief response boolean lists. So trying to respond to a brief with more than 10 requirements of any type always resulted in an error.


TheDoubleK:
:+1:

PR #390, opened 2016-05-05
Add data to lots table

TheDoubleK:
This is required to do a "proper" fix for the following bug:
https://www.pivotaltracker.com/story/show/118986461

By keeping relevant data in the lots table about the type of thing that gets submitted for that lot we can avoid having logic in the frontend that needs to know the specifics of different lots.  Instead the frontend can get everything it needs to know from the API.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -44,6 +44,7 @@ class Lot(db.Model):
     slug = db.Column(db.String, nullable=False, index=True)
     name = db.Column(db.String, nullable=False)
     one_service_limit = db.Column(db.Boolean, nullable=False, default=False)
+    data = db.Column(JSON)
 
     @property
     def allows_brief(self):
@@ -53,13 +54,15 @@ class Lot(db.Model):
         return '<{}: {}>'.format(self.__class__.__name__, self.name)
 
     def serialize(self):
-        return {
+        lot = {
             'id': self.id,
             'slug': self.slug,
             'name': self.name,
             'oneServiceLimit': self.one_service_limit,
             'allowsBrief': self.allows_brief,
         }
+        lot.update(dict(self.data.items()))
+        return lot
 
 
 class Framework(db.Model):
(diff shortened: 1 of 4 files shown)


app/models.py
```diff
@@ -53,13 +54,15 @@ def __repr__(self):
...
     def serialize(self):
-        return {
+        lot = {
             'id': self.id,
             'slug': self.slug,
             'name': self.name,
             'oneServiceLimit': self.one_service_limit,
             'allowsBrief': self.allows_brief,
         }
+        lot.update(dict(self.data.items()))
```
allait: Should we follow the same rules we use for services: start with `data = dict(self.data.items())` and then update with the column fields dictionary above, so that column values always override the JSON data if there's a collision?

TheDoubleK: Yes that makes sense.  Although we really shouldn't be getting any overlap.

migrations/versions/600_add_units_to_lots.py
```diff
@@ -0,0 +1,38 @@
...
+def upgrade():
+    op.add_column('lots', sa.Column('data', postgresql.JSON(), nullable=True))
+    op.execute("""
+        UPDATE lots
+          SET data = '{"unitSingular": "service", "unitPlural": "services"}'
+          WHERE  slug in ('saas', 'paas', 'iaas', 'scs', 'digital-outcomes', 'user-research-participants');
+    """)
+    op.execute("""
+        UPDATE lots
+          SET data = '{"unitSingular": "specialist", "unitPlural": "specialists"}'
```
allait: Is this what we do now? One digital-specialists service contains multiple specialists, I thought we used "service" to refer to the lot submission.

TheDoubleK: hmm, yes I think you're right.  I was thinking of the frameworks questions where we spilt into multiple questions on the basis of `specialist` vs. `service`, but maybe where this will be used `service` is still the thing.

I've also spotted we have things like `{{ "this digital specialist role" if lot == 'digital-specialists' else "a service in this category" }}`

Beginning to think this isn't going to be the end of our problems by any means.

allait:
Looks good once the tests are fixed 👍

TheDoubleK:
The tests were already fixed but got missed out of that last commit.  Honest :)

PR #391, opened 2016-05-05
[#118945593] Three new buyer domains for the list

TheDoubleK:
For this chore: https://www.pivotaltracker.com/story/show/118945593
Although it requires a release, so might be better labelled a feature.


allait:
👍

PR #392, opened 2016-05-06
Update schema for number of suppliers question with number type

allait:
Adds a migration to change brief numberOfSuppliers to integer.
If 'numberOfSuppliers' is a string parse the last number from the value and convert it to integer.

### DIFF
--- a/json_schemas/briefs-digital-outcomes-and-specialists-digital-outcomes.json
+++ b/json_schemas/briefs-digital-outcomes-and-specialists-digital-outcomes.json
@@ -106,8 +106,10 @@
       "type": "array"
     },
     "numberOfSuppliers": {
-      "minLength": 1,
-      "type": "string"
+      "exclusiveMaximum": false,
+      "maximum": 15,
+      "minimum": 3,
+      "type": "integer"
     },
     "organisation": {
       "maxLength": 100,
--- a/json_schemas/briefs-digital-outcomes-and-specialists-digital-specialists.json
+++ b/json_schemas/briefs-digital-outcomes-and-specialists-digital-specialists.json
@@ -96,8 +96,10 @@
       "type": "array"
     },
     "numberOfSuppliers": {
-      "minLength": 1,
-      "type": "string"
+      "exclusiveMaximum": false,
+      "maximum": 15,
+      "minimum": 3,
+      "type": "integer"
     },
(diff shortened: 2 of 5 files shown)


migrations/versions/600_change_numberofsuppliers_to_integer.py
```diff
@@ -0,0 +1,61 @@
+"""Change numberOfSuppliers to integer
+
+Revision ID: 600
+Revises: 590
+Create Date: 2016-05-06 11:37:59.204714
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '600'
```
TheDoubleK: This will need to bump to 610 now, sorry.

migrations/versions/600_change_numberofsuppliers_to_integer.py
```diff
@@ -0,0 +1,61 @@
...
+
+def upgrade():
+    conn = op.get_bind()
+    for brief in conn.execute(briefs.select()):
+        # Skip briefs with missing or integer 'numberOfSuppliers'
+        if brief.data.get('numberOfSuppliers') is None or isinstance(brief.data['numberOfSuppliers'], int):
+            continue
+
+        # Get the last number in the 'numberOfSuppliers' string
+        number_of_suppliers = re.search(r'(\d+)(?!.*\d)', brief.data['numberOfSuppliers'])
```
TheDoubleK: I'm guessing the "last group" is a requirement from looking at the actual data?

migrations/versions/600_change_numberofsuppliers_to_integer.py
allait: Updated

migrations/versions/600_change_numberofsuppliers_to_integer.py
allait: Yeah, that's based and tested on the values we have in production right now. I'm going to check that it's still fine right before merging and releasing, can't think of any other way to do it.

TheDoubleK:
Looks good - upgrade and downgrade tested locally and works. 

:+1:

TheDoubleK:
Not sure why that test was failing. Passes fine locally for me.  Kicked off Travis again... fingers crossed.

PR #395, opened 2016-05-13
Allow updating brief status from 'live' to 'draft'

allait:
We need to unpublish requirements if there's a problem with one of the questions (like point scores not being set on all requirements) so allowing to do this through the API.

The frontend apps set the status explicitly to 'live' when the use the status update apiclient method, but in the future we could add a separate '.publish_brief' method to make sure we don't allow users to accidentally unpublish briefs on their own.


pcraig3:
👍

PR #396, opened 2016-05-16
PSN allowed values change in schemas, from latest content

TheDoubleK:
This is also needed to make the SSP tests go green.


allait:
👍

PR #397, opened 2016-05-24
Add support for PaaS trial deployments

allait:
This is mostly changes from @robyoung's PaaS spike branch.

### Add gunicorn to requirements.txt for PaaS deploys  …
Add gunicorn to requirements.txt for running the app on Cloud Foundry
(possibly not ideal see [1]).

Explicitly add cffi to requirements so that Cloud Foundry spots it and
adds libffi.

There doesnt' seem to be an easy way to use a different requirements
file for Cloud Foundry, so we'll have to add the dependencies
everywhere.

[1] etianen/django-herokuapp#9

### Pull SQLAlchemy URI from VCAP_SERVICES  …
Cloud Foundry stores the connected services configuration in the
VCAP_SERVICES environment variable, so in order to connect to the
database instance we need to parse the URL value from the variable.

http://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#VCAP-SERVICES

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -2,6 +2,7 @@ from functools import wraps
 from flask import Flask
 from flask.ext.bootstrap import Bootstrap
 from flask.ext.sqlalchemy import SQLAlchemy
+import json
 
 import dmapiclient
 from dmutils import init_app, flask_featureflags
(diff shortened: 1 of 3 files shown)


TheDoubleK:
:+1: after a rebase

PR #399, opened 2016-05-25
Add phone number to user

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/119726147) story on Pivotal.

More work to come on the buyer front end to capture phone numbers on sign up.

This PR:

0. Updates the user model and adds a migration to add a phone_number column.
0. Adds phoneNumber to the serialize method on the User model.
0. Adds phoneNumber to the user json schema with a pattern to validate it. The patter check if the number is empty, or if it's a combination of digits, +'s, -'s, parenthesis or spaces. It also makes sure it's character length is in the range 6-20.
0. Updates the post /users view to create users with the phone number.
0. Updates one test and adds another to ensure users can be created with and without a phone number.
0. Updates an old test which was testing the wrong thing (creating an admin user when it should have been creating a buyer user)

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -113,7 +113,7 @@ def create_user():
     now = datetime.utcnow()
     user = User(
         email_address=json_payload['emailAddress'].lower(),
-        phone_number=json_payload.get('phoneNumber'),
+        phone_number=json_payload.get('phoneNumber') or None,
         name=json_payload['name'],
         role=json_payload['role'],
         password=password,
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -113,6 +113,7 @@ def create_user():
     now = datetime.utcnow()
     user = User(
         email_address=json_payload['emailAddress'].lower(),
+        phone_number=json_payload.get('phoneNumber'),
         name=json_payload['name'],
         role=json_payload['role'],
         password=password,
--- a/json_schemas/users.json
+++ b/json_schemas/users.json
@@ -21,7 +21,7 @@
       "type": "string"
     },
     "phoneNumber": {
-      "pattern": "^$|^\\+?([\\d\\s()-]){6,20}$",
+      "pattern": "^$|^\\+?([\\d\\s()-]){9,20}$",
       "type": "string"
     },
     "role": {
--- a/json_schemas/users.json
+++ b/json_schemas/users.json
@@ -20,6 +20,10 @@
       "minLength": 10,
       "type": "string"
     },
+    "phoneNumber": {
+      "pattern": "^$|^([\\d\\s()+-]){6,20}$",
+      "type": "string"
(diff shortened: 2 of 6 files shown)


app/main/views/users.py
```diff
@@ -113,6 +113,7 @@ def create_user():
     now = datetime.utcnow()
     user = User(
         email_address=json_payload['emailAddress'].lower(),
+        phone_number=json_payload.get('phoneNumber'),
```
allait: Should we do `phone_number=json_payload.get('phoneNumber') or None` here, so that we don't store empty strings as well as NULLs?

json_schemas/users.json
```diff
@@ -20,6 +20,10 @@
       "minLength": 10,
       "type": "string"
     },
+    "phoneNumber": {
+      "pattern": "^$|^([\\d\\s()+-]){6,20}$",
```
allait: I think we only expect an optional `+` at the beginning

tests/app/views/test_users.py
```diff
@@ -192,15 +192,16 @@ def test_can_post_a_buyer_user(self):
...
                     'password': '1234567890',
-                    'role': 'admin',
+                    'role': 'buyer',
                     'name': 'joe bloggs'}}),
             content_type='application/json')
 
         assert_equal(response.status_code, 201)
         data = json.loads(response.get_data())["users"]
-        assert_equal(data["emailAddress"], "joeblogs@email.com")
+        assert_equal(data["emailAddress"], "joeblogs@email.gov.uk")
```
TheDoubleK: Could also add an additional `assert` here to test the phone number is stored OK.

TheDoubleK:
Small comment on the test but looks good otherwise.

Wynndow:
Have updated the min phone number length to be 9 after conversations with Sarah and Roz. This way, the buyer will need to put in an area code to reach the 9 digit limit.

tests/app/views/test_users.py
```diff
@@ -231,6 +233,52 @@ def test_creating_buyer_user_with_good_email_domain_succeeds(self):
...
+            data=json.dumps({
+                'users': {
+                    'emailAddress': 'joeblogs@digital.cabinet-office.gov.uk',
+                    'phoneNumber': '123456',
+                    'password': '1234567890',
+                    'role': 'buyer',
+                    'name': 'joe bloggs'}}),
+            content_type='application/json')
+
+        assert response.status_code == 400
```
allait: Can we check what gets returned in `data['errors']` here?

Wynndow: `u "JSON was not a valid format. u'123456' does not match u'^$|^\\\\+?([\\\\d\\\\s()-]){9,20}$'"`

allait:
👍

PR #400, opened 2016-05-27
WIP add scripts/generate_model_diagram.py

risicle:
I thought it would be nice to be able to generate a new diagram at any time as things change.

This works, but `sadisplay` is not added to `requirements.txt` - I'm not sure if it should be added to `requirements.txt` or `requirements_for_test.txt`, if either.

Thoughts?


allait:
👍

PR #401, opened 2016-06-03
Remove mandatory evaluation types

allait:
### Update DOS evaluationType JSON schemas
Removes mandatory options and makes the question optional for all lots.

### Add a migration to drop mandatory assessment methods from brief data
Removes work history and written proposal from all briefs, since they're no longer a valid option for evaluationType and are added automatically to all briefs.

Downgrade will add the options to all briefs based on lot. Since we don't know if the brief had the option selected before the upgrade migration we add them to ALL briefs, including draft ones. This means we might end up adding it to briefs that didn't have it before, but that seems better than silently removing the option, as all briefs should have them selected in the end.

### DIFF
--- a/scripts/list_migrations.py
+++ b/scripts/list_migrations.py
@@ -4,8 +4,11 @@
 from __future__ import print_function
 
 import sys
+import warnings
 
 from alembic.script import ScriptDirectory
+
+warnings.simplefilter('error')
 
 
 def detect_heads(migrations):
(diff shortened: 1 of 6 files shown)


TheDoubleK:
The revision numbers need to be bumped 10 each, but after a rebase looks good.  The raw lot IDs in the downgrade make me slightly nervous, but as the lots were all added by migrations they will work fine.
:+1:

scripts/list_migrations.py
```diff
@@ -4,9 +4,11 @@
 from __future__ import print_function
 
 import sys
+import warnings
 
 from alembic.script import ScriptDirectory
 
+warnings.simplefilter('error')
```
TheDoubleK: This has angered PEP8.  Needs an extra empty line below.

Otherwise, nice catch!

PR #402, opened 2016-06-03
[#119645373] Briefs: Don't use location when deciding whether suppliers can apply

risicle:
This comes as two commits. The first simply removes the checks which prevent the brief being listed by the API and therefore applied for. The second removed the ability to filter briefs by location at all. This may or may not be desired, so submitted as two commits, the second can be discarded if needs be.

All tests pass locally at both revisions.

### DIFF
--- a/app/brief_utils.py
+++ b/app/brief_utils.py
@@ -30,7 +30,6 @@ def get_supplier_service_eligible_for_brief(supplier, brief):
         framework_slugs=[brief.framework.slug],
         statuses=["published"],
         lot_slug=brief.lot.slug,
-        location=brief.data["location"],
         role=brief.data["specialistRole"] if brief.lot.slug == "digital-specialists" else None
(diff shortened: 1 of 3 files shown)


risicle:
(This is for https://www.pivotaltracker.com/story/show/119645373)

risicle:
(second commit discarded)

pcraig3:
```
      -----------
    < Looks dope! >
      -----------
        / 
       / 
```
✌️ 😃  👍

risicle:
Guess I should wait till the frontend counterpart is ready before I commit this...

risicle:
Ok, on second thoughts, the current error message as it is isn't currently _wrong_, just a little over-specific, so it's probably not the end of the world if we have an intervening period of showing the old error message, and I want to get this off my books so will merge when it lets me.

PR #403, opened 2016-06-07
Update dmutils to 21.0.0

allait:
Moves DM_LOG_PATH to Live config, so that development and testing log to stderr.


TheDoubleK:
It's jumping a few major version changes, but the intermediate ones don't affect the API code at all.
:+1:

PR #405, opened 2016-06-09
Changed the url for the local API curl example to one that 200s

idavidmcdonald:
Previous url was returning a 404


TheDoubleK:
:+1: good to merge after a rebase.

PR #406, opened 2016-06-15
models: add comments - feedback welcome

risicle:
I've found these notes useful. What do people think? Are they correct? Anyone have any additions? Amendments? Anywhere where I've completely got the wrong end of the stick?

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -219,6 +219,7 @@ class ContactInformation(db.Model):
 class Supplier(db.Model):
     __tablename__ = 'suppliers'
 
+    # NOTE other tables tend to make foreign key references to `supplier_id` instead of this
     id = db.Column(db.Integer, primary_key=True)
 
     supplier_id = db.Column(db.BigInteger, Sequence('suppliers_supplier_id_seq'), index=True, unique=True,
@@ -377,9 +378,10 @@ class User(db.Model):
     ROLES = [
         'buyer',
         'supplier',
-        'admin',
-        'admin-ccs-category',
-        'admin-ccs-sourcing',
+        'admin',               # a general admin user, with permission to do most (but not all)
+                               # admin actions.
+        'admin-ccs-category',  # generally restricted to read-only access to admin views.
+        'admin-ccs-sourcing',  # can perform admin actions involving supplier acceptance.
     ]
 
     id = db.Column(db.Integer, primary_key=True)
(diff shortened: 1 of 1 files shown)


app/models.py
```diff
@@ -422,6 +433,9 @@ def validate_role(self, key, value):
 
     @property
     def locked(self):
+        """
+            Whether account has had too many recent login attempts
```
TheDoubleK: `too many recent failed login attempts`

app/models.py
```diff
@@ -399,7 +404,13 @@ class User(db.Model):
     password_changed_at = db.Column(db.DateTime, index=False, unique=False,
                                     nullable=False)
     logged_in_at = db.Column(db.DateTime, nullable=True)
+
+    # used to determine whether account is `locked`. field is reset upon successful login or can
+    # be reset manually to "unlock" an account.
     failed_login_count = db.Column(db.Integer, nullable=False, default=0)
+
+    # used by frontends to determine whether view access should be allowed, also
+    # used to police the validity of some operations by the api internally.
```
TheDoubleK: Interesting - I didn't really think the API cared about this internally. Can you remember where it's used (to save me doing the hard work to check it?)

app/models.py
```diff
@@ -1167,4 +1189,5 @@ def filter_null_value_fields(obj):
 
 
 def generate_new_service_id(framework_slug):
+    # FIXME framework_slug parameter ignored??
```
TheDoubleK: Yes we should fix this.  In the past there were different generation rules for different frameworks.

app/models.py
risicle: Actually "the validity of some operations" is misleading - it's places where e.g. an extra field is returned where the user is a supplier and so on mostly.

risicle: _Actually_, forget that - the place where it does that is in the `User` object serialization itself, which makes perfect sense. Can probably get away with removal of this sencence.

risicle:
Ok let's see if this todo works:

- [ ] remove "police validity" comment
- [ ] change "failed login attempts" language

app/models.py
```diff
@@ -422,6 +432,9 @@ def validate_role(self, key, value):
 
     @property
     def locked(self):
+        """
+            Whether account has had too many recent failed login attempts
+        """
```
pcraig3: Super nickpicky, but we don't care if they're recent.  

```Whether account has had too many failed login attempts``` is more accurate.

pcraig3:
Yeah, generally seems okay to me.

app/models.py
risicle: orly? ok.

TheDoubleK:
Yep, let's get this merged :+1:

PR #407, opened 2016-06-16
Add and expose framework_agreement_version field to framework model

idavidmcdonald:
Added framework_agreement_version field to framework model
Migration also sets special value of 'v1.0' for g-cloud-8 framework_agreement_version
Updated relevant test_all_frameworks_are_returned test


TheDoubleK:
Upgrade and downgrade tested.

All looks good :+1:

PR #408, opened 2016-06-20
[#121573503] API changes (2) - countersigned and signer_details fields

idavidmcdonald:
Added migration for Supplier_frameworks countersigning and signer_details fields

Added signerDetails and countersigned fields to SupplierFramework.serialize

Updated API endpoint to set/update the above fields and added cross validation between agreement_signed_at and countersigned_at fields

We decided that the signer_details field was best left as a mostly 'dumb' field which can accept any json input and won't presume anything about how it is going to be used i.e. won't do any automatic resetting of itself

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -413,11 +413,8 @@ def update_supplier_framework_details(supplier_id, framework_slug):
 
     if 'onFramework' in update_json:
         interest_record.on_framework = update_json['onFramework']
-    if 'agreementReturned' in update_json:
-        if update_json['agreementReturned']:
-            interest_record.agreement_returned_at = uniform_now
-        else:
-            interest_record.agreement_returned_at = None
+    if update_json.get('agreementReturned'):
+        interest_record.agreement_returned_at = uniform_now
     if update_json.get('countersigned'):
         interest_record.countersigned_at = uniform_now
     if 'signerDetails' in update_json:
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -418,7 +418,13 @@ def update_supplier_framework_details(supplier_id, framework_slug):
     if update_json.get('countersigned'):
         interest_record.countersigned_at = uniform_now
     if 'signerDetails' in update_json:
-        interest_record.signer_details = update_json["signerDetails"]
+        if update_json["signerDetails"] is None:
+            interest_record.signer_details = None
+        else:
+            interest_record.signer_details = interest_record.signer_details or {}
+            interest_record.signer_details.update(update_json["signerDetails"])
+            # a dummy assignment to force the validator to run. FIXME sort this out project-wide
+            interest_record.signer_details = interest_record.signer_details
 
     audit_event = AuditEvent(
         audit_type=AuditTypes.supplier_update,
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -5,7 +5,8 @@ from .. import main
 from ... import db
 from ...models import (
     Supplier, ContactInformation, AuditEvent,
-    Service, SupplierFramework, Framework
+    Service, SupplierFramework, Framework,
+    ValidationError
 )
 from ...validation import (
     validate_supplier_json_or_400,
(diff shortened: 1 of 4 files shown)


idavidmcdonald:
Related to the [API client pull request](https://github.com/alphagov/digitalmarketplace-apiclient/pull/26)

app/main/views/suppliers.py
```diff
@@ -286,6 +287,7 @@ def set_a_declaration(supplier_id, framework_slug):
     return jsonify(declaration=supplier_framework.declaration), status_code
```
TheDoubleK: I don't think we need this comment.

app/main/views/suppliers.py
```diff
@@ -408,13 +410,22 @@ def update_supplier_framework_details(supplier_id, framework_slug):
...
     if 'agreementReturned' in update_json:
         if update_json['agreementReturned']:
-            interest_record.agreement_returned_at = datetime.utcnow()
+            interest_record.agreement_returned_at = uniform_now
         else:
             interest_record.agreement_returned_at = None
+    if 'countersigned' in update_json:
+        if update_json['countersigned']:
+            interest_record.countersigned_at = uniform_now
+        else:
```
TheDoubleK: I think we should remove both the `else None` clauses from the date setting. These aren't actions that should ever be un-done add a matter of course.

Also, as this stands, when an agreement is marked as countersigned the agreement-returned-at will either be set to the current time or deleted, neither of which I'd the right thing.

TheDoubleK: Sorry, autocorrect, *as* a matter of course, *is* the right thing.

tests/app/views/test_suppliers.py
```diff
@@ -1368,18 +1386,76 @@ def test_agreement_returned_at_timestamp_cannot_be_set(self):
...
+                    'agreementReturned': True,
+                    'countersigned': True,
+                    'countersignedAt': '2013-13-13T00:00:00.000000Z',
+                }
+            )
+            assert_equal(response.status_code, 200)
+            data = json.loads(response.get_data())
+            assert_equal(data['frameworkInterest']['countersignedAt'], '2012-12-12T00:00:00.000000Z')
+
+    def test_agreement_returned_at_is_unset_when_agreement_returned_flag_is_false(self):
```
TheDoubleK: So we can do away with the unset tests, because wet shouldn't allow unsetting. But a test that setting one date leaves the value of the other date unchanged would be useful.

app/main/views/suppliers.py
risicle: Fair enough. Remove all the things and cut down on complex unnecessary behaviour.

tests/app/views/test_suppliers.py
risicle: Ok in the latest version we have tests that trigger each of the timestamps and assert that the _other_ timestamp value isn't affected in the returned value.

app/main/views/suppliers.py
```diff
@@ -408,13 +409,16 @@ def update_supplier_framework_details(supplier_id, framework_slug):
...
-        if update_json['agreementReturned']:
-            interest_record.agreement_returned_at = datetime.utcnow()
-        else:
-            interest_record.agreement_returned_at = None
+    if update_json.get('agreementReturned'):
+        interest_record.agreement_returned_at = uniform_now
+    if update_json.get('countersigned'):
+        interest_record.countersigned_at = uniform_now
+    if 'signerDetails' in update_json:
+        interest_record.signer_details = update_json["signerDetails"]
```
TheDoubleK: Thinking forward, we are going to need to add data to the `signerDetails` in future - when CCS do the countersigning.  So it might be better if this updated the existing signer details, rather than overwriting them completely, eg something like 
`interest_record.signer_details.update(update_json["signerDetails"])`

Would be nice to have a test around this too - i.e. a second call to the endpoint updating some existing details, and not losing any keys.

risicle: How would an API user "delete" a key from the json then? Would they set the key to json `null` and rely on `purge_nulls_from_data` turning that into a removal? Because we haven't added the necessary validator to the db field, trying to not overthink the field behaviour. (Some json fields have validators doing this, some don't - we guessed that ours shouldn't)

If we're going with this change should we add the json-cleaning validators to the model?

TheDoubleK: I'm not sure that we should ever need to remove a key from this field - because you can't un-sign a thing.  

But yes, I think we should probably add that validator, so that we get uniform behaviour across the API (i.e. send a null value for a key to delete it from the data).

risicle: It kinda feels like we should do be doing a "deep update" of the json here if we're going to update it but I note that no other json fields seem to get deep-updated so meh.

TheDoubleK: It depends if we have nested objects, and generally we don't (and I don't forsee that we will here either).

tests/app/views/test_suppliers.py
```diff
@@ -1356,6 +1371,26 @@ def test_adding_that_agreement_has_been_returned(self):
...
+                'digital-outcomes-and-specialists',
+                update={'countersigned': True}
+            )
+            assert_equal(response.status_code, 200)
+            data = json.loads(response.get_data())
+            assert_equal(data['frameworkInterest']['supplierId'], 0)
+            assert_equal(data['frameworkInterest']['frameworkSlug'], 'digital-outcomes-and-specialists')
+            assert_equal(data['frameworkInterest']['agreementReturned'], False)
+            assert_is(data['frameworkInterest']['agreementReturnedAt'], None)
+            assert_equal(data['frameworkInterest']['countersigned'], True)
```
pcraig3: In real life, we wouldn't expect this to happen _**but**_ it seems fair enough to test this functionality as long as our API allows it.

pcraig3:
Looks okay to me 👍

PR #409, opened 2016-06-20
Fix bug if multiple @'s in buyer email address

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/121730503) bug on Pivotal.
See [other related work](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/350) on the buyer front end.

If a buyer supplies an email address with two or more `@` symbols in, the current method for checking the domain of the address will fail with a ValueError. This modifies the method to deal with multiple @'s.

This allows a bad email address to pass through the domain check (if it has a valid domain; if not it will fail as usual), and fail gracefully when checked against the pattern defined in the json schema.

Hopefully this should never actually happen as the email address is validated in the buyer front end before this (see PR linked above).

### DIFF
--- a/app/validation.py
+++ b/app/validation.py
@@ -290,5 +290,5 @@ _BUYER_EMAIL_DOMAINS = load_buyer_email_domains('./data/buyer-email-domains.txt'
 
 
 def is_valid_buyer_email(email):
-    _, domain = email.split('@')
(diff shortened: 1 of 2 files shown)


TheDoubleK:
:+1:

PR #410, opened 2016-06-22
[#121428145] add oneoff script generate_g8_master_csv.py

risicle:
Appears to work correctly when used against staging endpoint. Don't know if this needs formal tests as it will be a oneoff? (this is https://www.pivotaltracker.com/story/show/121428145)

### DIFF
--- a/scripts/oneoff/generate_g8_master_csv.py
+++ b/scripts/oneoff/generate_g8_master_csv.py
@@ -1,0 +1,72 @@
+"""
+Generate a CSV (to stdout) with per-lot draft statistics for each supplier
+
+Usage:
+    scripts/oneoff/generate_g8_master_csv.py <data_api_url> <data_api_token>
+
+Example:
+    scripts/oneoff/generate_g8_master_csv.py http://api myToken
+"""
+import csv
+from sys import stdout
+from itertools import product, chain
+from collections import Counter, OrderedDict
+
+from docopt import docopt
+from dmapiclient import DataAPIClient
+
+_lot_slug_status_label = lambda lot_slug, status: "{}_{}".format(lot_slug, status)
+
+
+def generate_g8_master_csv(
+        data_api_url,
+        data_api_token,
+        outfile=stdout,
+        target_framework_slug="g-cloud-8",
+        target_statuses_labels=OrderedDict((
+            ("submitted", "completed",),
+            ("not-submitted", "draft",),
+        )),
+        target_lot_slugs=("saas", "paas", "iaas", "scs",),
+        ):
+    client = DataAPIClient(data_api_url, data_api_token)
+
+    writer = csv.DictWriter(outfile, lineterminator="\n", fieldnames=(
+        "supplier_id",
+        "supplier_name",
+        "application_status",
+        "declaration_status",
+        ) + tuple(_lot_slug_status_label(status, lot_slug) for status, lot_slug in product(
+            target_statuses_labels.values(), target_lot_slugs))
+    )
+    writer.writeheader()
+
+    # we iterate over each user but only need to emit each supplier once
+    seen_suppliers = set()
+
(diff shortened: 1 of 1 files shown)


TheDoubleK:
This should live in https://github.com/alphagov/digitalmarketplace-scripts - it's the new(ish) home for all our scripts, we never got around to moving the old ones from here.

I'll take a look at this now, but probably best to close this PR and move it to the other repository.

At some point we need to take an audit of and then mostly delete the scripts here, but it's never top of the priority list, you know how it is...

scripts/oneoff/generate_g8_master_csv.py
```diff
@@ -0,0 +1,72 @@
...
+        "declaration_status",
+        ) + tuple(_lot_slug_status_label(status, lot_slug) for status, lot_slug in product(
+            target_statuses_labels.values(), target_lot_slugs))
+    )
+    writer.writeheader()
+
+    # we iterate over each user but only need to emit each supplier once
+    seen_suppliers = set()
+
+    for user in client.export_users(target_framework_slug)["users"]:
```
TheDoubleK: There is also `client.get_interested_suppliers(target_framework_slug)["interestedSuppliers"]` that would give you a list of all the supplier IDs we need to do for the framework.

risicle:
Cath seems to be happy with the output of this now.

scripts/oneoff/generate_g8_master_csv.py
risicle: The reason I'm using the users endpoint is it allows me to piggyback on the `application_status` and `declaration_status` they return.

scripts/oneoff/generate_g8_master_csv.py
```diff
@@ -0,0 +1,72 @@
+"""
+Generate a CSV (to stdout) with per-lot draft statistics for each supplier
+
+Usage:
+    scripts/oneoff/generate_g8_master_csv.py <data_api_url> <data_api_token>
```
TheDoubleK: There is a handy function here https://github.com/alphagov/digitalmarketplace-scripts/blob/master/dmscripts/env.py#L1

That allows you to only need to put the name of the stage (plus valid token) as a command line arg (`dev|preview|staging|production`) rather than needing to know the full API URL.

scripts/oneoff/generate_g8_master_csv.py
```diff
@@ -0,0 +1,72 @@
...
+        target_statuses_labels=OrderedDict((
+            ("submitted", "completed",),
+            ("not-submitted", "draft",),
+        )),
+        target_lot_slugs=("saas", "paas", "iaas", "scs",),
+        ):
+    client = DataAPIClient(data_api_url, data_api_token)
+
+    writer = csv.DictWriter(outfile, lineterminator="\n", fieldnames=(
+        "supplier_id",
```
TheDoubleK: As the files are for CCS rather than developers, we could go with "nicer"* headings with capital letters and spaces.

* in the eye of the beholder, I know.

TheDoubleK:
Reviewed - couple of minor comments, but the output is right and I'm convinced it's doing all the right things.

Will thumbs it up once moved to the scripts repo.

scripts/oneoff/generate_g8_master_csv.py
TheDoubleK: Ah OK, yes I didn't notice that (or know that included that stuff in users!)

scripts/oneoff/generate_g8_master_csv.py
risicle: Cath was quite specific about the headings :shrug:

TheDoubleK: OK cool, let's stick with this.

risicle:
New PR is https://github.com/alphagov/digitalmarketplace-scripts/pull/52

PR #412, opened 2016-06-28
[PROPOSAL] mutable json validation fix through sqlalchemy.ext.mutable

risicle:
```
add, use DMMutableContainer (derived from the sqlalchemy.ext.mutable
extension) in attempt to make @validates more reliable

when @validates is used on a JSON field it doesn't pick up mutations
of the orm-mapped objects for re-validation. this implementation of
mutable containers does a little better, advertizing edits of the top-level
container to the validator, but is still not perfect as it doesn't catch
edits to deep-nested mutable objects - that might be a further exercise.
```

This is an alternate approach to #411. I was cooking up a variant that was properly recursive so didn't have the deep-nested limitations, but that looks like it will be tricky to get the `changed()` propagation to work.

The _advantage_ to this one though is that validation happens on attribute assignment (as we were assuming it did) and so should require no code changes. As can be seen by all tests passing.

### DIFF
--- a/app/fields.py
+++ b/app/fields.py
@@ -1,4 +1,7 @@
 from sqlalchemy.ext.mutable import Mutable, MutableDict
+from sqlalchemy.orm.attributes import set_attribute
+
+from six import iteritems
 
 # here we implement subclasses of json-style containers (dict, list) which advertize their mutations back to a parent
 # model instance when used for e.g. a JSON field. this allows validators to run a *little* more reliably.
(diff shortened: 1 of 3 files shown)


risicle:
(damn you pep8)

risicle:
Note last commit 60d02a1 - tests still pass following it.

risicle:
Ah... py3.

pcraig3:
This looks relevant.  It's a guy's blog post about doing almost exactly the same thing. http://variable-scope.com/posts/mutation-tracking-in-nested-json-structures-using-sqlalchemy

risicle:
Yup that takes it a step further and does the equivalent of what I've got bubbling away in my "recursive" branch to cope with deep-nested structures. Thought I'd keep it "simple" for the initial proposal...

risicle:
One thing the above link doesn't handle is circular references. An accidental circular reference in one of their containers would cause infinite recursion (eventually raise a max recursion exceeded exception)

risicle:
Hey, @pcraig3 remember I said I could come up with an implementation that had nested structure support but it would be more complicated...?

risicle:
The above implementation would require some specific tests written for it I think, though existing tests all pass for me.

PR #413, opened 2016-06-29
[#122321641] move Framework.framework_agreement_version into new framework_agreement_details json field

risicle:
migrating old framework_agreement_version values into "frameworkAgreementVersion" json attributes and emulating old serialization behaviour of framework_agreement_version.

Story https://www.pivotaltracker.com/story/show/122321641

This is going to need to be migration-rebased once #408 is merged.

Interested in feedback @allait, @pcraig3 on what the `serialize()` output should be for this. Currently I've emulated the previous api output, but was wondering if I should really just be dumping the json itself out as-is as `frameworkAgreementVersion`.

@TheDoubleK 's opinions on the matter:
```
<kevkeenoy> With other objects where we have a JSON blob of data we tend to unpack it I think.
<kevkeenoy> But this feels a bit different for some reason.
<kevkeenoy> That I can't really put my finger on.
<kevkeenoy> I guess because with eg services the data blob is the main thing.
<kevkeenoy> And other fields are essentially metadata
<kevkeenoy> But with this one the new thing isn't the whole thing - and feels like it might be better
    to leave it encapsulated as "all the countersigning info"
```

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -80,7 +80,7 @@ class Framework(db.Model):
     slug = db.Column(db.String, nullable=False, unique=True, index=True)
     name = db.Column(db.String(255), nullable=False)
     framework = db.Column(db.String(), index=True, nullable=False)
-    framework_agreement_version = db.Column(db.String(), nullable=True)
+    framework_agreement_details = db.Column(JSON, nullable=True)
     status = db.Column(db.String(),
                        index=True, nullable=False,
                        default='pending')
@@ -104,7 +104,7 @@ class Framework(db.Model):
             'name': self.name,
             'slug': self.slug,
             'framework': self.framework,
-            'frameworkAgreementVersion': self.framework_agreement_version,
+            'frameworkAgreementVersion': (self.framework_agreement_details or {}).get("frameworkAgreementVersion"),
             'status': self.status,
             'clarificationQuestionsOpen': self.clarification_questions_open,
             'lots': [lot.serialize() for lot in self.lots],
--- a/migrations/versions/670_move_framework_framework_agreement_version.py
+++ b/migrations/versions/670_move_framework_framework_agreement_version.py
@@ -1,0 +1,37 @@
+"""Move Framework.framework_agreement_version into new framework_agreement_details json field
+
+Revision ID: 670
+Revises: 660
+Create Date: 2016-06-29 11:54:27.719038
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '670'
+down_revision = '660'
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+def upgrade():
+    op.add_column('frameworks', sa.Column('framework_agreement_details', postgresql.JSON(), nullable=True))
+    op.execute("""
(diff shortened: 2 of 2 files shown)


pcraig3:
I don't really have a strong opinion on this, I think mostly because I don't know what else we're going to put into it. 
My gut reaction is that as long as we don't have a strong idea of what else is going to have to be in this thing, just return the whole thing as a nested JSON (kind of like what we're doing with `signer_details` at present).

risicle:
@pcraig3 except that's _technically_ an api-breaking change...

pcraig3:
We're the only users of our api currently. [Not sure how much it bothers me.](http://i.giphy.com/FooCq874TGUfu.gif)

risicle:
migration-rebased.

risicle:
huh... tests pass locally.

risicle:
The particularly weird thing is that the string that hypothesis is sending the BriefResponse has an `availability` that's too _long_ (101 chars) and it's returning the error code `under_character_limit`.

risicle:
ok tests pass now. weird.

TheDoubleK:
On the `under_character_limit` error message -  yes, some of our message keys (that one in particular) are counterintuitive.

risicle:
FWIW this is the test failure I've fixed in #415.

migrations/versions/660_move_framework_framework_agreement_version.py
```diff
@@ -0,0 +1,37 @@
...
+            framework_agreement_version IS NOT NULL and framework_agreement_version != ''
+    """)
+    op.drop_column('frameworks', 'framework_agreement_version')
+
+
+def downgrade():
+    op.add_column('frameworks', sa.Column('framework_agreement_version', sa.VARCHAR(), autoincrement=False, nullable=True))
+    op.execute("""
+        UPDATE frameworks SET
+            framework_agreement_version = framework_agreement_details->'frameworkAgreementVersion'
```
TheDoubleK: This downgrade doesn't quite work... for some reason the string that gets set after dereferencing `frameworkAgreementVersion` includes double quotes around the string.  I added a version for DOS to my local DB to check I wasn't going mad:
screen shot 2016-07-06 at 09 44 33

I think it needs the `->>` operator to work properly.

Other than that looking good, sorry it's taken so long to look at this.

TheDoubleK:
Yep let's go with this flat version for now and get it out.  Should be good to go with `->` changed to `->>` and a rebase.

migrations/versions/660_move_framework_framework_agreement_version.py
risicle: That's weird it worked for me but perhaps I broke it since I last tested it...

risicle: Yes I see what it was - I was just seeing the quotes in the listing and thinking "yup thats how postgres shows a string..." - should have been kicked out of the postgres conference yesterday with that kind of thinking. Doing the `->>`.

risicle:
Goddamn @idavidmcdonald stop getting your changes in and making me migration-rebase...

TheDoubleK:
Nice one :+1:

PR #414, opened 2016-07-01
Changes to signer_details

idavidmcdonald:
Renamed `signer_details` to `agreement_details` after discussion that this field should take more information about the general agreement rather than just being restricted to details about signing. This includes a database migration to do so and changes to the API and tests.

Also added back in functionality to unset agreement_returned which had previously been removed in [this pull request](https://github.com/alphagov/digitalmarketplace-api/pull/408). Functionality is now a little stricter and expects 'False' rather than something that is not true.


pcraig3:
👍  
quick quick!

PR #415, opened 2016-07-04
hypothesis example_listings: add max_size to availability and *Requirements strings

risicle:
sizes corresponding to size limits in json schemas. this should stop sporadic test failures.

cc @andrewchong, @allait (original author)

Also wondering if I should:
- expand the `alphabet` a bit to include spaces and possibly even some odd unicode characters.
- make `location` actually output one of the locations from the schema - I'm not sure how it validates with just some random strings...

### DIFF
--- a/tests/example_listings.py
+++ b/tests/example_listings.py
@@ -1,9 +1,13 @@
+# -*- coding: utf-8 -*-
 from hypothesis import settings
 from hypothesis.strategies import (
     fixed_dictionaries, lists,
     booleans, integers, text, none,
     composite, sampled_from, one_of,
     just)
+
+from functools import partial
+import re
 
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Is there a possibility that adding spaces to the dictionary could introduce more intermittent test failures if they happened to push things over a word limit in the schema?  Some unicode characters might not be a bad thing though.

risicle:
Ah yes word limit good point

risicle:
@TheDoubleK I've got around the word count problem by using hypothesis' `.filter()` feature.

TheDoubleK:
Looks good, tests still pass :)
I'm not 100% sure how these are used in the test suite, but clearly the location must never matched "for real".
:+1:

PR #416, opened 2016-07-05
agreementDetails validation and set frameworkAgreementVersion automatically

idavidmcdonald:
- Added JSON schema validation for `agreementDetails` which only allows 
`signerName`, `signerRole`, `signerUserId`, `frameworkAgreementVersion` fields, along with the required validation for each of those fields. In time `agreementDetails` will soon be used for storing countersigning fields as well but that is not in our story scope at the moment and so those fields have not yet been added to the schema. This schema validation is then used in our API endpoint.

- `agreementDetails` will only be set if the relevant framework has a `frameworkAgreementVersion`. This will allow the previous signing process to run as normal.

- The user process for adding fields to `agreementDetails` will not see all required fields set at the same time due to a multipage form / multistep process. Therefore we introduce `validate_agreement_details_data` function which allows us to override which fields are required (if not all as set in the JSON schema). This is a similar pattern that has been used in other apps. Example usage will allow us to set `signerName` and `signerRole` without having to set `uploaderUserID` until the next step.

- If `agreementReturned` is set for a supplier framework then we will set `agreementDetails.frameworkAgreementVersion` to that of the relevant framework's agreement version. 

- If `agreementReturned` is unset then we will also unset `agreementDetails`. 

- The endpoint will allow `agreementDetails.frameworkAgreementVersion` to be updated but only in the case when the agreement is already marked as returned. It is anticipated that this behaviour will not actually be needed but does allow us the ability just in case.

- Note, we used a slightly roundabout way of assigning `interest_record.agreement_details` due to SQLAlchemy not picking up internal changes to `agreementDetails`. (It may be possible to rewrite this in a more straightforward way if one of @risicle's two propsed pull requests are accepted.)

- We store the `uploaderUserId`. Later on in our apps we will use this data to show information related to the uploaderUser, such as their name and email address. So this can happen easily we have added to `SupplierFramework.serialize()` to include `uploaderUserName` and `uploaderUserEmail` in `agreementDetails` if there is a `uploaderUserId`.

- Tidying up of test code to move from using `assert_*` functions to use the `assert` keyword plainly i.e. using `assert a == b` rather than `assert_equal(a,b)`.

- Some older tests have had their method names updated so they reflect clearly what they are doing e.g. `test_pass_fail_update_creates_audit_event` to `test_changing_on_framework_to_passed_creates_audit_event`

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -422,9 +422,6 @@ def update_supplier_framework_details(supplier_id, framework_slug):
         interest_record_agreement_details_proposed = interest_record.agreement_details.copy() \
             if interest_record.agreement_details else {}
         interest_record_agreement_details_proposed.update(update_json.get('agreementDetails', {}))
-        interest_record_agreement_details_proposed = drop_foreign_fields(
-            interest_record_agreement_details_proposed, 'frameworkAgreementVersion'
-        )
 
         validate_agreement_details_data(
             interest_record_agreement_details_proposed,
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -408,6 +408,10 @@ def update_supplier_framework_details(supplier_id, framework_slug):
     if not interest_record:
         abort(404, "supplier_id '{}' has not registered interest in {}".format(supplier_id, framework_slug))
 
+    # `agreementDetails` shouldn't be passed in unless the framework has framework_agreement_details
+    if 'agreementDetails' in update_json and framework.framework_agreement_details is None:
+        abort(400, "Framework '{}' does not accept 'agreementDetails'".format(framework_slug))
+
     if (
             (framework.framework_agreement_details and framework.framework_agreement_details.get('frameworkAgreementVersion')) and  # noqa
             ('agreementDetails' in update_json or update_json.get('agreementReturned'))
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -5,18 +5,16 @@ from .. import main
 from ... import db
 from ...models import (
     Supplier, ContactInformation, AuditEvent,
-    Service, SupplierFramework, Framework,
-    ValidationError
+    Service, SupplierFramework, Framework
 )
 from ...validation import (
     validate_supplier_json_or_400,
     validate_contact_information_json_or_400,
-    validate_agreement_details_json_or_400,
     is_valid_string_or_400
 )
 from ...utils import pagination_links, drop_foreign_fields, get_json_from_request, \
     json_has_required_keys, json_has_matching_id, get_valid_page_or_1, validate_and_return_updater_request
-from ...supplier_utils import validate_and_return_supplier_request
+from ...supplier_utils import validate_and_return_supplier_request, validate_agreement_details_data
 from dmapiclient.audit import AuditTypes
 
 
@@ -402,9 +400,6 @@ def update_supplier_framework_details(supplier_id, framework_slug):
     json_has_required_keys(json_payload, ["frameworkInterest"])
     update_json = json_payload["frameworkInterest"]
 
-    if 'agreementDetails' in update_json:
-        validate_agreement_details_json_or_400(update_json['agreementDetails'])
-
     interest_record = SupplierFramework.query.filter(
         SupplierFramework.supplier_id == supplier.supplier_id,
         SupplierFramework.framework_id == framework.id
@@ -413,26 +408,46 @@ def update_supplier_framework_details(supplier_id, framework_slug):
     if not interest_record:
         abort(404, "supplier_id '{}' has not registered interest in {}".format(supplier_id, framework_slug))
 
-    uniform_now = datetime.utcnow()
+    if 'agreementDetails' in update_json:
+        required_fields = ['signerName', 'signerRole']
+        if update_json.get('agreementReturned'):
+            required_fields.append('uploaderUserId')
 
-    agreement_details = interest_record.agreement_details.copy() if interest_record.agreement_details else {}
+        # Make a copy of the existing agreement_details with our new changes to be added and validate this
+        # If invalid, 400
+        interest_record_agreement_details_proposed = interest_record.agreement_details.copy() \
+            if interest_record.agreement_details else {}
+        interest_record_agreement_details_proposed.update(update_json.get('agreementDetails'))
+
+        validate_agreement_details_data(
+            interest_record_agreement_details_proposed,
+            enforce_required=False,
+            required_fields=required_fields
+        )
+
+    uniform_now = datetime.utcnow()
 
     if 'onFramework' in update_json:
         interest_record.on_framework = update_json['onFramework']
     if 'agreementReturned' in update_json:
         if update_json["agreementReturned"] is False:
             interest_record.agreement_returned_at = None
-            if agreement_details:
-                agreement_details = None
+            interest_record.agreement_details = None
         else:
             interest_record.agreement_returned_at = uniform_now
-            agreement_details['frameworkAgreementVersion'] = framework.framework_agreement_version
     if update_json.get('countersigned'):
         interest_record.countersigned_at = uniform_now
-    if 'agreementDetails' in update_json:
-        agreement_details.update(update_json['agreementDetails'])
 
-    interest_record.agreement_details = agreement_details or None
+    # agreementDetails should only exist for g-cloud-8 and above (signified by having a framework_agreement_version)
+    if framework.framework_agreement_version:
+        agreement_details = interest_record.agreement_details.copy() if interest_record.agreement_details else {}
+
+        if update_json.get('agreementDetails'):
+            agreement_details.update(update_json['agreementDetails'])
+        if update_json.get('agreementReturned'):
+            agreement_details['frameworkAgreementVersion'] = framework.framework_agreement_version
+
+        interest_record.agreement_details = agreement_details or None
 
     audit_event = AuditEvent(
         audit_type=AuditTypes.supplier_update,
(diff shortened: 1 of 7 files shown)


app/main/views/suppliers.py
```diff
@@ -411,23 +415,24 @@ def update_supplier_framework_details(supplier_id, framework_slug):
...
 
+    agreement_details = interest_record.agreement_details.copy() if interest_record.agreement_details else {}
+
     if 'onFramework' in update_json:
         interest_record.on_framework = update_json['onFramework']
     if 'agreementReturned' in update_json:
         if update_json["agreementReturned"] is False:
             interest_record.agreement_returned_at = None
+            if agreement_details:
+                agreement_details.pop('frameworkAgreementVersion', None)
```
TheDoubleK: If we're clearing `agreementReturned` we probably want to scratch out `agreement_details` altogether (ie set to NULL), not just remove the `frameworkAgreementVersion`?

json_schemas/agreement-details.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "additionalProperties": false,
+  "minProperties": 1,
+  "properties": {
+    "signerName": {
```
TheDoubleK: Should name, role and user ID be required fields (cf required fields here: https://github.com/alphagov/digitalmarketplace-api/blob/master/json_schemas/users.json#L40)

tests/app/views/test_suppliers.py
```diff
@@ -1280,13 +1280,23 @@ def setup(self):
                 data=json.dumps({'updated_by': 'interested@example.com'}),
                 content_type='application/json')
 
+            # framework_agreement_version gets saved into the agreementDetails JSON blob of a supplier_framework
+            # when an agreement is returned
+            Framework.query.filter_by(slug='digital-outcomes-and-specialists').update(
```
TheDoubleK: Oh this is interesting... not all frameworks will have a `framework_agreement_version`.  In fact, only G-Cloud 8 will.

These "return agreement" routes all still need to work for older frameworks that don't have any framework agreement details, where there won't be any framework agreement details to refer to.

So (sorry!) it might be nice to leave the old-style returning agreements flow tests exactly as they were - to prove that the existing flow isn't broken - and then have a separate set of G-Cloud 8 returning agreements tests where there are framework agreement details.  These new tests should check the new agreement details stuff.

TheDoubleK: We don't necessarily need a full copy of all the tests for different frameworks, but anything using the framework details field should have a test both with and without the details existing.

json_schemas/agreement-details.json
idavidmcdonald: We can't think of a likely reason for the signer's email address being needed later on in the frontend app at the moment? Unless you think its more likely to happen than not, we will probably leave it for the moment.

TheDoubleK: OK maybe not then.  I was thinking of Laurence's admin app, but I think the email we will want to surface there is who clicked the "countersign" button.

idavidmcdonald: Have spoken with Laurence and looked at the admin app. You are right. In his prototype we want to show the name and the email of the user who uploaded the framework agreement. Will discuss with @pcraig3 if it is worth storing this information or getting it on the fly and come to a decision.

app/models.py
```diff
@@ -388,6 +391,17 @@ def serialize(self, data=None):
             "countersignedAt": countersigned_at,
         }, **(data or {}))
 
+        if self.agreement_details and self.agreement_details.get('uploaderUserId'):
+            user = User.query.filter(
+                User.id == self.agreement_details.get('uploaderUserId')
+            ).first()
+
+            if user:
```
TheDoubleK: Should this fail horribly if there's no user found with the user ID supplied?  Feels like that should be an error - we don't want to store "uploaderUserId" if it isn't associated with a user account.

tests/app/views/test_suppliers.py
```diff
@@ -1427,89 +1520,225 @@ def test_countersigned_at_timestamp_cannot_be_set(self):
...
+        agreement_details_payload = {
+            "signerName": "name",
+            "signerRole": "role",
+            "uploaderUserId": 1,
+        }
+        response = self.supplier_framework_update(
+            0, 'digital-outcomes-and-specialists',
+            update={'agreementDetails': agreement_details_payload})
+
+        # weird because it still returns a 200-level response even though it doesn't update anything
```
TheDoubleK: Is this the right thing to do?  Maybe it should be an error to send that info for a framework that doesn't take it, rather than just throwing away all the data and saying "OK"?

tests/app/views/test_suppliers.py
```diff
@@ -1427,89 +1520,225 @@ def test_countersigned_at_timestamp_cannot_be_set(self):
...
+        response = self.supplier_framework_update(
+            0, 'g-cloud-8',
+            update={'agreementDetails': agreement_details_payload})
+
+        data = json.loads(response.get_data())
+        assert response.status_code == 200
+        assert data['frameworkInterest']['agreementDetails'] == {
+            "signerName": "name",
+            "signerRole": "role",
+            "uploaderUserId": 999
```
TheDoubleK: Yeah this seems weird - if the user ID doesn't exist in our system we should complain I think.

app/models.py
pcraig3: There's actually a precedent for this: https://github.com/alphagov/digitalmarketplace-api/blob/master/app/models.py#L808

It will just not return the user's information if there's no user found.

pcraig3: Seems like, if we do _need_ that information (ie, we can't ignore it on the front-end), then it would make sense it would fail here rather than just having the frontend not find something it  expects to always be there.

I'll ask @laurenceberry.

tests/app/views/test_suppliers.py
pcraig3: :expressionless:

I think all of the if conditions are horrible enough already, but you're probably right.

app/models.py
pcraig3: @laurenceberry has said that we should just be doing the ignoring thing.

If you're considering the CSS admin user looking at a contract page, the choices seem to be:
- the frontend app fails with a 500 error after not finding the user's information
- there's a blank space under "Uploader name & email" in the admin app, where normally we would see a name and email.

It seems like, on the very outside chance that a user's ID was saved here but then their account gets deleted or something, that at least CCS could get some useful information out of the contract page rather than just having everything break.

pcraig3: So we're for keeping it in.

tests/app/views/test_suppliers.py
pcraig3: :/ 

I get the sentiment, but I don't think it's worth blocking the frontend app over. 

Maybe we could trigger a warning or something but keep the apps functioning normally?

app/models.py
TheDoubleK: This isn't just to do with what CCS will see though - if our app is sending a user ID that doesn't exist then something is fundamentally broken and we shouldn't proceed to save something wrong.
The case of serializing an audit event is not the same thing at all - that's deciding about what data to *show* that has already been stored in the database, and it's possible that some audit events might not have an associated user (peraps they have been deleted several years down the line).
But in this case it's about what info to save into our DB, and a user doing an action right now should definitely exist.

tests/app/views/test_suppliers.py
TheDoubleK: This *is* an update though.  I really think the update shouldn't save a bad user ID.

pcraig3: Yeah, that makes sense.  
We shouldn't save data that we know is bad at the point where we're saving it.

app/main/views/suppliers.py
```diff
@@ -409,25 +408,52 @@ def update_supplier_framework_details(supplier_id, framework_slug):
...
 
+    # `agreementDetails` shouldn't be passed in unless the framework has framework_agreement_details
+    if 'agreementDetails' in update_json and framework.framework_agreement_details is None:
+        abort(400, "Framework '{}' does not accept 'agreementDetails'".format(framework_slug))
+
+    if (
+            (framework.framework_agreement_details and framework.framework_agreement_details.get('frameworkAgreementVersion')) and  # noqa
+            ('agreementDetails' in update_json or update_json.get('agreementReturned'))
+    ):
+        if update_json.get('agreementDetails') and update_json['agreementDetails'].get('uploaderUserId'):
```
allait: Would this check be easier after `validate_agreement_details_data` since we know that data is there and user_id is integer etc?

pcraig3: We decided to check this before doing anything else because we only want to check for users when an id is actually passed in.  We wanted to avoid checking a user exists every time an update is submitted.

We _could_ move all of this code to line 442 (so that we know anything in the `upload_data` would have already been validated) but we would still have to check for the existence of an `uploaderUserId` in the passed-in data.

tests/app/views/test_suppliers.py
```diff
@@ -1427,89 +1520,226 @@ def test_countersigned_at_timestamp_cannot_be_set(self):
...
+            update={'agreementDetails': agreement_details_payload})
+
+        assert response.status_code == 400
+        data = json.loads(response.get_data())
+        strings_we_expect_in_the_error_message = [
+            'Framework', 'digital-outcomes-and-specialists', 'does not accept',  'agreementDetails']
+        for error_string in strings_we_expect_in_the_error_message:
+            assert error_string in data['error']
+
+    def test_setting_agreement_details_with_nonexistent_user_id_doesnt_return_user_details(self):
```
TheDoubleK: Nitpick... this test name isn't quite right any more... something along the lines of
`test_can_not_set_agreement_details_with_nonexistent_user_id`?

TheDoubleK:
minor nitpick, but I think this is good now :+1:

PR #418, opened 2016-07-18
Add upload_signed_agreement audit event type

allait:
Updates apiclient with the new audit event type.


idavidmcdonald:
👍 when tests green

PR #419, opened 2016-07-21
[#125881031]list_briefs view: add support for filtering by multiple (comma separated) lots & frameworks

risicle:
Though we don't have multiple brief-capable frameworks yet, code added to support multiple
framework filtering for reasons of symmetry, reducing the number of different implementations
of parameter handling we're using in this view.

This is for story https://www.pivotaltracker.com/story/show/125881031 "Filter catalogue of opportunities"

### DIFF
--- a/tests/app/views/test_briefs.py
+++ b/tests/app/views/test_briefs.py
@@ -521,6 +521,18 @@ class TestBriefs(BaseApplicationTest):
         assert res.status_code == 200
         assert len(data['briefs']) == 5, data['briefs']
 
+    def test_list_briefs_by_multiple_frameworks(self):
+        self.setup_dummy_briefs(1, status='draft')
+        self.setup_dummy_briefs(3, status='live', brief_start=2)
+
+        # we don't have multiple brief-capable frameworks in the test dataset yet so this test is a bit of a stub
+        # until we do. at least it tests framework slugs are properly split
+        res = self.client.get('/briefs?framework=digital-outcomes-and-specialists,analogue-outcomes-and-specialists')
+        data = json.loads(res.get_data(as_text=True))
+
+        assert res.status_code == 200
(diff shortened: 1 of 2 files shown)


tests/app/views/test_briefs.py
```diff
@@ -521,6 +521,18 @@ def test_list_briefs_by_framework(self):
         assert res.status_code == 200
         assert len(data['briefs']) == 5, data['briefs']
 
+    def test_list_briefs_by_multiple_frameworks(self):
+        self.setup_dummy_briefs(1, status='draft')
+        self.setup_dummy_briefs(3, status='live', brief_start=2)
+
+        # we, um, don't have multiple brief-capable frameworks in the test dataset yet so this test is a bit of a stub
```
TheDoubleK: I agree having a comment makes sense here, but it could be less informal in style.  I was going to suggest it would be nice to setup multiple brief-capable frameworks in the tests, but on reflection it's a bit of a headache for little gain at this point.

tests/app/views/test_briefs.py
```diff
@@ -552,6 +564,28 @@ def test_list_briefs_by_lot(self):
...
+        res = self.client.get('/briefs?lot=digital-specialists,user-research-participants')
+        data = json.loads(res.get_data(as_text=True))
+
+        assert res.status_code == 200
+        assert len(data['briefs']) == 5, data['briefs']
+        assert all(
+            (brief["lotSlug"] in ("digital-specialists", "user-research-participants",)) for brief in data['briefs']
+        )
+
+        # bonus test that malformed lot choices string still gives the same result
```
TheDoubleK: I don't think we need this comment so much though.

TheDoubleK:
Small comments left but essentially :+1:

PR #422, opened 2016-07-25
Specialist briefs can be open for 1 or 2 weeks.

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/123706751) story on Pivotal.

Will require PR's on [utils](https://github.com/alphagov/digitalmarketplace-utils/pull/275) and [apiclient](https://github.com/alphagov/digitalmarketplace-apiclient/pull/30) to be accepted first.

This PR is to update the brief model to use the new dates package in dmutils.

It also updates the json schema to allow the new requirementsLength key for digital specialist briefs.

It also allows the api to return a list of briefs in a human readable order by passing in a new flag. This is for the buyer frontend catalogue of briefs, which is now ordered by live briefs and then closed briefs. Within those groups they're ordered by published date. This is done from the api so that pagination doesn't break.

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -109,7 +109,11 @@ def get_brief(brief_id):
 
 @main.route('/briefs', methods=['GET'])
 def list_briefs():
-    briefs = Brief.query.order_by(Brief.published_at.desc(), Brief.id)
+    if request.args.get('human'):
+        briefs = Brief.query.order_by(Brief.status.desc(), Brief.published_at.desc(), Brief.id)
+    else:
+        briefs = Brief.query.order_by(Brief.published_at.desc(), Brief.id)
+
     page = get_valid_page_or_1()
 
     user_id = get_int_or_400(request.args, 'user_id')
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -112,7 +112,7 @@ def list_briefs():
     if request.args.get('human'):
         briefs = Brief.query.order_by(Brief.status.desc(), Brief.published_at.desc(), Brief.id)
     else:
-        briefs = Brief.query.order_by(Brief.published_at.desc(), Brief.id)
+        briefs = Brief.query.order_by(Brief.id)
 
     page = get_valid_page_or_1()
 
--- a/app/models.py
+++ b/app/models.py
@@ -931,16 +931,9 @@ class Brief(db.Model):
     def applications_closed_at(self):
         if self.published_at is None:
             return None
+        brief_publishing_date_and_length = self._build_date_and_length_data()
 
-        published_day = self.published_at.replace(hour=23, minute=59, second=59, microsecond=0)
-        requirements_length = self.data.get('requirementsLength')
-        data = {
-            'publishedAt': published_day,
-            'requirementsLength': requirements_length
-        }
-        closing_date = get_publishing_dates(data)['closing_date']
-
-        return closing_date
+        return get_publishing_dates(brief_publishing_date_and_length)['closing_date']
 
     @applications_closed_at.expression
     def applications_closed_at(cls):
@@ -954,31 +947,17 @@ class Brief(db.Model):
     def clarification_questions_closed_at(self):
         if self.published_at is None:
             return None
+        brief_publishing_date_and_length = self._build_date_and_length_data()
 
-        published_day = self.published_at.replace(hour=23, minute=59, second=59, microsecond=0)
-        requirements_length = self.data.get('requirementsLength')
-        data = {
-            'publishedAt': published_day,
-            'requirementsLength': requirements_length
-        }
-        closing_date = get_publishing_dates(data)['questions_close']
-
-        return closing_date
+        return get_publishing_dates(brief_publishing_date_and_length)['questions_close']
 
     @hybrid_property
     def clarification_questions_published_by(self):
         if self.published_at is None:
             return None
+        brief_publishing_date_and_length = self._build_date_and_length_data()
 
-        published_day = self.published_at.replace(hour=23, minute=59, second=59, microsecond=0)
-        requirements_length = self.data.get('requirementsLength')
-        data = {
-            'publishedAt': published_day,
-            'requirementsLength': requirements_length
-        }
-        closing_date = get_publishing_dates(data)['answers_close']
-
-        return closing_date
+        return get_publishing_dates(brief_publishing_date_and_length)['answers_close']
 
     @hybrid_property
     def clarification_questions_are_closed(self):
@@ -1034,6 +1013,15 @@ class Brief(db.Model):
         current_data.update(data)
 
         self.data = current_data
+
+    def _build_date_and_length_data(self):
+        published_day = self.published_at.replace(hour=23, minute=59, second=59, microsecond=0)
+        requirements_length = self.data.get('requirementsLength')
+
+        return {
+            'publishedAt': published_day,
+            'requirementsLength': requirements_length
+        }
 
     def serialize(self, with_users=False):
         data = dict(self.data.items())
--- a/app/models.py
+++ b/app/models.py
@@ -939,9 +939,9 @@ class Brief(db.Model):
     def applications_closed_at(cls):
         return sql_case([
             (cls.data['requirementsLength'].astext == '1 week', func.date_trunc('day', cls.published_at) + sql_cast(
-                '7 days 23:59:59', INTERVAL)),
+                '1 week 23:59:59', INTERVAL)),
             ], else_=func.date_trunc('day', cls.published_at) + sql_cast(
-                '14 days 23:59:59', INTERVAL))
+                '2 weeks 23:59:59', INTERVAL))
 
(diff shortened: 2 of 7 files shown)


tests/app/test_model.py
```diff
@@ -181,22 +181,62 @@ def test_applications_closed_at_is_none_for_drafts(self):
...
+    def test_closing_dates_are_set_with_published_at_when_no_requirementsLength(self):
         brief = Brief(data={}, framework=self.framework, lot=self.lot,
                       published_at=datetime(2016, 3, 3, 12, 30, 1, 2))
 
         assert brief.applications_closed_at == datetime(2016, 3, 17, 23, 59, 59)
         assert brief.clarification_questions_closed_at == datetime(2016, 3, 10, 23, 59, 59)
         assert brief.clarification_questions_published_by == datetime(2016, 3, 16, 23, 59, 59)
 
-    def test_query_brief_applications_closed_at_date(self):
+    def test_closing_dates_are_set_with_published_at_when_requirementsLength_is_two_weeks(self):
```
idavidmcdonald: `requirements_length` instead of `requirementsLength`

tests/app/test_model.py
```diff
@@ -181,22 +181,62 @@ def test_applications_closed_at_is_none_for_drafts(self):
...
+        brief = Brief(data={'requirementsLength': '2 weeks'}, framework=self.framework, lot=self.lot,
+                      published_at=datetime(2016, 3, 3, 12, 30, 1, 2))
+
+        assert brief.applications_closed_at == datetime(2016, 3, 17, 23, 59, 59)
+        assert brief.clarification_questions_closed_at == datetime(2016, 3, 10, 23, 59, 59)
+        assert brief.clarification_questions_published_by == datetime(2016, 3, 16, 23, 59, 59)
+
+    def test_closing_dates_are_set_with_published_at_when_requirementsLength_is_one_week(self):
+        brief = Brief(data={'requirementsLength': '1 week'}, framework=self.framework, lot=self.lot,
```
idavidmcdonald: requirements_length instead of requirementsLength

idavidmcdonald:
Apart from two nitpicky camel case comments, I think this looks good. 

After camel case changes done it will need the related dependencies to be merged, tests to pass but then it should be good 👍

json_schemas/briefs-digital-outcomes-and-specialists-digital-specialists.json
```diff
@@ -188,6 +194,7 @@
     "numberOfSuppliers",
     "organisation",
     "priceWeighting",
+    "requirementsLength",
```
idavidmcdonald: Does adding this new required field have any knock on effects for briefs which have already been published? If we need to edit them again our validation would fail.
Should we migrate already published briefs to include this field?

app/models.py
```diff
@@ -930,36 +931,53 @@ def applications_closed_at(self):
...
-        return self.applications_closed_at - timedelta(days=self.CLARIFICATION_QUESTIONS_PUBLISHED_DAYS)
+        published_day = self.published_at.replace(hour=23, minute=59, second=59, microsecond=0)
+        requirements_length = self.data.get('requirementsLength')
+        data = {
+            'publishedAt': published_day,
+            'requirementsLength': requirements_length
+        }
+        closing_date = get_publishing_dates(data)['answers_close']
+
+        return closing_date
```
pcraig3: This bit of code is almost identical to that used in both `applications_closed_at` and `clarification_questions_closed_at`.  Could you stick the code somewhere more general and then just implement the specific bit that's needed in each of these functions?

json_schemas/briefs-digital-outcomes-and-specialists-digital-specialists.json
pcraig3: (I think we should as a matter of correctness, even though I can't think of a clear example of us validating briefs again after they're published.)

allait: I'd do a migration. That's what we've been doing so far and I think in the long run it helps to keep related code simpler, as it doesn't have to deal with missing fields/different data types and so on.

Keeping in mind the decision we've made about separate migrations I think we can release this first and then do a migration for published briefs. If we do it the other way around there's a risk of a new specialist being published after the migration but before the new field is added.

This also affects unpublished briefs – for example if a brief was ready to publish before now it will requires more information. So it might be worth checking if any comms have to go out beforehand, if you haven't already.

Wynndow: How would I go about doing a migration? And does it just mean giving all published briefs a requirementsLength of 2 weeks?

app/models.py
Wynndow: @pcraig3 have extracted the code out to a separate function in the model. What do you think?

tests/app/test_model.py
Wynndow: yep, changed.

tests/app/test_model.py
Wynndow: changed!

app/main/views/briefs.py
```diff
@@ -109,7 +109,11 @@ def get_brief(brief_id):
 
 @main.route('/briefs', methods=['GET'])
 def list_briefs():
-    briefs = Brief.query.order_by(Brief.published_at.desc(), Brief.id)
+    if request.args.get('human'):
+        briefs = Brief.query.order_by(Brief.status.desc(), Brief.published_at.desc(), Brief.id)
+    else:
+        briefs = Brief.query.order_by(Brief.published_at.desc(), Brief.id)
```
allait: I think non-human sort order would be only `Brief.id`, since publishing could move a brief between pages, reordering everything else

json_schemas/briefs-digital-outcomes-and-specialists-digital-specialists.json
allait: Yeah, I think setting all published ones to 2 weeks is what we want.

Running `python application.py db migrate --rev-id <number> -m <message>` should create a new migration file. We use numeric migration IDs and increment each number by 10 (so if the current version is 680 then new migration revision would be 690).

[Here's an example of a migration that changes a brief field](https://github.com/alphagov/digitalmarketplace-api/blob/master/migrations/versions/610_change_numberofsuppliers_to_integer.py).

app/main/views/briefs.py
Wynndow: Cool, makes sense. Have changed to just order by id.

tests/app/views/test_briefs.py
```diff
@@ -596,6 +598,24 @@ def test_list_briefs_by_lot_and_status(self):
...
+        self.setup_dummy_briefs(1, data={'requirementsLength': '2 weeks'}, status=None, title='Second',
+                                published_at=(today + timedelta(days=-9)), brief_start=2)
+        self.setup_dummy_briefs(1, data={'requirementsLength': '1 week'}, status=None, title='Third',
+                                published_at=(today + timedelta(days=-8)), brief_start=3)
+        self.setup_dummy_briefs(1, data={'requirementsLength': '1 week'}, status=None,  title='Fourth',
+                                published_at=(today + timedelta(days=-2)), brief_start=4)
+
+        res = self.client.get('/briefs?human=True')
+        data = json.loads(res.get_data(as_text=True))
+        titles = map(lambda brief: brief['title'], data['briefs'])
```
pcraig3: This is going to break in Python3 😱 

It seems like [you want to wrap a `list()` around it](http://stackoverflow.com/questions/1303347/getting-a-map-to-return-a-list-in-python-3-x).

Wynndow: Good shout. Fixed it.

json_schemas/briefs-digital-outcomes-and-specialists-digital-specialists.json
Wynndow: I've got a PR in for the migration [here](https://github.com/alphagov/digitalmarketplace-api/pull/423). Wasn't sure about the downgrade though...

app/models.py
```diff
@@ -929,37 +930,33 @@ def validates_data(self, key, data):
...
-        return closing_time
+        return get_publishing_dates(brief_publishing_date_and_length)['closing_date']
 
     @applications_closed_at.expression
     def applications_closed_at(cls):
-        return func.date_trunc('day', cls.published_at) + sql_cast(
-            '%d days 23:59:59' % cls.APPLICATIONS_OPEN_DAYS, INTERVAL
-        )
+        return sql_case([
+            (cls.data['requirementsLength'].astext == '1 week', func.date_trunc('day', cls.published_at) + sql_cast(
```
allait: There's a chance that `INTERVAL` understands weeks directly, might be worth checking if you haven't yet. Would make it easier to add additional lengths in the future.

allait:
Small note about the interval, but everything else looks good.

Should be good to go once the utils change is merged and Travis is happy 👍

app/models.py
Wynndow: Cool, all done

PR #423, opened 2016-07-28
Create migration for specialist requirementLengths

Wynndow:
This PR is part of [this](https://www.pivotaltracker.com/story/show/123706751) story on Pivotal and has come about due to [this](https://github.com/alphagov/digitalmarketplace-api/pull/422) PR.

The migration is to add a `requirementsLength` value to published specialists briefs.

Currently the downgrade is blank as without adding a new flag somewhere, I'm not sure how the migrated briefs could be differentiated from actual briefs. Any advice more than welcome!


allait:
Yeah, I don't think we can do anything with the downgrade in this case, since the API change goes out first. 👍

PR #424, opened 2016-07-29
[#125881031]list_briefs view: add "meta"->"total" attribute to response for returning total result count

risicle:
update tests accordingly

Forgot I would need this for story https://www.pivotaltracker.com/story/show/125881031

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -139,20 +139,24 @@ def list_briefs():
     if user_id:
         return jsonify(
             briefs=[brief.serialize() for brief in briefs.all()],
-            links=dict()
+            links={},
         )
     else:
         briefs = briefs.paginate(
             page=page,
-            per_page=current_app.config['DM_API_BRIEFS_PAGE_SIZE'])
+            per_page=current_app.config['DM_API_BRIEFS_PAGE_SIZE'],
+        )
 
         return jsonify(
             briefs=[brief.serialize() for brief in briefs.items],
+            meta={
+                "total": briefs.total,
+            },
             links=pagination_links(
                 briefs,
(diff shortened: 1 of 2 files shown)


app/main/views/briefs.py
```diff
@@ -136,23 +136,31 @@ def list_briefs():
...
+
     if user_id:
         return jsonify(
             briefs=[brief.serialize() for brief in briefs.all()],
-            links=dict()
+            meta=meta,
+            links={},
         )
     else:
+        total_count = briefs.count()
```
allait: I don't think this is used anywhere

app/main/views/briefs.py
```diff
@@ -136,23 +136,31 @@ def list_briefs():
             status.strip() for status in request.args['status'].split(',')
             ))
 
+    meta = {
+        "total": briefs.count(),
```
allait: Pagination objects have a [`.total` attribute](http://flask-sqlalchemy.pocoo.org/2.1/api/#flask.ext.sqlalchemy.Pagination.total), so I think we can avoid doing a separate count.

app/main/views/briefs.py
risicle: Yup this should have got scrubbed

TheDoubleK:
Looks good after a rebase :+1:

pcraig3:
Should we do this for services too?  
@TheDoubleK, @andrewchong and I were counting a lot of services today by figuring out the number of pages and multiplying them by 100.

risicle:
@pcraig3 perhaps, but the services are fetched through the search api, so separate job. Actially I was talking to you about pagination on the search api the other day, wasn't I?

PR #421, opened 2016-07-25
[#119728323] Add migration to add `withdrawn_at` field to Brief model

idavidmcdonald:
Migration for [this story](https://www.pivotaltracker.com/story/show/119728323) to withdraw a publish brief.

This is a separate PR just for the migration, the rest of the story will follow as discussed last week.

PR #425, opened 2016-07-29
[#119728323] Withdraw a published brief

idavidmcdonald:
Pivotal story: [https://www.pivotaltracker.com/story/show/119728323](https://www.pivotaltracker.com/story/show/119728323)

#### Adds functionality for a brief's status to be "withdrawn"
The required database migration to add a new `withdrawn_at` field was added separately in [this previous pull request](https://github.com/alphagov/digitalmarketplace-api/pull/421) and has already been merged. In this PR, we add logic to the model for when and how a brief's status should be set as "withdrawn". We also add logic so we can run SQL queries that filter for briefs that are withdrawn.

#### Deprecates `update_brief_status` route and introduce `update_brief_status_by_action` route to handle publishing and withdrawing of briefs
We previously allowed a flexible way of setting a brief's status, using the `update_brief_status` route, where a payload contained the value of the new status you want to set. This flexibility was used in cases such as withdrawing a brief the old way (status from 'live' to 'draft'). Now we introduce a more formal way to withdraw a brief (status from 'live' to 'withdrawn'), we no longer want to be able to go from 'live' to 'draft'. Instead of having layers of logic to decide if we are able to set the requested status, we will introduce explicit routes to publish and withdraw briefs as these are the only two expected actions.

We leave the `update_brief_status` route intact (with a deprecated comment) for backwards compatibility. Usage of this route should be replaced by the newer `update_brief_status_by_action` route. After all usage is transferred to the newer route then we can remove the deprecated route.

#### Introduce `copy` brief route
There is a user need to republish a brief that is very similar to it's withdrawn original. Previously this was sometimes facilitated as briefs with no responses would have their status set to "draft" when being withdrawn. Now this is no longer allowed, we add a route to make a draft copy of a brief. This route has been made as a general copy route that will make a draft copy of any brief (regardless of the original's status).

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -160,8 +160,47 @@ def list_briefs():
         )
 
 
+@main.route('/briefs/<int:brief_id>/status', methods=['PUT'])
+def update_brief_status(brief_id):
+    """Route is deprecated. Use `update_brief_status_by_action` instead."""
+    updater_json = validate_and_return_updater_request()
+
+    json_payload = get_json_from_request()
+    json_has_required_keys(json_payload, ['briefs'])
+    brief_json = json_payload['briefs']
+    json_has_required_keys(brief_json, ['status'])
+
+    brief = Brief.query.filter(
+        Brief.id == brief_id
+    ).first_or_404()
+
+    if brief.framework.status != 'live':
+        abort(400, "Framework is not live")
+
+    if brief_json['status'] != brief.status:
+        brief.status = brief_json['status']
+
+        validate_brief_data(brief, enforce_required=True)
+
+        audit = AuditEvent(
+            audit_type=AuditTypes.update_brief_status,
+            user=updater_json['updated_by'],
+            data={
+                'briefId': brief.id,
+                'briefStatus': brief.status,
+            },
+            db_object=brief,
+        )
+
+        db.session.add(brief)
+        db.session.add(audit)
+        db.session.commit()
+
+    return jsonify(briefs=brief.serialize()), 200
+
+
 @main.route('/briefs/<int:brief_id>/<any(publish, withdraw):action>', methods=['PUT'])
-def update_brief_status(brief_id, action):
+def update_brief_status_by_action(brief_id, action):
     updater_json = validate_and_return_updater_request()
 
     brief = Brief.query.filter(
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -160,14 +160,9 @@ def list_briefs():
         )
 
 
-@main.route('/briefs/<int:brief_id>/status', methods=['PUT'])
+@main.route('/briefs/<int:brief_id>/publish', methods=['PUT'])
 def update_brief_status(brief_id):
     updater_json = validate_and_return_updater_request()
-
-    json_payload = get_json_from_request()
-    json_has_required_keys(json_payload, ['briefs'])
-    brief_json = json_payload['briefs']
-    json_has_required_keys(brief_json, ['status'])
 
     brief = Brief.query.filter(
         Brief.id == brief_id
@@ -176,8 +171,8 @@ def update_brief_status(brief_id):
     if brief.framework.status != 'live':
         abort(400, "Framework is not live")
 
-    if brief_json['status'] != brief.status:
-        brief.status = brief_json['status']
+    if brief.status != 'live':
+        brief.status = 'live'
 
         validate_brief_data(brief, enforce_required=True)
 
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -160,8 +160,8 @@ def list_briefs():
         )
 
 
-@main.route('/briefs/<int:brief_id>/publish', methods=['PUT'])
-def update_brief_status(brief_id):
+@main.route('/briefs/<int:brief_id>/<any(publish, withdraw):action>', methods=['PUT'])
+def update_brief_status(brief_id, action):
     updater_json = validate_and_return_updater_request()
 
     brief = Brief.query.filter(
@@ -171,16 +171,23 @@ def update_brief_status(brief_id):
     if brief.framework.status != 'live':
         abort(400, "Framework is not live")
 
-    if brief.status != 'live':
-        brief.status = 'live'
+    action_to_status = {
+        'publish': 'live',
+        'withdraw': 'withdrawn'
+    }
+    if brief.status != action_to_status[action]:
+        previousStatus = brief.status
+        brief.status = action_to_status[action]
 
-        validate_brief_data(brief, enforce_required=True)
+        if action == 'publish':
+            validate_brief_data(brief, enforce_required=True)
 
         audit = AuditEvent(
             audit_type=AuditTypes.update_brief_status,
             user=updater_json['updated_by'],
             data={
                 'briefId': brief.id,
+                'briefPreviousStatus': previousStatus,
                 'briefStatus': brief.status,
             },
             db_object=brief,
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -199,7 +199,7 @@ def update_brief_status(brief_id):
     return jsonify(briefs=brief.serialize()), 200
 
 
-@main.route('/briefs/<int:brief_id>/<any(publish, withdraw):action>', methods=['PUT'])
+@main.route('/briefs/<int:brief_id>/<any(publish, withdraw):action>', methods=['POST'])
 def update_brief_status_by_action(brief_id, action):
     updater_json = validate_and_return_updater_request()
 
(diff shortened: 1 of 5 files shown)


app/main/views/briefs.py
```diff
@@ -194,6 +195,90 @@ def update_brief_status(brief_id):
     return jsonify(briefs=brief.serialize()), 200
 
 
+@main.route('/briefs/<int:brief_id>/<any(publish, withdraw):action>', methods=['PUT'])
```
allait: We've used `POST` for publishing drafts endpoint, I think it makes sense to use it here too

app/main/views/briefs.py
```diff
@@ -194,6 +195,90 @@ def update_brief_status(brief_id):
...
+        abort(400, "Framework is not live")
+
+    action_to_status = {
+        'publish': 'live',
+        'withdraw': 'withdrawn'
+    }
+    if brief.status != action_to_status[action]:
+        brief.status = action_to_status[action]
+
+        validate_brief_data(brief, enforce_required=True)
```
allait: I think we only require this when going from draft to published. I'm not necessarily against rerunning the validation when withdrawing a brief, but it would commit us to _always_ running migrations on published briefs when we change the schemas.

app/main/views/briefs.py
```diff
@@ -194,6 +195,90 @@ def update_brief_status(brief_id):
...
+
+        audit = AuditEvent(
+            audit_type=AuditTypes.update_brief_status,
+            user=updater_json['updated_by'],
+            data={
+                'briefId': brief.id,
+                'briefStatus': brief.status,
```
allait: We should record the previous status here too, now that there's more than one option

app/main/views/briefs.py
```diff
@@ -194,6 +195,90 @@ def update_brief_status(brief_id):
...
+    ).first_or_404()
+
+    if brief.framework.status != 'live':
+        abort(400, "Framework is not live")
+
+    if brief.status != 'withdrawn':
+        abort(400, "Brief status is not 'withdrawn'")
+
+    new_brief_data = brief.data
+    new_brief_data['title'] = 'Copy of ' + brief.data['title']
```
allait: I suggest we follow the approach we took with drafts and encapsulate this logic in the [model's `.copy` method](https://github.com/alphagov/digitalmarketplace-api/blob/master/app/models.py#L741). When it comes to the title change, drafts append `" copy"` instead of `"Copy of "` and only when the addition won't overflow the maxLength set for the title field by the schema (otherwise the validation step will fail). I think we should do the same here.

app/main/views/briefs.py
```diff
@@ -194,6 +195,90 @@ def update_brief_status(brief_id):
...
+        )
+
+        db.session.add(brief)
+        db.session.add(audit)
+        db.session.commit()
+
+    return jsonify(briefs=brief.serialize()), 200
+
+
+@main.route('/briefs/<int:brief_id>/create-draft-from-withdrawn-brief', methods=['POST'])
```
allait: Can we call this `/copy` instead? The restriction on "withdrawn to draft" seems like it's not essential for the action and is similar to how we copy submitted drafts to non-submitted. We can keep it, or make sure we set the copy status to "draft" whatever the source.

app/main/views/briefs.py
```diff
@@ -194,6 +195,90 @@ def update_brief_status(brief_id):
...
+
+    new_brief_data = brief.data
+    new_brief_data['title'] = 'Copy of ' + brief.data['title']
+    new_brief = Brief(
+        data=new_brief_data,
+        framework=brief.framework,
+        lot=brief.lot,
+        users=brief.users
+    )
+    validate_brief_data(new_brief, enforce_required=True)
```
allait: The new brief is a draft and we didn't accept any additional data (apart from the title change we've applied), so we're mostly revalidating the original brief again. I don't think we have to do this.

The draft brief will have to go through validation when published, so the only difference is whether we fail to make a copy or the created draft has validation issues (which should be fine for missing fields, but could be a bit trickier for invalid values like overflowing title as we don't show a very useful error message when the validation fails on a field that doesn't belong to the current form)

app/models.py
```diff
@@ -967,19 +967,28 @@ def clarification_questions_are_closed(self):
     def status(self):
         if self.published_at is None:
             return 'draft'
+        elif self.withdrawn_at:
```
allait: This is a "should never happen" scenario, but I think `withdrawn_at` should be checked first. If we consider a brief that for some reason has a `withdran_at` but no `published_at`:
* With current logic, brief status is set to draft, which means the brief can be published, at which point it will get a `published_at` and go straight to withdrawn.
* If we make `withdrawn_at` the first check then the brief is considered withdrawn. For a withdrawn brief `published_at` is mostly informational, so we're still missing some data, but I'd say the status is correct.

app/models.py
```diff
@@ -989,6 +998,7 @@ def status(self, value):
     def status(cls):
         return sql_case([
             (cls.published_at.is_(None), 'draft'),
+            (cls.withdrawn_at.isnot(None), 'withdrawn'),
```
allait: Similar to the python `.status` implementation, this should be the first check

app/models.py
```diff
@@ -967,19 +967,28 @@ def clarification_questions_are_closed(self):
...
             return 'live'
         else:
             return 'closed'
 
     @status.setter
     def status(self, value):
-        if value == self.status:
+        if self.status == value:
             return
+        elif self.status == 'withdrawn':
```
allait: I think if we look at the possible status changes that we allow there are only two:

1. draft to live
2. live to withdrawn

(Live to draft is no longer valid since it was decided in the kickoff to withdraw everything. Closed and withdrawn briefs can't change status. Draft to withdrawn is pointless. No status can be changed to 'closed'.)

With that in mind, I think the best way to simplify this logic is to explicitly cover the two valid cases:

```python
# Removing the short-circuit broke some tests in the past
# might be worth checking if we can get rid of it now
if value == self.status:
    return

if value == 'live' and self.status == 'draft':
    self.published_at = datetime.utcnow()
elif value == 'withdrawn' and self.status == 'live':
    self.withdrawn_at = datetime.utcnow()
else:
    raise ValidationError("Cannot change brief status from '{}' to '{}'".format(self.status, value))
```

app/main/views/briefs.py
idavidmcdonald: Sorry, just looking to clarify this comment as not entirely sure what you are saying.

Are you saying that this route should 
- only be allowed for creating a draft brief from a withdrawn brief? 
- be more general and take any brief and make a copy which is a draft?

And if we wish to have some kind of restriction on the brief status, should this restriction be in the model or just in this route?

allait: There are two points here:

1. No matter what the restriction is I think the route should be called `/copy`, as that's the actual action
2. I don't see a reason to restrict it to 'withdrawn', so I'd say take any brief and make a 'draft' copy of it.

app/main/views/briefs.py
```diff
@@ -162,6 +162,7 @@ def list_briefs():
 
 @main.route('/briefs/<int:brief_id>/status', methods=['PUT'])
 def update_brief_status(brief_id):
+    """Route is deprecated. Use `update_brief_status_by_action` instead."""
```
risicle: Wonder if we should log a warning message in cases like this so an administrator can know if an old interface is still being used.

app/models.py
```diff
@@ -965,7 +965,9 @@ def clarification_questions_are_closed(self):
 
     @hybrid_property
     def status(self):
-        if self.published_at is None:
+        if self.withdrawn_at:
```
risicle: Any reason you're using `withdrawn_at`'s truthiness here instead of `is None` as is used against `published_at`?

tests/app/test_model.py
```diff
@@ -161,6 +162,20 @@ def test_query_live_brief(self):
 
             assert Brief.query.filter(Brief.status == 'draft').count() == 0
             assert Brief.query.filter(Brief.status == 'live').count() == 1
+            assert Brief.query.filter(Brief.status == 'withdrawn').count() == 0
```
risicle: Would be nice to test in a few places whether the python implementation and `expression` implementation of `status` always agree (these lines only test the `expression` implementation).

app/models.py
idavidmcdonald: Have changed to be slightly more consistent.

app/main/views/briefs.py
idavidmcdonald: Good idea but probably wont be necessary (looks like this route is only being used once across our apps so not much to keep track of)

tests/app/test_model.py
idavidmcdonald: Have added some simple checks for this, thanks.

allait:
:+1:🎉

PR #427, opened 2016-08-01
Wonderful new tests for the service indexing function.

pcraig3:
Trying to write tests for the index method,
which previously had no tests.

First problem I ran into was that the tests are testing
an actual `Service` object, so they I couldn’t just use
the existing DOS/G-Cloud fixtures.

So I went for the method to create dummy services.

Then, I needed a G-Cloud 8 framework, so I created a class
that inherits from `BaseApplicationTest` but adds a G-Cloud 8
framework in the setup and removes it in the teardown.

Unfortunately, my tests were *still* failing because the
initial statuses of the frameworks were all wrong.
G5 and G6 are the only live frameworks; G7, DOS, and G8 are
in standstill, pending, and live, respectively.

So this is where I completely lost the plot.
I expanded the test class in this really clunky way to:
- override all of the initial framework statuses
- revert them all to their initial statuses in the teardown

If we start changing stuff about frameworks between
tests, then other tests start to fail.

Basically, this is the fundamental problem.
Our tests are all sharing the same set of `Framework` database
objects, which we used to set and update in migrations but we
don't any longer.  Now that we've stopped, the default state of
all of our frameworks no longer reflect the world as it
presently exists.

It’s also pretty inconvenient to create a totally new service
object.  Before we have a `Service`, we need:
- a `Supplier`
- a `Framework`
- a `Lot` that belongs to that `Framework`

And there’s no straightforward way to create new `Framework` 
or `Lot` objects and then get rid of them at the end of the test.

This pull request sucks and I don’t think we should merge it.
But we should figure out what to do about these tests.

### DIFF
--- a/tests/app/helpers.py
+++ b/tests/app/helpers.py
@@ -313,3 +313,80 @@ class JSONUpdateTestMixin(JSONTestMixin):
 
         assert_equal(response.status_code, 400)
         assert_in("'updated_by' is a required property", response.get_data(as_text=True))
+
+
+class GCloud8ApplicationTest(BaseApplicationTest):
+
+    """
+    --- frameworks ---
+    1: g-cloud-6: live
+    2: g-cloud-4: expired
+    3: g-cloud-5: live
+    4: g-cloud-7: pending
+    5: digital-outcomes-and-specialists: coming
+    6: g-cloud-8: open
+
+    """
+
+    def setup(self):
+        self.new_framework_statuses = {
+            'digital-outcomes-and-specialists': 'live',
+            'g-cloud-8': 'live',
+            'g-cloud-7': 'live',
+            'g-cloud-6': 'expired',
+            'g-cloud-5': 'expired',
+            'g-cloud-4': 'expired',
+        }
+
+        super(GCloud8ApplicationTest, self).setup()
+        with self.app.app_context():
+            self.setup_g_cloud_8()
+
+            if self.new_framework_statuses:
+                self.initial_framework_statuses = self._new_framework_statuses(self.new_framework_statuses)
+
+    @staticmethod
+    def setup_g_cloud_8():
+        g_cloud_8 = Framework(
+            slug='g-cloud-8',
+            name='G-Cloud 8',
+            framework='g-cloud',
+            framework_agreement_details={'frameworkAgreementVersion': 'v1.0'},
+            status='open',
+            clarification_questions_open=False
+        )
+        db.session.add(g_cloud_8)
+        db.session.commit()
+
+    @staticmethod
+    def _new_framework_statuses(new_statuses):
+        initial_statuses = {}
+
+        frameworks = Framework.query.all()
+        for framework in frameworks:
+            # save the old status
+            initial_statuses[framework.slug] = framework.status
+
+            # overwrite it with a new status
+            if framework.slug in new_statuses:
+                framework.status = new_statuses[framework.slug]
+
+            db.session.add(framework)
+
+        db.session.commit()
+        return initial_statuses
+
+    def teardown(self):
+        with self.app.app_context():
+            if self.new_framework_statuses:
+                self._new_framework_statuses(self.initial_framework_statuses)
+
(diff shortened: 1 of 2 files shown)


allait:
Think about what is it that you're actually trying to test. `index_service` takes an object as an argument and makes a decision based on the properties of that object.

It doesn't try to read from the database, so the first thing we could try is pass in a `Service` object we've created right there, instead of adding it to the session, committing and reading back from the DB. We do this quite a bit [in the model tests](https://github.com/alphagov/digitalmarketplace-api/blob/master/tests/app/test_model.py#L656-L659) and this should allow you to create whatever Framework etc objects you need for the current test and throw them out without storing them in the DB session.

Going even further, how much does `index_service` care about having an actual model instance? The function takes an argument, so what if you pass in a mock object instead of the actual service? We don't often mock model instances since we're using the extensive SQLAlchemy APIs which are hard to fake. And we're actually interested in testing that we use the model methods correctly and they're doing what we expect, which mocking would obscure. But in this case we're only using a few attributes, so a mock object with `Mock(spec=Service)` could be a very clean approach.

Re. the shared frameworks in cases where we do need to set up some instances in the database. The approach that I've tried to start on before was [using py.test fixtures for test DB dependencies](https://github.com/alphagov/digitalmarketplace-api/commit/f4fdba652a1822698778c65df66a2639ec0fc5ad). Since fixtures are self-contained they take care of setting and resetting the state for each test. And since they can depend on other fixtures they could be layer so for example a fixture for a live service depends on a live framework and supplier fixtures. I still think it's the way to go, but I haven't had much time to work on it.

pcraig3:
Yes, pretty obvious in retrospect. 
Committing to a database totally isn't necessary.

pcraig3:
Spent a good while today looking at the pytest documentation and trying a bunch of different things to get fixtures to accept parameters defined in or near the actual tests that run them. It seems totally weird how passing parameters to fixtures works (ie, magically), and also a bit weird how you don't really see them being called. 
There's a decorator we can use to [`parameterize`](http://docs.pytest.org/en/latest/parametrize.html) fixtures, so that's what I've ended up going with.

Asked @risicle for help in thinking about how good/bad the things that I was trying to do were, and it seems to be going okay. 
Seems like the fixture for updating framework statuses can be generalised pretty cleanly, and that a fixture for adding/removing a G8 framework is something we don't really need to be generic.

I'm kind of thinking we can combine both of these fixtures to (a) create a G8 entry and then (b) update its status if we need to, but maybe this is a bit of a weird approach? (ie, it would probably screw up our tests if the teardowns weren't all executed in order.)

Anyway, I'm going to write the tests on the index methods and see how it shakes out.

[Progress so far is on this branch.](https://github.com/alphagov/digitalmarketplace-api/tree/pc-fix-framework-fixtures)

Anyway, everything I've done so far with the filters is

allait:
@pcraig3 I'm not really sure what you're trying to achieve there.

I think fixtures work quite well when they name the dependency they're providing (so if a test needs a live framework then it adds `live_framework` to the list of arguments) and having separate fixtures for live/expired dos/g-cloud frameworks removes a lot of guesswork around proper use from actual tests without making fixture file that big.

But then again, parametrised features would be a lot more useful for services.

In the meantime, what's the plan for the `index_services` tests? Are we going with mock-based tests?

pcraig3:
What do you think of [this](https://github.com/alphagov/digitalmarketplace-api/commit/56b3695a2f987fd4364c5f4590f35cecb0e79567)?

First test is probably the most crazy.  
(a) Mocks `search_api_client` so that we don't actually call the `.index()` method
(b) adds G-Cloud 8 to the database using a fixture (default status is `open`)
(c) `update_framework_status` fixture sets G8 to `live`
(d) Mock the `Service.serialize()` method in the body of the test because we don't actually want to do that. 

If you think this is a good way to go, I'll open a pull request.  Otherwise, if you'd prefer the idea of using separate fixtures (`live_dos_framework`, `live_g8_framework`, `expired_g6_framework`, etc), that's easy enough to put in instead.

PR #429, opened 2016-08-04
Update PaaS VCAP_SERVICES config load for new environment

allait:
New PaaS environment has a different name for the postgres service compared to the trial one, so we need to update the config code that parses the VCAP_SERVICES JSON.


TheDoubleK:
:+1:

PR #430, opened 2016-08-05
Admin can extract a list of buyers and their briefs.

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/124104161) story on Pivotal.

Admin's need to be able to extract a csv of buyers details and their briefs from the admin app. Rather than polling the api for every buyer, this sets up a new endpoint which returns them all in one hit.

The database query returns a row for every brief with buyer details duplicated, so there is also some logic to reduce the rows to one per buyer with all their briefs together.

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -6,7 +6,7 @@ from flask import jsonify, abort, request, current_app
 
 from .. import main
 from ... import db, encryption
-from ...models import User, AuditEvent, Supplier, Framework, SupplierFramework, DraftService
+from ...models import User, AuditEvent, Supplier, Framework, SupplierFramework, DraftService, Brief, BriefUser
 from ...utils import get_json_from_request, json_has_required_keys, \
     json_has_matching_id, pagination_links, get_valid_page_or_1, validate_and_return_updater_request
 from ...validation import validate_user_json_or_400, validate_user_auth_json_or_400, is_valid_buyer_email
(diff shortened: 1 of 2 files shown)


app/main/views/users.py
```diff
@@ -296,3 +296,32 @@ def check_supplier_role(role, supplier_id):
         abort(400, "'supplier_id' is required for users with 'supplier' role")
     elif role != 'supplier' and supplier_id:
         abort(400, "'supplier_id' is only valid for users with 'supplier' role, not '{}'".format(role))
+
+
+@main.route("/buyers/briefs", methods=["GET"])
```
allait: While this gives us exactly what we need, I feel like this might be a bit too standalone for the API. It's a new endpoint and a response type that's not tied to a model and unlikely to be used by anything else.

What if we extend `/briefs` with an additional GET parameter to inline user data instead of user IDs? This gives the admin app all the data needed to build the list of buyers with associated briefs. We'd still need to paginate through the briefs and the API responses will contain all brief data that we don't really need, but as long as it's not timing out efficiency doesn't seem like a crucial concern for this.

Alternatively, if we also want to include buyer users that don't have any briefs we could add a `role` filter to `/users` and make admin app request all briefs and all buyer users and join them in place. If that proves to be too slow we could look into adding a way to request a full users/brief list without pagination, which should cut down the number of requests. The main benefit (and I think it's an important one) compared to a custom endpoint then would be that these could be reused by other scripts/tasks.

app/main/views/users.py
```diff
@@ -296,3 +296,32 @@ def check_supplier_role(role, supplier_id):
...
+                format(brief.data.get('title'), brief.status) if brief else None
+            continue
+        last_email_address = buyer.email_address
+        buyer_rows.append({
+            'buyer_name': buyer.name,
+            'buyer_email': buyer.email_address,
+            'buyer_phone': buyer.phone_number,
+            'buyer_created': buyer.created_at,
+            'briefs': '{} - {}'.format(brief.data.get('title'), brief.status) if brief else ''
+        })
```
risicle: This is just a point of interest here because @allait wants a different approach taken, but what you're doing here can all be done _in_ the SQL using aggregations and `GROUP BY`. It's much more reliable, declarative and (usually much, much) faster to do it that way. https://www.postgresql.org/docs/9.5/static/tutorial-agg.html

risicle: Also, if you did need to do it in python, a much clearer & less error prone way do aggregations is using `itertools.groupby`

https://docs.python.org/2/library/itertools.html#itertools.groupby

Again, see me if you want to know more...

app/main/views/users.py
Wynndow: The admin wants all buyers, even if they don't have any briefs so I've gone for adding a role filter to /users and combining in the admin app. I've got a PR up [here](https://github.com/alphagov/digitalmarketplace-api/pull/431) for it so will close this PR.

PR #431, opened 2016-08-09
Adjust /briefs and /users for admin app buyer download

Wynndow:
This PR is part of [this](https://www.pivotaltracker.com/story/show/124104161) Pivotal story. 

This PR is alternative approach to the previos PR on the API for the same story, found [here](https://github.com/alphagov/digitalmarketplace-api/pull/430).

The specific endpoint set up in the previous PR was too specific and couldn't be reused for other means. This approach is hitting the api to get all briefs and all buyers and the combining them in the admin app.

To do this, the api needed to return a little more information. For briefs, it needed to return the user data along with the brief. For users, it needed a way of filtering just on buyers. This PR adds that functionality.

For both routes it also adds some metadata to the json being returned which gives the total number of paginated pages. The admin app will need to hit the api a certain number of times to make sure it's retrieved all briefs/buyers and this bit of meta allows that.

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -116,6 +116,8 @@ def list_briefs():
 
     page = get_valid_page_or_1()
 
+    with_users = request.args.get('with_users', 'false').lower() == 'true'
+
     user_id = get_int_or_400(request.args, 'user_id')
 
     if user_id:
@@ -138,7 +140,7 @@ def list_briefs():
 
     if user_id:
         return jsonify(
-            briefs=[brief.serialize() for brief in briefs.all()],
+            briefs=[brief.serialize(with_users) for brief in briefs.all()],
             links={},
         )
     else:
@@ -148,7 +150,7 @@ def list_briefs():
         )
 
         return jsonify(
-            briefs=[brief.serialize() for brief in briefs.items],
+            briefs=[brief.serialize(with_users) for brief in briefs.items],
             meta={
                 "total": briefs.total,
             },
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -63,6 +63,15 @@ def list_users():
             links={}
         )
 
+    role = request.args.get('role')
+    if role:
+        if role in ['buyer']:
+            user_query = user_query.filter(
+                User.role == role
+            )
+        else:
+            abort(400, 'Invalid user role: {}'.format(role))
+
     supplier_id = request.args.get('supplier_id')
     if supplier_id is not None:
         try:
--- a/tests/app/views/test_users.py
+++ b/tests/app/views/test_users.py
@@ -1065,6 +1065,29 @@ class TestUsersGet(BaseUserTest):
             "supplier_id '{}' not found".format(non_existent_supplier),
             response.get_data(as_text=True))
 
+    def test_only_buyers_returned_with_role_param_set_to_buyer(self):
+        self._post_user({
+            "emailAddress": "Chris@gov.uk",
+            "name": "Chris",
+            "role": "buyer",
+            "password": "minimum10characterpassword"
+        })
+
+        response = self.client.get('/users?role=buyer')
+        data = json.loads(response.get_data())
+        buyer = data['users'][0]
+
+        assert response.status_code == 200
+        assert len(data['users']) == 1
+        assert buyer['name'] == 'Chris'
+
+    def test_400_returned_if_role_param_is_not_valid(self):
+        response = self.client.get('/users?role=incorrect')
+        data = response.get_data(as_text=True)
+
+        assert response.status_code == 400
+        assert_in("Invalid user role: incorrect", data)
+
 
 class TestUsersExport(BaseUserTest):
     framework_slug = None
--- a/tests/app/views/test_briefs.py
+++ b/tests/app/views/test_briefs.py
@@ -2,6 +2,7 @@ import json
 from datetime import datetime
 from datetime import timedelta
 
+import pytest
 import mock
 from ..helpers import BaseApplicationTest, COMPLETE_DIGITAL_SPECIALISTS_BRIEF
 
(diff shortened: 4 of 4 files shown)


app/main/views/briefs.py
```diff
@@ -116,6 +116,11 @@ def list_briefs():
 
     page = get_valid_page_or_1()
 
+    if request.args.get('with_users') == 'true':
```
idavidmcdonald: This could be shortened to:

`with_users = (request.args.get('with_users') == 'true')`

app/main/views/users.py
```diff
@@ -76,15 +82,18 @@ def list_users():
...
+    paginated_users = user_query.paginate(
         page=page,
         per_page=current_app.config['DM_API_SERVICES_PAGE_SIZE'],
     )
 
     return jsonify(
-        users=[user.serialize() for user in users.items],
+        users=[user.serialize() for user in paginated_users.items],
+        meta={
+            "total_pages": paginated_users.pages
```
idavidmcdonald: Any idea why we don't have a "total" key in our meta like for briefs? I'm guessing it's not been needed but just wondering if there was an explanation for the inconsistency?

app/main/views/users.py
```diff
@@ -63,6 +63,12 @@ def list_users():
             links={}
         )
 
+    only_buyers = request.args.get('only_buyers')
```
idavidmcdonald: Could it be worth making this functionality more general so you can effectively filter by any user role rather than just for the buyer role? (feel free to say no)

app/main/views/briefs.py
Wynndow: Yep, that's a much better way of doing it. Have changed.

app/main/views/users.py
Wynndow: I'm in two minds about this. On the one hand it may be useful and wouldn't be too hard to implement. On the other hand, it's not functionality that's needed anywhere so I'm not convinced it should be done. Maybe if a story comes up that needs to filter by a different user role it could be implemented then?

app/main/views/users.py
Wynndow: No idea. I can only assume it was added to briefs because it was needed somewhere!

pcraig3:
So, as a general thing, I think the opportunity's ripe here to do some commit squashing (or learn about it, if you haven't done it before(!)).  Should be able to knock you down from 9 commits 11 commits to 4 or 5.

app/main/views/users.py
pcraig3: I said this yesterday in person, but there was a moment a last year when @minglis (our old  technical architect) said that "APIs always get bigger -- never smaller".  
He was arguing against generalising something that we didn't need to be generalised yet. 

I think keep this `only_buyers` switch instead of passing in a role.  Right now, `buyers` are the only kinds of users we're looking to bulk export for this story.

pcraig3: Also, as a style thing, you could make this a bit nicer by replicating what you've already done with your `with_users` variable.

Something like:

```py
only_buyers = request.args.get('only_buyers') == 'true'
```

app/main/views/users.py
pcraig3: @risicle added it [here](https://github.com/alphagov/digitalmarketplace-api/pull/424/files).  After doing some when we were indexing services as part of making G8 live, I think it would be good to have `total`s in all of our `list_*` routes.

tests/app/views/test_users.py
```diff
@@ -1065,6 +1066,45 @@ def test_returns_404_for_nonexistent_supplier(self):
             "supplier_id '{}' not found".format(non_existent_supplier),
             response.get_data(as_text=True))
 
+    def test_only_buyers_returned_with_only_buyers_arg_set_true(self):
```
pcraig3: This test and, especially, the one following it, are super weird in that they are expecting users to be returned that aren't mentioned in the test itself. 

Not sure this is the time to fix this particular issue, but just making a note that it sucks.

app/main/views/briefs.py
```diff
@@ -142,21 +144,22 @@ def list_briefs():
...
             per_page=current_app.config['DM_API_BRIEFS_PAGE_SIZE'],
         )
 
         return jsonify(
-            briefs=[brief.serialize() for brief in briefs.items],
+            briefs=[brief.serialize(with_users) for brief in paginated_briefs.items],
             meta={
-                "total": briefs.total,
+                "total": paginated_briefs.total,
+                "total_pages": paginated_briefs.pages
```
pcraig3: Maybe we should just call this `pages`? 
Seems like it probably means the same thing and it might be more succinct.

app/main/views/users.py
allait: I'm with David on this one. `role=...` filter seems like a straightforward solution. We could only allow `role=buyer` right now (which would fit with the "APIs don't get smaller" but will be nicer to extend to other roles without breaking backwards compatibility), but for example finding all admin or admin-ccs users for the environment is something I had to do in the past, so an endpoint would definitely be useful.

app/main/views/users.py
TheDoubleK: Is `total_pages` the same as `last_page`? Elsewhere where we do pagination we have links for previous, next and last pages. I guess a count is different from a URL, but not sure it is necessary?

app/main/views/users.py
TheDoubleK: Also agree a role filter makes sense, we have filters like this elsewhere where a comma-separated list of values we want can be passed in (eg for services on frameworks I think).

app/main/views/users.py
Wynndow: @TheDoubleK The link to the last page looks something like this `http://localhost:5000/users?page=25&only_buyers=True` so I guess I could extract the total pages from there. Is that better than just passing the number in the json?

tests/app/views/test_users.py
```diff
@@ -1065,6 +1065,30 @@ def test_returns_404_for_nonexistent_supplier(self):
...
+        response = self.client.get('/users?role=buyer')
+        data = json.loads(response.get_data())
+        buyer = data['users'][0]
+
+        assert response.status_code == 200
+        assert len(data['users']) == 1
+        assert buyer['name'] == 'Chris'
+
+    def test_400_returned_if_role_param_is_not_valid(self):
+
```
idavidmcdonald: Blank line in here we don't need

Wynndow: It's gone!

idavidmcdonald:
good work 👍

app/main/views/briefs.py
```diff
@@ -148,7 +150,7 @@ def list_briefs():
         )
 
         return jsonify(
-            briefs=[brief.serialize() for brief in briefs.items],
+            briefs=[brief.serialize(with_users) for brief in briefs.items],
```
pcraig3: Super-duper nitpick (I know it's already been merged), but I think using a keyword argument parameter might be a tiny bit clearer here.  

ie, given that `with_users` is `True` or `False`, I think that calling `brief.serialize(True)` seems less intuitive than `brief.serialize(with_users=True)`. (Even though I get that what it would actually look like is `brief.serialize(with_users=with_users)`.)

Really not a big deal.

tests/app/views/test_users.py
```diff
@@ -1065,6 +1065,29 @@ def test_returns_404_for_nonexistent_supplier(self):
...
+        response = self.client.get('/users?role=buyer')
+        data = json.loads(response.get_data())
+        buyer = data['users'][0]
+
+        assert response.status_code == 200
+        assert len(data['users']) == 1
+        assert buyer['name'] == 'Chris'
+
+    def test_400_returned_if_role_param_is_not_valid(self):
+        response = self.client.get('/users?role=incorrect')
```
pcraig3: Also tiny nitpick, but it might be good to test a role that we might expect to be valid (ie, `supplier` or `admin`) just to be explicit that these aren't supposed to work.

Wynndow: Yeah, those are pretty solid points. Next time I've got a PR i'll try and remember to update them!

PR #432, opened 2016-08-11
Add tests for `index_service` function; improve framework-setting fixtures

pcraig3:
Added tests for the `index_services` function that no longer indexes DOS services.

In the process, tried to ameliorate the problem where our tests are relying on Frameworks and Lots to be in the database but not actually taking responsibility for them.

After quite a lot of exploring potential solutions (you can see them if you look at the various commits), I've gone with a fairly straightforward way where we can easily create fixtures for certain frameworks in certain states.  Existing frameworks will have their status temporarily modified and new frameworks will be temporarily created.  Not passing parameters in anymore from the tests themselves because, while it would slim the code for the fixtures, it would fatten the code for the actual unit tests.  Instead, I'm just using meaningfully named fixture functions and tried to lower the overhead as much as possible when creating new ones.  It's the most standard solution (the others mostly involved some degree of hacking about), and pretty much an extension of what @allait started with.

I think all existing tests relying on frameworks could be made to pull in a fixture they need instead of the tacit expectation of a framework already in the database, but I don't really think I can justify doing that right now.  

Would be nice to refactor the tests as we go though.

***

Actually, as an addendum to what I said about how all existing tests could be refactored to pull in a fixture they needed -- this is true for the frameworks in isolation, but it doesn't have much to say about more complex relationships between objects.  For example, if I want an `open` `DOS` framework and I want to have a supplier with a `SupplierFramework` record with `agreementReturnedAt` set to a specific time, currently everything but the framework setup is just a manual process that happens in the test.  Maybe that's okay, but it feels kind of like a limitation we would prefer to have a workaround for.

### DIFF
--- a/tests/app/views/test_suppliers.py
+++ b/tests/app/views/test_suppliers.py
@@ -1264,6 +1264,59 @@ class TestRegisterFrameworkInterest(BaseApplicationTest, JSONUpdateTestMixin):
             assert_equal(data['frameworks'], ['digital-outcomes-and-specialists'])
 
 
+class TestSupplierFrameworkResponse(BaseApplicationTest):
+
+    def setup(self):
+        super(TestSupplierFrameworkResponse, self).setup()
+
+        self.setup_dummy_suppliers(1)
+        self.setup_dummy_user(1, role='supplier')
+        self.supplier_id = 0
+        self.framework_slug = 'g-cloud-8'
+
+    def test_get_supplier_framework_info(self, open_g8_framework):
+        framework_id = json.loads(
+            self.client.get('frameworks/{}'.format(self.framework_slug)).get_data()
+        )['frameworks']['id']
+
+        # add SupplierFramework record to the database
+        with self.app.app_context():
+            answers = SupplierFramework(
+                supplier_id=self.supplier_id, framework_id=framework_id,
+                declaration={'an_answer': 'Yes it is'},
+                on_framework=True,
+                agreement_returned_at=datetime(2015, 10, 10, 10, 10, 10),
+                countersigned_at=datetime(2015, 11, 12, 13, 14, 15),
+                agreement_details={
+                    u'signerName': u'thing',
+                    u'signerRole': u'thing',
+                    u'uploaderUserId': 20
+                },
+            )
+            db.session.add(answers)
+            db.session.commit()
+
+        # Get back the SupplierFramework record
+        response = self.client.get(
+            '/suppliers/{}/frameworks/{}'.format(self.supplier_id, self.framework_slug))
+
+        data = json.loads(response.get_data())
+        assert response.status_code, 200
+        assert data['frameworkInterest']['supplierId'] == self.supplier_id
+        assert data['frameworkInterest']['frameworkSlug'] == self.framework_slug
+        assert data['frameworkInterest']['declaration'] == {'an_answer': 'Yes it is'}
+        assert data['frameworkInterest']['onFramework'] is True
+        assert data['frameworkInterest']['agreementReturned'] is True
+        assert data['frameworkInterest']['agreementReturnedAt'] == '2015-10-10T10:10:10.000000Z'
+        assert data['frameworkInterest']['countersigned'] is True
+        assert data['frameworkInterest']['countersignedAt'] == '2015-11-12T13:14:15.000000Z'
+        assert data['frameworkInterest']['agreementDetails'] == {
+            'signerName': 'thing',
+            'signerRole': 'thing',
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -44,6 +43,99 @@ def teardown():
...
+        def teardown():
+            with app.app_context():
+                framework = Framework.query.filter(Framework.slug == slug).first()
+                Framework.query.filter(Framework.id == framework.id).delete()
+                db.session.commit()
+
+    request.addfinalizer(teardown)
+
+
+def _framework(request, app, slug, status, framework, framework_agreement_details=None):
```
idavidmcdonald: Although I like the consistency of this with the other function names, I can't help but feel the function name should be more descriptive, given that `_framework` tells very little about what this function is doing. Not sure though, what do you think?

tests/app/views/test_suppliers.py
```diff
@@ -1264,25 +1264,25 @@ def test_can_get_registered_frameworks_for_a_supplier(self):
...
 
     def setup(self):
-        super(TestSupplierFrameworkUpdates, self).setup()
+        super(TestSupplierFrameworkResponse, self).setup()
 
         self.setup_dummy_suppliers(1)
         self.setup_dummy_user(1, role='supplier')
+        self.supplier_id = 0
+        self.framework_slug = 'g-cloud-6'
+        # can't do this unless the framework already exists, unfortunately
```
idavidmcdonald: Are you saying that this line requires g cloud 6 to already exist? Found this comment a little unclear

pcraig3: Yeah, that's the idea. 

Can't get a `g-cloud-8` framework id because there's I guess the setup step for the class happens before the test fixture is execute (or at least that's my assumption).

tests/conftest.py
pcraig3: yeaahhhhh, fine.

>.<

tests/app/views/test_suppliers.py
risicle: seems like adding the word `FIXME` to this comment would make it less ambiguous

tests/app/views/test_suppliers.py
```diff
@@ -1350,6 +1316,36 @@ def test_get_supplier_framework_info(self):
             'uploaderUserId': 20
         }
 
+
+class TestSupplierFrameworkUpdates(BaseApplicationTest, JSONUpdateTestMixin):
+    method = "post"
+    endpoint = "/suppliers/0/frameworks/digital-outcomes-and-specialists"
```
idavidmcdonald: Sorry, could you please explain what this is for? Haven't been able to quite work it out. I see it somehow relates to JSONUpdateTestMixin but it's interesting that this is for DOS but not all of this class's tests are about DOS (at least one is doing things with G-cloud 8).

tests/conftest.py
risicle: Yeah `_base_framework` or `_framework_base` or something might be better.

risicle:
Apart from a couple of naming quibbles i think :+1:

tests/app/views/test_suppliers.py
```diff
@@ -1350,6 +1316,38 @@ def test_get_supplier_framework_info(self):
             'uploaderUserId': 20
         }
 
+
+class TestSupplierFrameworkUpdates(BaseApplicationTest, JSONUpdateTestMixin):
+    # These variables are for the JSONUpdateTestMixin, which checks we get standardised
+    # error messages when we POST bad JSON to this route (the SupplierFramework updates route)
```
idavidmcdonald: Sorry, thanks for the explanation but it probably doesn't need to actually be included as a comment (at least not here)

tests/conftest.py
```diff
@@ -44,6 +43,113 @@ def teardown():
...
+        framework_agreement_details=False
+    )
+
+
+# G8
+
+
+@pytest.fixture()
+def open_g8_framework(request, app):
+    _g8_framework(request, app, status='open')
```
risicle: Hold on, where is the teardown for all these happening?

tests/conftest.py
```diff
@@ -44,6 +43,113 @@ def teardown():
...
+        status=status,
+        framework='dos',
+        framework_agreement_details=False
+    )
+
+
+# G8
+
+
+@pytest.fixture()
```
risicle: still not able to use `pytest.yield_fixture` for these?

pcraig3: Tried it and tests started breaking.  

Didn't really look into it.

Feel free to refactor if you like!

PR #433, opened 2016-08-18
[#128509823] Supplier can be eligible for closed requirements

TheDoubleK:
Part 2 of this bug fix: https://www.pivotaltracker.com/story/show/128509823

The existing condition that briefs should be 'live' before checking eligibility was there to avoid possible key errors trying to match suppliers against incomplete draft briefs.

We need the frontend to be able to check eligibility for closed briefs, so it's better to exclude 'draft' briefs here rather than only allowing 'live'.

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -352,7 +352,7 @@ def list_brief_services(brief_id):
     brief = Brief.query.filter(
         Brief.id == brief_id
     ).filter(
-        Brief.status == "live"
+        Brief.status != "draft"
     ).first_or_404()
 
     supplier_id = get_int_or_400(request.args, 'supplier_id')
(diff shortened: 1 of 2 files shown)


app/main/views/briefs.py
```diff
@@ -352,7 +352,7 @@ def list_brief_services(brief_id):
     brief = Brief.query.filter(
         Brief.id == brief_id
     ).filter(
-        Brief.status == "live"
+        Brief.status != "draft"
```
idavidmcdonald: We wish to allow the "withdrawn" case?

TheDoubleK: For this API endpoint I think yes - the restriction is only there to avoid KeyErrors on incomplete drafts.  It's up to the frontend apps to decide whether to show users something relating to a withdrawn brief, but I don't think there's any reason to stop the API from making the info about it available to the frontend apps.

idavidmcdonald:
👍 Nice 💯

PR #434, opened 2016-08-18
[#127806845]SupplierFramework: add migration adding agreed_variations json field

risicle:
For story https://www.pivotaltracker.com/story/show/127806845

Code following in separate PR

### DIFF
--- a/migrations/versions/700_add_agreed_variations_to_SupplierFramework.py
+++ b/migrations/versions/700_add_agreed_variations_to_SupplierFramework.py
@@ -1,0 +1,22 @@
+"""add agreed_variations to SupplierFramework
+
+Revision ID: 700
+Revises: 690
+Create Date: 2016-08-12 15:52:48.366958
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '700'
+down_revision = '690'
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+def upgrade():
+    op.add_column('supplier_frameworks', sa.Column('agreed_variations', postgresql.JSON(), nullable=True))
+
+
+def downgrade():
+    op.drop_column('supplier_frameworks', 'agreed_variations')
--- a/migrations/versions/710_add_g_cloud_8_variation_1.py
+++ b/migrations/versions/710_add_g_cloud_8_variation_1.py
@@ -1,0 +1,31 @@
+"""Add g-cloud-8 variation 1
+
+Revision ID: 710
+Revises: 700
+Create Date: 2016-08-19 16:11:38.493586
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '710'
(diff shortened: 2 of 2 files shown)


TheDoubleK:
The migration code looks good, I've run it up and down and it works. 

But it would be nice to have a commit message that explains what this is for, and also that explains why we are going with a new column rather than putting variation details into the existing `agreement_details` JSON (I know there was an argument for it, but for the life of me I can't remember now what it was).

risicle:
Will do

migrations/versions/700_add_agreed_variations_to_SupplierFramework.py
```diff
@@ -0,0 +1,22 @@
...
+# revision identifiers, used by Alembic.
+revision = '700'
+down_revision = '690'
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+def upgrade():
+    op.add_column('supplier_frameworks', sa.Column('agreed_variations', postgresql.JSON(), nullable=True))
```
allait: Why are we splitting the column for `supplier_frameworks` from agreement details but not from `frameworks` table? It feels like they should be similar.

risicle: Because adding a new variation is expected to be a rare manual thing and not really expected to be race-y.

risicle: (read the extended commit message for why i considered choice of field to be related to raciness)

allait: I don't really see a race condition risk in this case and it feels like the agreement_variation updates are limited by the attached schemas, but I suppose keeping the fields separate makes it simpler.

risicle: So then beyond the race thing, it was realized that of course we want a specific audit event for variation agreement and just relying on `update_supplier_framework_details` and sniffing to see what the caller was _actually_ doing would start making `update_supplier_framework_details` a bit weirder and more complex than it already is, so having a separate endpoint was seen as a good thing. So if we're going to have a separate endpoint it's probably better to not let the fields be manually dickable with too.

risicle: I should probably add that reason to the comment too.

risicle: When it comes to schemas, something the schemas can't do is enforce that `agreedVariations` entries reference a `variation` that actually exists on the `Framework`.

risicle: And the schemas can't enforce not doing something odd like removing an agreement entry. Just that the end result is valid according to the schema.

TheDoubleK:
OK let's do it :+1;

PR #435, opened 2016-08-18
[#127806845]Framework variations

risicle:
Story https://www.pivotaltracker.com/story/show/127806845

Corresponding migration in PR #434, **don't be surprised if the tests don't pass initially** because this branch is currently missing the migration.

A single endpoint added for allowing a caller to agree to a variation. Variations are referenced by "slug". Valid variations are looked up against `Framework.framework_agreement_details["variations"]`, which is expected to be a dict of the form:

```
{
            "banana": {
                "createdAt": "2016-06-06T20:01:34.000000Z",
            },
            "toblerone": {
                "createdAt": "2016-07-06T21:09:09.000000Z",
            }
},
```

For two example variation slugs `banana` and `toblerone`. No methods for updating the framework variations are currently included - this can probably be done manually using existing general purpose endpoints.

`SupplierFramework`'s serialization gains a new `agreedVariations` key which will contain a dict looking something like:

```
{
    "banana": {
            "agreedUserId": 2,
            "agreedUserEmail": "test+2@digital.gov.uk",
            "agreedUserName": "my name",
            "agreedAt": "2016-06-06T00:00:00.000000Z",
        }
}
```

For a supplier that has agreed to variation `banana`. Variations which are not agreed to should have no entry in this dict.

To agree to a variation, a caller should `PUT` to the endpoint `/suppliers/<int:supplier_id>/frameworks/<framework_slug>/variation/<variation_slug>` json of the form:

```
           {
                "updated_by": "<updater>",
                "agreedVariation": {
                    "agreedUserId": 1234,
                },
            }
```

Examples of all of this in action can be seen in the tests.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 from flask import jsonify, abort, request, current_app
+from six import text_type
 from sqlalchemy.exc import IntegrityError, DataError
 from .. import main
 from ... import db
@@ -12,10 +13,19 @@ from ...validation import (
     validate_contact_information_json_or_400,
     is_valid_string_or_400
 )
-from ...utils import pagination_links, drop_foreign_fields, get_json_from_request, \
-    json_has_required_keys, json_has_matching_id, get_valid_page_or_1, validate_and_return_updater_request
+from ...utils import (
+    pagination_links,
+    drop_foreign_fields,
+    get_json_from_request,
+    json_has_required_keys,
+    json_only_has_required_keys,
+    json_has_matching_id,
+    get_valid_page_or_1,
+    validate_and_return_updater_request,
+)
 from ...supplier_utils import validate_and_return_supplier_request, validate_agreement_details_data
 from dmapiclient.audit import AuditTypes
+from dmutils.formats import DATETIME_FORMAT
 
 
 @main.route('/suppliers', methods=['GET'])
@@ -471,3 +481,80 @@ def update_supplier_framework_details(supplier_id, framework_slug):
         return jsonify(message="Database Error: {0}".format(e)), 400
 
     return jsonify(frameworkInterest=interest_record.serialize()), 200
+
+
+@main.route('/suppliers/<int:supplier_id>/frameworks/<framework_slug>/variation/<variation_slug>', methods=['PUT'])
+def agree_framework_variation(supplier_id, framework_slug, variation_slug):
+
+    framework = Framework.query.filter(
+        Framework.slug == framework_slug
+    ).first_or_404()
+
+    supplier = Supplier.query.filter(
+        Supplier.supplier_id == supplier_id
+    ).first_or_404()
+
+    # TODO SELECT FOR UPDATE?
+    # (technically we should be doing a FOR UPDATE of SupplierFramework and a FOR SHARE of Framework)
+    interest_record = SupplierFramework.query.filter(
+        SupplierFramework.supplier_id == supplier.supplier_id,
+        SupplierFramework.framework_id == framework.id
+    ).first()
+    if not interest_record:
+        abort(404, "supplier_id '{}' has not registered interest in {}".format(supplier_id, framework_slug))
+
+    if not interest_record.on_framework:
+        abort(404, "supplier_id '{}' is not on framework {}".format(supplier_id, framework_slug))
+
+    # acceptable variation_slugs must have corresponding createdAt entry in variations dict
+    if not (framework.framework_agreement_details or {}).get("variations", {}).get(variation_slug, {}).get("createdAt"):
+        abort(404, "Unknown variation '{}' for framework {}".format(variation_slug, framework_slug))
+
+    if (interest_record.agreed_variations or {}).get(variation_slug, {}).get("agreedAt"):
+        abort(400, "supplier_id '{}' has already agreed to variation '{}'.".format(supplier_id, variation_slug))
+
+    json_payload = get_json_from_request()
+    updater_json = validate_and_return_updater_request()
+    json_has_required_keys(json_payload, ["agreedVariations"])
+    json_only_has_required_keys(json_payload["agreedVariations"], ["agreedUserId"])
+
+    user = User.query.filter(User.id == json_payload["agreedVariations"]["agreedUserId"]).first()
+    if not user:
+        abort(400, "No user found with id '{}'".format(json_payload["agreedVariations"]["agreedUserId"]))
+    if supplier_id != user.supplier_id:
+        abort(403, "user '{}' isn't associated with supplier_id '{}'".format(
+            json_payload["agreedVariations"]["agreedUserId"],
+            supplier_id,
+        ))
+
+    agreed_variations = interest_record.agreed_variations.copy() if interest_record.agreed_variations else {}
+    agreed_variations[variation_slug] = agreed_variations.get(variation_slug, {})
+    agreed_variations[variation_slug].update({
+        "agreedAt": datetime.utcnow().strftime(DATETIME_FORMAT),
+        "agreedUserId": user.id,
+    })
+    interest_record.agreed_variations = agreed_variations
+
+    audit_event = AuditEvent(
+        audit_type=AuditTypes.agree_framework_variation,
+        user=updater_json['updated_by'],
+        data={
+            'supplierId': supplier.supplier_id,
+            'frameworkSlug': framework_slug,
+            'variationSlug': variation_slug,
+            'update': json_payload["agreedVariations"],
+        },
+        db_object=interest_record,
+    )
+
+    try:
+        db.session.add(interest_record)
+        db.session.add(audit_event)
+        db.session.commit()
+    except IntegrityError as e:
+        db.session.rollback()
+        return jsonify(message="Database Error: {0}".format(e)), 400
+
+    return jsonify(
+        agreedVariations=SupplierFramework.serialize_agreed_variation(agreed_variations[variation_slug]),
+    ), 200
--- a/tests/app/views/test_suppliers.py
+++ b/tests/app/views/test_suppliers.py
@@ -3,12 +3,14 @@ from flask import json
 import pytest
 from freezegun import freeze_time
 from nose.tools import assert_equal, assert_in, assert_is_not_none, assert_true, assert_is
+from random import randint
+
+from dmapiclient.audit import AuditTypes
 
 from app import db
 from app.models import Supplier, ContactInformation, AuditEvent, \
     SupplierFramework, Framework, DraftService, Service
 from ..helpers import BaseApplicationTest, JSONTestMixin, JSONUpdateTestMixin
-from random import randint
 
 
 class TestGetSupplier(BaseApplicationTest):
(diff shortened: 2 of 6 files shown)


TheDoubleK:
Ah, unforeseen consequence of having the migration separate is the failing tests here that are expecting the new column to exist.

allait:
This branch should be on top of the migration branch, which would solve the tests issue. And you'd have to rebase it anyway once migration is merged.

risicle:
Yup, hence it would be nice to get the migration merged soon.

risicle:
right ok ill include the migration

tests/app/views/test_suppliers.py
```diff
@@ -1749,3 +1753,261 @@ def test_changing_on_framework_to_passed_creates_audit_event(self, open_dos_fram
...
+
+            assert audit_events[0].user == "test123"
+            assert audit_events[0].data["supplierId"] == 1
+            assert audit_events[0].data["frameworkSlug"] == "g-cloud-8"
+            assert audit_events[0].data["variationSlug"] == "banana"
+            assert audit_events[0].data["update"] == {
+                "agreedUserId": 1,
+            }
+
+    def test_agree_variation_playpen(
```
idavidmcdonald: This test name could be more descriptive please

risicle: So, yeah, this is named this because it actually tests a bunch of things together, so I found it hard to give a more specific name. And yes, in a perfect world where you have time to write a test for every possible call in every possible existing state tests would be split up, but a lot of the things tested here rely on the state built up early on in the test. The longer the test, the more rich & complex the existing state becomes, enabling the testing of more complex scenarios...

If you can think of a better name, cool. It was previously named `test_agree_variation_various`.

tests/app/views/test_suppliers.py
```diff
@@ -1749,3 +1753,261 @@ def test_changing_on_framework_to_passed_creates_audit_event(self, open_dos_fram
...
+        assert response4_json["agreedVariation"] == expected_variation_banana
+
+        response5 = self.client.get("/suppliers/2/frameworks/g-cloud-8")
+        assert response5.status_code == 200
+        assert json.loads(response5.get_data())["frameworkInterest"]["agreedVariations"] == {
+            "banana": expected_variation_banana,
+            "toblerone": expected_variation_toblerone,
+        }
+
+        # check we can't agree a second time
```
idavidmcdonald: It might be good to split this test out into a separate one as it seems like it's testing a separate bit of functionality

tests/conftest.py
```diff
@@ -19,137 +22,321 @@ def app(request):
...
+        request,
+        app,
+        live_g8_framework_2_variations,
+        user_role_supplier,
+        ):
+    _supplierframework_fixture_inner(request, app, sf_kwargs={"on_framework": True})
+
+
+@pytest.fixture()
+def live_g8_framework_2_variations_suppliers_on_framework_with_alt(
```
idavidmcdonald: It took me a little while to figure out what "with alt" means. After following the code trail I now believe it means using a "second" or "alternative" supplier. I'm wondering if there could be a better way to describe this?

risicle: Yes... you can see I was already finding the fixture names ridiculously long and trying to find ways to compress them...

tests/app/views/test_suppliers.py
risicle: I know, it would be nice to be able to address each bit of functionality separately for running partial tests, but in the real world functionality doesn't exist in isolation in the same way... if I were to "split up" these tests, to get the same sort of richness of state in the testing I would really have to end up _copying_, then modifying them. Pushed for time, but I think I'm at least going to change the name back to `_various` and put a comment explaining the rationale.

TheDoubleK:
This looks good I think.

Am I right in thinking that the only way to add a variation to the Framework is through a migration?  Not sure if we we want the `update_framework` endpoint to allow setting a new variation?

tests/conftest.py
```diff
@@ -19,137 +22,321 @@ def app(request):
...
+        request,
+        app,
+        live_g8_framework_2_variations,
+        user_role_supplier,
+        ):
+    _supplierframework_fixture_inner(request, app)
+
+
+@pytest.fixture()
+def live_g8_framework_2_variations_suppliers_on_framework(
```
allait: I think this is needlessly specific. There's no known side-effects for having framework variations in other tests, so I think we could add them to `live_g8_framework` fixture

tests/conftest.py
allait: I feel like this is not very useful as a fixture. If you think about test dependencies, I'm finding it hard to imagine a lot of tests would need this exact thing. What they do seem to need is a supplier that's on framework, a user for that supplier and a user for a different supplier. The first two could be fixtures (with user depending on supplier, which in turn depends on framework), the last one could probably be created by the test itself.

tests/conftest.py
```diff
@@ -19,137 +22,321 @@ def app(request):
...
+    return _supplier_fixture_inner(request, app, supplier_kwargs={"supplier_id": 2})
+
+
+# Users
+
+
+@pytest.fixture()
+def user_role_supplier(request, app, supplier_basic):
+    return _user_fixture_inner(request, app, user_kwargs={
+        "id": 1,
```
allait: Is there a reason to hardcode these IDs? I think there should be a way to let the DB create a new record and make the new ID available to the tests through the fixture value. This also makes me question the value of _alt version, but I'd have to think about how you could make the test require two user records at once.

tests/conftest.py
risicle: I do have a specific test for what happens with a missing `variations` field.

risicle:
@TheDoubleK 
> Not sure if we we want the update_framework endpoint to allow setting a new variation?

Don't we? I was assuming that we would. And I thought we were trying to get away from populating the database in migrations...

TheDoubleK:
Well if we do want it, we kind of need it in this PR I think.

risicle:
How did we set `{"frameworkAgreementVersion": "v1.0"}`?

TheDoubleK:
Think it was here: https://github.com/alphagov/digitalmarketplace-api/blob/master/migrations/versions/640_add_framework_agreement_version_to_.py#L20

risicle:
Ah shag.

tests/conftest.py
allait: I think if it's a specific test that needs a different record then the setup should be handled by the test itself. Fixtures are only useful if they're required by multiple tests.

app/main/views/suppliers.py
```diff
@@ -471,3 +481,80 @@ def update_supplier_framework_details(supplier_id, framework_slug):
...
+        abort(404, "supplier_id '{}' has not registered interest in {}".format(supplier_id, framework_slug))
+
+    if not interest_record.on_framework:
+        abort(404, "supplier_id '{}' is not on framework {}".format(supplier_id, framework_slug))
+
+    # acceptable variation_slugs must have corresponding createdAt entry in variations dict
+    if not (framework.framework_agreement_details or {}).get("variations", {}).get(variation_slug, {}).get("createdAt"):
+        abort(404, "Unknown variation '{}' for framework {}".format(variation_slug, framework_slug))
+
+    if (interest_record.agreed_variations or {}).get(variation_slug, {}).get("agreedAt"):
```
TheDoubleK: I'm not sure this should be an API error here - could it just return the existing `agreedVariation` details? You're generally supposed to be able to PUT multiple times to the same thing I think...

app/main/views/suppliers.py
```diff
@@ -471,3 +481,80 @@ def update_supplier_framework_details(supplier_id, framework_slug):
...
+    try:
+        db.session.add(interest_record)
+        db.session.add(audit_event)
+        db.session.commit()
+    except IntegrityError as e:
+        db.session.rollback()
+        return jsonify(message="Database Error: {0}".format(e)), 400
+
+    return jsonify(
+        agreedVariation=SupplierFramework.serialize_agreed_variation(agreed_variations[variation_slug]),
```
TheDoubleK: I don't think we should have the singular `agreedVariation` here and plural `agreedVariations` in the `serialize` method. 
Both should be `agreedVariations`.  We consistently use the plural everywhere else I think, including where only a single thing is being returned, e.g:
https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/briefs.py#L201
https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/briefs.py#L274
https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/drafts.py#L71
https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/services.py#L189
https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/suppliers.py#L131

TheDoubleK:
Apologies for late-breaking comments but I've just run into the issues.

app/main/views/suppliers.py
TheDoubleK: Thinking about this a bit more, I don't think the API should choke on a perfectly valid request to update some data - it should be up to the frontend apps to make the decision on whether a user can perform an action - if the frontend app thinks it's OK then the API call should succeed and set the data as requested.

app/main/views/suppliers.py
TheDoubleK: Also, thinking about this a bit more (insomnia's a bitch), I think the route should return the entire `SupplierFramework` object, not just the `agreedVariations` part, as that is the model object that's referenced by the URL and that is updated by the call, i.e. `return jsonify(frameworkInterest=interest_record.serialize())`

Other routes that update this record do that (except PUTting a declaration which only returns the declaration, but I think that's probably doing the wrong thing).

TheDoubleK:
@allait's comments about the test fixtures are still relevant, but in the interests of getting this piece of work out perhaps it can be sorted out later :+1:

risicle:
Yes, I am partially in agreement with @allait re fixtures - it's certainly something that needs to be hammered out down the road.

app/main/views/suppliers.py
```diff
@@ -471,3 +481,80 @@ def update_supplier_framework_details(supplier_id, framework_slug):
         return jsonify(message="Database Error: {0}".format(e)), 400
 
     return jsonify(frameworkInterest=interest_record.serialize()), 200
+
+
+@main.route('/suppliers/<int:supplier_id>/frameworks/<framework_slug>/variation/<variation_slug>', methods=['POST'])
```
allait: So, if we are going with `POST`, then the URL would be `'/suppliers/<int:supplier_id>/frameworks/<framework_slug>/variation`, so that it's clear that we create a new variation. This is what we do with briefs/services/users etc. But recognising that changing that would require a lot of changes to the endpoint, maybe we should stick with PUT.

risicle:
(agh also @idavidmcdonald made a good point about postfixing fixture names with `_alt` being quite vague, but let's get this out...)

app/main/views/suppliers.py
risicle: Ohh yes I think I see what you mean.

Damn the `POST` change has already made it into the apiclient.

risicle: I was also thinking that if we were being really rest-y it would be good to accept `GET` to this endpoint which would just return the response (or... 404? if not agreed) bit it wouldn't really add anything that you don't get from the supplierframework endpoint anyway.

allait: There's no harm in adding it, but I think we tend to skip endpoints we know we probably won't use.

PR #436, opened 2016-08-22
[#127807225] Add "variations agreed" to framework user export

TheDoubleK:
Required for this story: https://www.pivotaltracker.com/story/show/127807225

This extra field is required for Admin csv file download, to enable admin to see who has and who hasn't agreed to framework variations.

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -254,6 +254,7 @@ def export_users_for_framework(framework_slug):
         application_status = ''
         application_result = ''
         framework_agreement = ''
+        variations_agreed = ''
 
         # if framework is pending, live, or expired
         if framework.status != 'open':
@@ -277,6 +278,7 @@ def export_users_for_framework(framework_slug):
             else:
                 application_result = 'pass' if sf.on_framework else 'fail'
             framework_agreement = bool(sf.agreement_returned_at)
+            variations_agreed = sf.agreed_variations.keys() if sf.agreed_variations else []
 
         user_rows.append({
             'user_email': u.email_address,
@@ -285,7 +287,8 @@ def export_users_for_framework(framework_slug):
             'declaration_status': declaration_status,
             'application_status': application_status,
             'framework_agreement': framework_agreement,
-            'application_result': application_result
+            'application_result': application_result,
+            'variations_agreed': variations_agreed
         })
 
     return jsonify(users=[user for user in user_rows])
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -278,7 +278,7 @@ def export_users_for_framework(framework_slug):
             else:
                 application_result = 'pass' if sf.on_framework else 'fail'
             framework_agreement = bool(sf.agreement_returned_at)
-            variations_agreed = list(sf.agreed_variations.keys()) if sf.agreed_variations else []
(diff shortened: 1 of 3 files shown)


tests/app/helpers.py
```diff
@@ -280,6 +280,15 @@ def set_framework_status(self, slug, status):
...
 
+    def set_framework_variation(self, slug):
+        Framework.query.filter_by(slug=slug).update({
+            'framework_agreement_details': {
+                "frameworkAgreementVersion": "v1.0",
+                "variations": {"1": {"createdAt": "2016-08-19T15:31:00.000000Z"}}
+            }
+        })
+        db.session.commit()
+
```
risicle: Now here's where the two approaches to test fixtures start overlapping with each other. (@pcraig3  &) I created a bunch of pytest fixtures that amongst other things creates frameworks with variations (see `conftest.py`)

tests/app/views/test_users.py
```diff
@@ -1393,6 +1408,25 @@ def test_400_response_if_framework_is_coming(self):
...
         assert response.status_code == 400
 
+    def test_response_agreed_contract_variation(self):
+        self._setup()
+        self._put_complete_declaration()
+        self._post_complete_draft_service()
+        self._post_result(True)
+        self._post_framework_agreement()
+        self._set_framework_variation()
+        self._put_variation_agreement()
```
risicle: I was under the impression we generally try to set up existing database state in these tests through loading things directly into the database rather than successive api calls...?

app/main/views/users.py
```diff
@@ -277,6 +278,7 @@ def export_users_for_framework(framework_slug):
             else:
                 application_result = 'pass' if sf.on_framework else 'fail'
             framework_agreement = bool(sf.agreement_returned_at)
+            variations_agreed = ', '.join(sf.agreed_variations.keys()) if sf.agreed_variations else ''
```
Wynndow: Just a thought, instead of returning an empty string if there are no agreed variations, it might be easier to read for an admin if the string 'None' is returned here, or something similar.

tests/app/views/test_users.py
TheDoubleK: This is adding a single test for the newly added field to the 400+ line `TestUsersExport` class that already exists to test this endpoint.  I have just followed the pattern that all the existing tests in this class already use.

tests/app/helpers.py
TheDoubleK: Discussed in real life: moving to use the fixture approach here would mean rewriting the whole set of tests for this endpoint and there doesn't seem to be value in doing that right now.

app/main/views/users.py
TheDoubleK: Cath would be the person to ask, but yeah that seems reasonable, will change now.

TheDoubleK: Actually, looking at it again all the other fields get an empty string if there's no value (e.g. `application_status`, `application_result`, `framework_agreement`) so it seems reasonable to leave blank for no variations agreed either (also bearing in mind that most frameworks don't have any variations, so a column of "None" might look a bit much when nothing would be expected there anyway).

Wynndow: Fair do's!

Wynndow:
👍

PR #437, opened 2016-08-23
[#121577487] [Parked] SupplierFramework: accept countersignedUserId and countersignedOfficialName values, in agreement_details, handle in update endpoint and serialization

risicle:
This has no tests yet and is not ready for further consideration (waiting on discussion about audit event behaviour and existing supplier framework update endpoint behaviour). But it's here so we have a PR# to refer to.


risicle:
This is too outdated to be useful now

PR #438, opened 2016-08-25
[#128752313] Create more generic framework update route

pcraig3:
Currently, we can't do many updates to existing Framework database records except through migrations (and we'd like to move away from those).
The updates we _can_ do are to a framework's status and whether or not the period for clarification questions are open.

As part of the countersigning thing, we need to store the current countersigner's name in the Framework.framework_agreement_details.  Instead of doing a migration, it'll be nicer to ship out a more generic 'update' endpoint that can update the agreement_details as well as the statuses and clarification question thing (which is what this is).

This PR:
- Does the general framework update route
- Adds a `framework-agreement-details` schema for things that we currently know about (namely, the `frameworkAgreementVersion` and `variations`)

This PR does not:
- Add in a `countersignerName` field (or whatever) to the `framework_agreement_details`.  
  - (That story can be found [here](https://www.pivotaltracker.com/story/show/128843567).)

***

Worth pointing out that in order to update the existing `framework_agreement_details`, we need to include every key in the update data.  This may possibly be too restrictive sometime in the future, but for now, our one framework that has the agreement details also has an entry for all the keys.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -14,6 +14,7 @@ from ...utils import (
     get_json_from_request, json_has_required_keys, json_only_has_required_keys,
     validate_and_return_updater_request,
 )
+from ...framework_utils import validate_framework_agreement_details_data
 
 
 @main.route('/frameworks', methods=['GET'])
@@ -79,6 +80,12 @@ def get_framework(framework_slug):
 
 @main.route('/frameworks/<framework_slug>', methods=['POST'])
 def update_framework(framework_slug):
+    attribute_whitelist = {
+        'status': 'status',
+        'clarificationQuestionsOpen': 'clarification_questions_open',
+        'frameworkAgreementDetails': 'framework_agreement_details'
+    }
+
     updater_json = validate_and_return_updater_request()
     framework = Framework.query.filter(
         Framework.slug == framework_slug
@@ -86,11 +93,26 @@ def update_framework(framework_slug):
 
     json_payload = get_json_from_request()
     json_has_required_keys(json_payload, ['frameworks'])
-    json_only_has_required_keys(json_payload['frameworks'], ['status', 'clarificationQuestionsOpen'])
+
+    if not json_payload['frameworks']:
+        abort(400, "Framework update expects a payload")
+
+    invalid_keys = (set(json_payload['frameworks'].keys()) - set(attribute_whitelist.keys()))
+    if invalid_keys:
+        abort(400, "Invalid keys for framework update: '{}'".format("', '".join(invalid_keys)))
+
+    if 'frameworkAgreementDetails' in json_payload['frameworks']:
+        # all frameworkAgreementDetails keys must be present or the update will fail
+        validate_framework_agreement_details_data(
+            json_payload['frameworks']['frameworkAgreementDetails'],
+            enforce_required=True
+        )
+
+    for whitelisted_key, value in attribute_whitelist.items():
+        if whitelisted_key in json_payload['frameworks']:
+            setattr(framework, value, json_payload['frameworks'][whitelisted_key])
 
     try:
-        framework.status = json_payload['frameworks']['status']
-        framework.clarification_questions_open = json_payload['frameworks']['clarificationQuestionsOpen']
         db.session.add(framework)
         db.session.add(
             AuditEvent(
--- a/app/models.py
+++ b/app/models.py
@@ -137,7 +137,16 @@ class Framework(db.Model):
 
     @validates('clarification_questions_open')
     def validates_clarification_questions_open(self, key, value):
-        if self.status not in ('coming', 'open', 'pending') and value:
+        # Note that because of an undefined order of setting attributes on this model
+        # (ie, between this `clarification_questions_open` and `status`)
+        # it _is_ possible to be in a state where `clarification_questions_open` is `True`
+        # and the new status is not one of the allowed ones
+        # if `self.status` *at the time of the update* is one of the allowed ones
+        if (
+            self.status is not None and
+            self.status not in ('coming', 'open', 'pending')
+            and value is True
+        ):
             raise ValidationError("Clarification questions are only permitted while the framework is open")
         return value
 
--- a/app/models.py
+++ b/app/models.py
@@ -129,11 +129,17 @@ class Framework(db.Model):
 
     slug_pattern = re.compile("^[\w-]+$")
 
-    @validates("slug")
+    @validates('slug')
     def validates_slug(self, key, slug):
         if not self.slug_pattern.match(slug):
             raise ValidationError("Invalid slug value '{}'".format(slug))
         return slug
+
+    @validates('clarification_questions_open')
+    def validates_clarification_questions_open(self, key, value):
+        if self.status not in ('coming', 'open', 'pending') and value:
+            raise ValidationError("Clarification questions are only permitted while the framework is open")
+        return value
 
     def __repr__(self):
         return '<{}: {} slug={}>'.format(self.__class__.__name__, self.name, self.slug)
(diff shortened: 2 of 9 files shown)


app/main/views/frameworks.py
```diff
@@ -79,18 +81,37 @@ def get_framework(framework_slug):
...
+        abort(400, "Invalid keys for framework update: '{}'".format("', '".join(invalid_keys)))
+
+    if 'frameworkAgreementDetails' in json_payload['frameworks']:
+        validate_framework_agreement_details_data(
+            json_payload['frameworks']['frameworkAgreementDetails'],
+            enforce_required=False
+        )
+
+    for key in json_payload['frameworks']:
+        setattr(framework, inflection.underscore(key), json_payload['frameworks'][key])
```
risicle: I can see that you've got the whitelist, but this feels like a dangerous line of code to have in here. Would feel a lot safer if instead it iterated through the `attribute_whitelist` and only set those attrs it knew about.

risicle: Also a neater way of doing this is
```
for key, value in iteritems(json_payload["frameworks"]):
    setattr(framework, inflection.underscore(key), value)
```
(importing `iteritems` from `six`)

app/models.py
```diff
@@ -129,12 +129,18 @@ def validates_framework(self, key, framework):
...
-    @validates("slug")
+    @validates('slug')
     def validates_slug(self, key, slug):
         if not self.slug_pattern.match(slug):
             raise ValidationError("Invalid slug value '{}'".format(slug))
         return slug
 
+    @validates('clarification_questions_open')
+    def validates_clarification_questions_open(self, key, value):
+        if self.status not in ['coming', 'open', 'pending'] and value:
```
risicle: My little piece of dogma here: don't use a list where a tuple will do.

app/main/views/frameworks.py
allait: Yeah, I agree about iterating over `attribute_whitelist`. And I'm not sure about the use of `inflection` here, I think doing something like
```python
attribute_whitelist = {
    'clarificationQuestionsOpen': 'clarification_questions_open',
    ...
]
```
while a bit repetitive is less likely to fail in surprising ways.

json_schemas/framework-agreement-details.json
```diff
@@ -0,0 +1,20 @@
...
+  "$schema": "http://json-schema.org/schema#",
+  "additionalProperties": false,
+  "minProperties": 1,
+  "properties": {
+    "frameworkAgreementVersion": {
+      "minLength": 1,
+      "type": "string"
+    },
+    "variations": {
+      "type": "object"
```
risicle: You could probably be a bit more restrictive here, no?

tests/app/views/test_frameworks.py
```diff
@@ -202,62 +200,151 @@ class TestUpdateFramework(BaseApplicationTest, JSONUpdateTestMixin):
...
+                })
+
+                assert response.status_code == 200
+                data = json.loads(response.get_data())['frameworks']
+                # frameworkAgreementDetails doesn't show up in the serialized framework
+                if isinstance(value, dict):
+                    for nested_key, nested_value in value.items():
+                        assert data[nested_key] == nested_value
+                else:
+                    assert data[key] == value
```
risicle: I personally don't think it's good enough to just check the response from a modifying call. I would also want to perform a secondary `GET` and "check it's stuck".

Very easy to add the check - you can probably just compare equality to the `POST`-returned response.

app/main/views/frameworks.py
```diff
@@ -79,18 +80,38 @@ def get_framework(framework_slug):
...
+    if invalid_keys:
+        abort(400, "Invalid keys for framework update: '{}'".format("', '".join(invalid_keys)))
+
+    if 'frameworkAgreementDetails' in json_payload['frameworks']:
+        validate_framework_agreement_details_data(
+            json_payload['frameworks']['frameworkAgreementDetails'],
+            enforce_required=False
+        )
+
+    for whitelisted_key in attribute_whitelist:
```
risicle: Still not quite sure why you're not doing `for key, value in attribute_whitelist.items():` here and instead doing a second lookup.

app/main/views/frameworks.py
```diff
@@ -79,18 +80,38 @@ def get_framework(framework_slug):
...
+
+    invalid_keys = (set(json_payload['frameworks'].keys()) - set(attribute_whitelist.keys()))
+    if invalid_keys:
+        abort(400, "Invalid keys for framework update: '{}'".format("', '".join(invalid_keys)))
+
+    if 'frameworkAgreementDetails' in json_payload['frameworks']:
+        validate_framework_agreement_details_data(
+            json_payload['frameworks']['frameworkAgreementDetails'],
+            enforce_required=False
+        )
```
pcraig3: Worth noting that this never validates using all of the required fields.
Means that you could have a `framework_agreement_details` like:
```
{ 
  "frameworkAgreementVersion": "v1.0"
}
```

And then update it with:
```
{
  "countersignerName": "Dan Saxby"
}
```

And then wipe out your `"frameworkAgreementVersion"`.

Other than setting `enforce_required=True` on this validate method, I don't know how to require some things but not others.

tests/app/views/test_frameworks.py
```diff
@@ -202,62 +200,169 @@ class TestUpdateFramework(BaseApplicationTest, JSONUpdateTestMixin):
...
-                                        content_type="application/json")
+            for key, value in valid_attributes_and_values.items():
+                response = self.post_framework_update({
+                    key: value
+                })
+
+                assert response.status_code == 200
+                post_data = json.loads(response.get_data())['frameworks']
+
+                # frameworkAgreementDetails doesn't show up in the serialized framework
```
idavidmcdonald: Ah I see what you've done here. The comment could be a little clearer though, to show that the contents of frameworkAgreementDetails are being transformed into a different format in the serialize method rather than they are not included.

Feel free to improve but one suggestion could be:
frameworkAgreementDetails is not included in Framework.serialize itself but instead each key, value in  frameworkAgreementDetails are included as separate keys in the return from Framework.serialize

app/models.py
```diff
@@ -129,12 +129,25 @@ def validates_framework(self, key, framework):
...
     def validates_slug(self, key, slug):
         if not self.slug_pattern.match(slug):
             raise ValidationError("Invalid slug value '{}'".format(slug))
         return slug
 
+    @validates('clarification_questions_open')
+    def validates_clarification_questions_open(self, key, value):
+        # Note that because of a race condition between this key and `status`, it _is_ possible to be in a state where
+        # `clarification_questions_open` is `True` and the new status is not one of the allowed ones if `self.status`
+        # *at the time of the update* is one of the allowed ones
```
idavidmcdonald: Well spotted. Is there anything we can/want to do about this or are we happy to just be aware of it?

app/models.py
```diff
@@ -129,12 +129,25 @@ def validates_framework(self, key, framework):
...
-    @validates("slug")
+    @validates('slug')
     def validates_slug(self, key, slug):
         if not self.slug_pattern.match(slug):
             raise ValidationError("Invalid slug value '{}'".format(slug))
         return slug
 
+    @validates('clarification_questions_open')
+    def validates_clarification_questions_open(self, key, value):
+        # Note that because of a race condition between this key and `status`, it _is_ possible to be in a state where
```
risicle: It's not technically a race condition - it all happens in one thread of execution, it's just the `__init__` doesn't know it should set field `A` before field `B` or vice versa.

idavidmcdonald:
👍 

Paul and I just spoke about the fact that a Framework when serialised has a different interface to the interface to update a framework. An example is you receive `frameworkAgreementVersion` as a top level key but you update it by sending it contained in `frameworkAgreementDetails`. This behaviour seems a little bit unintuitive and might be technical debt. We think it's acceptable to continue this for the moment but maybe something to think about for the future (may need some changes to front end apps if we are changing our interface)

app/models.py
risicle: The only thing I think could be _done_ for this is building some mechanism that deferred validation for a particular code section while updates are being made (I was thinking a context manager).

PR #443, opened 2016-09-01
[#128843567] Add countersignature details to framework table

pcraig3:
This pull request adds the `countersignerName` key to `Framework.framework_agreement_details`, and then returns it in the `serialize()` method if it finds it.

Planning to add just a string that says `"Dan Saxby"` to G8 in all environments once this is merged and deployed.

### DIFF
--- a/json_schemas/framework-agreement-details.json
+++ b/json_schemas/framework-agreement-details.json
@@ -3,6 +3,10 @@
   "additionalProperties": false,
   "minProperties": 1,
   "properties": {
+    "countersignerName": {
+      "minLength": 1,
+      "type": "string"
+    },
     "frameworkAgreementVersion": {
       "minLength": 1,
       "type": "string"
--- a/app/models.py
+++ b/app/models.py
@@ -105,6 +105,9 @@ class Framework(db.Model):
             'name': self.name,
             'slug': self.slug,
             'framework': self.framework,
+            'countersignerName': (
+                self.framework_agreement_details or {}
+            ).get("countersignerName"),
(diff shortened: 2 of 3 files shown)


json_schemas/framework-agreement-details.json
```diff
@@ -3,6 +3,10 @@
   "additionalProperties": false,
   "minProperties": 1,
   "properties": {
+    "countersignerDetails": {
```
idavidmcdonald: Does it make sense to call this field `countersignerDetails`, which kind of implies its a dictionary full of information (like `signerDetails`) when at the moment it is just a string for the name of the counter signer. Would something like `countersignerName` be more appropriate?

pcraig3: What do you think, @allait?

We have some other `{x}Details` keys as well as some other `{x}Name` keys, so both have some kind of precedent. 

I thought we were going under the assumption that we were expecting to have to store more countersigner stuff sometime, so calling this `countersignerDetails` at this point was forward-looking.
If not though, `countersignerName` is probably fine as long as it doesn't clash with anything we're going to be storing with our framework agreements.

allait: I think since it's a key inside `framework_agreement_details` we can add additional ones beside it if we need to store more data in the future. `countersignedName` makes a little bit more sense if it's a string, I'd expect a dictionary object in `{x}Details`.

idavidmcdonald:
👍  💯

TheDoubleK:
Will merge once functional tests are green.

PR #444, opened 2016-09-05
FrameworkAgreement initial routes

idavidmcdonald:
Add `serialize` and some updating/validation methods to the  `FrameworkAgreement` model

---

Migration to add countersigned fields (`countersigned_agreement_details`, `countersigned_agreement_path`, `countersigned_agreement_returned_at`). Note, this migration should cause no issues being deployed contained in this pull request and has therefore been included as it makes our life easier.

---

When serializing a `SupplierFramework`, it will look for the first related `FrameworkAgreement`. If it can't find one (which it shouldn't at the moment as we have not migrated any data to the `framework_agreements` table) then it will use the original framework agreement details stored in the `SupplierFramework` object. We have kept the `SupplierFramework` interface the same. We are currently working under the assumption that there will be only one framework agreement per supplier framework as a halfway step in our refactoring journey. This can then later be changed to allow multiple framework agreements per supplier framework.

---

We added initial routes for framework agreements. It was decided to have several routes that handle different actions that might be applied to a framework agreement. This means we can avoid having one massive update method (similar to what we currently have in the `SupplierFramework`).

We added:

- A simple `get_framework_agreement` route that returns a serialized `FrameworkAgreement` by ID
- An `update_framework_agreement` route. At the moment only `signedAgreementDetails` and `signedAgreementPath` fields can be updated through this route. You can not sign an agreement using this route. For validation, we do not enforce the order of fields to follow the user flow in the front end app. Instead we simply validate the fields against the agreementDetails JSON schema. We will rely on the (to be written) `sign` route to ensure that a `FrameworkAgreement` is in a valid state to be signed. We make the behaviour of this route follow that of the `SupplierFramework` so we can swap out the old functionality and use this new model and nothing should break (for example you can resign framework agreements at the moment). The next step of refactoring will then break this change by introducing multiple framework agreements.

Our current plan is to then add:

- A route to sign a framework agreement
- A route to countersign a framework agreement

---

Add `json_has_keys` util for validating if JSON data contains all required keys and restricts the list of keys that are allowed by using optional keys.

---

We added a useful `@fixture_params` decorator for parametrizing fixtures, in particular those that aren't used directly by the test, so when making fixtures that depend on one another we can
still set parameters on the base ones. Although we haven't ended up using the decorator in this pull request, we anticipate it will be useful in the future.


tests/app/views/test_agreements.py
```diff
@@ -0,0 +1,264 @@
...
+                    'agreement': {
+                        'signerAgreementDetails': {
+                            'signerName': 'name',
+                            'signerRole': 'role',
+                        }
+                    }
+                }),
+            content_type='application/json')
+
+        assert res.status_code == 400
```
TheDoubleK: Might be nice to test that the error message matches what we expect - i.e. that the request has failed for the right reason.

tests/app/views/test_agreements.py
```diff
@@ -0,0 +1,264 @@
...
+        data = json.loads(res.get_data(as_text=True))
+        assert data['agreement'] == {
+            'id': agreement_id,
+            'supplierId': supplier_framework['supplier_id'],
+            'frameworkId': supplier_framework['framework_id'],
+            'signedAgreementDetails': {
+                'signerName': 'name',
+                'signerRole': 'role',
+            }
+        }
```
TheDoubleK: Is it worth also doing a subsequent GET here to check that the update has actually saved, rather than just testing the initial response to the POST? (Thanks to @risicle I have a feeling this is something we're generally guilty of not testing enough on our update routes.)

tests/app/views/test_agreements.py
```diff
@@ -0,0 +1,264 @@
...
+        })
+
+        assert res.status_code == 200
+        data = json.loads(res.get_data(as_text=True))
+        assert data['agreement'] == {
+            'id': agreement_id,
+            'supplierId': supplier_framework['supplier_id'],
+            'frameworkId': supplier_framework['framework_id'],
+            'signedAgreementPath': '/example.pdf'
+        }
```
TheDoubleK: Similarly for other update tests here - it would be nice to do a subsequent GET to see the saved changes.

tests/app/views/test_agreements.py
```diff
@@ -0,0 +1,264 @@
...
+    def test_400_if_signed_agreement_details_contains_some_random_key(self, supplier_framework):
+        agreement_id = self.create_agreement(supplier_framework)
+        res = self.post_agreement_update(agreement_id, {
+            'signedAgreementDetails': {
+                'signerName': 'name',
+                'randomKey': 'value',
+            }
+        })
+
+        assert res.status_code == 400
```
TheDoubleK: Again, might be nice to check for the "right" failure condition.

tests/app/views/test_agreements.py
```diff
@@ -0,0 +1,264 @@
...
+            'signedAgreementDetails': {'details': 'here'},
+            'signedAgreementPath': 'path',
+            'signedAgreementReturnedAt': '2016-10-01T01:01:01.000000Z',
+            'countersignedAgreementDetails': {'countersigneddetails': 'here'},
+            'countersignedAgreementReturnedAt': '2016-11-01T01:01:01.000000Z',
+            'countersignedAgreementPath': '/example.pdf'
+        }
+
+
+class TestUpdateFrameworkAgreement(BaseApplicationTest):
```
TheDoubleK: There don't seem to be any checks that the correct audit event has been created on update... maybe update tests should GET audit events following the update to check them.

app/models.py
```diff
@@ -467,7 +489,43 @@ class FrameworkAgreement(db.Model):
...
+
+    def serialize(self):
+        return purge_nulls_from_data({
+            'id': self.id,
+            'supplierId': self.supplier_id,
+            'frameworkId': self.framework_id,
+            'signedAgreementDetails': self.signed_agreement_details,
+            'signedAgreementPath': self.signed_agreement_path,
+            'signedAgreementReturnedAt': (
+                self.signed_agreement_returned_at and
```
TheDoubleK: Doesn't this (and the `countersignedAgreementReturnedAt`) give you a boolean rather than a date in the serialized output?

app/models.py
```diff
@@ -429,14 +429,33 @@ def serialize(self, data=None):
...
             "onFramework": self.on_framework,
-            "agreementReturned": bool(agreement_returned_at),
-            "agreementReturnedAt": agreement_returned_at,
-            "agreementDetails": self.agreement_details,
-            "countersigned": bool(countersigned_at),
-            "countersignedAt": countersigned_at,
             "agreedVariations": agreed_variations,
         }, **(data or {}))
 
+        if self.framework_agreements:
```
TheDoubleK: Where does `self.framework_agreements` come from? Does it match the `supplier_id` and `framework_id` in the framework_agreements table through SQLAlchemy and the magic of foreign keys?

app/models.py
```diff
@@ -429,14 +429,33 @@ def serialize(self, data=None):
...
+            agreement = self.framework_agreements[0]
```
TheDoubleK: If there are multiple matching entries for `framework_agreements` what guarantees that the first in the list is the most recent?

tests/app/views/test_suppliers.py
```diff
@@ -1320,6 +1320,136 @@ def test_get_supplier_framework_info(self, open_g8_framework):
             'uploaderUserId': 20
         }
 
+    def test_get_supplier_framework_returns_framework_agreement(self, live_example_framework):
```
TheDoubleK: It feels like there should be some testing around always getting the most recent signed agreement details when there are multiple signed agreements saved.

app/models.py
idavidmcdonald: Yep, done by SQLAlchemy using a backref in the `FrameworkAgreement` model

app/models.py
idavidmcdonald: Sorry, I forgot to mention in the pull request that we are planning on getting everything working under the current assumption that the relationship will be one to one. We were trying to keep the refactoring steps small and will then need to change this as a later step.

tests/app/views/test_suppliers.py
idavidmcdonald: Same as above, currently working under the assumption of only one framework agreement per supplier framework.

app/main/views/agreements.py
```diff
@@ -0,0 +1,71 @@
...
+@main.route('/agreements/<int:agreement_id>', methods=['POST'])
+def update_framework_agreement(agreement_id):
+    framework_agreement = FrameworkAgreement.query.filter(FrameworkAgreement.id == agreement_id).first_or_404()
+
+    json_payload = get_json_from_request()
+    updater_json = validate_and_return_updater_request()
+    json_has_required_keys(json_payload, ["agreement"])
+    update_json = json_payload["agreement"]
+
+    json_has_keys(update_json, optional_keys=['signedAgreementDetails', 'signedAgreementPath'])
```
risicle: It really is good to get into the habit of using tuples for sequences that are intended to be ephemeral.

allait: I think it's a balance. If we're talking about a global value (like a class attribute) then immutability is more important than semantics. In this case though we have a homogeneous sequence of elements (with a very limited scope), which is what lists are for.

tests/app/views/test_agreements.py
idavidmcdonald: Thanks, have added. And it ended up being a very good idea since it turned out one of our tests was failing for the wrong reason and we had forgotten to add some functionality (only allow frameworks with frameworkAgreementVersion to update signedAgreementDetails) to our view.

tests/app/views/test_agreements.py
idavidmcdonald: Done

tests/app/views/test_agreements.py
idavidmcdonald: Have added a test to check an audit event was created

tests/app/views/test_agreements.py
idavidmcdonald: Added

tests/app/views/test_agreements.py
idavidmcdonald: Done

app/main/views/agreements.py
```diff
@@ -0,0 +1,78 @@
...
+
+    if (
+        framework_agreement.signed_agreement_returned_at
+        and ('signedAgreementDetails' in update_json or 'signedAgreementPath' in update_json)
+    ):
+        abort(400, "Can not update signedAgreementDetails or signedAgreementPath if agreement has been signed")
```
idavidmcdonald: @allait Am I right in thinking that this check should be removed? If we wish to do the smallest refactoring step possible and move all existing behaviour into our new model then this behaviour should be retained (because we currently allow users to restart the process in our frontend)

tests/app/views/test_agreements.py
TheDoubleK: Nice to know I wasn't just being picky :)

app/models.py
TheDoubleK: Magic! :tophat: :rabbit2:

app/models.py
TheDoubleK: OK... might it be worth adding a `TODO` or `FIXME` comment here so it doesn't get overlooked?

idavidmcdonald: Think we will be OK, there is a reasonable chunk of work to introduce the functionality of multiple framework agreements and should be focussed around this line of code so it shouldn't get missed.

TheDoubleK:
Think this is good to go after a rebase :+1:

PR #445, opened 2016-09-06
Add `.cache/` to gitignore

idavidmcdonald:
image

`.cache/` has been showing in my git diff recently (not sure why, Rob and I hypothesised that it could be due to a new version of pytest). I have added it to the gitignore so it's not accidentally committed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ venv
 *.pyc
 __pycache__
 version_label
+.cache/


risicle:
:+1:

Would check to see if it's worth adding to other repos too...

idavidmcdonald:
Closed due to suggestion to instead add to my global gitignore

PR #448, opened 2016-09-13
FrameworkAgreement: add lazy join to backref of FrameworkAgreement-SupplierFramework relationship too

risicle:
This should fix timeouts of `get_framework_suppliers` calls.

Though I do slightly wonder if it's a bit dangerous to blanket-add this join implicitly to all queries. It may add unnecessary weight to `SupplierFramework` queries that don't need `FrameworkAgreement` information. But... then again... I suppose seeing as the information is pulled in by `serialize`, any time we're going to be getting a `SupplierFramework` to `serialize` we're going to need it. Meh.


allait:
👍

PR #449, opened 2016-09-14
Save framework agreement information in `framework_agreements` table

idavidmcdonald:
#### Saving data
We continue to update framework agreement information through the update supplier framework route, however this no longer stores the data in the `supplier_frameworks` table but instead in the `framework_agreements` table. 

If a record already exists in the `framework_agreement` table then we will update it. If a record does not yet exist in the `framework_agreement` table, then we will create a new record using any existing data from the `supplier_frameworks` table. This enforces a one to one relationship between a `SupplierFramework` and a `FrameworkAgreement`.

This will give us framework agreement information stored across both of the tables for the moment (managed by `SupplierFramework.serialize()` to make sure we access the correct information we need). The next step will be to do a database migration and move all remaining information from the `supplier_frameworks` table to the `framework_agreements` table.

#### Squashed a few bugs related to having information stored in two locations
With framework agreement information stored in both tables, there were a few knock on effects that needed to be dealt with, such as exporting user information and filtering framework suppliers by if they have returned their agreement. See individual commits for further information.

===

### FrameworkAgreements
- Change from exposing a `frameworkId` to exposing a `frameworkSlug` to keep consistency with the rest of our API
- Added new routes to create and sign a `FrameworkAgreement` in `views/agreements`

===

### Unfortunately
Sorry, commits ended up getting quite tangled up, in particular around tests, and I wasn't easily able to rebase this into two separate pull requests! Also, please do take a look at the full commit messages as they contain more information about the changes made.

### DIFF
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -1,9 +1,10 @@
+from datetime import datetime
 from flask import jsonify, abort
 from .. import main
 from sqlalchemy.exc import IntegrityError
 from dmapiclient.audit import AuditTypes
 from ...models import (
-    db, FrameworkAgreement, AuditEvent
+    AuditEvent, db, FrameworkAgreement, User
 )
 
 from ...utils import (
@@ -65,6 +66,57 @@ def update_framework_agreement(agreement_id):
             'supplierId': framework_agreement.supplier_id,
             'frameworkSlug': framework_agreement.supplier_framework.framework.slug,
             'update': update_json},
+        db_object=framework_agreement
+    )
+
+    try:
+        db.session.add(framework_agreement)
+        db.session.add(audit_event)
+        db.session.commit()
+    except IntegrityError as e:
+        db.session.rollback()
+        return jsonify(message="Database Error: {0}".format(e)), 400
+
+    return jsonify(agreement=framework_agreement.serialize())
+
+
+@main.route('/agreements/<int:agreement_id>/sign', methods=['POST'])
+def sign_framework_agreement(agreement_id):
+    framework_agreement = FrameworkAgreement.query.filter(FrameworkAgreement.id == agreement_id).first_or_404()
+    framework_agreement_details = framework_agreement.supplier_framework.framework.framework_agreement_details
+
+    json_payload = get_json_from_request()
+    updater_json = validate_and_return_updater_request()
+    update_json = None
+
+    if framework_agreement_details and framework_agreement_details.get('frameworkAgreementVersion'):
+        json_has_required_keys(json_payload, ["agreement"])
+        update_json = json_payload["agreement"]
+
+        json_has_keys(update_json, required_keys=['signedAgreementDetails'])
+
+        framework_agreement.update_signed_agreement_details_from_json(update_json['signedAgreementDetails'])
+        framework_agreement.update_signed_agreement_details_from_json(
+            {'frameworkAgreementVersion': framework_agreement_details['frameworkAgreementVersion']}
+        )
+        validate_agreement_details_data(
+            framework_agreement.signed_agreement_details,
+            enforce_required=True
+        )
+
+        user = User.query.filter(User.id == update_json['signedAgreementDetails']['uploaderUserId']).first()
+        if not user:
+            abort(400, "No user found with id '{}'".format(update_json['signedAgreementDetails']['uploaderUserId']))
+
+    framework_agreement.signed_agreement_returned_at = datetime.utcnow()
+
+    audit_event = AuditEvent(
+        audit_type=AuditTypes.sign_agreement,
+        user=updater_json['updated_by'],
+        data=dict({
+            'supplierId': framework_agreement.supplier_id,
+            'frameworkSlug': framework_agreement.supplier_framework.framework.slug,
+        }, **({'update': update_json} if update_json else {})),
         db_object=framework_agreement
     )
 
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -17,6 +17,8 @@ from ...utils import (
 from ...supplier_utils import validate_agreement_details_data
 
 
+# This route not currently used and not yet supported by API client
+# It will be brought into use following migration of franmework agreement data
 @main.route('/agreements', methods=['POST'])
 def create_framework_agreement():
     json_payload = get_json_from_request()
@@ -174,8 +176,7 @@ def sign_framework_agreement(agreement_id):
             enforce_required=True
         )
 
-        user = User.query.filter(User.id == update_json['signedAgreementDetails']['uploaderUserId']).first()
-        if not user:
+        if not User.query.filter(User.id == update_json['signedAgreementDetails']['uploaderUserId']).first():
             abort(400, "No user found with id '{}'".format(update_json['signedAgreementDetails']['uploaderUserId']))
 
     framework_agreement.signed_agreement_returned_at = datetime.utcnow()
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -4,7 +4,7 @@ from .. import main
 from sqlalchemy.exc import IntegrityError
 from dmapiclient.audit import AuditTypes
 from ...models import (
-    AuditEvent, db, FrameworkAgreement, User
+    AuditEvent, db, FrameworkAgreement, SupplierFramework, User
 )
 
 from ...utils import (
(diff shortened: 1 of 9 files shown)


app/main/views/agreements.py
```diff
@@ -16,6 +17,76 @@
 from ...supplier_utils import validate_agreement_details_data
 
 
+@main.route('/agreements', methods=['POST'])
```
TheDoubleK: Is it worth adding a comment that this route isn't currently used, and is not yet supported by the API Client?

app/main/views/agreements.py
```diff
@@ -16,6 +17,76 @@
...
+        db_object=framework_agreement
+    )
+
+    db.session.add(audit_event)
+    db.session.commit()
+
+    return jsonify(agreement=framework_agreement.serialize()), 201
+
+
+# TODO Do we really need this route?
```
TheDoubleK: Feels like we're more likely to need a route to get by supplier_id and framework_slug rather than directly by ID.  But let's leave it how it is until we work out what we actually need.

app/main/views/agreements.py
```diff
@@ -77,3 +148,54 @@ def update_framework_agreement(agreement_id):
...
+        framework_agreement.update_signed_agreement_details_from_json(update_json['signedAgreementDetails'])
+        framework_agreement.update_signed_agreement_details_from_json(
+            {'frameworkAgreementVersion': framework_agreement_details['frameworkAgreementVersion']}
+        )
+        validate_agreement_details_data(
+            framework_agreement.signed_agreement_details,
+            enforce_required=True
+        )
+
+        user = User.query.filter(User.id == update_json['signedAgreementDetails']['uploaderUserId']).first()
```
TheDoubleK: This `user` isn't used again after this. Not sure if it needs to be?  If not then don't bother assigning it, just check it exists.

app/main/views/frameworks.py
```diff
@@ -216,16 +216,21 @@ def get_framework_suppliers(framework_slug):
...
         SupplierFramework.framework_id == framework.id
-    )
+    ).outerjoin(SupplierFramework.framework_agreements)
 
     if agreement_returned is not None:
         if convert_to_boolean(agreement_returned):
             supplier_frameworks = supplier_frameworks.filter(
-                SupplierFramework.agreement_returned_at.isnot(None)
-            ).order_by(SupplierFramework.agreement_returned_at.desc())
+                SupplierFramework.agreement_returned_at.isnot(None) |
```
TheDoubleK: Is there a reason for using the bitwise `|` and `&` operators rather than `or` and `and`? I can see they will work OK here, but so would the regular boolean operators I think?

app/main/views/suppliers.py
```diff
@@ -449,35 +484,35 @@ def update_supplier_framework_details(supplier_id, framework_slug):
...
         audit_type=AuditTypes.supplier_update,
         user=updater_json['updated_by'],
-        data={'supplierId': supplier.supplier_id, 'frameworkSlug': framework_slug, 'update': update_json},
-        db_object=supplier
+        data={
+            'supplierId': supplier.id,
+            'frameworkSlug': framework.slug,
+            'update': update_json
+        },
+        db_object=framework_agreement
```
TheDoubleK: Having a `framework_agreement` as the object of a `supplier_update` feels a bit weird. We discussed this already though I think and decided it was the best way?

tests/app/views/test_agreements.py
```diff
@@ -337,6 +459,192 @@ def test_400_if_signed_agreement_details_contains_empty_strings_as_values(self,
...
+            }
+
+    def test_can_not_sign_framework_agreement_that_has_no_signer_name(self, supplier_framework):
+        agreement_id = self.create_agreement(
+            supplier_framework,
+            signed_agreement_details={'signerRole': 'role'},
+            signed_agreement_path='/example.pdf'
+        )
+        res = self.sign_agreement(agreement_id, {'signedAgreementDetails': {'uploaderUserId': 20}})
+        assert res.status_code == 400
```
TheDoubleK: Can we test the failure messages on these 400 tests? There are lots of ways we might get a 400 so it's nice to be sure it's the right failure that we expect.

tests/app/views/test_agreements.py
```diff
@@ -337,6 +459,192 @@ def test_400_if_signed_agreement_details_contains_empty_strings_as_values(self,
...
+            supplier_framework,
+            signed_agreement_details={'signerName': 'name', 'signerRole': 'role'},
+            signed_agreement_path='/example.pdf'
+        )
+        res = self.sign_agreement(agreement_id, {'signedAgreementDetails': {'uploaderUserId': 20}})
+        assert res.status_code == 400
+
+
+class TestSignFrameworkAgreementThatHasNoFrameworkAgreementVersion(BaseFrameworkAgreementTest):
+    def sign_agreement(self, agreement_id):
```
TheDoubleK: It might be worth adding a comment to the test class to explain why we need this - I.e for pre G8 frameworks.

tests/app/views/test_frameworks.py
```diff
@@ -586,6 +579,50 @@ def test_list_suppliers_with_agreements_returned(self):
...
+
+            # Check SupplierFramework.agreement_returned_at is not used (indicated by supplier 4
+            # appearing before supplier 3 in our results)
+            supplier_ids = [item['supplierId'] for item in data['supplierFrameworks']]
+            assert supplier_ids[0] > supplier_ids[1]
+
+    def test_list_suppliers_with_agreements_returned_includes_those_with_only_supplier_framework(self):
+        with self.app.app_context():
+            # Set supplier 2 to have their agreement returned information only set in the SupplierFramework table
+            db.session.execute(
```
TheDoubleK: How do we know there's nothing in the FrameworkAgreement table for this supplier? It doesn't seem obvious to me from the test setup here. Does our also need to execute a delete from the FrameworkAgreement table to be sure there are no entries there?

TheDoubleK: "It", not "our". Github seems to have removed the "edit comment" function :/

tests/app/views/test_suppliers.py
```diff
@@ -1450,25 +1435,23 @@ def test_get_supplier_framework_returns_countersigned_framework_agreement(self,
             'agreedVariations': {}
         }
 
+    def test_get_supplier_framework_info_non_existent_by_framework(self, supplier_framework):
```
TheDoubleK: Can we drop the `by` from the test name?

tests/app/views/test_suppliers.py
```diff
@@ -1450,25 +1435,23 @@ def test_get_supplier_framework_returns_countersigned_framework_agreement(self,
...
-    method = "post"
-    endpoint = "/suppliers/0/frameworks/digital-outcomes-and-specialists"
+        assert response.status_code == 404
 
-    def setup(self):
-        super(TestSupplierFrameworkUpdates, self).setup()
-        self.setup_dummy_suppliers(1)
-        self.setup_dummy_user(1, role='supplier')
-        self.supplier_id = 0
+    def test_get_supplier_framework_info_non_existent_by_supplier(self, supplier_framework):
```
TheDoubleK: And here?

tests/app/views/test_suppliers.py
```diff
@@ -1480,69 +1463,53 @@ def supplier_framework_interest(self, framework_slug, update):
...
             update={'onFramework': True}
         )
         assert response.status_code == 200
         data = json.loads(response.get_data())
-        assert data['frameworkInterest']['supplierId'] == 0
-        assert data['frameworkInterest']['frameworkSlug'], 'digital-outcomes-and-specialists'
+        assert data['frameworkInterest']['supplierId'] == supplier_framework['supplierId']
+        assert data['frameworkInterest']['frameworkSlug'] == supplier_framework['frameworkSlug']
         assert data['frameworkInterest']['onFramework'] is True
-        assert data['frameworkInterest']['agreementReturned'] is False
```
TheDoubleK: Wee losing a lot of asserts here... Is it worth adding these things to the fixture and keeping the asserts in? Or is that a bit pointless?

tests/app/views/test_suppliers.py
```diff
@@ -1480,69 +1463,53 @@ def supplier_framework_interest(self, framework_slug, update):
...
         data = json.loads(response.get_data())
-        assert data['frameworkInterest']['supplierId'] == 0
-        assert data['frameworkInterest']['frameworkSlug'], 'digital-outcomes-and-specialists'
+        assert data['frameworkInterest']['supplierId'] == supplier_framework['supplierId']
+        assert data['frameworkInterest']['frameworkSlug'] == supplier_framework['frameworkSlug']
         assert data['frameworkInterest']['onFramework'] is False
 
     def test_can_set_agreement_returned_without_agreement_details_for_framework_with_no_agreement_version(
-            self, open_dos_framework
+            self, supplier_framework
```
TheDoubleK: How do we know there's no framework agreement version? It isn't obvious from the test setup here.

TheDoubleK: I think this is probably fine as is, but I do think using fixtures like this loses some clarity from tests.

tests/app/views/test_suppliers.py
```diff
@@ -1689,31 +1629,32 @@ def test_setting_signer_details_and_then_returning_agreement(self, open_g8_frame
             "frameworkAgreementVersion": "v1.0",
         }
 
-    def test_can_not_set_agreement_details_on_frameworks_without_framework_agreement_version(self, open_dos_framework):
+    def test_can_not_set_agreement_details_on_frameworks_without_framework_agreement_version(self, supplier_framework):
```
TheDoubleK: Again, we lose some clarity of what's going on with this substitution.

tests/app/views/test_suppliers.py
```diff
@@ -1879,12 +1826,86 @@ def test_changing_on_framework_to_passed_creates_audit_event(self, open_dos_fram
...
+            assert audit.data['frameworkSlug'] == supplier_framework['frameworkSlug']
             assert audit.data['update']['onFramework'] is True
             assert audit.data['update']['agreementReturned'] is True
 
 
+class TestSupplierFrameworkAgreementsDataMigration(BaseApplicationTest):
+    # Tests for whilst we have framework agreement data saved in both the
+    # supplier frameworks and framework agreements table. When framework agreement
+    # data is fully migrated from the supplier frameworks table to the
+    # framework_agreements table then these tests can be removed
```
TheDoubleK: Nice.

tests/app/views/test_agreements.py
```diff
@@ -327,6 +328,192 @@ def test_400_if_signed_agreement_details_contains_empty_strings_as_values(self,
...
+                'signedAgreementReturnedAt': '2016-12-12T00:00:00.000000Z'
+            }
+
+    def test_can_not_sign_framework_agreement_that_has_no_signer_name(self, supplier_framework):
+        agreement_id = self.create_agreement(
+            supplier_framework,
+            signed_agreement_details={'signerRole': 'role'},
+            signed_agreement_path='/example.pdf'
+        )
+        res = self.sign_agreement(agreement_id, {'signedAgreementDetails': {'uploaderUserId': 20}})
```
idavidmcdonald: Good spot

tombye:
👍  for fixing the thing I commented on.

tests/app/views/test_agreements.py
```diff
@@ -1,23 +1,156 @@
...
+            data=json.dumps(
+                {
+                    'updated_by': 'interested@example.com',
+                    'agreement': agreement_data
+                }),
+            content_type='application/json')
+
+    def test_can_create_framework_agreement(self, supplier_framework):
+        res = self.post_create_agreement(
+            supplier_id=supplier_framework['supplierId'],
```
pcraig3: This is the magic of fixtures, @TheDoubleK!

Once you define a fixture (ie, [`supplier_framework`](https://github.com/alphagov/digitalmarketplace-api/blob/master/tests/conftest.py#L41)), then you can simply pass it into any test as an argument. 

[There's a simple example on the first page of the docs](http://doc.pytest.org/en/latest/fixture.html#fixtures-as-function-arguments).

✨ 🔮 ✨

PR #451, opened 2016-09-22
Keep SupplierFrameworks consistent with FrameworkAgreements

TheDoubleK:
While we are using a combination of both tables to store agreement details we should keep the data in both places consistent.

This will fix a bug with unsetting `agreementReturnedAt` where it gets updated in FrameworkAgreement but not in SupplierFrameworks, and so continues to be returned by the list framework agreements endpoint for `agreement_returned=true` even though it has been unset.


idavidmcdonald:
Well hunted down Kev 👍

TheDoubleK:
Fingers crossed...

PR #452, opened 2016-09-22
Deprecate SupplierFramework fields

idavidmcdonald:
This pull request can not be merged until: 
 - [x] https://github.com/alphagov/digitalmarketplace-api/pull/455
has been deployed.

`agreement_returned_at`, `agreement_details` and `countersigned_at` have been deprecated for the `supplier_frameworks` database table. These details are now being stored in the `framework_agreements` table.

Any existing information in these deprecated fields has already been migrated to the `framework_agreements` table.

As we no longer should be using these fields, we remove their direct usage around the codebase. Note, that the `SupplierFramework` serialized interface remains exactly the same as it will retrieve this information from the `framework_agreements` table.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -441,7 +441,6 @@ def update_supplier_framework_details(supplier_id, framework_slug):
     if 'agreementDetails' in update_json and framework.framework_agreement_details is None:
         abort(400, "Framework '{}' does not accept 'agreementDetails'".format(framework_slug))
 
-    # Try and find an existing FrameworkAgreement to update, if not then create one based on data from SupplierFramework
     framework_agreement = FrameworkAgreement.query.filter(
         FrameworkAgreement.supplier_id == supplier.supplier_id,
         FrameworkAgreement.framework_id == framework.id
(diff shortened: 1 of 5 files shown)


app/main/views/suppliers.py
```diff
@@ -441,7 +441,7 @@ def update_supplier_framework_details(supplier_id, framework_slug):
     if 'agreementDetails' in update_json and framework.framework_agreement_details is None:
         abort(400, "Framework '{}' does not accept 'agreementDetails'".format(framework_slug))
 
-    # Try and find an existing FrameworkAgreement to update, if not then create one based on data from SupplierFramework
+    # Try and find an existing FrameworkAgreement to update, if not then create one
```
TheDoubleK: We probably don't need this comment at all now - seems straightforward enough from the code.  Although if we're going to block creation of a new FrameworkAgreement by this route in future it might be worth mentioning that in a comment.

app/models.py
```diff
@@ -328,10 +328,13 @@ class SupplierFramework(db.Model):
                              primary_key=True)
     declaration = db.Column(JSON)
     on_framework = db.Column(db.Boolean, nullable=True)
+    agreed_variations = db.Column(JSON)
+
+    # The following three fields are deprecated as this information now lives in the
```
TheDoubleK: Maybe word this comment more strongly - e.g. `The following three fields are deprecated, no longer updated, may contain outdated data and MUST NOT BE USED. The framework_agreements table is now the only the source of truth for this data.`

TheDoubleK:
Good stuff - I had a check through API code and couldn't find anywhere else these fields are used so looks like this should be it :+1: (with or without suggested changes to comments in code - your call).

PR #453, opened 2016-09-23
Revert data migration, fix finding framework agreements

pcraig3:
### 	Revert "Migrate framework agreements to `framework_agreements`"

This reverts commit 54368e9.

We haven't been returning the correct framework agreement in all
cases, so we're reverting the migration so that we can release
the correct version of the code through to production.

### 	Filter FrameworkAgreement by `supplier_id` and `framework_id`

Before, we were only looking for `FrameworkAgreements` by
supplier_id, which meant we might be getting back an agreement
for the right supplier but for the wrong framework.

This commit will fix this.

### Others

In a couple of places, we've been using `supplier.id` where we should be using `supplier.supplier_id`.  Easy fixes.  Means a bunch of our audit events have the wrong supplier id, but not any longer.


TheDoubleK:
All looks good :+1:

pcraig3:
🚅 🚋 🚋

PR #454, opened 2016-09-23
Add tests for bugfixes in #453

pcraig3:
Adds tests for two of the bugfixes shipped out in pull request #453.

Specifically,
- checks that the correct `supplier_id` is added to the audit event created on a generic framework update
- makes sure that the wrong `FrameworkAgreement` will not be updated 

***

In the latter case, the problem was that the previous code would look for any `FrameworkAgreement` belonging to a supplier and update it.  This meant that if 
- a framework agreement existed for a supplier
- a framework agreement update came in for a _different_ framework
the framework agreement for the first framework would always be returned and updated.  A new framework agreement would never be created.

This test verifies that even if a framework agreement for another framework exists, a new framework agreement is always created.

### DIFF
--- a/tests/app/views/test_suppliers.py
+++ b/tests/app/views/test_suppliers.py
@@ -1810,17 +1810,15 @@ class TestSupplierFrameworkUpdates(BaseApplicationTest):
         assert data['frameworkInterest']['onFramework'] is False
         assert data['frameworkInterest']['agreementReturned'] is False
 
-    def test_updating_framework_agreement_creates_audit_event(self, supplier_framework):
+    def test_updating_framework_agreement_creates_audit_event(self, live_example_framework, supplier_framework):
         self.supplier_framework_interest(
             supplier_framework,
             update={'signerName': "Josh Moss", 'signerRole': "The Boss"}
         )
         with self.app.app_context():
-            framework = Framework.query.filter(Framework.slug == supplier_framework['frameworkSlug']).first()
-
             framework_agreement = FrameworkAgreement.query.filter(
                 FrameworkAgreement.supplier_id == supplier_framework['supplierId'],
-                FrameworkAgreement.framework_id == framework.id
+                FrameworkAgreement.framework_id == live_example_framework['id']
             ).first()
 
             audit = AuditEvent.query.filter(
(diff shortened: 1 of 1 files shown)


tests/app/views/test_suppliers.py
```diff
@@ -1810,10 +1810,35 @@ def test_changing_on_framework_from_passed_to_failed(self, supplier_framework):
...
         assert data['frameworkInterest']['agreementReturned'] is False
 
+    def test_updating_framework_agreement_creates_audit_event(self, supplier_framework):
+        self.supplier_framework_interest(
+            supplier_framework,
+            update={'signerName': "Josh Moss", 'signerRole': "The Boss"}
+        )
+        with self.app.app_context():
+            print(supplier_framework)
+            framework = Framework.query.filter(Framework.slug == supplier_framework['frameworkSlug']).first()
```
allait: Can you get this information from the `framework` fixture instead of querying?

tests/app/views/test_suppliers.py
```diff
@@ -1810,10 +1810,35 @@ def test_changing_on_framework_from_passed_to_failed(self, supplier_framework):
         assert data['frameworkInterest']['onFramework'] is False
         assert data['frameworkInterest']['agreementReturned'] is False
 
+    def test_updating_framework_agreement_creates_audit_event(self, supplier_framework):
+        self.supplier_framework_interest(
+            supplier_framework,
+            update={'signerName': "Josh Moss", 'signerRole': "The Boss"}
+        )
+        with self.app.app_context():
+            print(supplier_framework)
```
allait: Probably don't need this

pcraig3:
💀

pcraig3:
After talking with @idavidmcdonald, it sounds like both of these tests are no longer needed.

- The first tests tests the second of the two audit events that are being created in the [`update_supplier_framework_details`](https://github.com/alphagov/digitalmarketplace-api/blob/5032d43b0ba2bb881eca2f5e58a7360fdb66a1b5/app/main/views/suppliers.py#L499) route, but apparently this route is being hacked up over the next two days.
- The second test makes sure that the correct framework is returned in [`the same route`](https://github.com/alphagov/digitalmarketplace-api/blob/5032d43b0ba2bb881eca2f5e58a7360fdb66a1b5/app/main/views/suppliers.py#L444).  However, since we're moving towards a world where a SupplierFramework has multiple FrameworkAgreements, the logic this test relies on is also being cut up and redistributed.

PR #456, opened 2016-09-26
Add `signed_agreement_put_on_hold_at` migration 

idavidmcdonald:
Used to signify when a signed framework agreement is on hold when being reviewed for countersigning.

### DIFF
--- a/migrations/versions/750_framework_agreement_on_hold_at.py
+++ b/migrations/versions/750_framework_agreement_on_hold_at.py
@@ -1,0 +1,22 @@
+"""framework agreement on hold at
+
+Revision ID: 750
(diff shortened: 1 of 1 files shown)


allait:
I know this one is probably safe, but to keep the habit, should we release the migration separately?

idavidmcdonald:
When you say the migration separately you are referring to the just the migration file and therefore not the one line added to the model?

allait:
@idavidmcdonald yeah, the migration file should go first, before the model change

TheDoubleK:
:+1: but let's hold off merging just until the previous code change is through, and do everything properly one step at a time.

PR #457, opened 2016-09-26
Add Framework agreement status and route to set `FrameworkAgreement` as on hold

idavidmcdonald:
- Adds the `signed_agreement_put_on_hold` field to the `FrameworkAgreement` model from the above mentioned pull request and it's serialization.
- Adds a `status` property to a `FrameworkAgreement`. This is done in a similar way to the `status` field for `Brief`'s.
- Adds a route to set a framework agreement as `on hold`.
- Add `agreementStatus` field to `SupplierFramework.serialize`

#### Decided not needed/in scope (although may be added later)
- Set `FrameworkAgreement fields`, such as `signed_agreement_returned_at` or `signed_agreement_put_on_hold_at` using the `status` field? The `Brief` model currently does this.
- Add ability to query framework agreements using the `status` field. Again, the `Brief` model currently does this.
- Add `signedAgreementPutOnHold` field to `SupplierFramework.serialize`

### DIFF
--- a/tests/app/views/test_agreements.py
+++ b/tests/app/views/test_agreements.py
@@ -50,6 +50,7 @@ class TestCreateFrameworkAgreement(BaseApplicationTest):
         assert res_agreement_json['id'] > 0
         assert res_agreement_json['supplierId'] == supplier_framework['supplierId']
         assert res_agreement_json['frameworkSlug'] == supplier_framework['frameworkSlug']
+        assert res_agreement_json['status'] == 'draft'
 
         res2 = self.client.get('/agreements/{}'.format(res_agreement_json['id']))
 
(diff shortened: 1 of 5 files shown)


TheDoubleK:
Let's add the FA status as `framework_agreement_status` in the SupplierFramework serialization - we're going to need it in the frontend apps.

tests/app/views/test_agreements.py
```diff
@@ -657,5 +691,89 @@ def test_can_re_sign_framework_agreement(self, supplier_framework):
...
+
+        assert res.status_code == 400
+        error_message = json.loads(res.get_data(as_text=True))['error']
+        assert error_message == "Framework agreement must have status 'signed' to be put on hold"
+
+    @fixture_params('live_example_framework', {'framework_agreement_details': {'frameworkAgreementVersion': 'v1.0'}})
+    def test_can_not_put_countersigned_framework_agreement_on_hold(self, supplier_framework):
+        agreement_id = self.create_agreement(
+            supplier_framework,
+            countersigned_agreement_returned_at=datetime(2016, 10, 1)
```
TheDoubleK: Could we have both `signed` and `countersigned` timestamps in here - to prove it's the presence of countersignature rather than absence of signature that causes the error?

TheDoubleK:
:+1: once Travis is happy

PR #459, opened 2016-09-27
Return `agreementId` in `SupplierFramework.serialize`

idavidmcdonald:
So given a SupplierFramework we will know the ID of the relevant FrameworkAgreement. This is beneficial now for when we have a 1-1 relationship but will also be used when we have a 1-n relationship as the serialized SupplierFramework will always include the most relevant FrameworkAgreement.


TheDoubleK:
:+1:

PR #460, opened 2016-09-27
[#130991743] Remove application_status from framework users export

allait:
Current implementation of application_status check requires N+1 queries to request draft services for each framework supplier. This is timing out due to the number of suppliers in production.

As a temporary workaround, since we're mainly interested in exports for live frameworks at the moment we're removing application status values from the export.

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -244,35 +244,18 @@ def export_users_for_framework(framework_slug):
         User.active.is_(True)
     ).all()
 
-    submitted_draft_counts_per_supplier = {}
     user_rows = []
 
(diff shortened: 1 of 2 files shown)


TheDoubleK:
:+1:
Yep let's get this in and worry later about how to do it properly... not sure if making application status a property on `supplier_framework` will help as it will still require the same amount of database lookups, just wrapped with syntactic sugar.  Another way might be storing and managing a separate count-of-submitted-services field, but feels a bit mad.

TheDoubleK:
If we do want to do that though we'd ideally need to get it in before DOS2 opens, thought could do a migration easily enough later I guess.

PR #461, opened 2016-09-29
Add temporary countersign route needed for one off script

idavidmcdonald:
Temporary route needed so we can set a user defined timestamp for an agreement to be countersigned (the real route that is still to be created will not allow the user to define the timestamp but will take the current time). This route will also allow us to set the countersigned path. After a one off refactoring script is run then this route will no longer be needed and we can delete it.

### DIFF
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -237,3 +237,53 @@ def put_signed_framework_agreement_on_hold(agreement_id):
         return jsonify(message="Database Error: {0}".format(e)), 400
 
     return jsonify(agreement=framework_agreement.serialize())
+
+
+@main.route('/agreements/<int:agreement_id>/countersign-script', methods=['POST'])
+def temp_countersign_time_and_path(agreement_id):
+    """Temporary countersign route needed for one off script
+
+    Temporary route needed so we can set a user defined timestamp for an agreement
+    to be countersigned (the real route that is still to be created will not allow
+    the user to define the timestamp but will take the current time). This route
+    will also allow us to set the countersigned path. After a one off refactoring script
+    is run then this route will no longer be needed and we can delete it.
+    """
+    framework_agreement = FrameworkAgreement.query.filter(FrameworkAgreement.id == agreement_id).first_or_404()
+
+    json_payload = get_json_from_request()
+    updater_json = validate_and_return_updater_request()
+    json_has_required_keys(json_payload, ["agreement"])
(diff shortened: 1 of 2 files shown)


app/main/views/agreements.py
```diff
@@ -237,3 +237,53 @@ def put_signed_framework_agreement_on_hold(agreement_id):
...
+    updater_json = validate_and_return_updater_request()
+    json_has_required_keys(json_payload, ["agreement"])
+    update_json = json_payload["agreement"]
+
+    json_has_keys(
+        update_json,
+        required_keys=['countersignedAgreementPath', 'countersignedAgreementReturnedAt']
+    )
+
+    if framework_agreement.status != 'signed':
```
TheDoubleK: Could it also be "on hold"?  I guess not for now...

app/main/views/agreements.py
```diff
@@ -237,3 +237,53 @@ def put_signed_framework_agreement_on_hold(agreement_id):
...
+        abort(400, "Framework agreement must have status 'signed' to be countersigned")
+
+    framework_agreement.countersigned_agreement_returned_at = update_json['countersignedAgreementReturnedAt']
```
TheDoubleK: Do we not need to do anything to convert the string into a timestamp or anything?  The tests pass, so I guess not, but feels a bit like magic!

tests/app/views/test_agreements.py
```diff
@@ -778,3 +778,69 @@ def test_can_not_put_framework_agreement_on_hold_that_has_no_framework_agreement
...
+    def test_temp_countersign(self, supplier_framework):
+        agreement_id = self.create_agreement(
+            supplier_framework,
+            signed_agreement_returned_at=datetime(2016, 10, 1),
+        )
+
+        res = self.temp_countersign(
+            agreement_id,
+            'example/path/file.pdf',
+            '2016-11-01T00:00:00.000000Z'
```
TheDoubleK: Would it be useful to have a comment on the route about the time format required for it to work properly?

TheDoubleK:
Think this will do us after a rebase :+1:

app/main/views/agreements.py
idavidmcdonald: Yes, I wasn't sure how exactly the datetime stuff was going to work (like what it was expecting from the frontend). It appears to work for the tests and I've just double checked manually with psql to make sure it's all working as expected. Magic!

PR #462, opened 2016-09-29
[#128669753] Add migration to add G8 contract variation agreed-by-CCS details

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/128669753

**CONTRACT VARIATION IS GO! GO! GO!** 

We are almost ready for the contract variation to come into place.

This migration needs to run when the time comes to set countersigner details on the variation.

This was the neatest way I could figure to make the update readable but not add a load of whitespace and newlines into the database json field.  Open to suggestions if you can see a way to make it prettier.  (My first version was all on one line, which was kind-of fine...)


TheDoubleK:
Going to do this through the new API endpoint instead.

PR #463, opened 2016-09-29
Simplify users export query and avoid calling sf.serialize

allait:
Export endpoint is still timing out in production. Looking at the recent changes it seems that serializing supplier_framework records to check for framework agreement status might be slowing down the API response.

The query time in production is ~6s, so it doesn't feel like it's the main cause of 30s timeouts.

This simplifies the database query and replaces `sf.serialize` with an attribute check.

#### Add `SupplierFramework.current_framework_agreement` property
Allows looking up the current framework agreement for the supplier.
Right now the only way to access this is either using `agreements[0]` from the view (which would break in the future) or calling `supplier_framework.serialize()`.

Adding a property encapsulates the current agreement logic in a model property, which should make it easy to use in other views.

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -234,19 +234,19 @@ def export_users_for_framework(framework_slug):
     if framework.status == 'coming':
         abort(400, 'framework not yet open')
 
-    supplier_frameworks_and_suppliers_and_users = db.session.query(
-        SupplierFramework, Supplier, User
+    supplier_frameworks_and_users = db.session.query(
+        SupplierFramework, User
     ).join(
-        Supplier, User, Framework
+        User, SupplierFramework.supplier_id == User.supplier_id
     ).filter(
-        Framework.slug == framework_slug
+        SupplierFramework.framework == framework
     ).filter(
         User.active.is_(True)
     ).all()
 
     user_rows = []
 
-    for sf, s, u in supplier_frameworks_and_suppliers_and_users:
+    for sf, u in supplier_frameworks_and_users:
 
         # always get the declaration status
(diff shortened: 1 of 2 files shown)


app/models.py
```diff
@@ -422,7 +427,7 @@ def serialize(self, data=None):
         }, **(data or {}))
 
         if self.framework_agreements:
```
idavidmcdonald: Should we still check `self.framework_agreements` here or should we now check `self.current_framework_agreement`?

app/main/views/users.py
```diff
@@ -260,13 +260,13 @@ def export_users_for_framework(framework_slug):
                 application_result = 'no result'
             else:
                 application_result = 'pass' if sf.on_framework else 'fail'
-            framework_agreement = bool(sf.serialize()['agreementReturnedAt'])
+            framework_agreement = bool(getattr(sf.current_framework_agreement, 'signed_agreement_returned_at', None))
```
idavidmcdonald: You could use the `status` property of a framework agreement instead of the `signed_agreement_returned_at` if we want our code to be more reliant on the status field. (This way is also fine though)

allait: I'd have to check for both 'signed', 'countersigned' and 'on_hold' though. It seems that checking the attribute directly makes more sense in this case right now. Maybe if we change the column to show the status at some point the status will be helpful though.

app/models.py
allait: Good point. Updated with `agreement` check

app/main/views/users.py
idavidmcdonald: Good point

PR #464, opened 2016-09-30
Remove temporary countersign agreement route

idavidmcdonald:
Route was only needed for a one off script and can now be removed.


TheDoubleK:
:fire: :+1: :fire:

PR #466, opened 2016-10-04
Create endpoint for countersigning framework_agreements

Wynndow:
For (this story on Pivotal)[https://www.pivotaltracker.com/story/show/121577487]

This PR creates and endpoint on the API to allow framework_agreements to be countersigned.

The details of the official countersigner for the framework need to be recorded however we're not entirely sure how that's going to work at present. The approach taken for now is to look in the `framework_agreement_details` on the framework and copy `agreementCountersignerDetails` if it exists and update it with the logged in user (to record who actually clicked the button). This may need to change in future.

The other point of note is that if a framework_agreement is on hold, that timestamp needs resetting. This is done by just setting it to `None`

Next up is to hook up the button on the admin app to hit this endpoint.

### DIFF
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -285,7 +285,7 @@ def approve_for_countersignature(agreement_id):
         data={
             'supplierId': framework_agreement.supplier_id,
             'frameworkSlug': framework_agreement.supplier_framework.framework.slug,
-            'status': 'countersigned'
+            'status': 'approved'
         },
         db_object=framework_agreement
     )
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -265,7 +265,7 @@ def countersign_agreement(agreement_id):
     framework_agreement.countersigned_agreement_returned_at = datetime.utcnow()
 
     countersigner_details = framework_agreement_details.get('agreementCountersignerDetails', {})
-    countersigner_details.update(updater_json)
+    countersigner_details.update({'updatedBy': updater_json['updated_by']})
     framework_agreement.countersigned_agreement_details = countersigner_details
 
     audit_event = AuditEvent(
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -264,7 +264,7 @@ def countersign_agreement(agreement_id):
     framework_agreement.signed_agreement_put_on_hold_at = None
     framework_agreement.countersigned_agreement_returned_at = datetime.utcnow()
 
-    countersigner_details = framework_agreement_details.get('agreementCountersignerDetails', {})
+    countersigner_details = {'countersignerName': framework_agreement_details.get('countersignerName', '')}
     countersigner_details.update({'updatedBy': updater_json['updated_by']})
     framework_agreement.countersigned_agreement_details = countersigner_details
 
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -247,6 +247,7 @@ def put_signed_framework_agreement_on_hold(agreement_id):
 
     return jsonify(agreement=framework_agreement.serialize())
 
+
 @main.route('/agreements/<int:agreement_id>/countersign', methods=['POST'])
 def countersign_agreement(agreement_id):
     framework_agreement = FrameworkAgreement.query.filter(FrameworkAgreement.id == agreement_id).first_or_404()
@@ -263,7 +264,7 @@ def countersign_agreement(agreement_id):
     framework_agreement.signed_agreement_put_on_hold_at = None
     framework_agreement.countersigned_agreement_returned_at = datetime.utcnow()
 
-    countersigner_details = framework_agreement_details.get('agreement_countersigner_details', {})
+    countersigner_details = framework_agreement_details.get('agreementCountersignerDetails', {})
     countersigner_details.update(updater_json)
     framework_agreement.countersigned_agreement_details = countersigner_details
 
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -253,6 +253,10 @@ def countersign_agreement(agreement_id):
     framework_agreement = FrameworkAgreement.query.filter(FrameworkAgreement.id == agreement_id).first_or_404()
     framework_agreement_details = framework_agreement.supplier_framework.framework.framework_agreement_details
 
+    json_payload = get_json_from_request()
+    json_has_required_keys(json_payload, ['userId'])
+    updater_id = json_payload['userId']
+
     updater_json = validate_and_return_updater_request()
 
     if framework_agreement.status not in ['signed', 'on hold']:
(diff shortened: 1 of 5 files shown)


app/main/views/agreements.py
```diff
@@ -237,3 +237,45 @@ def put_signed_framework_agreement_on_hold(agreement_id):
...
+def countersign_agreement(agreement_id):
+    framework_agreement = FrameworkAgreement.query.filter(FrameworkAgreement.id == agreement_id).first_or_404()
+    framework_agreement_details = framework_agreement.supplier_framework.framework.framework_agreement_details
+
+    updater_json = validate_and_return_updater_request()
+
+    if framework_agreement.status not in ['signed', 'on hold']:
+        abort(400, "Framework agreement must have status 'signed' or 'on hold' to be countersigned")
+
+    if not framework_agreement_details or not framework_agreement_details.get('frameworkAgreementVersion'):
```
Wynndow: Cool. Have updated.

tests/app/views/test_agreements.py
```diff
@@ -778,3 +778,125 @@ def test_can_not_put_framework_agreement_on_hold_that_has_no_framework_agreement
...
+                    'updated_by': 'chris@example.com'
+                }),
+            content_type='application/json')
+
+    @fixture_params(
+        'live_example_framework', {
+            'framework_agreement_details': {
+                'frameworkAgreementVersion': 'v1.0',
+                'agreementCountersignerDetails': {
+                    'countersignerName': 'The Boss'}
```
Wynndow: Yep, moved when I adjusted for the correct key for countersignerName

app/main/views/agreements.py
```diff
@@ -237,3 +237,56 @@ def put_signed_framework_agreement_on_hold(agreement_id):
...
+        if framework_agreement_details.get('countersignerName'):
+            countersigner_details.update({
+                'countersignerName': framework_agreement_details['countersignerName']
+            })
+        if framework_agreement_details.get('countersignerRole'):
+            countersigner_details.update({
+                'countersignerRole': framework_agreement_details['countersignerRole']
+            })
+
+    countersigner_details.update({'updatedById': updater_id})
```
TheDoubleK: This is specifically the admin user who approved the agreement to be countersigned, and so `updatedById` seems too generic to me (this isn't just about updating a record, but specifically saying "this is good to go"). For uploaded agreements we have `agreedUserId` - this key might be better as `approvedByUserId` rather than `updatedById`.

app/main/views/agreements.py
```diff
@@ -237,3 +237,56 @@ def put_signed_framework_agreement_on_hold(agreement_id):
         return jsonify(message="Database Error: {0}".format(e)), 400
 
     return jsonify(agreement=framework_agreement.serialize())
+
+
+@main.route('/agreements/<int:agreement_id>/countersign', methods=['POST'])
+def countersign_agreement(agreement_id):
```
TheDoubleK: Part of me is starting to prefer the term `approve` rather than `countersign` for this action, as it records an admin user approving the countersignature, rather than the countersigning itself.  Maybe one for discussion, or maybe we just stick with it how it is rather than introducing yet another term.

But maybe adding `approved` as an additional status would clear up the logic in the frontend that checks for a countersigned file path rather than for the status of the FA.  If we added `approved` as a status we could have `approved` = good to go but no countersigned file exists yet and `countersigned` = file has been generated and made available to the supplier.

app/main/views/agreements.py
```diff
@@ -237,3 +237,56 @@ def put_signed_framework_agreement_on_hold(agreement_id):
...
+    countersigner_details.update({'updatedById': updater_id})
+    framework_agreement.countersigned_agreement_details = countersigner_details
+
+    audit_event = AuditEvent(
+        audit_type=AuditTypes.countersign_agreement,
+        user=updater_json['updated_by'],
+        data={
+            'supplierId': framework_agreement.supplier_id,
+            'frameworkSlug': framework_agreement.supplier_framework.framework.slug,
+            'status': 'countersigned'
```
TheDoubleK: Again, this could be `approved` if we decided to go that way.

app/models.py
```diff
@@ -469,6 +469,15 @@ def serialize(self, data=None):
                 supplier_framework['agreementDetails']['uploaderUserName'] = user.name
                 supplier_framework['agreementDetails']['uploaderUserEmail'] = user.email_address
 
+        if supplier_framework['countersignedDetails'] and supplier_framework['countersignedDetails'].get('updatedById'):
+            user = User.query.filter(
+                User.id == supplier_framework['countersignedDetails']['updatedById']
+            ).first()
+
+            if user:
+                supplier_framework['countersignedDetails']['updaterUserName'] = user.name
```
TheDoubleK: Similarly, these would become `agreedByUserName` and `agreedByUserEmail`,

tests/app/views/test_agreements.py
```diff
@@ -778,3 +778,217 @@ def test_can_not_put_framework_agreement_on_hold_that_has_no_framework_agreement
...
+
+        with freeze_time('2016-10-02'):
+            on_hold_res = self.client.post(
+                '/agreements/{}/on-hold'.format(agreement_id),
+                data=json.dumps(
+                    {
+                        'updated_by': 'interested@example.com'
+                    }),
+                content_type='application/json')
+        assert on_hold_res.status_code == 200
```
TheDoubleK: It might be nice to look for the "on hold" status in the JSON response here - just to be doubly sure it is on hold when being countersigned.

tests/app/views/test_agreements.py
```diff
@@ -778,3 +778,217 @@ def test_can_not_put_framework_agreement_on_hold_that_has_no_framework_agreement
...
+            'id': agreement_id,
+            'supplierId': supplier_framework['supplierId'],
+            'frameworkSlug': supplier_framework['frameworkSlug'],
+            'status': 'countersigned',
+            'signedAgreementReturnedAt': '2016-10-01T00:00:00.000000Z',
+            'countersignedAgreementReturnedAt': '2016-10-03T00:00:00.000000Z',
+            'countersignedAgreementDetails': {'updatedById': '1234'}
+        }
+
+    @fixture_params(
```
TheDoubleK: From the `test_can_not_countersign_unsigned_framework_agreement` test above it looks like this fixture will fit on one line.

tests/app/views/test_agreements.py
```diff
@@ -778,3 +778,217 @@ def test_can_not_put_framework_agreement_on_hold_that_has_no_framework_agreement
...
+        'live_example_framework', {
+            'framework_agreement_details': {
+                'frameworkAgreementVersion': 'v1.0'
+            }
+        }
+    )
+    def test_can_countersign_framework_agreement_with_agreement_version_but_no_name_or_role(self, supplier_framework):
```
TheDoubleK: This shouldn't ever happen, but yep I think we should allow it rather than blocking.

app/main/views/agreements.py
Wynndow: Yeah, I think this is probably a good distinction to make, there doesn't seem to be another way of referring to that particular set of agreements.

Where is the frontend checking for a file path?

TheDoubleK: In the new supplier frontend stuff, in PR at the moment.

app/main/views/agreements.py
Wynndow: Yeah, that makes sense. I thought it looked a little off when I was writing it but didn't come up with anything better.

tests/app/views/test_agreements.py
Wynndow: Will do.

tests/app/views/test_agreements.py
Wynndow: Ah yes. It was bigger, but I removed some unneeded stuff and didn't move it to one line.

Wynndow:
@TheDoubleK I've made the changes you've mentioned above plus the other stuff we've talked about this morning.

TheDoubleK:
Good stuff - nice one, Chris :+1:

PR #467, opened 2016-10-05
Add countersigner details to variation JSON schema

TheDoubleK:
Required to countersign the variation.

### DIFF
--- a/json_schemas/framework-agreement-details.json
+++ b/json_schemas/framework-agreement-details.json
@@ -17,6 +17,15 @@
         "properties": {
           "createdAt": {
             "type": "string"
+          },
+          "countersignedAt": {
+            "type": "string"
+          },
+          "countersignerName": {
+            "type": "string"
+          },
+          "countersignerRole": {
+            "type": "string"
           }
         },
         "required": [


pcraig3:
Actually, one thing I forgot to say is that I've already put in a [`countersignerName`](https://github.com/alphagov/digitalmarketplace-api/blob/c92c5d843755a55d05c2d525436b61041bee014e/json_schemas/framework-agreement-details.json#L6) field into  the `Framework.framework_agreement_details`, so maybe it's worth matching up the names a bit more closely so that we're not super confused.

What do you think?

pcraig3:
Matches the [`signerName`](https://github.com/alphagov/digitalmarketplace-api/blob/fd2ef8098220bba7b3a703701e6e6b463abe9935/json_schemas/agreement-details.json#L10) stuff we have for `agreement_details`.

TheDoubleK:
I'm not sure.  `countersignedBy` feels more like a definite thing done at a specific time, whereas the `countersignerName` is the person who notionally signs agreements as admin presses the green button. The `signerName` thing might be better as `signedBy`, but it's too late to change that at this point.

I don't think having two keys with the exact same name here will make it less confusing though - i.e. two `countersignerName`s doesn't feel like a win.

TheDoubleK:
OK I've updated it.  Will make frontend changes now.

pcraig3:
👍

PR #469, opened 2016-10-05
Add countersignerRole to frameworkDetails schema

TheDoubleK:
We need to capture both name and role for populating countersigned files - this is the thing we need to complete details in:

screen shot 2016-10-05 at 16 57 27


idavidmcdonald:
👍

PR #472, opened 2016-10-12
Add 'agreement' key to expected json data to approve an agreement

idavidmcdonald:
All API calls should have a top level key that represents the object that is being changed. We had forgotten the 'agreement' key and had been sending the framework agreement data in the top level so this has been fixed to be consistent with the rest of our app.

Note, this route is not yet being used in production so it is not an issue to change our API interface like this.

### DIFF
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -254,8 +254,10 @@ def approve_for_countersignature(agreement_id):
     framework_agreement_details = framework_agreement.supplier_framework.framework.framework_agreement_details
 
     json_payload = get_json_from_request()
-    json_has_required_keys(json_payload, ['userId'])
-    approved_by_user_id = json_payload['userId']
+    json_has_required_keys(json_payload, ['agreement'])
+    update_json = json_payload["agreement"]
+    json_has_keys(update_json, required_keys=['userId'])
+    approved_by_user_id = update_json['userId']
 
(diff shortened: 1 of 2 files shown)


app/main/views/agreements.py
```diff
@@ -254,8 +254,9 @@ def approve_for_countersignature(agreement_id):
     framework_agreement_details = framework_agreement.supplier_framework.framework.framework_agreement_details
 
     json_payload = get_json_from_request()
-    json_has_required_keys(json_payload, ['userId'])
-    approved_by_user_id = json_payload['userId']
+    json_has_required_keys(json_payload, ['agreement'])
+    json_has_required_keys(json_payload['agreement'], ['userId'])
```
TheDoubleK: In other routes we only use `json_has_required_keys` to check the top-level object key, and then use `json_has_keys` to check the keys in the nested object.  I'm not 100% sure why it's done like that, but feels like it would be nice to use the same pattern here for consistency, e.g. something like this would match the patter used elsewhere:
```
json_has_required_keys(json_payload, ["agreement"])
update_json = json_payload["agreement"]
json_has_keys(update_json, required_keys=['userId'])
```

idavidmcdonald: Thanks, have updated as such

TheDoubleK:
:+1:

PR #473, opened 2016-10-13
**DOS2: Add DOS2 schemas PLUS BUGFIX for single-service-lots

TheDoubleK:
These schemas are built from the latest (pull-requested) content - version 3.3.0 (see https://github.com/alphagov/digitalmarketplace-frameworks/pull/315).

These supercede the versions in @allait's branch from the spike (https://github.com/alphagov/digitalmarketplace-api/tree/open-dos2)

See latest commit for a relatively big bugfix for single-service lots.  For the test I tried very hard to use the fixtures rather than set up the DOS2 framework in the database, but the fixtures can't set up framework backrefs to lots (at least, I couldn't make it work).  I think adding it directly to the session is fine, and as it is added with an ID > 100 it gets removed during the test teardown() anyway and so won't interfere with other tests.


tests/app/helpers.py
```diff
@@ -247,6 +247,25 @@ def setup_dummy_services_including_unpublished(self, n):
             )
             db.session.commit()
 
+    def setup_dos_2_framework(self, status='open', clarifications=True):
```
TheDoubleK: No - I left a note in the PR message - also needs associated Lots to be set up which I couldn't figure out how to do with fixtures.

idavidmcdonald:
👍

PR #474, opened 2016-10-13
Add "current" framework agreement and filter by status

TheDoubleK:
This adds the "current_framework_agreement" to the supplier_framework and allows filtering the list of supplier_frameworks by their status (signed, countersigned, on-hold).

We currently include draft agreements if they exist, in order to keep the signing flow working.  Once the signing flow is updated to use FrameworkAgreemnts directly, rather than updating via the SupplierFramework, drafts can be excluded.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -360,22 +360,16 @@ class SupplierFramework(db.Model):
             if self.framework_agreements[-1].status == 'draft':
                 return self.framework_agreements[-1]
 
-            def get_most_recent_time(framework_agreement):
-                if framework_agreement.countersigned_agreement_returned_at:
-                    return framework_agreement.countersigned_agreement_returned_at
-                else:
-                    return framework_agreement.signed_agreement_returned_at
-
             most_recently_signed_or_countersigned = self.framework_agreements[-1]
-            most_recent_time = get_most_recent_time(most_recently_signed_or_countersigned)
+            most_recent_time = most_recently_signed_or_countersigned.most_recent_signature_time
 
             for fa in self.framework_agreements:
                 if fa.status == "draft":
                     continue
 
-                if get_most_recent_time(fa) > most_recent_time:
+                if fa.most_recent_signature_time > most_recent_time:
                     most_recently_signed_or_countersigned = fa
-                    most_recent_time = get_most_recent_time(fa)
+                    most_recent_time = fa.most_recent_signature_time
 
             return most_recently_signed_or_countersigned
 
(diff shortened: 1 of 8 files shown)


app/models.py
```diff
@@ -342,8 +342,42 @@ class SupplierFramework(db.Model):
...
+        After the supplier frontend has been refactored so that the signing flow deals with a framework agreement rather
+        than a suqpplier framework then we will be able to ignore drafts and not include them in the current framework
+        agreement
+        """
         if self.framework_agreements:
-            return self.framework_agreements[0]
+            if self.framework_agreements[-1].status == 'draft':
+                return self.framework_agreements[-1]
+
+            def get_most_recent_time(framework_agreement):
```
allait: This feels like it should be a method/property on `FrameworkAgreement`

TheDoubleK: Makes sense - we've moved it to a property.

TheDoubleK:
@allait - waiting for the afternoon while other people sorted out the problem with pycparser and released a new version has made tests run green, so think this is hopefully good to go now?

allait:
@TheDoubleK it looks good to me :+1:. I'm having a hard time evaluating the queries though. Seeing how we had some performance issues when joining the agreements table before, do we have a plan for testing this before going to prod?

TheDoubleK:
I know what you mean - the filtering by status is a worry and I'm not sure how we test it before it gets to prod.  The query will get simpler very soon, and the plan is at that point to create an SQL version of the "current framework agreement" query that should speed things up a fair bit.
I'm tempted to ship it as is to see how the page performs with our production data (the G8 list page page is not linked from anywhere in the app yet).  If it breaks then we wait until the refactor and SQL query exists before we add the link for CCS users to get to the page.  If it works then we can give them access sooner rather than later.

I appreciate that this may be considered a bit gung-ho, so am open to other suggestions as to how to test it on big data.

TheDoubleK:
@allait - one more commit pushed that might be worth a quick look over, plus any thoughts you have about shipping this with the possibility that it could time out in production would be appreciated before I merge it.

TheDoubleK:
Another option would be to copy data from prod to staging to check it, as there's a lot of G8 agreements returned in prod already.  But we'd need to update the migration job I think because agreement tables aren't included in the export AFAIK.

allait:
@TheDoubleK do we know of any buyer/supplier pages that are using the list of supplier frameworks? The only one I can think of is "Suppliers A-Z", but it's using a different query in `/suppliers`.

If you can think of any, what I suggest we could do is look at the SQL query API runs for that page and run that query on a production DB with time/explain. Otherwise, this should be good to go.

TheDoubleK:
No, I'm pretty sure it is only the admin app list that does this (and possibly some of our scripts).  All other pages just request individual supplier_framework records, there's generally no need to list all suppliers.

PR #475, opened 2016-10-20
DEMO: approximately turn SupplierFramework.current_framework_agreement into a db.relationship and FrameworkAgreement.most_recent_signature_time into a hybrid_property

risicle:
This should be much faster and more flexible, however it doesn't yet cover the "most recent being a draft" case. I think that _may_ be possible, but I don't know how long you're going to be keeping the `draft` condition around...

Don't merge. Tests don't all pass because they expect the "most recent draft" behaviour.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -6,7 +6,7 @@ from flask import current_app
 from flask_sqlalchemy import BaseQuery
 from six import string_types, iteritems
 
-from sqlalchemy import asc, desc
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Yep we don't want the drafts in, they will go very soon (like a day or so).

TheDoubleK:
I'm going to close this PR as this is in the branch I'm currently working on.

PR #477, opened 2016-10-24
[#129838783] Add `submitted_at` migration to `brief_responses` table

idavidmcdonald:
Represents when a brief response has been submitted and is therefore a complete response, as opposed to one that may be in a draft form. Note, we chose not to use the word "published" as a brief response is not published publicly.

This field can then be used to decide the status of a brief response.


allait:
👍

PR #478, opened 2016-10-24
BUGFIX - "approved" agreements have been returned

TheDoubleK:
The `agreement_returned` filter was missing the `approved` status.  This fixes it.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -230,7 +230,7 @@ def get_framework_suppliers(framework_slug):
         agreement_returned = request.args.get('agreement_returned')
         if agreement_returned is not None:
             if convert_to_boolean(agreement_returned):
-                requested_statuses = requested_statuses or ('signed', 'on-hold', 'countersigned')
+                requested_statuses = requested_statuses or ('signed', 'on-hold', 'approved', 'countersigned')
             else:
                 supplier_frameworks = [sf for sf in supplier_frameworks if sf.current_framework_agreement is None
                                        or (sf.current_framework_agreement
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -238,6 +238,7 @@ def get_framework_suppliers(framework_slug):
                                            )
                                        ]
                 # TODO: The 'or' clause here can be deleted once drafts are excluded from current_framework_agreement
+                # TODO: Much of the logic here would be better done querying with SQL than manipulating with Python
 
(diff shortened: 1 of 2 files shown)


tests/app/views/test_frameworks.py
```diff
@@ -845,6 +845,28 @@ def test_list_suppliers_by_multiple_statuses_2(self, live_g8_framework):
...
+            response = self.client.get(
+                '/frameworks/g-cloud-8/suppliers?agreement_returned=true'
+            )
+
+            assert response.status_code == 200
+            data = json.loads(response.get_data())
+            assert len(data['supplierFrameworks']) == 7
+            self.assert_supplier_ids(data, (3, 4, 5, 6, 7, 8, 9))
+
+    def test_list_suppliers_by_agreement_returned_false(self, live_g8_framework):
```
idavidmcdonald: These two new tests you've added are good but it looks to slightly duplicate tests on lines 577, 588. Maybe we could merge these together?

It looks like the test on 577 has an important assertion which could be absorbed into these new tests.
The test on 588 looks to not be doing much different and could probably be absorbed fully here?

app/main/views/frameworks.py
```diff
@@ -230,7 +230,7 @@ def get_framework_suppliers(framework_slug):
         agreement_returned = request.args.get('agreement_returned')
         if agreement_returned is not None:
             if convert_to_boolean(agreement_returned):
-                requested_statuses = requested_statuses or ('signed', 'on-hold', 'countersigned')
+                requested_statuses = requested_statuses or ('signed', 'on-hold', 'approved', 'countersigned')
```
risicle: Side note there's an awful lot of python stuff going on around here that is probably better done in sql.

risicle:
I agree with @idavidmcdonald 's test-related comment but if you just want to get this is in I'll :+1:

app/main/views/frameworks.py
TheDoubleK: Agreed - there is an API "tidy-up" to be done once we can safely ignore draft agreements everywhere and I'll address at least some of the python-vs-sql-ness then.

tests/app/views/test_frameworks.py
TheDoubleK: Good point, will sort it out.

TheDoubleK:
Rebased and updated tests accordingly.
I'll take the existing thumbs-up once Travis is happy.

PR #479, opened 2016-10-25
Add a status to brief responses

Wynndow:
For [this story on Pivotal](https://www.pivotaltracker.com/story/show/129838783).

As part of the story around changing the supplier flow for applying to a brief, briefs can be in an unsubmitted state (draft). 

Brief responses can now have a submitted_at field, which is used to deduce their status. 

We provide a submit brief response route to mark a brief response as submitted.

Next steps:
- Current front end one page flow will use the submit brief response route whenever a brief response is submitted in addition to the create brief response route it already uses.
- A migration will be done on all old brief responses to mark them as submitted.
- Brief responses will need to not include drafts when being listed.

### DIFF
--- a/app/main/views/brief_responses.py
+++ b/app/main/views/brief_responses.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 from flask import jsonify, abort, request, current_app
 from sqlalchemy.exc import IntegrityError, DataError
 
@@ -42,7 +44,7 @@ def create_brief_response():
 
     brief_service = get_supplier_service_eligible_for_brief(supplier, brief)
     if not brief_service:
-        abort(400, "Supplier not eligible")
+        abort(400, "Supplier is not eligible to apply to this brief")
 
     # Check if brief response already exists from this supplier
     if BriefResponse.query.filter(BriefResponse.supplier == supplier, BriefResponse.brief == brief).first():
@@ -80,6 +82,54 @@ def create_brief_response():
     db.session.commit()
 
     return jsonify(briefResponses=brief_response.serialize()), 201
+
+
+@main.route('/brief-responses/<int:brief_response_id>/submit', methods=['POST'])
+def submit_brief_response(brief_response_id):
+    updater_json = validate_and_return_updater_request()
+
+    brief_response = BriefResponse.query.filter(
+        BriefResponse.id == brief_response_id
+    ).first_or_404()
+
+    brief = brief_response.brief
+    supplier = brief_response.supplier
+    brief_service = get_supplier_service_eligible_for_brief(supplier, brief)
+    if not brief_service:
+        abort(400, "Supplier is not eligible to apply to this brief")
+
+    if brief_response.status != 'draft':
+        abort(400, "Brief response must be a draft")
+
+    if brief.framework.status != 'live':
+        abort(400, "Brief framework must be live")
+
+    brief_role = brief.data["specialistRole"] if brief.lot.slug == "digital-specialists" else None
+    service_max_day_rate = brief_service.data[brief_role + "PriceMax"] if brief_role else None
+
+    brief_response.validate(max_day_rate=service_max_day_rate)
+
+    brief_response.submitted_at = datetime.utcnow()
+
+    audit = AuditEvent(
+        audit_type=AuditTypes.submit_brief_response,
+        user=updater_json['updated_by'],
+        data={
+            'briefResponseId': brief_response.id
+        },
+        db_object=brief_response,
+    )
+
+    db.session.add(brief_response)
+    db.session.add(audit)
+
+    try:
+        db.session.commit()
+    except IntegrityError as e:
+        db.session.rollback()
+        abort(400, e.orig)
+
+    return jsonify(briefResponses=brief_response.serialize()), 200
 
 
 @main.route('/brief-responses/<int:brief_response_id>', methods=['GET'])
--- a/tests/app/views/test_brief_response.py
+++ b/tests/app/views/test_brief_response.py
@@ -1,7 +1,9 @@
 import json
 import pytest
+import mock
 
 from hypothesis import given
+from freezegun import freeze_time
 
 from ..helpers import BaseApplicationTest, JSONUpdateTestMixin
 from ... import example_listings
@@ -211,7 +213,7 @@ class TestCreateBriefResponse(BaseBriefResponseTest, JSONUpdateTestMixin):
         })
 
         assert res.status_code == 400
-        assert 'Supplier not eligible' in res.get_data(as_text=True)
+        assert 'Supplier is not eligible to apply to this brief' in res.get_data(as_text=True)
 
     def test_cannot_respond_to_a_brief_that_isnt_live(self, live_dos_framework):
         with self.app.app_context():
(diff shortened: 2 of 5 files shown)


TheDoubleK:
The implementation looks fine I think, but a coupe of questions:
1. what is the benefit in having this `status` property when it's just the existence or not of a submitted timestamp?  Can't we just look for the timestamp?
2. is `created_at` of any relevance to anything any more?  If not then we could change the `created_at` to be called `submitted_at` rather than store two timestamps - this would be fine for existing brief responses with existing `created_at` dates because currently creating and submitting are the same thing.

idavidmcdonald:
Really good questions Kev... Status feels like a nice abstraction, consistent with our other models, but you raise a valid point of could we have the same functionality with much less code.
@allait any thoughts? Chris and I aren't quite sure which way to lean

idavidmcdonald:
Chatted with Alexey. We agreed that we will keep the status field as it's a nice abstraction and consistent.

We also decided to change the status from "submitted" to "live" as this is more consistent across our models.

Should be good for a review/thumbs up if thats OK? @TheDoubleK

TheDoubleK:
"Live" seems a really odd terminology to use for brief responses. I think we talk about framework agreements being submitted, and that makes much more sense to me. I'd say services and briefs can be "live" but not responses.

idavidmcdonald:
After much thought and discussion we have reverted back to "submitted".

TheDoubleK:
Like I said before, implementation looks good :+1:

idavidmcdonald:
After chatting with Alexey, we have removed a commit that sets the `submitted_at` time whenever a brief response is created. Instead we will create a new route that 'submits' a brief response and get the front end to call it instead of letting the API do it automatically.

I will assume your thumbs up still stands Kev (unless I hear otherwise shortly).

Wynndow:
@TheDoubleK Have removed your thumbs up, as after chatting with Alexey we're gonna take a slightly different approach.

idavidmcdonald:
@TheDoubleK Changes complete, ready for final review please (we've added a new route since you've last looked at it).

app/main/views/brief_responses.py
```diff
@@ -82,6 +84,54 @@ def create_brief_response():
...
+    json_payload = get_json_from_request()
+    updater_json = validate_and_return_updater_request()
+
+    brief_response = BriefResponse.query.filter(
+        BriefResponse.id == brief_response_id
+    ).first_or_404()
+
+    brief = brief_response.brief
+    supplier = brief_response.supplier
+    brief_service = get_supplier_service_eligible_for_brief(supplier, brief)
```
TheDoubleK: Actually it's not even doing that check is it - should there also be
```
if not brief_service:
    abort(400, "Supplier is not eligible to apply to this brief")
```

app/main/views/brief_responses.py
```diff
@@ -82,6 +84,54 @@ def create_brief_response():
...
+
+    if brief_response.status != 'draft':
+        abort(400, "Brief response must be a draft")
+
+    if brief.framework.status != 'live':
+        abort(400, "Brief framework must be live")
+
+    brief_role = brief.data["specialistRole"] if brief.lot.slug == "digital-specialists" else None
```
TheDoubleK: I'm really wary of these hard-coded DOS-specific lot names and question keys.  The API shouldn't need to know about these things.

It doesn't really seem scalable to brief building being able to work for "any framework that needs buyers to specify requirements", which is supposed to be the goal.  (I know these are already in the `create_brief_response` route too, but I think it's worth mentioning again that at some point we're going to need to find a way to avoid doing this, perhaps somehow passing information from the frontend app that tell the API what general-purpose rules to apply).

app/main/views/brief_responses.py
```diff
@@ -82,6 +84,54 @@ def create_brief_response():
...
+    service_max_day_rate = brief_service.data[brief_role + "PriceMax"] if brief_role else None
```
TheDoubleK: With future generalisation in mind, maybe this should be something like `service_max_price` or `service_price_limit`?

app/main/views/brief_responses.py
```diff
@@ -82,6 +84,54 @@ def create_brief_response():
...
+
+    brief_response.validate(max_day_rate=service_max_day_rate)
```
TheDoubleK: Same here, although then you'd need to change the validation method in the model... not sure how much hassle that would be?

app/main/views/brief_responses.py
```diff
@@ -82,6 +84,54 @@ def create_brief_response():
...
+
+    db.session.add(brief_response)
+    db.session.add(audit)
+
+    try:
+        db.session.flush()
+    except IntegrityError as e:
+        db.session.rollback()
+        abort(400, e.orig)
+    db.session.commit()
```
TheDoubleK: I think we generally have a `try... except` around our commits.  I don't see the advantage of doing the flush here first?

tests/app/views/test_brief_response.py
```diff
@@ -324,6 +325,110 @@ def test_day_rate_should_be_less_than_service_max_price(self, live_dos_framework
         assert data == {'error': {'dayRate': 'max_less_than_min'}}
 
 
+class TestSubmitBriefResponse(BaseBriefResponseTest):
+
+    def setup_existing_brief_response(self, brief_response_data):
```
TheDoubleK: I think our use of the convention is pretty patchy in our code, and my grasp of the convention is patchy at best, but should these internal helper methods be `_setup_existing_brief_response` and `_submit_brief_response`?

tests/app/views/test_brief_response.py
```diff
@@ -324,6 +325,110 @@ def test_day_rate_should_be_less_than_service_max_price(self, live_dos_framework
...
+        data = json.loads(repeat_res.get_data(as_text=True))
+        assert data == {'error': 'Brief response must be a draft'}
+
+    @given(example_listings.brief_response_data())
+    def test_can_not_submit_a_brief_response_for_a_framework_that_is_not_live(
+        self, live_dos_framework, brief_response_data
+    ):
+        self.setup_existing_brief_response(brief_response_data)
+
+        with self.app.app_context():
```
TheDoubleK: It might be neater here to have an `expired_dos_framework` fixture to pass in, rather than passing in the live one and then updating it?

tests/app/views/test_brief_response.py
```diff
@@ -324,6 +325,110 @@ def test_day_rate_should_be_less_than_service_max_price(self, live_dos_framework
...
+            assert data == {'error': 'Brief framework must be live'}
+
+    @given(example_listings.brief_response_data())
+    def test_can_not_submit_an_invalid_brief_response(
+        self, live_dos_framework, brief_response_data
+    ):
+        self.setup_existing_brief_response(brief_response_data)
+
+        with self.app.app_context():
+            # New data to make brief_response invalid. When we build the update route we will use it instead.
```
TheDoubleK: This should probably be a `TODO` comment if it's something to do later.

tests/app/views/test_brief_response.py
idavidmcdonald: The framework has to be live so we can set up the existing brief response (when we do the create brief response call).

app/main/views/brief_responses.py
idavidmcdonald: Actioned.

app/main/views/brief_responses.py
idavidmcdonald: Good point about flexibility/generalisation. I think for the time being we won't change it but will be something to look at when we want to put DOS2 live.

app/main/views/brief_responses.py
idavidmcdonald: Good point, have removed the flush and moved the commit into the try block.

tests/app/views/test_brief_response.py
idavidmcdonald: Actioned

idavidmcdonald:
Thanks for the good comments Kev. We've addressed most of them. We haven't done anything about making it more generic for different frameworks but this will be something we need to think about in the future.

TheDoubleK:
Nice one :+1:

PR #480, opened 2016-10-25
[129838783] Add migration that sets `submitted_at` for old brief responses

idavidmcdonald:
For [this pivotal story](https://www.pivotaltracker.com/story/show/129838783)

For any brief response that does not have a `submitted_at` time, we add one based on the `created_at` time. As we will now be looking at `submitted_at` rather than `created_at` to indicate a submitted brief response, we need to migrate all older brief responses to have a `submitted_at` time.

This pull request should not be merged until [this pull request](https://github.com/alphagov/digitalmarketplace-api/pull/479) has been deployed.
#### Before

screen shot 2016-10-25 at 11 22 16
#### After (note the order has swapped)

screen shot 2016-10-25 at 11 23 27

### DIFF
--- a/migrations/versions/770_set_submitted_at_for_old_brief_responses.py
+++ b/migrations/versions/770_set_submitted_at_for_old_brief_responses.py
@@ -1,0 +1,27 @@
+"""set submitted at for old brief responses
+
+Revision ID: 770
+Revises: 760
+Create Date: 2016-10-25 11:10:53.245586
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '770'
+down_revision = '760'
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+def upgrade():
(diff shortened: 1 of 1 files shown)


allait:
This should go after the frontend apps are setting `submitted_at` for all new responses. Otherwise we'll end up with new records that went through the old process and don't have the timestamp.

idavidmcdonald:
This line should be sufficient in the pull request that will go out beforehand?
https://github.com/alphagov/digitalmarketplace-api/blob/84f96ed575a5798a74aca0957ea0a693e259fef1/app/main/views/brief_responses.py#L56

idavidmcdonald:
[Related front end pull request](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/545) that sets submitted at for all new responses is up.

TheDoubleK:
:+1: (even though it means I will have to renumber my migration)

PR #482, opened 2016-10-28
Update brief response

idavidmcdonald:
For [this pivotal story](https://www.pivotaltracker.com/story/show/129838783)

- Validate only requested page questions when creating brief response
- Add update brief response route
- Filter brief responses list by status (by default, only including submitted responses)

#### Relies on
- [x] [Supplier front end pull request](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/545) that sends back a list of `page_questions` when creating a brief response.
- [x] [API migration](https://github.com/alphagov/digitalmarketplace-api/pull/480) that sets `submitted_at` for all existing brief responses. When migration is merged then will need to rebase.

### DIFF
--- a/tests/app/views/test_brief_response.py
+++ b/tests/app/views/test_brief_response.py
@@ -72,6 +72,13 @@ class BaseBriefResponseTest(BaseApplicationTest):
             data=json.dumps({
                 'updated_by': 'test@example.com',
                 'briefResponses': data,
+                'page_questions': [
+                    'respondToEmailAddress',
+                    'essentialRequirements',
+                    'niceToHaveRequirements',
+                    'availability',
+                    'dayRate'
+                ]
             }),
             content_type='application/json'
         )
@@ -100,6 +107,46 @@ class TestCreateBriefResponse(BaseBriefResponseTest, JSONUpdateTestMixin):
         assert data['briefResponses']['supplierName'] == 'Supplier 0'
         assert data['briefResponses']['briefId'] == self.brief_id
 
+    def test_create_new_brief_response_with_no_page_questions(self, live_dos_framework):
+        res = self.client.post(
+            '/brief-responses',
+            data=json.dumps({
+                'updated_by': 'test@example.com',
+                'briefResponses': {
+                    'briefId': self.brief_id,
+                    'supplierId': 0
+                },
+                'page_questions': []
+            }),
+            content_type='application/json'
+        )
+
+        data = json.loads(res.get_data(as_text=True))
+
+        assert res.status_code == 201, data
+        assert data['briefResponses']['supplierName'] == 'Supplier 0'
+        assert data['briefResponses']['briefId'] == self.brief_id
+
+    def test_create_new_brief_response_with_missing_response_to_page_question_will_error(self, live_dos_framework):
+        res = self.client.post(
+            '/brief-responses',
+            data=json.dumps({
+                'updated_by': 'test@example.com',
+                'briefResponses': {
+                    'briefId': self.brief_id,
+                    'supplierId': 0,
+                    'respondToEmailAddress': 'email@email.com'
+                },
+                'page_questions': ['respondToEmailAddress', 'availability']
+            }),
+            content_type='application/json'
+        )
+
+        data = json.loads(res.get_data(as_text=True))
+
+        assert res.status_code == 400
+        assert data == {'error': {'availability': 'answer_required'}}
+        
     @given(example_listings.brief_response_data())
     def test_create_brief_response_creates_an_audit_event(self, live_dos_framework, brief_response_data):
         res = self.create_brief_response(dict(brief_response_data, **{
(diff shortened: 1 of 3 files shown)


tests/app/views/test_brief_response.py
```diff
@@ -367,26 +387,140 @@ def test_day_rate_should_be_less_than_service_max_price(self, live_dos_framework
         assert data == {'error': {'dayRate': 'max_less_than_min'}}
 
 
-class TestSubmitBriefResponse(BaseBriefResponseTest):
+class TestUpdateBriefResponse(BaseBriefResponseTest):
+    def _create_brief_response(self):
```
Wynndow: This function is really similar to the one defined on the `BaseBriefResponseTest` class. Could they be reduced to one, passing in the different kwargs as needed?

tombye:
The commit adding the migration is also in https://github.com/alphagov/digitalmarketplace-api/pull/480 so does it need to be here?

idavidmcdonald:
Will need rebasing when the migration goes in itself.

tests/app/views/test_brief_response.py
```diff
@@ -326,26 +389,140 @@ def test_day_rate_should_be_less_than_service_max_price(self, live_dos_framework
...
+    def test_update_brief_response_that_does_not_exist_will_404(self, live_dos_framework):
+        res = self._update_brief_response(100, {'respondToEmailAddress': 'newemail@email.com'})
+        assert res.status_code == 404
+
+    def test_can_not_update_brief_response_for_framework_that_is_not_live(self, live_dos_framework):
+        self._create_brief_response()
+
+        with self.app.app_context():
+            # Change live framework to be expired
+            db.session.execute("UPDATE frameworks SET status='expired' WHERE slug='digital-outcomes-and-specialists'")
```
tombye: I might be misunderstanding this but does this framework status get reset after this test?

Looking at the database teardown that's part of the `BaseApplicationTest` I'm reading it isn't touched. If that's the case this should be reset at the end of this test.

idavidmcdonald: After some digging, it looks like it does. 

Because this test class uses the `live_dos_framework` fixture, the next test that is run after this will make sure that there is a live dos framework available in the database. 

The way this is done is quite smart, that because dos already exists in the database, it will update the dos record in the database as opposed to creating a new one. Thus setting dos back to 'live'. (see https://github.com/alphagov/digitalmarketplace-api/blob/master/tests/conftest.py#L114).

This isn't perfect however because this behaviour only happens because subsequent tests use the `live_dos_framework` fixture. If this was the last test that used that fixture and then the next tests were expected dos to be live then I'm not sure they would pass. This isn't the case, so this currently won't cause any issues. Do we think it's worth doing to be extra safe?

tombye: Didn't realise the test method used the [magic pytest fixtures feature](https://github.com/alphagov/digitalmarketplace-api/blob/master/tests/conftest.py#L295). Given that knowledge this isn't the problem I thought it was 👍

tombye:
With the rebase and Chris's comment on the tests dealt with, this LGMT 👍

tests/app/views/test_brief_response.py
```diff
@@ -100,6 +129,47 @@ def test_create_new_brief_response(self, live_dos_framework, brief_response_data
         assert data['briefResponses']['supplierName'] == 'Supplier 0'
         assert data['briefResponses']['briefId'] == self.brief_id
 
+    def test_create_new_brief_response_with_only_one_page_question(self, live_dos_framework):
```
idavidmcdonald: I think instead the usage case will need to be that we create a brief response with no page questions (as in creating the brief response when you click the start application button rather than when you submit your first page of questions)

tests/app/views/test_brief_response.py
```diff
@@ -326,26 +388,124 @@ def test_day_rate_should_be_less_than_service_max_price(self, live_dos_framework
...
+        res = self.create_brief_response({
             'briefId': self.brief_id,
             'supplierId': 0,
-        }))
-        assert res.status_code == 201
-
+            'essentialRequirements': [False, False, False, False, False],
+            'niceToHaveRequirements': [False, False, False, False, False],
+            'availability': u'a',
+            'respondToEmailAddress': 'supplier@email.com',
```
idavidmcdonald: Would be good to start with a less complete brief response as this is is currently valid (or is it, not sure if it requires a max_day_rate or not to be valid?)

tests/app/views/test_brief_response.py
```diff
@@ -326,26 +388,124 @@ def test_day_rate_should_be_less_than_service_max_price(self, live_dos_framework
...
+        with mock.patch('app.main.views.brief_responses.get_supplier_service_eligible_for_brief') as mock_patch:
+            mock_patch.return_value = None
+
+            res = self._update_brief_response(self.brief_response_id, {'respondToEmailAddress': 'newemail@email.com'})
+            data = json.loads(res.get_data(as_text=True))
+
+            assert res.status_code == 400
+            assert data == {'error': 'Supplier is not eligible to apply to this brief'}
+
+    def test_can_not_update_brief_response_that_has_already_been_submitted(self, live_dos_framework):
```
idavidmcdonald: Note to chris, should be able to use setup dummy brief response here

PR #483, opened 2016-10-31
Bugfix (temporary) - ordering framework agreements

TheDoubleK:
The SQLAlchemy `.order_by` clause was not behaving as expected - rather than ordering on the already-joined framework_agreements table it was doing a second outer join with no relationships specified and ordering on that, which gives undefined results.  (Sometimes it works and sometimes it doesn't).

This TEMPORARILY replaces the broken `order-by` by ordering the SupplierFrameworks using Python, in advance of fixing the API to exclude drafts from `current_framework_agreement`, at which point we can properly sort this using SQL once an expression for `current_framework_agreement` can be more easily defined.

The `signed_agreement_date_or_millenium` method is there to prevent errors when trying to order draft agreements that have `None` returned-at timestamps - it instead considers al drafts to be a product of the millenium bug.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -212,14 +212,10 @@ def get_framework_suppliers(framework_slug):
         Framework.slug == framework_slug
     ).first_or_404()
 
-    # Listing agreements is something done for Admin only (suppliers only retrieve their individual agreements)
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Not going to do this now.

PR #484, opened 2016-11-03
API tidy up - FrameworkAgreements

TheDoubleK:
Now that FrameworkAgreements are updated directly by the frontend apps the API should no longer allow setting and unsetting agreement details through the SupplierFramework update route, so that functionality is removed here.

This also means that we should no longer include "draft" agreements in the SupplierFramework response - only signed/on-hold/countersigned agreements are counted as possible "current framework agreements".  This also means that there is no need to order FAs by id, as that was only used for selecting the "most recent" draft to include.

Three now-unused fields have been removed from the SupplierFramework model - database migration to come in separate PR.  These code changes will need to go in before the migration though.

I have replaced the `current_framework_agreement` property with a `db.relationship`, but I haven't been able to get it to order SupplierFrameworks by current_framework_agreement date.

We have agreed that for now we will not order the SupplierFrameworks returned from the list endpoint, and will instead do the ordering in the Admin frontend where the list response is used.

I have added `TODO`s as it would be nice to get this working at some point in the near future.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -232,12 +232,8 @@ def get_framework_suppliers(framework_slug):
             if convert_to_boolean(agreement_returned):
                 requested_statuses = requested_statuses or ('signed', 'on-hold', 'approved', 'countersigned')
             else:
-                supplier_frameworks = [sf for sf in supplier_frameworks if sf.current_framework_agreement is None
-                                       or (sf.current_framework_agreement
-                                           and sf.current_framework_agreement.status == 'draft'
-                                           )
-                                       ]
-                # TODO: The 'or' clause here can be deleted once drafts are excluded from current_framework_agreement
+                supplier_frameworks = [sf for sf in supplier_frameworks if sf.current_framework_agreement is None]
+
                 # TODO: Much of the logic here would be better done querying with SQL than manipulating with Python
 
         if requested_statuses:
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -212,14 +212,13 @@ def get_framework_suppliers(framework_slug):
         Framework.slug == framework_slug
     ).first_or_404()
 
+    # TODO: Work out a way to order by current_framework_agreement.signed_agreement_returned_at
     # Listing agreements is something done for Admin only (suppliers only retrieve their individual agreements)
     # CCS always want to work from the oldest returned date to newest, so order by ascending date
+
+    # TODO: Work out a way to query.filter() on current_framework_agreement rather than manipulating with Python
     supplier_frameworks = SupplierFramework.query.filter(
         SupplierFramework.framework_id == framework.id
-    ).outerjoin(
-        SupplierFramework.framework_agreements
-    ).order_by(
-        FrameworkAgreement.signed_agreement_returned_at.asc()
     )
 
     if request.args.get('agreement_returned') is not None or request.args.get('status') is not None:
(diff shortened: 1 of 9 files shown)


app/models.py
```diff
@@ -1144,24 +1136,24 @@ def applications_closed_at(cls):
                 '2 weeks 23:59:59', INTERVAL))
 
     @hybrid_property
-    def clarification_questions_closed_at(self):
-        if self.published_at is None:
+    def clarification_questions_closed_at(self_or_cls):
```
idavidmcdonald: Not that experienced with the difference between property and hybrid properties but is another option here to knock it back to being a property? Therefore it won't even be possible to try and use it as a SQL expression?

app/main/views/frameworks.py
```diff
@@ -232,13 +231,7 @@ def get_framework_suppliers(framework_slug):
...
                 requested_statuses = requested_statuses or ('signed', 'on-hold', 'approved', 'countersigned')
             else:
-                supplier_frameworks = [sf for sf in supplier_frameworks if sf.current_framework_agreement is None
-                                       or (sf.current_framework_agreement
-                                           and sf.current_framework_agreement.status == 'draft'
-                                           )
-                                       ]
-                # TODO: The 'or' clause here can be deleted once drafts are excluded from current_framework_agreement
-                # TODO: Much of the logic here would be better done querying with SQL than manipulating with Python
+                supplier_frameworks = [sf for sf in supplier_frameworks if sf.current_framework_agreement is None]
```
risicle: You probably *can* use the ORM to filter these out can't you?

app/models.py
```diff
@@ -330,48 +330,22 @@ class SupplierFramework(db.Model):
...
+        order_by=lambda: desc(FrameworkAgreement.most_recent_signature_time),
+        primaryjoin=lambda: and_(
+            SupplierFramework.supplier_id == FrameworkAgreement.supplier_id,
+            SupplierFramework.framework_id == FrameworkAgreement.framework_id,
+            FrameworkAgreement.status != "draft"
+        ),
+        distinct_target_key=True,
+        uselist=False,
+        viewonly=True,
+    )
```
risicle: Hmmm... come to think of it do we need to specify that this should be an outer join? How else would it be able to return a row for a `SupplierFramework` that had no `FrameworkAgreement`s?

tests/app/test_model.py
```diff
@@ -894,74 +898,105 @@ def test_supplier_has_to_be_associated_with_a_framework(self):
 
     def test_supplier_fails_if_not_associated_with_a_framework(self):
```
risicle: Does this test name relate to what the test does anymore?

risicle:
I'm currently reading the sql output of this new `relationship` to check it does the right thing. It uses a more complex sql construction than I was expecting which is why it's taking a bit of time.

risicle:
Ok I've posted the working, `non_primary` `mapper` approach to the branch `ris-framework-agreement-final-final-really-final-api-changes` though I suggest possibly squashing some of the dead-end approaches out might be better than using that branch as-is so we don't have too much dirty misleading history lying around.

risicle:
Also note I'm now pushing a bunch more things to that branch, not all of which you may want to stick in this PR, so take note & be selective...

TheDoubleK:
OK I've pulled in changes from @risicle to get the current_framework_agreement relationship working properly and rebased to remove the old broken version.  Think this should be good to go now.

risicle:
*Ahem* *cough* :+1: *cough*

PR #485, opened 2016-11-07
Add migration to drop now-unused supplier_framework columns

TheDoubleK:
This will need to be merged AFTER:
 - [x] https://github.com/alphagov/digitalmarketplace-api/pull/484

The recent addition of the framework_agreement table means that these columns
are no longer needed. (They are no longer referred to anywhere in the API code.)

For the downgrade I've spent much of the morning experimenting with trying to get the "current framework agreement" data back in to the SupplierFramework table but had no real success.  Rather than lose more time I'd rather have the downgrade reinstate the dropped columns, and if we ever need to do it we can script an update to populate them with the most recent data.

### DIFF
--- a/migrations/versions/780_remove_unused_cols.py
+++ b/migrations/versions/780_remove_unused_cols.py
@@ -1,0 +1,30 @@
+"""Remove agreement_returned_at, countersigned_at and agreement_details
+   columns from supplier_framework table as they are no longer used
+
+Revision ID: 780
+Revises: 770
+Create Date: 2016-11-07 10:14:00.000000
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '780'
+down_revision = '770'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.drop_column('supplier_frameworks', 'agreement_returned_at')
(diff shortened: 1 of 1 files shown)


allait:
I think we should check that whatever's in the current_framework_ageement in production matches the supplier_framework columns before rolling this out.

And there seems to be a lot of failing tests, so this will probably end up being two separate pull requests.

TheDoubleK:
The tests shouldn't fail at all once #484 is merged.  The data will no longer match if any updates have been made to framework agreements since moving to the new system - these columns aren't touched by the code any more and haven't been kept in step with the framework_agreements table.

TheDoubleK:
#484 is merged now and this rebased.

Here is the script that did the data migration from these SupplierFramework columns to the FrameworkAgreement table a month or so ago: https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/oneoff/save-signed-agreement-path.py

I'm not sure there's much to be gained by cross-referencing data, but let's discuss on Monday.

allait:
Of course, I forgot we've stopped writing to the old table. We probably should make a manual production snapshot before applying this though. Let's do this 👍

PR #486, opened 2016-11-14
Fix initial project setup

danielknell:
- cleared out unused code from bootstrap.sh
- updated bootstrap.sh to use make file for installing dependencies and running migrations

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -7,19 +7,6 @@
 
 set -o pipefail
 
-function display_result {
-  RESULT=$1
-  EXIT_STATUS=$2
-  TEST=$3
-
-  if [ $RESULT -ne 0 ]; then
-    echo -e "\033[31m$TEST failed\033[0m"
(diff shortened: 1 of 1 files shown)


idavidmcdonald:
Looks reasonable 👍 
Will need rebasing on top of master (that might get rid of the third commit too which is a master merge)

galund:
(Learning, learning...) Why rebase? git's really good at merging, rebasing all the time is making things like wot svn did, isn't it?

allait:
@galund one of the reasons is forcing Travis to rerun the tests with the current master. We've had situations before where an open PR was passing tests when it was created, but not when merged into the current master. Pushing the branch forces Travis to rerun and requiring the branch to be up to date makes sure that things don't get merged accidentally without being synchronised with `master`.

Also, linear history is much nicer in the long run. Otherwise you end up with [Guitar Hero](https://twitter.com/HenryHoffman/status/694184106440200192).

PR #487, opened 2016-11-14
[#134017265]get_framework_suppliers rewrite

risicle:
Story https://www.pivotaltracker.com/story/show/134017265

I'd be very interested to see how this performs on the live data. If it's not faster I'll eat some clothing item.

Now orders primarily by `current_framework_agreement.most_recent_signature_time`, with `supplier_id` as secondary disambiguator.

Tests are also better mocking timestamps in a more realistic way.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -224,6 +224,9 @@ def get_framework_suppliers(framework_slug):
 
     supplier_frameworks = supplier_frameworks.filter(
         SupplierFramework.framework_id == framework.id
+    ).order_by(
+        cfa.most_recent_signature_time.asc().nullsfirst(),
+        SupplierFramework.supplier_id,
     )
 
     #
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -212,32 +212,37 @@ def get_framework_suppliers(framework_slug):
         Framework.slug == framework_slug
     ).first_or_404()
 
-    # TODO: Work out a way to order by current_framework_agreement.signed_agreement_returned_at
-    # Listing agreements is something done for Admin only (suppliers only retrieve their individual agreements)
-    # CCS always want to work from the oldest returned date to newest, so order by ascending date
+    # we're going to need an alias for the current_framework_agreement join
+    cfa = orm.aliased(SupplierFramework._CurrentFrameworkAgreement)
+    supplier_frameworks = SupplierFramework.query.outerjoin(
+        cfa, SupplierFramework.current_framework_agreement
+    ).options(
+        orm.contains_eager(
+            SupplierFramework.current_framework_agreement, alias=cfa
+        )
+    )
 
-    # TODO: Work out a way to query.filter() on current_framework_agreement rather than manipulating with Python
-    supplier_frameworks = SupplierFramework.query.filter(
+    supplier_frameworks = supplier_frameworks.filter(
         SupplierFramework.framework_id == framework.id
     )
 
-    if request.args.get('agreement_returned') is not None or request.args.get('status') is not None:
-        requested_statuses = None
-        if request.args.get('status') is not None:
-            requested_statuses = request.args.get('status').split(',')
+    #
+    # bit of an oddly designed interface here in that the two filterable parameters aren't
+    # really orthogonal. implementing them as such anyway for clarity.
+    #
 
-        agreement_returned = request.args.get('agreement_returned')
-        if agreement_returned is not None:
-            if convert_to_boolean(agreement_returned):
-                requested_statuses = requested_statuses or ('signed', 'on-hold', 'approved', 'countersigned')
-            else:
-                supplier_frameworks = [sf for sf in supplier_frameworks if sf.current_framework_agreement is None]
+    agreement_returned = request.args.get('agreement_returned')
+    if agreement_returned is not None:
+        supplier_frameworks = supplier_frameworks.filter(
+            # using the not-nullable FrameworkAgreement.id as a proxy for testing row null-ness
+            cfa.id.isnot(None) if convert_to_boolean(agreement_returned) else cfa.id.is_(None)
+        )
 
-        if requested_statuses:
-            supplier_frameworks = [
-                sf for sf in supplier_frameworks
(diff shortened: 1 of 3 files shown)


app/main/views/frameworks.py
```diff
@@ -212,32 +212,40 @@ def get_framework_suppliers(framework_slug):
         Framework.slug == framework_slug
     ).first_or_404()
 
-    # TODO: Work out a way to order by current_framework_agreement.signed_agreement_returned_at
-    # Listing agreements is something done for Admin only (suppliers only retrieve their individual agreements)
-    # CCS always want to work from the oldest returned date to newest, so order by ascending date
+    # we're going to need an alias for the current_framework_agreement join
```
galund: Is it obvious enough why we need to alias? (It isn't to me, but that's cos I know nothing at this point :) )

risicle: - [ ] I'll improve the language.

app/main/views/frameworks.py
```diff
@@ -212,32 +212,40 @@ def get_framework_suppliers(framework_slug):
...
+            SupplierFramework.current_framework_agreement, alias=cfa
+        )
+    )
 
-    # TODO: Work out a way to query.filter() on current_framework_agreement rather than manipulating with Python
-    supplier_frameworks = SupplierFramework.query.filter(
+    supplier_frameworks = supplier_frameworks.filter(
         SupplierFramework.framework_id == framework.id
+    ).order_by(
+        cfa.most_recent_signature_time.asc().nullsfirst(),
```
TheDoubleK: Hmm, there's a subtle distinction here between ordering on the `most_recent_signature_time` (which could be time of signing or of countersigning), and the `signed_agreement_returned_at` time of the `cfa`.  Doing it by `most_recent_signature_time` as we have here will move a supplierFramework to the bottom of the list as soon as it's countersigned, whereas doing it by `signed_agreement_returned_at` will keep the same ordering as CCS get on with the countersigning.

I'm not sure which makes more sense.  Maybe it's fine as it is?  What do you think?

tests/app/views/test_frameworks.py
```diff
@@ -538,186 +539,211 @@ class TestGetFrameworkSuppliers(BaseApplicationTest):
...
             assert response.status_code == 200
             data = json.loads(response.get_data())
-            assert len(data['supplierFrameworks']) == 1
+            assert tuple(sf["supplierId"] for sf in data["supplierFrameworks"]) == (0,)
             # Ten G8 suppliers
             response = self.client.get('/frameworks/g-cloud-8/suppliers')
             assert response.status_code == 200
             data = json.loads(response.get_data())
-            assert len(data['supplierFrameworks']) == 10
+            assert tuple(sf["supplierId"] for sf in data["supplierFrameworks"]) == (0, 1, 2, 10, 5, 9, 4, 8, 3, 7, 6)
```
TheDoubleK: This feels like we could do with one comment somewhere here in the tests that describes what the expected order is, e.g. something like  "supplierFrameworks are returned in order of ID if they don't have a framework agreement returned, and from oldest to newest returned if they do".  Just to make it clear what the tests are checking.

app/main/views/frameworks.py
```diff
@@ -212,32 +212,40 @@ def get_framework_suppliers(framework_slug):
         Framework.slug == framework_slug
     ).first_or_404()
 
-    # TODO: Work out a way to order by current_framework_agreement.signed_agreement_returned_at
-    # Listing agreements is something done for Admin only (suppliers only retrieve their individual agreements)
-    # CCS always want to work from the oldest returned date to newest, so order by ascending date
```
risicle: - [ ] Yup, re-add.

app/main/views/frameworks.py
risicle: That's a point. I had just considered it a spec to follow and didn't question it. Come to think of it if ordered by `most_recent_signature_time` it will also screw around with the "next" order...

app/main/views/frameworks.py
```diff
@@ -212,32 +212,40 @@ def get_framework_suppliers(framework_slug):
...
-            else:
-                supplier_frameworks = [sf for sf in supplier_frameworks if sf.current_framework_agreement is None]
-
-        if requested_statuses:
-            supplier_frameworks = [
-                sf for sf in supplier_frameworks
-                if sf.current_framework_agreement and sf.current_framework_agreement.status in requested_statuses
-                ]
+    #
+    # bit of an oddly designed interface here in that the two filterable parameters aren't
```
risicle: - [ ] Meh ok, I just stick them in to make it clear I'm not commenting any particular line. Can also re-word.

app/main/views/frameworks.py
galund: Maybe I can ask you to explain this one in person tomorrow - there is much to SQLAlchemy to learn, and I might as well start with the hard bits :-)

(Where I'd got to in my rtfm'ing... )

- http://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html#using-contains-eager-to-load-a-custom-filtered-collection-result doesn't seem to mention needing the aliases thing
- I don't really get why you need contains_eager for this anyway, if we're just loading a single related entity then can't we use joinedload?

risicle: We don't actually need to tell it to `joinedload` - specifying the relationship on the model with `lazy="joined"` will make it do that by default. What we _need_ to do is reuse this join for a filter and/or an ordering. To do that we need to grab a handle on it to be able to refer to it in `filter` and `order_by` clauses.

The _effect_ of this collection of lines is to bind the python-alias `cfa` to that existing join.

What `contains_eager` _literally_ does is say - "this query here - this already _has_ an equivalent join to what `.current_framework_agreement` asks for, and it's in the join we've 'aliased' to `cfa` (using the previous `.outerjoin` call), so.. please reuse it instead of creating yet another join to satisfy what `.current_framework_agreement` wants".

app/main/views/frameworks.py
risicle: - [ ] Yup

TheDoubleK:
Looks god - :+1: after a rebase

PR #488, opened 2016-11-16
Build error messages for dynamic_list questions

pcraig3:
Errors for dynamic_list questions need to be linked to the field
that is failing the validation: either the primary question or a
follow-up question (depending on the value of the primary question).
eg if we answer 'yes' to the question about having the essential
requirement, we then have to have a value for 'evidence'.

To achieve this, we're putting one extra key into the `errors_map`
dictionary indicating which field failed the validation.

 --

This pull request adds capacity without changing existing behaviour, but notably, it doesn't have any tests written for it. :sweat_smile: :sweat_smile: :sweat_smile:

Should be fine.

### DIFF
--- a/app/validation.py
+++ b/app/validation.py
@@ -156,15 +156,40 @@ def get_validation_errors(validator_name, json_data,
     errors = validator.iter_errors(json_data)
     form_errors = []
     for error in errors:
-        if error.path:
+        # validate follow-up questions are answered: eg evidence given for a 'yes' nice-to-have
+        if error.validator == 'oneOf':
+            key, index = error.path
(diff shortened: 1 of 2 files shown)


TheDoubleK:
Are tests out of the question? Feels like they might be useful.

allait:
@TheDoubleK I've added some tests. 

Since there's no schemas with dynamic list or followup questions yet we need to use a new approach for testing oneOf validator errors. Instead of running get_validation_error, we split the translation logic into a separate function and test that using manually made ValidationError instances.

allait:
👍

PR #489, opened 2016-11-17
SupplierFramework.serialize: add with_users kwarg

risicle:
This should speed things up considerably when called with `with_users=False`.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -250,8 +250,9 @@ def get_framework_suppliers(framework_slug):
             cfa.status.in_(status.split(","))
         )
 
+    with_users = convert_to_boolean(request.args.get("with_users", default="false"))
     return jsonify(supplierFrameworks=[
-        supplier_framework.serialize() for supplier_framework in supplier_frameworks
+        supplier_framework.serialize(with_users=with_users) for supplier_framework in supplier_frameworks
     ])
 
 
--- a/app/models.py
+++ b/app/models.py
@@ -388,8 +388,8 @@ class SupplierFramework(db.Model):
         }
 
     @staticmethod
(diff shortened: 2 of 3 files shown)


TheDoubleK:
Looks good - let's do it :+1:

app/models.py
```diff
@@ -403,8 +403,11 @@ def serialize_agreed_variation(agreed_variation):
             "agreedUserEmail": user.email_address,
         })
 
-    def serialize(self, data=None):
-        agreed_variations = {k: self.serialize_agreed_variation(v) for k, v in iteritems(self.agreed_variations or {})}
+    def serialize(self, data=None, with_users=True):
```
allait: This should be `with_users=False` by default

app/main/views/frameworks.py
```diff
@@ -250,8 +250,9 @@ def get_framework_suppliers(framework_slug):
             cfa.status.in_(status.split(","))
         )
 
+    with_users = convert_to_boolean(request.args.get("with_users", default="false"))
```
allait: I don't think this flag should be here if it's not used by any of the frontend apps. This listing is very expensive, so if it's not actually used then there's no reason to have it here.

app/models.py
```diff
@@ -388,8 +388,8 @@ def get_service_counts(supplier_id):
         }
 
     @staticmethod
-    def serialize_agreed_variation(agreed_variation):
-        if not agreed_variation.get("agreedUserId"):
+    def serialize_agreed_variation(agreed_variation, with_users=True):
```
allait: `with_users=False` by default. The reason we add flags for these things is that these operations are expensive for a list of objects, so they must require and explicit opt-in. Otherwise it's easy to trigger them accidentally.

PR #490, opened 2016-11-17
Add framework status to API /_status

allait:
To figure out current framework statuses for the given environment you either need access to the API token or you'd have to look through a number of frontend pages to infer the status from.

Framework status is a part of almost every request to the API, so it should always be available for a working API instance and it makes sense to add it to the /_status page.

Adding it to the /_status page creates an easier way to get the list of all framework statuses.


TheDoubleK:
Nice.  I wondered for a second if whether or not clarification questions is open should also be included, but it probably isn't of much value.
:+1:

PR #491, opened 2016-11-17
Allow filtering users by any existing user role

allait:
We often need to check for admin accounts on different environments, so having an easy way to filter users in the API seems useful.


TheDoubleK:
I can't remember why we restricted this in the first place, but makes sense to me :+1:

PR #492, opened 2016-11-17
SupplierFramework.serialize: set with_users kwarg default to False

risicle:
This should placate @allait 

All views except `get_supplier_frameworks_info` should now return no extended user details.


allait:
Yay! 🎉

One little note: `serialize_agreed_variation` still defaults to `with_users=True` and that method appears to be called from one of the agreement views. Could you check if that's used and flip this flag too?

risicle:
Good catch - the only thing that hits that endpoint doesn't even take heed of the return value. Will change default.

allait:
👍

PR #494, opened 2016-11-25
get_framework_suppliers tests: merge all of these tests into "subtest"s called from a single test for improved execution speed

risicle:
Runtime on my laptop is now ~32s, still no miracle but a lot better than before. Should fix these properly some day.


allait:
👍

PR #495, opened 2016-12-05
API to support both legacy and new data schemas for supplier evidence

idavidmcdonald:
When applying for a brief using the current (legacy) flow and for the upcoming new flow (multipage), we will be storing different data based on the flow. Therefore our API when it receives data to create, update or submit a brief response, must validate the input data against the correct schema.

We have the following three cases for the `FEATURE_FLAGS_NEW_SUPPLIER_FLOW` feature flag:
- creating/updating/submitting a brief response when the feature flag is off
- creating/updating/submitting a brief response for a brief that was created before the feature flag
- creating/updating/submitting a brief response for a brief that was created after the feature flag

For the first two cases, we should be validating our data against the legacy brief response schema. For the later case we should be validating against the new brief response schema.

### Pivotal stories
- https://www.pivotaltracker.com/story/show/129839965
- https://www.pivotaltracker.com/story/show/129843611
- https://www.pivotaltracker.com/story/show/131617615

### Notes
- Not all pages have been completed as per [the prototype ](https://dm-prototype.cloudapps.digital/structure/opportunities/dos-opportunities). There will be more work to come which will do things like adding further new fields to JSON schemas. For the moment, we have only introduced the `essentialRequirementsMet` question for the new flow instead of `essentialRequirements`.
- After we complete our two transition period from the old flow to the new flow, then we will no longer require functionality to create/update/submit a brief response that has data matching the legacy schema. At this point we will be able to remove this functionality from the API.
- The tests are a bit gnarly as we have tried to keep the intention of the original tests but now they cover all three feature flag cases mentioned above. You may find it easier to not use the git diff as it's hard to read and compare. I suggest opening the old and new version side by side in new tabs.


tests/app/views/test_brief_response.py
```diff
@@ -607,6 +738,103 @@ def test_can_not_submit_an_invalid_brief_response(
...
+            'availability': u'a',
+            'respondToEmailAddress': 'supplier@email.com'
+        }
+
+        create_res = self.create_brief_response(data=non_legacy_brief_response_data)
+        assert create_res.status_code == 201
+
+        brief_response_id = json.loads(create_res.get_data(as_text=True))['briefResponses']['id']
+
+        self.app.config["FEATURE_FLAGS_NEW_SUPPLIER_FLOW"] = self.feature_flag_date
```
tombye: Could we mark this with a comment? I know you mentioned it above (a duplicate of the last sentence from that would do) but it would really help the reading to have that info next to the code.

tests/app/views/test_brief_response.py
```diff
@@ -607,6 +738,103 @@ def test_can_not_submit_an_invalid_brief_response(
...
+        assert data['error']['essentialRequirements'] == "answer_required"
+
+
+class TestSubmitBriefReponseWhenFeatureFlagIsOff(TestSubmitBriefReponseForBriefCreatedBeforeFeatureFlag):
+
+    feature_flag_date = False
+
+    def setup(self):
+        super(TestSubmitBriefReponseWhenFeatureFlagIsOff, self).setup()
+        self.app.config["FEATURE_FLAGS_NEW_SUPPLIER_FLOW"] = self.feature_flag_date
```
tombye: Can we have a comment here explaining what the flag is being set to?

I'd quite like it for all the other times we set it tbh, so it's clear what's the flag is being set to.

An alternative is having two attributes on one of the base classes, something like `date_one_week_ago`, `date_one_week_in_the_future`. Would mean we could write things like:

`self.app.config["FEATURE_FLAGS_NEW_SUPPLIER_FLOW"] = self.date_one_week_in_the_future`

tests/app/views/test_brief_response.py
```diff
@@ -550,12 +683,8 @@ def test_can_not_submit_a_brief_response_that_already_been_submitted(self, live_
...
-    def test_can_not_submit_a_brief_response_for_a_framework_that_is_not_live(
-        self, live_dos_framework, brief_response_data
-    ):
-        self._setup_existing_brief_response(brief_response_data)
-
+    def test_can_not_submit_a_brief_response_for_a_framework_that_is_not_live(self, live_dos_framework):
+        self._setup_existing_brief_response()
         with self.app.app_context():
             # Change live framework to be expired
             db.session.execute("UPDATE frameworks SET status='expired' WHERE slug='digital-outcomes-and-specialists'")
```
idavidmcdonald: It needs to be live for `_setup_existing_brief_response` to work

tests/app/views/test_brief_response.py
```diff
@@ -449,37 +525,20 @@ def test_can_not_update_brief_response_if_supplier_is_ineligible_for_brief(self,
...
-            res = self._update_brief_response(brief_response_id, {'respondToEmailAddress': 'newemail@email.com'})
-            assert res.status_code == 400
-            data = json.loads(res.get_data(as_text=True))
-            assert data == {'error': 'Brief response must be a draft'}
-
-    def test_can_not_update_brief_response_with_invalid_content(self, live_dos_framework):
-        res = self._update_brief_response(
-            self.brief_response_id, {'essentialRequirements': [False, False, False, False]}
+        # Create dummy brief_response which has been submitted
+        brief_response_id = self.setup_dummy_brief_response(
```
idavidmcdonald: You can override this behaviour by calling `setup_dummy_brief_response(submitted_at=None)` so not sure if it makes sense to rename it (I'm 50 50 on that). 

We could maybe rename it to something better like `setup_brief_response_fixture` though? Although maybe the word "dummy" is useful as it implies you are using example brief response data rather than passing in brief response data?

Words are hard...

tests/app/views/test_brief_response.py
```diff
@@ -449,37 +525,20 @@ def test_can_not_update_brief_response_if_supplier_is_ineligible_for_brief(self,
...
         )
+
+        # Update brief response
+        res = self._update_brief_response(brief_response_id, {'respondToEmailAddress': 'newemail@email.com'})
         assert res.status_code == 400
         data = json.loads(res.get_data(as_text=True))
-        assert data == {'error': {'essentialRequirements': 'answer_required'}}
+        assert data == {'error': 'Brief response must be a draft'}
 
     def test_update_brief_response_with_missing_response_to_page_question_will_error(self, live_dos_framework):
```
idavidmcdonald: To be clear, are you suggesting that this test name should be `test_update_brief_response_with_missing_answer_to_page_question_will_error`?

tombye: Yes, assuming that still describes what the test does (which is how I read it).

tests/app/views/test_brief_response.py
tombye: Looking at it like that, I'm not sure this needs fixing. There's a precedent with services of us assuming it's published unless it's explicitly called a draft.

Words are hard and I'm not convinced it's worth the effort here. Consider it a mute point.

tests/app/views/test_brief_response.py
idavidmcdonald: Have implemented

tests/app/views/test_brief_response.py
idavidmcdonald: Have added comments to every setting of the config feature flag. I've also moved the actual dates themselves into the base class to keep the code more DRY.

tombye:
Thanks for making the changes and sorry to be pedantic but it doesn't look like the `feature_flag_date` variable is needed any more, could we lose it?

idavidmcdonald:
Great spot Tom. Most situations it can be removed and I have done so. Have rebased this change into the last commit.

tombye:
Looks good to me now 👍

PR #496, opened 2016-12-09
Move preview new supplier flow feature flag well into the future

idavidmcdonald:
We do this so functional tests that rely on current behaviour i.e. the legacy brief response flow, won't break on preview. If the feature flag is on then the legacy flow will break. By setting the flag well into the future, we can still use it for testing by individually manipulating briefs and setting their published time after the feature flag.


allait:
👍

PR #498, opened 2016-12-13
Make bootstrap.sh idempotent

danielknell:
currently overwrites the virtualenv if none is active which breaks pip

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -8,7 +8,7 @@
 set -o pipefail
 
 if [ ! $VIRTUAL_ENV ]; then
-  virtualenv ./venv
+  make virtualenv
   . ./venv/bin/activate
 fi
 


allait:
This is a bit mutually recursive at this point. `make bootstrap` is calling `bootstrap.sh` which is then calling `make virtualenv` ... I think we should get rid of the script and move the steps to the Makefile.

danielknell:
the only bit not really makefile-esque is the database creation, might work ok with a few `createdb || true` lines to stop duplicate dbs breaking things, that or checking first which is likely to get ugly.

danielknell:
@allait any steer on what would be preferable? 

`[ .... long check code ... ] && createdb ...` 

check the db exists before creating it which is more proper but will be horrible readability wise iirc

or

`createdb ... || true` 

short and sweet, will prevent the db existing causing make to bail, but will result in things not bailing if the db fails to create in the first place.

allait:
@danielknell I'm leaning towards the first one to get a slightly more helpful error message. But either one is fine in case you have a strong preference.

PR #499, opened 2016-12-15
Sqlalchemy upgrade

galund:
(The change to the `test_stats` unit test was purely to be able to see which bit was breaking, there should be no functional difference.)

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -278,7 +278,7 @@ class Supplier(db.Model):
 
     companies_house_number = db.Column(db.String, index=False, unique=False, nullable=True)
 
-    clients = db.Column(JSON, default=list)
+    clients = db.Column(JSON(none_as_null=True), default=list)
 
     # Drop this method once the supplier front end is using SupplierFramework counts
     def get_service_counts(self):
--- a/app/models.py
+++ b/app/models.py
@@ -1317,7 +1317,7 @@ class BriefResponse(db.Model):
     __tablename__ = 'brief_responses'
 
     id = db.Column(db.Integer, primary_key=True)
-    data = db.Column(JSON, nullable=False)
+    data = db.Column(JSON(none_as_null=True), nullable=False)
 
     brief_id = db.Column(db.Integer, db.ForeignKey('briefs.id'), nullable=False)
     supplier_id = db.Column(db.Integer, db.ForeignKey('suppliers.supplier_id'), nullable=False)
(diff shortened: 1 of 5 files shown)


tests/app/views/test_frameworks.py
```diff
@@ -453,54 +453,61 @@ def test_stats(self):
...
-            ],
-            u'supplier_users': [
-                {u'count': 4, u'recent_login': False},
-                {u'count': 2, u'recent_login': None},
-                {u'count': 5, u'recent_login': True},
-            ]
-        })
+        json_response = json.loads(response.get_data())
+
+        services = json_response[u'services']
```
allait: What's the difference between this and the previous approach? This looks less clear to me

galund: When the test failed, all I got was a JSON blob marked as different, and it was hard to tell which bit had changed without copying and pasting somewhere to do a diff, and even then diffing JSON is a bit dubious... so checking each separate bit, gives a much clearer indication of which bit of code we are test / is broken. In the successful case it's no different, in the failure case you get easier debugging / more detail.

allait: Don't you get a pytest diff though? Which shows you which list elements/keys are different.
I think apart form being a much more error prone way to write the assertion this also hides the whole error. An error in the query can affect multiple counts and it's not obvious to me that it's more clear what happened from a single item assertion.

app/models.py
```diff
@@ -722,7 +733,7 @@ class ServiceTableMixin(object):
     # Service publishing time.
     service_id = db.Column(db.String, index=True, unique=True, nullable=False)
 
-    data = db.Column(JSON)
+    data = db.Column(JSON, nullable=False, default=dict)
```
allait: Why the default? It feels to me that the service should never be an empty dictionary, so this is just hiding a potential error.

galund: This just felt like the less risky change (not knowing all the possible ways to create that row). Happy to change if you're confident that an empty dict is an error just as much as NULL would be.

allait: For services I'm pretty sure they are, since all endpoints should try to assign something to `data`. Tests I'm not sure about, so if there are a lot failing without this then maybe it's not a big deal to keep the default. Up to you.

allait:
Looks good 👍

(But let's make sure API is fully deployed before we merge this in case we need to hold this in preview for a bit)

app/models.py
galund: Cool, have removed the default as it doesn't seem to affect the tests.

galund:
Let me know when seems like a good moment to merge, probably when we're both in the office I guess :) cheers

PR #500, opened 2016-12-15
Sqlalchemy upgrade json null migration

galund:
(To land after the SQLAlchemy 1.1 upgrade has happened and is working.)


allait:
Migration looks good 👍

risicle:
@galund do you want to land this? I'll target a migration I'm working on to be rev id 800 anyway assuming you'll get to it before me...

galund:
AIUI this shouldn't land until after the sql-alchemy migration has actually gone live (and been considered stable enough not have to back it out). So I think assume it will not land on your schedule, cheers!

risicle:
Mine actually doesn't have a tremendously urgent schedule and it itself will have to bubble through the pipeline so it may not make it in till after this anyway... it just means i've got to keep correcting the auto migration builder which keeps insisting on including the latent json migrations.

PR #502, opened 2016-12-16
Contact information validation

galund:
Allow dropping of 'links' JSON fields in one (recursive) step 
 - by making this generic, we can ensure it works when the data
   is not as expected, e.g. in [#134425439] where we need to
   iterate over a list (of contact information datasets), that
   may (given invalid data) not actually be a list;
 - the actual validation (i.e. checking that the contact information
   is a list) can now be done by the schema-validator. This is
   much tidier than having to add additional schema-like checks
   up-front to make sure we don't crash when stripping 'foreign
   fields' such as 'links'
 - also test cases

### DIFF
--- a/app/supplier_utils.py
+++ b/app/supplier_utils.py
@@ -9,11 +9,7 @@ def validate_and_return_supplier_request(supplier_id=None):
     json_has_required_keys(json_payload['suppliers'], ['contactInformation'])
 
     # remove unnecessary fields
-    json_payload['suppliers'] = drop_foreign_fields(json_payload['suppliers'], ['links'])
-    json_payload['suppliers']['contactInformation'] = [
-        drop_foreign_fields(contact_data, ['links'])
-        for contact_data in json_payload['suppliers']['contactInformation']
-    ]
+    json_payload['suppliers'] = drop_foreign_fields(json_payload['suppliers'], ['links'], recurse=True)
 
     if supplier_id:
         validate_supplier_json_or_400(json_payload['suppliers'])
--- a/app/utils.py
+++ b/app/utils.py
@@ -92,12 +92,31 @@ def json_only_has_required_keys(data, keys):
         abort(400, "Invalid JSON must only have {} keys".format(keys))
 
 
-def drop_foreign_fields(json_object, list_of_keys):
-    json_object = json_object.copy()
-    for key in list_of_keys:
-        json_object.pop(key, None)
-
+def drop_foreign_fields(json_object, list_of_keys, recurse=False):
+    json_object = keyfilter_json(json_object, lambda k: k not in list_of_keys, recurse)
     return json_object
+
+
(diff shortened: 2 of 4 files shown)


app/utils.py
```diff
@@ -92,14 +92,33 @@ def json_only_has_required_keys(data, keys):
         abort(400, "Invalid JSON must only have {} keys".format(keys))
 
 
-def drop_foreign_fields(json_object, list_of_keys):
-    json_object = json_object.copy()
-    for key in list_of_keys:
-        json_object.pop(key, None)
-
+def drop_foreign_fields(json_object, list_of_keys, recurse=False):
```
allait: I'm not sure if there's a need for the nested drop_foreign_fields.

For one, we don't have many nested questions we accept in writes. Contact information might be the only one, and it should've been a separate endpoint to create  contact information objects for a supplier to begin with. The only reason it's nested was to save time when were rolling this out.

And the way `drop_foreign_fields` is used everywhere else doesn't really allow it to be safely applied to nested objects. `list_of_keys` is a collection of keys we don't want to store for the root dictionary, but it's not necessarily the same for the nested objects. I believe the way we're approaching this now is calling this function from the model method, which allows us to list the keys we want removed once for each model.

app/supplier_utils.py
```diff
@@ -9,11 +9,7 @@ def validate_and_return_supplier_request(supplier_id=None):
     json_has_required_keys(json_payload['suppliers'], ['contactInformation'])
 
     # remove unnecessary fields
-    json_payload['suppliers'] = drop_foreign_fields(json_payload['suppliers'], ['links'])
-    json_payload['suppliers']['contactInformation'] = [
-        drop_foreign_fields(contact_data, ['links'])
-        for contact_data in json_payload['suppliers']['contactInformation']
```
allait: So would `json_payload['suppliers'].get('contactInformation', [])` fix the issue?

galund: I don't think so, because the crash is when 'contactInformation' is there, but is not a list (or, worse, it's some kind of iterable that doesn't contain a dictionary)

app/utils.py
galund: I was careful to make sure that the default was not to recurse, which should be the same behaviour as before. (And we do have test coverage there.)

I agree with 'no necessarily the same', but that's why the recursion should be optional, and ISTM that you need to know the structure in order to know whether and how to call this, regardless of how you implement the stripping at each level. In other words, whether the recursion is implemented manually (as currently) or automatically (per this PR), you have to know what the structure is in order to know whether stripping a given set of keys is a valid thing to do.

All that said, I think that the stripping-before-validation thing is a bit weird: ISTM it would be better to change the schema, to allow for optional links, and make this whole 'strip-out-the-links' thing just go away? Otherwise, we basically have two schema: one properly documented (but without links), and one that we actually use (that has links in it).

allait:
👍

PR #504, opened 2016-12-23
NO-TICKET_TestAuditEvents#test_should_get_audit_event

benvand:
`digitalmarketplace-api/migrations/versions/460_add_audit_events_for_g4_framework.py` introduces an audit event into the database. `test_should_get_audit_event` Does the same and then asserts a single audit event exists.

The reason this _was_ working is that several other tests, in assorted `teardown` methods, drop and create tables. The table _was_ being blanked at some point in some prior another test's teardown.

I've removed this tests dependency on any other test.

To reproduce:
`py.test  tests/app/views/test_audits.py::TestAuditEvents::test_should_get_audit_event`

Contents:
* Order imports per https://www.python.org/dev/peps/pep-0008/#imports
* Create `add_audit_event` method which returns `id`
* Assert full expected data in `test_should_get_audit_event`
* refactor to use new function

### DIFF
--- a/tests/app/views/test_audits.py
+++ b/tests/app/views/test_audits.py
@@ -47,9 +47,10 @@ class TestAuditEvents(BaseApplicationTest):
             db.session.commit()
 
     def test_only_one_audit_event_created(self):
-        count = AuditEvent.query.count()
-        self.add_audit_event()
-        assert AuditEvent.query.count() == count + 1
+        with self.app.app_context():
+            count = AuditEvent.query.count()
+            self.add_audit_event()
+            assert AuditEvent.query.count() == count + 1
 
     def test_should_get_audit_event(self):
         aid = self.add_audit_event(0)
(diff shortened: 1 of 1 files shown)


tests/app/views/test_audits.py
```diff
@@ -37,14 +47,20 @@ def add_audit_events_with_db_object(self):
...
+        expected = {
+            'links': {'self': mock.ANY},
+            'type': 'supplier_update',
+            'acknowledged': False,
+            'user': '0',
+            'data': {'request': 'data'},
+            'id': aid,
+            'createdAt': mock.ANY
+        }
+        assert expected in data['auditEvents']
```
pcraig3: Should this also be asserting that only one exists?

benvand: I don't see that as being this tests job. The aim of this test is to assert that a newly created audit event is returned in the get request to the `/audit-events` end-point.
We shouldn't be assuming that the database is completely clean. We can but it'll be a quite radical overhaul of the tests. And even if we did want to assume and assert that we have a completely clean db at the start of each test then it would be better to do it somewhere else. 

I've just added a test that asserts that only one audit event is created when we're creating one. Don't want to be audit the creation of our audit events :)

pcraig3: Sure, makes sense.

pcraig3:
👍

PR #507, opened 2017-01-03
Validate dynamic list `essentialRequirements`

idavidmcdonald:
#### Work done
Pivotal story - https://www.pivotaltracker.com/story/show/129841565

- Update JSON schemas for new brief response flow including the `essentialRequirements` dynamic list question
- Validate `essentialRequirements` question, see test for an example of how the validation works. Note, if you do not answer the evidence question on the front end then we send an empty dict to the API which is covered in our test.
- Validate correct number of essential requirements in brief response as per [this story](https://www.pivotaltracker.com/story/show/136045527)

#### Further thought
- Not sure on if we are happy with validation messages for when booleans are passed instead of objects. This is not expected input from the front end but someone could post this directly to the API.

```
"essentialRequirements": [
        {"evidence": "valid evidence"},
        True
],
```
gives validation error:
```
{u'essentialRequirements': u"True is not of type u'object'"}
```

Is it worth the effort of trying to make it 
```
{
    'essentialRequirements': [
        {
            'error': u"True is not of type u'object'",
            'index': 1
        },
    ]
}
```

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1370,7 +1370,6 @@ class BriefResponse(db.Model):
         )
 
         if (
-            legacy and
             (required_fields and 'essentialRequirements' in required_fields or enforce_required) and
             'essentialRequirements' not in errs and
             len(self.data.get('essentialRequirements', [])) != len(self.brief.data['essentialRequirements'])
--- a/app/utils.py
+++ b/app/utils.py
@@ -136,17 +136,15 @@ def display_list(l):
 
 
 def strip_whitespace_from_data(data):
-    for key, value in data.items():
-        if isinstance(value, list):
-            # Strip whitespace and remove empty items from lists
-            data[key] = list(
-                filter(lambda x: x != '',
-                       map(lambda x: x.strip() if isinstance(x, string_types) else x, value))
-            )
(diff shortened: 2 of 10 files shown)


app/models.py
```diff
@@ -1357,7 +1357,6 @@ def validate(self, enforce_required=True, required_fields=None, max_day_rate=Non
         )
 
         if (
-            legacy and
```
allait: Ah, makes sense.

app/utils.py
```diff
@@ -136,17 +136,15 @@ def display_list(l):
...
-            # Strip whitespace from strings
-            data[key] = value.strip()
-    return data
+    """Recursively strips whitespace and removes empty items from lists"""
+    if isinstance(data, string_types):
+        return data.strip()
+    elif isinstance(data, dict):
+        return {key: strip_whitespace_from_data(value) for key, value in data.items()}
+    elif isinstance(data, list):
+        return list(filter(lambda x: x != '', map(lambda x: strip_whitespace_from_data(x), data)))
```
idavidmcdonald: Think you are right, have updated.

PR #508, opened 2017-01-03
135223377-application_status

benvand:
This adds a method on Framework to get all supplier id's who have managed to successfully complete inputting a service.
This info is added to the `export_users_for_framework` view in `views/users.py`.

It also adds a new `py.test` fixture in `conftest.py` which supplies a draft service.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -119,6 +119,19 @@ class Framework(db.Model):
             'lots': [lot.serialize() for lot in self.lots],
         }
 
+    def get_supplier_ids_for_completed_service(self):
+        """Only suppliers whose service has a status of submitted or failed."""
+        unpack_list_of_lists = lambda l: set(item for sublist in l for item in sublist)
+        results = db.session.query(
+            DraftService
+        ).filter(
+            DraftService.status.in_(('submitted', 'failed')),
+            DraftService.framework_id == self.id
+        ).with_entities(
+            DraftService.supplier_id
+        ).distinct()
+        return unpack_list_of_lists(results)
+
     @validates('status')
     def validates_status(self, key, value):
         if value not in self.STATUSES:
--- a/tests/app/test_model.py
+++ b/tests/app/test_model.py
@@ -12,7 +12,9 @@ from app.models import (
     Supplier, SupplierFramework, FrameworkAgreement,
     Brief, BriefResponse,
     ValidationError,
-    BriefClarificationQuestion
+    BriefClarificationQuestion,
+    DraftService,
+    FrameworkLot
 )
 
 from .helpers import BaseApplicationTest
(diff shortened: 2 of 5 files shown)


tests/app/test_model.py
```diff
@@ -894,6 +896,120 @@ def test_lot_data_is_serialized(self):
...
+                    with mock.patch('app.models.url_for', autospec=lambda i, **values: 'test.url/test'):
+                        ds = DraftService.from_service(Service.query.filter(Service.service_id == service_id).first())
+                        ds.status = service_status
+                    db.session.add(ds)
+                    db.session.commit()
+            # Assert that only those suppliers whose service_status list contains failed and/ or submitted are returned.
+            framework = Framework.query.filter(Framework.id == open_example_framework['id']).first()
+            assert sorted(framework.get_supplier_ids_for_completed_service()) == [0, 2, 4]
+
+    def teardown(self):
```
allait: Doesn't the database get cleared after each test? Why do we need additional cleanup here?

benvand: Not entirely.
https://github.com/alphagov/digitalmarketplace-api/blob/master/tests/app/helpers.py#L279-L279

This is why all the py.test fixtures have their own teardown. I want to explain a fix for this on Friday but for this PR I decided to make use of as much existing code as possible.

benvand: Thought of a work around though.

app/models.py
```diff
@@ -5,6 +5,7 @@
 from flask import current_app
 from flask_sqlalchemy import BaseQuery
 from six import string_types, iteritems
+from sqlalchemy import Index
```
benvand: True, wonder if it's possible to make a rule in the pep8 linter for that

app/models.py
```diff
@@ -119,6 +120,19 @@ def serialize(self):
             'lots': [lot.serialize() for lot in self.lots],
         }
 
+    def get_supplier_ids_for_completed_service(self):
+        """Only suppliers whose service has a status of submitted or failed."""
+        unpack_list_of_lists = lambda l: tuple(item for sublist in l for item in sublist)
```
benvand: Done

tests/app/test_model.py
benvand: So the problem is that the db doesn't get wiped completely. As far as I understand it, test objects should be created with ids > 99. They'll then get wiped. It would be better to wipe everything though. Otherwise if you select a framework by slug for instance 
`dos = Framework.query.filter(Framework.slug == 'digital-outcomes-and-specialists').first()` we don't know what state it's going to be in or what's been done to it.

benvand: But I've managed to remove the setup and teardown by using one of the other fixtures.

allait: Some of the tables (like Framework and Lots) are populated by fixtures, and since we only run the migrations once per test suite we can't wipe them after every tests. That's why some of the tables are not wiped and fixtures that are creating rows in those tables need a separate teardown.

Everything else should be deleted though.

allait:
Looks good 👍

PR #509, opened 2017-01-04
135223377-part-2-csv-header-change

benvand:
* Change header
* Update tests

As per comment here from @cathrooney 
https://www.pivotaltracker.com/story/show/135223377/comments/157671415
> As an extra thing, can we change the column header for 'user_email' to be 'email address'. I have to rename it every time to use it in Notify!


allait:
👍

PR #511, opened 2017-01-10
137170913-TESTING-structure-changes

benvand:
#### Change test directory structure to increase clarity

* `tests/app` dir shadowed flask 'builtin' name app. Changed to application to avoid name clashes.
* Moved `tests/app/helpers.py` -> `tests/helpers.py`.
  * These helpers are more visible in the root of the test dir
  * Promote reuse and expansion.
* Split `helpers.py` into `helpers.py` and `bases.py`
  * This code was becoming enormous an multi purpose.
* Moved db setup code from `tests.app.__init__` to `conftest.py`
  * This code is called at the start ove every session. It shouldn't be nested.


allait:
Should we merge `tests/application` into `tests/app`? It fees like it's testing the routes defined in `app.main` anyway.

tests/helpers.py
```diff
@@ -314,43 +243,19 @@ def set_framework_variation(self, slug):
...
-
-        assert_equal(response.status_code, 400)
-        assert_in(b'Invalid JSON',
-                  response.get_data())
-
-    def test_invalid_content_type_causes_failure(self):
-        response = self.open(
-            data='{"services": {"foo": "bar"}}')
+class HelpersMixin(object):
+    def string_to_time_to_string(self, value, from_format, to_format):
```
allait: These methods aren't really using `self` so maybe we should make them functions?

PR #512, opened 2017-01-12
Nice to haves

Wynndow:
For this story on Pivotal [https://www.pivotaltracker.com/story/show/129841633](https://www.pivotaltracker.com/story/show/129841633).

This PR is to update the API to accept the new style nice-to-have requirements questions data structure.

The old style boolean_list (legacy) data structure was a list of booleans with no evidence:

`{"niceToHaveRequirements": [True, False, True, False, True]}`

The new style dymanic_list is a list of dictionaries with one or two keys:

`{"niceToHaveRequirements": [{"yesNo": False}, {"yesNo": True, "evidence": "Awesome evidence"}]}`

This requires new schemas for the three DOS lots and some careful handling of validation. If a boolean was included in the dynamic_list data instead of a dictionary it did odd things. This could only happen if posting to the API directly and won't occur under normal use. We handle this unlikely scenario in this pull request.


tombye:
Looks good to me 👍

PR #517, opened 2017-01-23
[#134795163]get_framework_suppliers: make inclusion of declaration section optional

risicle:
This seems like the most sensible way to deal with the issue in story https://www.pivotaltracker.com/story/show/134795163

:man_shrugging:

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -250,8 +250,13 @@ def get_framework_suppliers(framework_slug):
             cfa.status.in_(status.split(","))
         )
 
+    with_declarations = convert_to_boolean(request.args.get("with_declarations", "true"))
+
     return jsonify(supplierFrameworks=[
-        supplier_framework.serialize(with_users=False) for supplier_framework in supplier_frameworks
+        supplier_framework.serialize(
+            with_users=False,
+            with_declaration=with_declarations,
+        ) for supplier_framework in supplier_frameworks
     ])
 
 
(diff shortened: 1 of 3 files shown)


app/main/views/frameworks.py
```diff
@@ -213,14 +214,14 @@ def get_framework_suppliers(framework_slug):
     ).first_or_404()
 
     # we're going to need an "alias" for the current_framework_agreement join
-    # so that we can refer to it in later clauses
+    # so that we can refer to it in later clausesris
```
TheDoubleK: Rlly?

app/main/views/frameworks.py
```diff
@@ -251,7 +252,10 @@ def get_framework_suppliers(framework_slug):
         )
 
     return jsonify(supplierFrameworks=[
-        supplier_framework.serialize(with_users=False) for supplier_framework in supplier_frameworks
+        supplier_framework.serialize(
+            with_users=False,
+            with_declaration=convert_to_boolean(request.args.get("with_declarations", "true")),
```
TheDoubleK: Would it be worth pulling the `convert_to_boolean` out before this so it doesn't get called for every supplier_framework?

app/main/views/frameworks.py
risicle: Shit! I explicitly meant to yank that out...

app/main/views/frameworks.py
risicle: good idea

PR #521, opened 2017-02-01
Remove scripts to import services/suppliers, and stop using the import routes in our unit tests

pcraig3:
This pull request is a bit sprawling, but I've tried to keep the commit history fairly clean (at least, after the first couple of commits).

This pull request removes the scripts mentioned [in the dm-manual](https://github.gds/gds/digitalmarketplace-manual/blob/dce04fba1dc968d3a817179e14b2d9174c0a49f8/developer-setup.rst#importing-stuff) (note: this is being rewritten), and removes all instances of using our import route in our unit tests.  It turns out that creating a service/supplier with a hardcoded id is really convenient for testing, so our import routes were being used in lots of test setups, and are still used extensively in the `jenkins` branch of our functional tests.  On account of the functional tests, I've left the import routes intact as well as the tests that specifically test them, but otherwise I've minimised the rest of our code's reliance on them.  Additionally, [the `apiclient` no longer has helper methods to call these routes](https://github.com/alphagov/digitalmarketplace-apiclient/pull/62), so that should stop anyone from trying to use them in the future.

Here's what's in this pull request:

- Removed `import_services_from_api.py` and `import_suppliers_from_api.py` scripts
- Refactored all setup methods that were relying on `import_supplier` or `import_service`
  - expanded the `create_dummy_service` and `create_dummy_suppliers` method to better cope with our test setup
- Ran the [nose2pytest](https://github.com/pytest-dev/nose2pytest) script in test files that I was modifying. (Replaces all `assert_*` nose functions with py.test's `assert`s.)
- Tweaked an error message returned when trying to create a supplier user with no `supplierId`
- Added a few comments about our 'editing services by creating new drafts' behaviour that we don't use


tests/main/views/test_users.py
```diff
@@ -336,61 +322,48 @@ def test_can_not_post_an_admin_ccs_user(self):
...
-
-    def test_can_post_a_supplier_user(self):
-        with self.app.app_context():
-            db.session.add(
-                Supplier(supplier_id=1, name=u"Supplier 1")
-            )
-            db.session.commit()
+        assert "'admin-ccs' is not one of" in error
 
+    def test_can_post_a_supplier_user(self, supplier_basic):
```
TheDoubleK: Where does `supplier_basic` come from? It looks like it's just an ID?  In which case we could call it that?

tests/main/views/test_users.py
```diff
@@ -632,20 +605,19 @@ def test_new_password_is_not_audited(self):
...
-            assert_equal(data['auditEvents'][0]['type'], 'update_user')
-            assert_equal(
-                data['auditEvents'][0]['data']['update']['password'],
-                'updated'
-            )
+            assert len(data['auditEvents']) == 1
+            assert data['auditEvents'][0]['type'] == 'update_user'
+            assert (
+                data['auditEvents'][0]['data']['update']['password'] ==
+                'updated')
```
TheDoubleK: I might drop the closing bracket here to the next line.

TheDoubleK:
One thing I noticed is that in quite a few places the new `assert`s have the value as the first thing, which reads a bit weird, e.g. 
```
assert 200 == response.status_code
assert 3 == len(data['suppliers'])
assert 0 == len(data['suppliers']['clients'])
assert 404 == response.status_code
assert 'Active' == data['suppliers'][0]['name']
```

I prefer the more usual pattern of having the thing we want to test first, and the value as the second argument to the `==`, e.g. 
```
assert response.status_code == 201
assert len(data['suppliers']) == 2
```

app/main/views/drafts.py
```diff
@@ -230,6 +230,11 @@ def publish_draft_service(draft_id):
         abort(400, "Cannot re-publish a submitted service")
 
     if draft.service_id:
+        """
+        publishing the draft with a "service_id" updates the original service -- it doesn't create a new one
```
pcraig3: that's what u think kev

tests/main/views/test_drafts.py
```diff
@@ -2,24 +2,21 @@
...
+class TestDraftServices(BaseApplicationTest, FixtureMixin):
     service_id = None
     updater_json = None
     create_draft_json = None
 
     def setup(self):
         super(TestDraftServices, self).setup()
 
-        payload = load_example_listing("G6-IaaS")
+        payload = load_example_listing("G6-SaaS")
```
pcraig3: I changed the lot to see what would break, to be honest. And not much does. 
So I figured `SaaS` was more normal, somehow.

There's a bunch of tests that create `Service` objects directly, and when they do they usually do something like

tests/main/views/test_drafts.py
```diff
@@ -48,14 +45,12 @@ def setup(self):
...
-            '/services/%s' % self.service_id,
-            data=json.dumps(
-                {'updated_by': 'joeblogs',
-                 'services': payload}),
-            content_type='application/json')
+            self.setup_dummy_service(
+                service_id=self.service_id,
+                **payload
+            )
+            db.session.commit()
```
pcraig3: You're right, Kev. It looks like most of them do do the commit at the end of the method (in [FixtureMixin](https://github.com/alphagov/digitalmarketplace-api/blob/a1f9f4dce3d93d53f11bcce575329ff8e635d525/tests/helpers.py#L43)), so maybe it's worth adding it.

tests/main/views/test_services.py
```diff
@@ -1506,59 +1280,57 @@ def test_add_a_new_service_with_status_disabled(self):
...
-            assert_almost_equal(now, service.created_at,
-                                delta=timedelta(seconds=2))
+                assert service.data[key] == payload[key]
+            assert abs(service.created_at - service.updated_at) <= timedelta(seconds=0.5)
+            assert abs(now - service.created_at) <= timedelta(seconds=2)
 
     def test_when_service_payload_has_mismatched_id(self):
         response = self.client.put(
-            '/services/1234567890123456',
+            '/services/{}1'.format(self.service_id),
```
pcraig3: Sure, I'll make it more obvious it's random.

tests/main/views/test_suppliers.py
```diff
@@ -13,68 +12,70 @@
...
 
 class TestGetSupplier(BaseApplicationTest, FixtureMixin):
+    supplier = supplier_id = None
+
     def setup(self):
         super(TestGetSupplier, self).setup()
 
         with self.app.app_context():
-            payload = load_example_listing("Supplier")
+            payload = load_example_listing("supplier_creation")
```
pcraig3: Yeah, we used the `Supplier` fixture file in loads of places because we could "PUT" it into the database and then hardcode the id everywhere thereafter.  

I agree that it's probably possible to turn `"supplier_creation"` into `"Supplier"` (and drop the existing supplier), but at a certain point, you just stop going down these rabbit holes.

(I'll have a look though, see how hard it is.)

tests/main/views/test_users.py
pcraig3: [It's a fixture, Kev](https://github.com/alphagov/digitalmarketplace-api/blob/a1f9f4dce3d93d53f11bcce575329ff8e635d525/tests/conftest.py#L390).  Define it in `conftest.py` and then pass it in as a parameter to any method.

Amazing, no?

I _could_ have used the setup_dummy method again, but it felt like a lot of overhead when we just wanted something simple (plus there's that whole faff of having to add 2 of them because [the first supplier added by `setup_dummy_suppliers()` has an id of 0](https://github.com/alphagov/digitalmarketplace-api/blob/a1f9f4dce3d93d53f11bcce575329ff8e635d525/tests/helpers.py#L114), which is rejected when [trying to create a supplier user](https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/users.py#L294)).

If you think it's not simple though, I can change it to use the `setup_dummy_suppliers()` method.

tests/main/views/test_users.py
pcraig3: Yeah, this is a weird formatting thing that comes out of `nose2pytest`.  I cleaned up most of them but not all of them.

pcraig3:
Yeah, it does feel nicer.  I'll see if it's a big hassle to fix them all.

tests/main/views/test_drafts.py
TheDoubleK: Yep fine, makes sense, I was just wondering.

app/main/views/drafts.py
TheDoubleK: LOLz

tests/main/views/test_users.py
TheDoubleK: No it's fine. Amazing... I'm still to be convinced.

app/main/views/drafts.py
pcraig3:
```python
 if draft.service_id:
        """
        me: publishing a draft rn thanks
        martyn from a year ago: (lol nope)
        me: wat
        """
```

pcraig3:
Fixed some of it but not all of it.

*fixed*
- sentences begin with cApTiAl lEtTeRs
- db.commit at the end of the dummy_service method
- reduced cognitive load
- got rid of multiline asserts with hanging brackets
- flipped round `200 == status` asserts

*not fixed*
- still using SaaS services
- didn't combine any of the supplier json files
- didn't remove the fixtures I added to the user tests

TheDoubleK:
I'm gonna take your word for it you did all those things properly. Too long to go back in.
:+1:

PR #523, opened 2017-02-01
Move config into the future

idavidmcdonald:
If we don't get our new flow released within a week then on my head be it...


pcraig3:
pasted image at 2017_02_01 13_35

👍

PR #525, opened 2017-02-02
[#138323337]add migration: Add SupplierFramework.prefill_declaration_from_framework_id column and constraints

risicle:
Story https://www.pivotaltracker.com/story/show/138323337

Accompanying model changes will follow with the feature once I've tidied up the commits.

**Edit:** feature PR https://github.com/alphagov/digitalmarketplace-api/pull/526

PR #526, opened 2017-02-02
[#138323337] add prefillDeclarationFromFrameworkSlug property to SupplierFramework

risicle:
This sits on top of https://github.com/alphagov/digitalmarketplace-api/pull/525

Story https://www.pivotaltracker.com/story/show/138323337

It was decided that we would reuse/repurpose the `update_supplier_on_framework` endpoint to be a more general `update_supplier_framework`, which now also accepts `reusesAnswersFromFrameworkSlug` being updated. Tests should have better coverage too.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -17,6 +17,7 @@ from ...utils import (
     pagination_links,
     drop_foreign_fields,
     get_json_from_request,
+    json_has_keys,
     json_has_required_keys,
     json_only_has_required_keys,
     json_has_matching_id,
@@ -395,7 +396,7 @@ def register_framework_interest(supplier_id, framework_slug):
 
 
 @main.route('/suppliers/<supplier_id>/frameworks/<framework_slug>', methods=['POST'])
-def update_supplier_on_framework(supplier_id, framework_slug):
+def update_supplier_framework(supplier_id, framework_slug):
     framework = Framework.query.filter(
         Framework.slug == framework_slug
     ).first_or_404()
@@ -408,7 +409,7 @@ def update_supplier_on_framework(supplier_id, framework_slug):
     updater_json = validate_and_return_updater_request()
     json_has_required_keys(json_payload, ["frameworkInterest"])
     update_json = json_payload["frameworkInterest"]
-    json_only_has_required_keys(update_json, ["onFramework"])
+    json_has_keys(update_json, optional_keys=("onFramework", "prefillDeclarationFromFrameworkSlug",))
 
     interest_record = SupplierFramework.query.filter(
         SupplierFramework.supplier_id == supplier.supplier_id,
@@ -418,7 +419,21 @@ def update_supplier_on_framework(supplier_id, framework_slug):
     if not interest_record:
         abort(404, "supplier_id '{}' has not registered interest in {}".format(supplier_id, framework_slug))
 
-    interest_record.on_framework = update_json['onFramework']
+    if "onFramework" in update_json:
+        interest_record.on_framework = update_json["onFramework"]
+    if "prefillDeclarationFromFrameworkSlug" in update_json:
+        if update_json["prefillDeclarationFromFrameworkSlug"] is not None:
+            prefill_declaration_from_framework_id = db.session.query(SupplierFramework.framework_id).filter(
+                SupplierFramework.framework.has(Framework.slug == update_json["prefillDeclarationFromFrameworkSlug"]),
+                SupplierFramework.supplier_id == supplier.supplier_id,
+            ).scalar()
+            if prefill_declaration_from_framework_id is None:
+                abort(400, "Supplier hasn't registered interest in a framework with slug '{}'".format(
+                    update_json["prefillDeclarationFromFrameworkSlug"]
+                ))
+        else:
+            prefill_declaration_from_framework_id = None
+        interest_record.prefill_declaration_from_framework_id = prefill_declaration_from_framework_id
 
     audit_event = AuditEvent(
         audit_type=AuditTypes.supplier_update,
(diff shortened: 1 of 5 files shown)


risicle:
@samuelhwilliams noted that perhaps a better name for this variable might be `prefillAnswersFromFrameworkSlug` and I think people generally agree with that...

risicle:
or even `prefillDeclarationFromFrameworkSlug`... `declarationPrefilledFromFrameworkSlug`?

risicle:
Oh look it's all magically changed name.

app/main/views/suppliers.py
```diff
@@ -418,7 +419,19 @@ def update_supplier_on_framework(supplier_id, framework_slug):
...
+    if "onFramework" in update_json:
+        interest_record.on_framework = update_json["onFramework"]
+    if "prefillDeclarationFromFrameworkSlug" in update_json:
+        if update_json["prefillDeclarationFromFrameworkSlug"] is not None:
+            prefill_declaration_from_framework_id = db.session.query(SupplierFramework.framework_id).filter(
+                SupplierFramework.framework.has(Framework.slug == update_json["prefillDeclarationFromFrameworkSlug"]),
+                SupplierFramework.supplier_id == supplier.supplier_id,
+            ).scalar()
+            if prefill_declaration_from_framework_id is None:
+                abort(400, "Unknown framework slug '{}'".format(update_json["prefillDeclarationFromFrameworkSlug"]))
```
TheDoubleK: It's not quite "Unknown", more "Supplier has not applied for framework slug..."

risicle: Oh that's interesting - I wonder how my change to this message got squeezed out in my rebase. It was supposed to be `Supplier hasn't registered interest in a framework with slug {}`, which is a bit of a cheat because what I'm doing is using the same error for both "unknown framework" and "not registered interest".

migrations/versions/830_add_supplierframework_prefill_declaration.py
```diff
@@ -0,0 +1,26 @@
+"""Add SupplierFramework.prefill_declaration_from_framework_id column and constraints
```
TheDoubleK: The migration part of this should go in its own separate pull request and get released first on its own.

risicle: It is, https://github.com/alphagov/digitalmarketplace-api/pull/525 - it's just that this PR is sitting on top of that because otherwise travis won't pass.

tests/conftest.py
```diff
@@ -222,17 +222,24 @@ def _framework_fixture_inner(request, app, slug, **kwargs):
 
 
 def _supplierframework_fixture_inner(request, app, sf_kwargs=None):
+    """
```
TheDoubleK: This is pretty sweet, if slightly mindbending :clap:

risicle: Really? I don't like it. I'm just cramming it in there until someone factory boys our api tests...

tests/main/views/test_suppliers.py
```diff
@@ -1511,7 +1516,38 @@ def supplier_framework_interest(self, supplier_framework, update):
                 }),
             content_type='application/json')
 
-    def test_adding_supplier_has_passed(self, supplier_framework):
+    @staticmethod
+    def _refetch_serialized_sf(supplier_framework):
```
TheDoubleK: These also seem really useful.  Nice.

tests/main/views/test_suppliers.py
```diff
@@ -1511,7 +1516,38 @@ def supplier_framework_interest(self, supplier_framework, update):
...
+            ).first(),
+            AuditEvent.type == "supplier_update",
+        ).order_by(AuditEvent.created_at.desc()).first()
+
+    @classmethod
+    def _assert_and_return_audit_event(cls, supplier_framework):
+        # must be performed within an app context
+        audit = cls._latest_supplier_update_audit_event(supplier_framework['supplierId'])
+        assert audit.type == "supplier_update"
+        assert audit.user == "interested@example.com"
```
TheDoubleK: Might it be worth leaving the user check up to the calling test?  I can imagine tests might want to test with different users, maybe?

risicle: I hadn't really intended for this to be used outside `TestSupplierFrameworkUpdates` - if someone finds it useful they can yank it out when they transfer it I guess?

tests/main/views/test_suppliers.py
```diff
@@ -1521,83 +1557,197 @@ def test_adding_supplier_has_passed(self, supplier_framework):
...
+        assert response.status_code == 400
+        error_message = json.loads(response.get_data(as_text=True))['error']
+        assert error_message == "Invalid JSON should not have 'agreementReturned' keys"
+
+        with self.app.app_context():
+            # check nothing has changed on db
+            assert supplier_framework == self._refetch_serialized_sf(supplier_framework)
+            assert self._latest_supplier_update_audit_event(supplier_framework["supplierId"]) is None
+
+    def test_setting_prefill_declaration_from_framework_happy_path(
```
TheDoubleK: `test_setting_and_unsetting...` ?
As this test does both.

tests/main/views/test_suppliers.py
TheDoubleK: Fair enough, yep.

tests/conftest.py
TheDoubleK: Well, it does a job in an interesting way (i.e. I wouldn't have thought of doing it like this).

tests/main/views/test_suppliers.py
risicle: yup

TheDoubleK:
Test name isn't a deal breaker - fit it up if you like - and the rest you've addressed already.
:+1:

PR #527, opened 2017-02-03
update dos framework.framework to be 'digital-outcomes-and-specialists'

danielknell:
We need to rename `framework.framework` for dos frameworks. It will be renamed from 'dos' to 'digital-outcomes-and-specialists'.

This code change precedes the migration. There will be similar pull requests on the buyer and supplier frontend to ensure backwards compatbility. After the migration we can then remove 'dos' from our apps.


TheDoubleK:
Looks legit :+1:

PR #528, opened 2017-02-03
update g9 schemas using first draft of hosting/software questions

samuelhwilliams:
Updated g9 schemas for cloud hosting/software/support, generated from first draft of questions. This will allow viewing the flow on preview to see the current state and spot things which need changing.

### DIFF
--- a/json_schemas/services-g-cloud-9-cloud-software.json
+++ b/json_schemas/services-g-cloud-9-cloud-software.json
@@ -2,19 +2,61 @@
   "$schema": "http://json-schema.org/schema#",
   "additionalProperties": false,
   "properties": {
-    "allTheThingsToSee": {
-      "items": {
-        "enum": [
-          "option_one",
-          "option_two"
-        ]
-      },
-      "maxItems": 2,
-      "minItems": 1,
-      "type": "array",
-      "uniqueItems": true
+    "API": {
+      "type": "boolean"
     },
-    "productBenefits": {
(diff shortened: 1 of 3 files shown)


json_schemas/services-g-cloud-9-cloud-software.json
```diff
@@ -2,18 +2,348 @@
...
+      "minLength": 1,
+      "type": "string"
+    },
+    "incidentManagementApproach": {
+      "minLength": 1,
+      "type": "string"
+    },
+    "incidentManagementType": {
+      "enum": [
+        "Conforms to a recognised standard, for example, CSA CCM v3.0 or ISO/IEC 27035:2011 or SSAE-16 / ISAE 3402",
```
TheDoubleK: Should we have a think about whether it would be better to store short values rather than the entire label text for some of these checkboxes and radio button questions?  I know it's what we have at the moment, but it feels like we're missing a trick.

samuelhwilliams: I agree, it would be nice to do this.

TheDoubleK: Just not right now? 😛

TheDoubleK:
Assuming these are from the latest content PR I just thumbsed-up, :+1:

Be aware that merging this will break the supplier frontend until new content is pulled in there, so maybe get that PR ready to go before merging this. (Buyer frontend will be fine to lag behind for now).

PR #530, opened 2017-02-06
SupplierFramework model: make prefill_declaration_from_supplier_framework relationship viewonly

risicle:
A quick fix following https://github.com/alphagov/digitalmarketplace-api/pull/526

This should silence a warning from sqlalchemy about a collision of multiple
relationships being able to write to `supplier_id` - we probably don't want
people being able to inadvertantly flip the `supplier_id` through setting
`prefill_declaration_from_supplier_framework` anyway


TheDoubleK:
Nice sleuthing :+1:

risicle:
(it suggested the fix in the error message)

PR #531, opened 2017-02-06
138323285-Add_application_close_date_to_framerwork

benvand:
#### 138323285-Add_application_close_date_to_framerwork

* Add application_close_date to framework model
* Update existing test

### Requires https://github.com/alphagov/digitalmarketplace-api/pull/531

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -96,16 +96,22 @@ class Framework(db.Model):
     name = db.Column(db.String(255), nullable=False)
     framework = db.Column(db.String(), index=True, nullable=False)
     framework_agreement_details = db.Column(JSON, nullable=True)
-    status = db.Column(db.String(),
-                       index=True, nullable=False,
-                       default='pending')
+    status = db.Column(
+        db.String(),
+        index=True, nullable=False,
+        default='pending'
+    )
     clarification_questions_open = db.Column(db.Boolean, nullable=False, default=False)
     lots = db.relationship(
-        'Lot', secondary="framework_lots",
-        lazy='joined', innerjoin=False,
+        'Lot',
+        secondary="framework_lots",
+        lazy='joined',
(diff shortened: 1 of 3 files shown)


migrations/versions/840_add_application_close_date_for_framework.py
```diff
@@ -0,0 +1,31 @@
+"""Add `application_close_date` for Framework
```
TheDoubleK: The migration should go in a separate PR from the code, so it can be released on its own.

migrations/versions/840_add_application_close_date_for_framework.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = '840'
+down_revision = '830'
+
+
+
+def upgrade():
+    op.add_column('frameworks', sa.Column('application_close_date', sa.Date(), nullable=True))
```
TheDoubleK: All our other timestamps are `sa.DateTime()` - I know we only need the Date for the current use-case, but feels like it wouldn't hurt to keep it the same here.

In terms of the time of day to set - I think DOS closed at 3pm, G8 and DOS2 at 5pm.

app/models.py
```diff
@@ -113,21 +118,19 @@ def get_lot(self, lot_slug):
         )
 
     def serialize(self):
+        acd = self.application_close_date.strftime(DATE_FORMAT) if self.application_close_date else ''
```
TheDoubleK: This could just be in-lined in the serialisation?  It's only used once, and everything else is in-line, so should be fine?

benvand:
New PR for these changes to clean up commits:
https://github.com/alphagov/digitalmarketplace-api/pull/539

PR #532, opened 2017-02-07
138323285-Add_application_close_date_to_framerwork MIGRATION

benvand:
* Migration

### DIFF
--- a/migrations/versions/840_add_application_close_date_for_framework.py
+++ b/migrations/versions/840_add_application_close_date_for_framework.py
@@ -4,7 +4,8 @@ Revises: 830
 Create Date: 2017-02-06 11:09:26.852142
 """
 from alembic import op
-from datetime import datetime, tzinfo
+from datetime import datetime
+from dateutil import tz
 import sqlalchemy as sa
 
 
@@ -21,6 +22,8 @@ frameworks_table = sa.table(
     sa.Column('application_close_date', sa.DateTime)
 )
 
+tz_utc = tz.tzoffset('utc', 0)
+
 
 def upgrade():
     op.add_column('frameworks', sa.Column('application_close_date', sa.DateTime(), nullable=True))
@@ -31,9 +34,9 @@ def upgrade():
 
     fields = ('slug', 'application_close_date', 'allow_declaration_reuse')
     new_values = (
-        ('digital-outcomes-and-specialists', datetime(2016, 1, 1, 15, tzinfo=tzinfo('utc')), False),
-        ('digital-outcomes-and-specialists-2', datetime(2017, 1, 16, 17, tzinfo=tzinfo('utc')), True),
-        ('g-cloud-8', datetime(2016, 6, 1, 17, tzinfo=tzinfo('utc')), True),
+        ('digital-outcomes-and-specialists', datetime(2016, 1, 1, 15, tzinfo=tz_utc), False),
+        ('digital-outcomes-and-specialists-2', datetime(2017, 1, 16, 17, tzinfo=tz_utc), True),
+        ('g-cloud-8', datetime(2016, 6, 1, 17, tzinfo=tz_utc), True),
 
     )
     new_values = [dict(zip(fields, i)) for i in new_values]
--- a/migrations/versions/840_add_application_close_date_for_framework.py
+++ b/migrations/versions/840_add_application_close_date_for_framework.py
@@ -1,0 +1,37 @@
+"""Add `application_close_date` for Framework
+Revision ID: 840
+Revises: 830
+Create Date: 2017-02-06 11:09:26.852142
+"""
+from alembic import op
+from datetime import datetime
+import sqlalchemy as sa
+
+from dmutils.formats import DATETIME_FORMAT, EUROPE_LONDON
+
+# revision identifiers, used by Alembic.
+revision = '840'
+down_revision = '830'
+
+
+
+def upgrade():
+    op.add_column('frameworks', sa.Column('application_close_date', sa.DateTime(), nullable=True))
+    op.add_column('frameworks', sa.Column('allow_declaration_reuse', sa.Boolean(), nullable=False, server_default='false'))
+
+    for framework_slug, a_c_date, declaration_reuse in (
+        ('digital-outcomes-and-specialists', datetime(2016, 1, 1, 15, tzinfo=EUROPE_LONDON), False),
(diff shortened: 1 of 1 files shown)


migrations/versions/840_add_application_close_date_for_framework.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def upgrade():
+    op.add_column('frameworks', sa.Column('application_close_date', sa.DateTime(), nullable=True))
+    op.add_column('frameworks', sa.Column('allow_declaration_reuse', sa.Boolean(), nullable=False, server_default='false'))
+
+    for framework_slug, a_c_date, declaration_reuse in (
+        ('digital-outcomes-and-specialists', datetime(2016, 1, 1, 15, tzinfo=EUROPE_LONDON), False),
+        ('digital-outcomes-and-specialists-2', datetime(2017, 1, 16, 17, tzinfo=EUROPE_LONDON), True),
+        ('g-cloud-8', datetime(2016, 6, 1, 17, tzinfo=EUROPE_LONDON), True),
```
risicle: I _suggest_ it's best to do these in UTC and avoid timezone handling entirely because it gets really weird.

migrations/versions/840_add_application_close_date_for_framework.py
```diff
@@ -0,0 +1,37 @@
...
+
+    ):
+        query_string = """UPDATE frameworks set application_close_date = '{a_c_date}'::DATE where slug = '{framework_slug}'"""
+        op.execute(query_string.format(framework_slug=framework_slug, a_c_date=a_c_date.strftime(DATETIME_FORMAT)))
```
risicle: Even though this is just a migration, it's really bad form to use standard string formatting with sql - I'm sure op.execute has its own (safe) parameter handling that would also remove the requirement for any of this `::DATE` stuff.

Wynndow: @risicle I've just done a similar thing in a migration I've got a PR open for. Why is doing this bad form?

risicle: https://xkcd.com/327/

Though it's not user submitted input here, it's generally a very had habit to get into.

Your database adaptor (`psycopg2` here) will also be able to do type conversion for you if you use its built-in string parametrization.

Wynndow: Ah, I see. What do you make of this (I've updated it after reading your comment and doing some googling): https://github.com/alphagov/digitalmarketplace-api/blob/migrate-draft-dos1-briefs-to-draft-dos2-briefs/migrations/versions/840_migrate_draft_dos1_briefs_to_draft_dos2.py#L44

risicle: ...pointless?

risicle: I'm not sure why you're searching by `name` and not `slug`, secondly, it shouldn't be problematic as it's coming from an id field that should already be constrained to an integer. But still not something to encourage. ORMs exist. Database adaptors can do proper safe parametrization. Don't ignore them.

risicle: Hold on, you're using adaptor-formatting, I'm just not used to seeing the sqlalchemy notation for it.

risicle: Hold on again, im not sure what notation that is but I'm not used to seeing it.

benvand: @risicle 
Explicitly no parameter handling:
https://github.com/ideadevice/alembic/blob/master/alembic/operations.py#L831

I side with the 'Parting Words' section of the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html ) view when it comes to why 'local' code style is important
(See migrations 430, 610, 160 and 340)

Also also:
> really bad form 
> very bad habit
are erring on the negative side of ways to address a colleagues contribution. I've found I catch more flies with honey with this sort of thing. I do appreciate you looking at PRs though, thanks :100: 

Changed the dates, it doesn't make a lot of odds to the end user, I don't mind either way, one less import.
Changed the migration. Not very in line with the codebase but does avoid strings.

@Wynndow 
As a standard for this sort of ORM data migration you'd generally either import the model (less good) or craft your own model (gooder) with the required fields. That way you never have to write any SQL (easy to mess up), you ensure that the ORM knows what's going on with the state of the db (because you're using its model) _and_ you remain consistent with the rest of your codebase with regards to accessing/ updating db records (easy for other people to read).
This Github answer explains the 'how' quite well:
http://stackoverflow.com/a/24623979

@here I crafted my own to see how it worked out :)

Wynndow: @benvand That's awesome. Thanks very much, that has gone in my bookmarks!

risicle: > Explicitly no parameter handling:

Alembic has an *offline* mode?? Well found. I guess this means we should be using things like temporary tables where we need to more complicated things then...

> more flies with honey

I think I'm just trying to emphasize how even though it doesn't matter in this case it's something to avoid because it trips peoples internal alarms. If we are indeed forced to do it this way it might be worth leaving a comment to assure those of us who are borderline robots.

It's not meant as a huge criticism - *I* have a bunch of "really really bad habits" in my coding style - we all do.

Wynndow: What's the difference between being offline and online?

risicle:
Ooh you figured out how to use orm queries in alembic migrations... will have to refer to this in future. :+1:

risicle:
(oh, provided you squash out or recomment the "foo" changesets)

migrations/versions/840_add_application_close_date_for_framework.py
risicle: In offline mode, alembic wants to be able to build essentially a standalone sql script for a whole series of migrations that can run without having to call back into python to e.g. manipulate values.

risicle: (so e.g. if we wanted to "note down" some values retrieved from a query for later use, the way to do it to keep the offline mode happy would probably be `CREATE TEMPORARY TABLE AS ...` and then pulling the values back from there in the later statements)

benvand:
Made another PR in the end as this branch got polluted with commits
https://github.com/alphagov/digitalmarketplace-api/pull/535

benvand:
Made another PR in the end as this branch got polluted with commits
#535

PR #533, opened 2017-02-07
Add brief response schemas for DOS2

idavidmcdonald:
For this pivotal story - https://www.pivotaltracker.com/story/show/137518153

### DIFF
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -489,105 +489,108 @@ def test_max_price_larger_than_min_does_not_overwrite_previous_errors():
 
 
 def test_brief_response_essential_requirements():
-    assert get_validation_errors(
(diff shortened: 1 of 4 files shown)


tests/test_validation.py
```diff
@@ -523,7 +523,7 @@ def test_brief_response_essential_requirements():
 
 
 def test_brief_response_nice_to_have_requirements():
-    schema_name = "brief-responses-digital-outcomes-and-specialists-digital-specialists"
+    schema_name = "brief-responses-digital-outcomes-and-specialists-2-digital-specialists"
```
Wynndow: If we previously had tests for both essentials and nice to have's for DOS1, should we not have tests for both for DOS2?

PR #534, opened 2017-02-07
Migrate draft DOS1 briefs to draft DOS2 briefs

Wynndow:
NOT TO BE MERGED UNTIL DOS2 IS LIVE AND DOS1 IS CLOSED FOR BRIEF SUBMISSIONS

For this story on Pivotal [https://www.pivotaltracker.com/story/show/137359003](https://www.pivotaltracker.com/story/show/137359003)

When DOS1 is closed for brief submissions and DOS2 opens, any supplier
who has a draft DOS1 brief will effectively lose it, as they'll be
unable to publish it.

This migration will move all draft DOS1 briefs to be draft DOS2 briefs,
allowing them to be published. Which framework the brief
goes into doesn't affect buyers - they are eligible by default.

This will need to be run immediately after DOS1 is closed so that no
buyer attempt to publish their DOS1 brief in the period between DOS1
closing and the migration happening.

UPDATE:
This migration was breaking all of the tests. To make this migration agnostic of the environment, it checks the id's of the DOS1 and DOS2 frameworks and uses those id's to update the rows it needs to. DOS1 was created with a migration so it is present in the db when this migration runs. DOS2 however was set up via the API and so it NOT present. Doing a `fetchall()` on the response from the database for an SQL statement that didn't find anything returns a list, and trying to access an index of that list is what broke the tests.

To fix this, I've set the migration up to bail out if it doesn't get a response from querying the id of DOS2. This feels a bit weird. There's another option, which is to import `current_app` from flask and check the environment variable in the config. If it's `test` then skip the migration. Again, feels a bit weird.

Another option would be to hard code the framework id's depending on the environment (accessed as mentioned above), so that the database doesn't need to be queried in the migration. It then wouldn't find any frameworks with that id when trying to perform the update so shouldn't kick an error.

Any thoughts are more than welcome.

### DIFF
--- a/migrations/versions/840_migrate_draft_dos1_briefs_to_draft_dos2.py
+++ b/migrations/versions/840_migrate_draft_dos1_briefs_to_draft_dos2.py
@@ -1,0 +1,26 @@
+"""Migrate draft DOS1 briefs to draft DOS2 briefs
+
+Revision ID: 840
+Revises: 830
+Create Date: 2017-02-07 15:31:50.715832
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '840'
+down_revision = '830'
+
+from alembic import op
+
+def upgrade():
+    # Change framework of draft DOS1 briefs from DOS1 (framework_id == 5) to DOS2 (framework_id == 7)
+    op.execute("""
+        UPDATE briefs
+        SET framework_id = 7
+        WHERE framework_id = 5 AND published_at IS NULL
+    """)
+
+
+def downgrade():
+    # No downgrade
+    pass
--- a/migrations/versions/840_migrate_draft_dos1_briefs_to_draft_dos2.py
+++ b/migrations/versions/840_migrate_draft_dos1_briefs_to_draft_dos2.py
@@ -18,13 +18,13 @@ def upgrade():
 
     dos1_res = conn.execute("""
         SELECT id FROM frameworks
-        WHERE name = 'Digital Outcomes and Specialists'
+        WHERE slug = 'digital-outcomes-and-specialists'
     """)
     dos1_framework_id = dos1_res.fetchall()[0][0]
 
     dos2_res = conn.execute("""
         SELECT id FROM frameworks
-        WHERE name = 'Digital Outcomes and Specialists 2'
+        WHERE slug = 'digital-outcomes-and-specialists-2'
     """)
 
     # DOS2 framework was not created with a migration, it was created via the API. This means that when the tests run
@@ -66,8 +66,6 @@ def upgrade():
         SET framework_id = :dos2_id
         WHERE framework_id = :dos1_id AND published_at IS NULL
     """), **framework_ids)
-
-
 
 def downgrade():
     # No downgrade
--- a/migrations/versions/840_migrate_draft_dos1_briefs_to_draft_dos2.py
+++ b/migrations/versions/840_migrate_draft_dos1_briefs_to_draft_dos2.py
@@ -42,10 +42,31 @@ def upgrade():
     }
 
     conn.execute(text("""
+        INSERT INTO audit_events (
+            type,
+            created_at,
+            "user",
+            data,
+            object_type,
+            object_id,
+            acknowledged
+        ) SELECT
+            'update_brief_framework_id',
+            NOW(),
+            'Migration 840',
+            '{ "previousFrameworkId": :dos1_id, "newFrameworkId": :dos2_id }',
+            'Brief',
+            briefs.id,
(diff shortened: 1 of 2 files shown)


migrations/versions/840_migrate_draft_dos1_briefs_to_draft_dos2.py
```diff
@@ -0,0 +1,26 @@
...
+revision = '840'
+down_revision = '830'
+
+from alembic import op
+
+def upgrade():
+    # Change framework of draft DOS1 briefs from DOS1 (framework_id == 5) to DOS2 (framework_id == 7)
+    op.execute("""
+        UPDATE briefs
+        SET framework_id = 7
```
idavidmcdonald: Are we able to rely on framework ID's being consistent across local, preview, staging and prod?

Wynndow: Hmm....let me check

Wynndow: Nope. Turns out that the above would be fine for staging/prod, but for some reason DOS2 has id 8 on preview.

Wynndow: Have updated to make it work regardless.

migrations/versions/840_migrate_draft_dos1_briefs_to_draft_dos2.py
```diff
@@ -0,0 +1,47 @@
...
+    dos2_framework_id = query_results[0][0]
+
+    op.execute("""
+        UPDATE briefs
+        SET framework_id = {}
+        WHERE framework_id = {} AND published_at IS NULL
+    """.format(dos2_framework_id, dos1_framework_id))
+
+
+def downgrade():
```
TheDoubleK: You could have a downgrade that takes all DOS2 and moves them to DOS? But it could only realistically be applied for a very short time after the upgrade...

migrations/versions/840_migrate_draft_dos1_briefs_to_draft_dos2.py
```diff
@@ -0,0 +1,47 @@
...
+down_revision = '830'
+
+from alembic import op
+
+def upgrade():
+    conn = op.get_bind()
+
+    dos1_res = conn.execute("""
+        SELECT id FROM frameworks
+        WHERE name = 'Digital Outcomes and Specialists'
```
TheDoubleK: I'd have gone with slug rather than name, but fine :)

TheDoubleK:
FWIW, the test thing is weird, but better than hard coded ID's. Even if Ids ok for live environments it wouldn't match my dev database.

Wynndow:
Thanks Kev. I'm gonna have a look at what happens if they try to publish DOS 1 brief when it's closed. I suspect the api will just error and they won't lose their draft - but I'll make sure.

TheDoubleK:
Also, just spotted this comment from Rob on a different PR, which is relevant here: https://github.com/alphagov/digitalmarketplace-api/pull/532#discussion_r100059980

Wynndow:
Yeah thanks Kev. I've had a chat with him and updated the migration

Wynndow:
@TheDoubleK After chatting with @allait it was decided it'd be a good idea to add audit events for the migrations. Could you take a look a the new stuff? Cheers!

TheDoubleK:
Ah, but then your select for the audit events won't work, will it?

Wynndow:
Yep, need to make the event before changing.

Wynndow:
@TheDoubleK Also, I forgot to mention, the audit type I've used in the migration is new. I've got a PR up on the apiclient to add it to the list there. Even though the script never touches the api client of the list of audit types, I thought it a good idea to add it to keep that list canonical?

[https://github.com/alphagov/digitalmarketplace-apiclient/pull/66](https://github.com/alphagov/digitalmarketplace-apiclient/pull/66)

TheDoubleK:
Yep. But then this migration should probably reference the new constant rather than hard coding, sorry!

allait:
@TheDoubleK I don't think migrations should use the constant. I think it creates a similar problem to using models in migration code: old migrations should run no matter what, so eg if we later decide to rename this type and replace it with something else and write a migration for it this one should still run, even though the constant has changed.

TheDoubleK:
Hmm, I can see the point, but also you could argue that once we have an audit type defined that  exists in the database we shouldn't remove it from our code in future.  But yep OK let's leave this as it is.

PR #537, opened 2017-02-10
BUGFIX-repair-migration

benvand:
* Remove tzinfo, postgres converts to utc anyway

FYI was going to use this to get around our tzinfo problem: 
```
utc = type('utc', (tzinfo,), {'tzname': lambda self, i: 'UTC', 'utcoffset': lambda self, i: timedelta(0)})()
```
but then I remembered I'm not a bad person.

### DIFF
--- a/migrations/versions/840_add_application_close_date_for_framework.py
+++ b/migrations/versions/840_add_application_close_date_for_framework.py
@@ -5,7 +5,6 @@ Create Date: 2017-02-06 11:09:26.852142
 """
 from alembic import op
 from datetime import datetime
-from dateutil import tz
 import sqlalchemy as sa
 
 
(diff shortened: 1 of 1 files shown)


migrations/versions/840_add_application_close_date_for_framework.py
```diff
@@ -34,9 +31,9 @@ def upgrade():
 
     fields = ('slug', 'application_close_date', 'allow_declaration_reuse')
     new_values = (
-        ('digital-outcomes-and-specialists', datetime(2016, 1, 1, 15, tzinfo=tz_utc), False),
-        ('digital-outcomes-and-specialists-2', datetime(2017, 1, 16, 17, tzinfo=tz_utc), True),
-        ('g-cloud-8', datetime(2016, 6, 1, 17, tzinfo=tz_utc), True),
+        ('digital-outcomes-and-specialists', datetime(2016, 1, 1, 15), False),
```
TheDoubleK: `digital-outcomes-and-specialists` should be True.  Other than that looks good.

TheDoubleK: Merge once that's changed and :crossed_fingers:

PR #539, opened 2017-02-14
138323285-Add_application_close_date_to_framework

benvand:
* Add application_close_date to framework model
* Update existing test

OLD PR:
https://github.com/alphagov/digitalmarketplace-api/pull/531

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -95,16 +95,22 @@ class Framework(db.Model):
     name = db.Column(db.String(255), nullable=False)
     framework = db.Column(db.String(), index=True, nullable=False)
     framework_agreement_details = db.Column(JSON, nullable=True)
-    status = db.Column(db.String(),
-                       index=True, nullable=False,
-                       default='pending')
+    status = db.Column(
+        db.String(),
+        index=True, nullable=False,
+        default='pending'
+    )
     clarification_questions_open = db.Column(db.Boolean, nullable=False, default=False)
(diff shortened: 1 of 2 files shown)


app/models.py
```diff
@@ -119,16 +125,16 @@ def serialize(self):
...
-                self.framework_agreement_details or {}
-            ).get("frameworkAgreementVersion"),
+            'countersignerName': (self.framework_agreement_details or {}).get("countersignerName"),
+            'frameworkAgreementVersion': (self.framework_agreement_details or {}).get("frameworkAgreementVersion"),
             'variations': (self.framework_agreement_details or {}).get("variations", {}),
             'status': self.status,
             'clarificationQuestionsOpen': self.clarification_questions_open,
             'lots': [lot.serialize() for lot in self.lots],
+            'application_close_date': (
+                self.application_close_date.strftime(DATETIME_FORMAT) if self.application_close_date else ''
```
TheDoubleK: None of our other nullable datetime fields do this (empty string if no date) - they seem to just return a null if there's no date.  I'm not saying that's the right thing, but there's a benefit in being consistent how we treat null things.

benvand: Ah like at 
https://github.com/alphagov/digitalmarketplace-api/blob/master/app/models.py#L497-L497
best to stay consistent, I'll get that changed

TheDoubleK:
:+1: after a rebase

PR #542, opened 2017-02-16
Follow up questions

allait:
### Update brief responses schemas with new followup rules

Changes to the followup schema generation preserve the existing
behaviour but move some of the rules around to work when used in
the API.

Eg dropping the lead-in question.id from the oneOf schema required
lists allows us to use page_questions and partial service updates,
which raised validation errors otherwise. Instead, it moves the
lead-in question to the parent schema's required list, since if it's
not optional then it's required in both cases.

Another change is dropping the property description for the followup
question from the oneOf subschema. The property is described in the
parent schema, which has access to the question type to generate a
correct one, so the only thing we need to add is to mark it as
required.

### Add validation errors for G9 follow-up questions

Adds JSON schema error translation for top-level oneOf validator
errors, which are used by the G9 schemas.

Adds tests using the existing G9 followup question.

### DIFF
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -646,7 +646,7 @@ def test_g9_followup_questions():
             "freeVersionLink": "https://gov.uk",
         }
         yield check, schema_name, ["freeVersionTrialOption"], data, {
-            'freeVersionLink': u'{} is not valid under any of the given schemas'.format(data)
+            '_form': [u'{} is not valid under any of the given schemas'.format(data)]
         }
 
         data = {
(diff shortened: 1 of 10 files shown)


tests/test_validation.py
```diff
@@ -593,6 +593,115 @@ def test_brief_response_nice_to_have_requirements():
         assert "object" in error_messages
 
 
+def test_g9_followup_questions():
+
+    def check(schema, required_fields, data, errors):
```
idavidmcdonald: Nice on cutting down duplication with this

PR #543, opened 2017-02-16
Allow create/update/submit response for expired frameworks

Wynndow:
Part of this story: [https://www.pivotaltracker.com/story/show/139283847](https://www.pivotaltracker.com/story/show/139283847)

When we expire DOS1, suppliers will still need to be able to apply to live DOS1 briefs. This means that the api needs to allow brief responses to be created/updated/published for frameworks who's status is `expired`.

This does that.

Also see this PR on the supplier frontend: [https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/619](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/619)

### DIFF
--- a/tests/main/views/test_brief_response.py
+++ b/tests/main/views/test_brief_response.py
@@ -261,12 +261,16 @@ class CreateBriefResponseSharedTests(BaseBriefResponseTest, JSONUpdateTestMixin)
         assert res.status_code == 400
         assert 'Supplier is not eligible to apply to this brief' in res.get_data(as_text=True)
 
-    def test_cannot_create_a_brief_response_if_framework_status_is_not_live_or_expired(self, expired_dos_framework):
+    def test_cannot_create_a_brief_response_if_framework_status_is_not_live_or_expired(self, live_dos_framework):
+        framework_id = live_dos_framework['id']
         for framework_status in ['coming', 'open', 'pending', 'standstill']:
             with self.app.app_context():
                 db.session.execute(
-                    "UPDATE frameworks SET status=:status WHERE slug='digital-outcomes-and-specialists'",
-                    {'status': framework_status},
+                    "UPDATE frameworks SET status=:status WHERE id = :framework_id",
+                    {
+                        'status': framework_status,
+                        'framework_id': framework_id,
+                    },
                 )
 
                 res = self.create_brief_response()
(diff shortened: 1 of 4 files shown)


tests/main/views/test_brief_response.py
```diff
@@ -252,6 +261,20 @@ def test_cannot_create_brief_response_when_supplier_isnt_eligible(self, live_dos
         assert res.status_code == 400
         assert 'Supplier is not eligible to apply to this brief' in res.get_data(as_text=True)
 
+    def test_cannot_create_a_brief_response_if_framework_status_is_not_live_or_expired(self, expired_dos_framework):
+        for framework_status in ['coming', 'open', 'pending', 'standstill']:
+            with self.app.app_context():
+                db.session.execute(
+                    "UPDATE frameworks SET status=:status WHERE slug='digital-outcomes-and-specialists'",
```
idavidmcdonald: I remember this sort of thing being talked about in Ben's talk regarding changing our database state and not rolling it back mid unit test. I believe the plan was to move to some package that would help with this... Can't remember exactly what the outcome was and whether we now advised against changing the frameworks or we should try and do teardowns...

Another option might be to target `expired_dos_framework` by ID rather than slug so you know you are only changing one of our fixtures rather than the real DOS framework in the DB. 

I don't remember the exacts but could be worth having a quick chat with Ben if needbe.

Wynndow: Had a chat with Ben, gonna go down the route of targeting the fixture and updating it. It'll reset itself every time it's used by a test so no issues!

idavidmcdonald:
What you've written looks good, however it would be good to make our tests a bit stricter for briefs ( not just brief responses) by checking you can't do things like create/update a brief for an expired framework (at the moment we only test with an open framework).

https://github.com/alphagov/digitalmarketplace-api/blob/master/tests/main/views/test_briefs.py#L102

Wynndow:
Yep, I'll take a look

Wynndow:
@idavidmcdonald Have updated some tests to check all invalid framework states. Also tidied up that  test file. It was one massive class of tests. Now it's a few large classes.

idavidmcdonald:
👍 niiiiiice

PR #544, opened 2017-02-23
Update g9 schemas from frameworks 6.2.0

samuelhwilliams:
Refreshes json schemas for G9 from latest on dm-frameworks.


TheDoubleK:
Wow the followups really add to the size of the schemas!  And to think, our very first schemas were handcrafted, imagine if we still had to do that! 😱 

All looks good :+1:

PR #545, opened 2017-02-28
Remove validation for 'old flow' brief responses

idavidmcdonald:
As the old supplier applies for a brief response has naturally deprecated, we will no longer need to validate brief responses against the legacy schemas. Therefore we can remove the code to do this and related tests.

Note, our feature flags will also no longer be needed as they were only used to drive choosing the correct manifest to validate against.


TheDoubleK:
screen shot 2017-03-03 at 16 27 03

That's a whole lot of :boom: but could there be even more...?

Are the legacy brief response JSON schema files also redundant now?  Could delete them at the same time if they're not used elsewhere.

idavidmcdonald:
Good question. Although by that argument should we also remove all schemas for DOS1 briefs and DOS1 brief responses (in 1 weeks time when people can no longer apply to DOS1 briefs published just before the framework expires). 

Similarly, we have a bunch of g5/6 schemas kicking about....

TheDoubleK:
Those other old schemas could still be called by our code though - e.g. dev running G6 open or whatever.

The `legacy` ones here will have no way of ever being called from the code though as far as I can tell - even if DOS was live again?

idavidmcdonald:
Good point, you couldn't even hit the API with the old style data and get it to save.

The only thing that makes me hesitant is that the legacy style data does exist and will be around forever and some of our pages will need to support that. So if you are a dev, you may need to put some of this legacy data in your database so you can test locally. Similar with preview and staging environments. It feels like somewhere we should have the schema for this data defined so people can refer to it. Also, the frameworks repo does not generate the legacy schemas so once these go we will just have the old yaml files to refer too. I guess the frontend define legacy brief responses for it's tests so we will still have a record there but that's it. 

So now I've collected my thoughts... it does feel OK to remove the schemas and I will do so but just wanted to make sure that won't have a negative impact anywhere else. Setting up this 'legacy data' for testing could be a bit of an issue but that's because we remove the API route rather than the schema it self.

TheDoubleK:
:+1: after a rebase

PR #547, opened 2017-03-03
Updated schema for G9

galund:
As generated from frameworks v6.4.1


TheDoubleK:
Assuming these come straight out of the script in frameworks repo :+1:

pcraig3:
We get different diffs depending on if we're running in python versions 2 or 3, but the absolute number of changed lines is the same.

[There's a pivotal story been created about this](https://www.pivotaltracker.com/story/show/141021383).

PR #548, opened 2017-03-06
Update G9 JSON schemas from frameworks 6.4.4

samuelhwilliams:
* As per title.


TheDoubleK:
Are these final changes as far as we know?

TheDoubleK:
:+1:

samuelhwilliams:
As far as we know; yes. As far as we should expect; no.

PR #551, opened 2017-03-08
Allow agreements to be unapproved for countersigning

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/141299435

CCS have pushed the big green button by mistake for some DOS2 agreements, so we need a way to "unapprove" them.

I *really* wanted to use `disapprove` rather than `unapprove` but it seemed too weird.  Open to suggestions for better words than `unapprove` though if you have any.


galund:
If there was a UI for this, I'd be writing "Cancel Approval", but I think "unapprove" is fine for back-end :)

risicle:
Of course I have to note the odd double negative of `/approve` -> `unapprove` -> `True`

PR #552, opened 2017-03-16
[#141778727]Framework agreement details expansion

risicle:
Story https://www.pivotaltracker.com/story/show/141778727

Ok, so, we're going to need to be able to actually *use* a whole load more keys in the `frameworkAgreementDetails` dictionary and pulling out specific ones and flattening them into the toplevel returned dictionary isn't going to cut it - at least having them returned under a specific `frameworkAgreementDetails` object gives us an amount of namespacing for these relatively obscure variables.

Another obstacle is that `frameworkAgreementDetails` itself has a json schema. Should we enforce the format of these new, obscure, variables or just open the schema to any `additionalProperties`? This strawman PR proposes the former.

Thoughts?

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -124,16 +124,19 @@ class Framework(db.Model):
             'name': self.name,
             'slug': self.slug,
             'framework': self.framework,
-            'countersignerName': (self.framework_agreement_details or {}).get("countersignerName"),
-            'frameworkAgreementVersion': (self.framework_agreement_details or {}).get("frameworkAgreementVersion"),
-            'variations': (self.framework_agreement_details or {}).get("variations", {}),
             'status': self.status,
             'clarificationQuestionsOpen': self.clarification_questions_open,
             'lots': [lot.serialize() for lot in self.lots],
             'application_close_date': (
                 self.application_close_date and self.application_close_date.strftime(DATETIME_FORMAT)
             ),
-            'allow_declaration_reuse': self.allow_declaration_reuse
+            'allow_declaration_reuse': self.allow_declaration_reuse,
+            'frameworkAgreementDetails': self.framework_agreement_details or {},
+            # the following are specific extracts of the above frameworkAgreementDetails which were previously used but
+            # should now possibly be deprecated:
+            'countersignerName': (self.framework_agreement_details or {}).get("countersignerName"),
+            'frameworkAgreementVersion': (self.framework_agreement_details or {}).get("frameworkAgreementVersion"),
+            'variations': (self.framework_agreement_details or {}).get("variations", {}),
         }
 
     def get_supplier_ids_for_completed_service(self):
--- a/json_schemas/framework-agreement-details.json
+++ b/json_schemas/framework-agreement-details.json
@@ -15,6 +15,54 @@
       "minLength": 1,
       "type": "string"
     },
+    "frameworkRefDate": {
+      "minLength": 1,
+      "type": "string"
+    },
+    "frameworkStartDate": {
+      "minLength": 1,
+      "type": "string"
+    },
+    "frameworkEndDate": {
+      "minLength": 1,
+      "type": "string"
+    },
+    "frameworkExtensionLength": {
+      "minLength": 1,
+      "type": "string"
+    },
+    "frameworkURL": {
+      "minLength": 1,
+      "type": "string"
+    },
+    "signaturePageNumber": {
+      "type": "integer",
+      "minimum": 1
+    },
+    "pageTotal": {
+      "type": "integer",
+      "minimum": 1
+    },
+    "contractNoticeNumber": {
+      "minLength": 1,
+      "type": "string"
+    },
+    "lotOrder": {
+      "type": "array",
+      "minItems": 1,
+      "items": {
+        "minLength": 1,
+        "type": "string"
+      }
+    },
+    "lotDescriptions": {
+      "type": "object",
+      "minProperties": 1,
+      "additionalProperties": {
+        "minLength": 1,
+        "type": "string"
+      }
+    },
     "variations": {
       "additionalProperties": {
         "additionalProperties": false,
--- a/tests/main/views/test_frameworks.py
+++ b/tests/main/views/test_frameworks.py
@@ -324,18 +324,50 @@ class TestUpdateFramework(BaseApplicationTest, JSONUpdateTestMixin):
 
     def test_schema_validation_for_framework_agreement_details(self, open_example_framework):
         invalid_framework_agreement_details = [
-            # should be a string
-            {'frameworkAgreementVersion': 1},
-            # cannot be empty
-            {'frameworkAgreementVersion': ""},
-            # should be an object
-            {'variations': 1},
-            # object must have 'createdAt' key
-            {'variations': {"created_at": "today"}},
-            # countersigner cannot be empty
-            {'countersignerName': ""},
+            # frameworkAgreementVersion should be a string
+            {
+                'variations': {},
+                'frameworkAgreementVersion': 1,
+            },
+            # can't have a numeric lotDescription
+            {
+                'variations': {},
+                'frameworkAgreementVersion': "1",
+                'lotDescriptions': {"test-lot": 4321},
+            },
+            # can't have empty lotOrder
+            {
+                'variations': {},
+                'frameworkAgreementVersion': "1",
+                'lotOrder': [],
+            },
+            # frameworkAgreementVersion cannot be empty
+            {
+                'variations': {},
+                'frameworkAgreementVersion': "",
+            },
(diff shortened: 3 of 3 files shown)


risicle:
Further questions: in the schema I've just copied the formats of the date strings used in https://github.com/alphagov/digitalmarketplace-scripts/blob/01bf1272d0981d9725bbf951e82b46c25131518e/scripts/generate-framework-agreement-counterpart-signature-pages.py#L14 though I'm really thinking if we're going to be specifying these we should probably be doing so in a standardized format and not just the string-format that happens to get used in the place ...where it's used.

app/models.py
```diff
@@ -124,16 +124,19 @@ def serialize(self):
...
-            'frameworkAgreementVersion': (self.framework_agreement_details or {}).get("frameworkAgreementVersion"),
-            'variations': (self.framework_agreement_details or {}).get("variations", {}),
             'status': self.status,
             'clarificationQuestionsOpen': self.clarification_questions_open,
             'lots': [lot.serialize() for lot in self.lots],
             'application_close_date': (
                 self.application_close_date and self.application_close_date.strftime(DATETIME_FORMAT)
             ),
-            'allow_declaration_reuse': self.allow_declaration_reuse
+            'allow_declaration_reuse': self.allow_declaration_reuse,
```
allait: I don't know why the field above is breaking the convention, but JSON response keys should be in camel-case.

json_schemas/framework-agreement-details.json
```diff
@@ -15,6 +15,57 @@
       "minLength": 1,
       "type": "string"
     },
+    "frameworkRefDate": {
```
allait: What does 'Ref' stand for? Can we expand this key name or is it already used somewhere?

json_schemas/framework-agreement-details.json
```diff
@@ -15,6 +15,57 @@
...
+      "minLength": 1,
+      "type": "string",
+      "patterm": "^[0-9-]+$"
+    },
+    "frameworkStartDate": {
```
allait: I feel like these things should be part of the Framework itself. I think @benvand was looking into this at some point.

json_schemas/framework-agreement-details.json
```diff
@@ -15,6 +15,57 @@
...
+      "minLength": 1,
+      "type": "string",
+      "patterm": "^[0-9/]+$"
+    },
+    "frameworkEndDate": {
+      "minLength": 1,
+      "type": "string",
+      "patterm": "^[0-9/]+$"
```
allait: patter**n**

allait: Also, maybe this should be using the built-in `date-time` format (https://spacetelescope.github.io/understanding-json-schema/reference/string.html)

app/models.py
risicle: I noticed this too but that's not part of this PR - think this is just a trailing comma that got added.

json_schemas/framework-agreement-details.json
risicle: good point - @TheDoubleK - this was just transposed from the script - thoughts on better naming?

json_schemas/framework-agreement-details.json
risicle: If you're reasonably sure we won't want to be able to override these dates just for the purposes of output on the signature page (@TheDoubleK wasn't so sure) I'd agree.

json_schemas/framework-agreement-details.json
risicle: goddamn, and I probably agree.

app/models.py
allait: Ah, of course, my bad. Never mind then.

json_schemas/framework-agreement-details.json
```diff
@@ -15,6 +15,57 @@
       "minLength": 1,
       "type": "string"
     },
+    "frameworkRefDate": {
+      "minLength": 1,
+      "type": "string",
+      "patterm": "^[0-9-]+$"
```
TheDoubleK: Typo: `patterm`

json_schemas/framework-agreement-details.json
TheDoubleK: It's a date that becomes part of the document reference on the framework agreement. It doesn't really make any sense outside of the context of being a thing that goes into the placeholder.  It's formatted differently from dates that go in to the text as dates rather than as part of the document reference number.

TheDoubleK: So to answer your question `dateForframeworkAgreementReferenceNumber` would be more descriptive, but I'm not sure it really helps :/

json_schemas/framework-agreement-details.json
TheDoubleK: I'd prefer to keep this framework agreement data dumb - i.e. it's just a collection of the exact strings we want to patch into the agreement document template.  

It feels like overkill to start saving actual timestamps and having the script to generate agreements know where to find the various timestamps and how they should be formatted for different placeholders in the template.  And what if the format changes from one framework agreement to the next?  We still need to be able to generate the right thing for all frameworks.  Just storing the strings in a convenient blob feels like the best thing.

json_schemas/framework-agreement-details.json
TheDoubleK: I don't think it should - we don't want an RFC compliant date.  We want the exact string as it should appear in the framework agreement document.  Arguably we shouldn't enforce any particular formatting here - just a non-empty string.

TheDoubleK: Actually yes, I'd argue that we don't want to specify any regex or format for any of these strings.  These are just strings to go into placeholders, we don't care in the DB what format they are.

risicle: Valid opinion. Thoughts @allait ?

galund: FWIW, I think the highest priority would be to avoid duplication (opportunity for inconsistency) of data. Ideally, all these dates would be in some non-ambiguous machine-readable format, formatted for display on the front end, and 'human readable date strings' in the framework data would _not_ be a thing.

TheDoubleK: * There is no duplication here at all - we don't store these dates anywhere else in the database. (We do store the "applications closed" date, but that is none of the dates here.)

* The data here (i.e. in the frameworkAgreementDetails block) is *not* intended to be used anywhere at all on the frontend.  It is for use by the scripts that generate the signature page PDFs.

* The main requirement of those PDFs is that they *exactly* match the page in the full framework agreement document, so storing the *exact* strings that need to go in there makes sense.

* I'm not sure in what sense these are ambiguous at all?

TheDoubleK:
I think it's worth explaining the exact intent of these changes.

We have two scripts that we use to generate framework agreement pages (unsigned and countersigned versions).

Currently these scripts take a bunch of data at the command line from a local JSON file - data that is needed to pass to the template for the agreement signature page with the right framework name, document reference number, etc. - e.g. see the docstring here for details:
See here: https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/generate-framework-agreement-signature-pages.py

That was cobbled together so we could quickly make the DOS2 agreement pages, but there is no "source of truth" about what those values need to be for making the documents.  So this is intended to have the required details for the agreement in the database rather than in a file on my machine that I put together with Tobi at the last minute.

I strongly feel that these "generate agreements" scrips should not have to know about how to format dates for different fields.  They should just get the exact strings they need to patch into the agreement page template.  Doing things this way also means we can concurrently support different formats for reference numbers, dates, etc. on the page: we just need to make sure the per-framework details are right and then we can mix it up between frameworks as much as we like.  (I'm not saying we should be mixing it up, but you know the whims of lawyers, designers, etc...)

allait:
@TheDoubleK yeah, I think that makes sense. I misread the PR title and was under the impression that we're storing a copy of this on each `frameworkAgrement` before I realised this is a field on the Framework model.

I agree with @galund that consistency is a concern. We've seen the unpleasant effects of small variations in framework data when it's not coming from a single source (like the lot numbering in the agreement letters for example) and I think that's something we need to keep in mind.

Ideally, we'd have these dates stored on a framework and then the API could add the keys to the agreement details to present the same data in a different format. But that's something we can do in the future once we have an actual use for these dates somewhere else and understand a bit more how these formats could change in the agreements.

So if @TheDoubleK is happy with this it LGTM.

TheDoubleK:
Yep works for me :+1:

TheDoubleK:
I've put a bug in our Icebox about the camelCase keys: https://www.pivotaltracker.com/story/show/141950783

risicle:
I'll remove the `parrern` restrictions on the date fields before merging.

PR #559, opened 2017-04-04
33-update-date-format-for-supplier-s-start-date

benvand:
## Update JSON from YAML

* Use `digitalmarketplace-frameworks/scripts.generate-schemas.py` to create the below updates.

https://trello.com/c/MzEa87hS/33-update-date-format-for-supplier-s-start-date

### DIFF
--- a/json_schemas/briefs-digital-outcomes-and-specialists-digital-specialists.json
+++ b/json_schemas/briefs-digital-outcomes-and-specialists-digital-specialists.json
@@ -154,8 +154,7 @@
       "type": "string"
     },
     "startDate": {
-      "maxLength": 100,
-      "minLength": 1,
+      "format": "date",
       "type": "string"
     },
     "summary": {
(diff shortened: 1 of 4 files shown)


idavidmcdonald:
We are going to run a migration that will delete start dates for any draft briefs.

However @benvand, we should have a conversation about roll out of this.
We will need to 
- deploy the buyer front end app
- deploy this PR for the API with the new schemas
- deploy the  migration

There was previously discussion about allowing people to not publish for 30 minutes between deploys so we don't have to deal with backwards compatibility but we need to have a think about if this is possible/acceptable given that we might have to wait for 2 deploys, give it a test run and make sure the app doesn't completely fall over midway.

json_schemas/briefs-digital-outcomes-and-specialists-digital-specialists.json
```diff
@@ -154,8 +154,7 @@
       "type": "string"
     },
     "startDate": {
-      "maxLength": 100,
-      "minLength": 1,
+      "format": "date",
```
idavidmcdonald: Again, if we don't change the DOS1 version in the frameworks repo then this doesn't need to change.

PR #561, opened 2017-04-07
40 delete start dates on draft requirements

benvand:
## 40-delete-start-dates-on-draft-requirements

Single migration to delete dos2 draft brief start dates that will no
longer be in the correct format.

Context: Change of date format means draft requirement dates will not
longer be in the right format to be published.

Delete start dates on all draft requirements so buyers can re-enter
dates before they are published.

Acceptance criteria
Given that I have entered a latest start date in my draft requirement,
when I try to publish my requirement, then I need to re-enter the latest
start date.

https://trello.com/c/BA5KSAvm/40-delete-start-dates-on-draft-requirements


idavidmcdonald:
👍

PR #563, opened 2017-04-10
33-update-date-format-for-supplier-s-start-date

benvand:
## Update JSON from YAML

* Use digitalmarketplace-frameworks/scripts.generate-schemas.py to create
the below updates.

### DIFF
--- a/json_schemas/briefs-digital-outcomes-and-specialists-2-digital-outcomes.json
+++ b/json_schemas/briefs-digital-outcomes-and-specialists-2-digital-outcomes.json
@@ -154,8 +154,7 @@
       "type": "string"
     },
     "startDate": {
-      "maxLength": 100,
-      "minLength": 1,
+      "format": "date",
       "type": "string"
     },
     "successCriteria": {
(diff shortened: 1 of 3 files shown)


benvand:
## NEW BRANCH AND PR. BELOW ARE OLD COMMENTS

benvand:
>>> @TheDoubleK 
How will this be rolled out? Existing drafts with start dates already set will fail to validate against this new schema and so won't ever be able to be published.

We'd need an either/or schema to allow old or new formats.

And as drafts can potentially hang around for years before they get published I'm not sure how we'll ever manage to deprecate the old format?

benvand:
@idavidmcdonald 
> We are going to run a migration that will delete start dates for any draft briefs.
However @benvand, we should have a conversation about roll out of this.
We will need to 
- deploy the buyer front end app
- deploy this PR for the API with the new schemas
- deploy the  migration
There was previously discussion about allowing people to not publish for 30 minutes between deploys so we don't have to deal with backwards compatibility but we need to have a think about if this is possible/acceptable given that we might have to wait for 2 deploys, give it a test run and make sure the app doesn't completely fall over midway.

benvand:
> @idavidmcdonald 
Looks good. I'll thumbs up when the rest is good and we've confirmed our deployment plan.

PR #564, opened 2017-04-10
[#139472449]export_users_for_framework: lazyload a lot of relationships we don't need

risicle:
This *might* fix story https://www.pivotaltracker.com/story/show/139472449

At some point we need to look at how much we overuse eager loading of relationships. It gets more than a bit silly.

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 from dmapiclient.audit import AuditTypes
 from sqlalchemy import func
+from sqlalchemy.orm import lazyload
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Confirmed that outputs are identical to previously.

Also did some rudimentary profiling locally ands this seems to have made it a *lot* faster ⏩ 
Average times before and after, in some unspecified units of time:
G7  286908 -> 023105
G8  071800 -> 024538
G9  076668 -> 021988
DOS 274108 -> 022142
:boom:

Let's get this in and release it ASAP :clap:

:+1:

PR #566, opened 2017-04-27
Add 4 new buyer domains

CrystalPea:
Trello card: https://trello.com/c/Ag8OTGAk/386-add-additional-buyer-email-domains-for-buyer-accounts


idavidmcdonald:
👍 
Nice Pea. Will need a rebase but then should be good!

PR #567, opened 2017-04-27
Remove start dates for the second time from draft dos2 briefs

idavidmcdonald:
This is identical to the previous migration but will be run again
to cover any draft briefs with invalid dates that could have
appeared during the previous API rollout process (after the previous
migration but before the code propogated fully to the ec2 instances).

https://trello.com/c/BA5KSAvm/40-delete-start-dates-on-draft-requirements


benvand:
👍

PR #568, opened 2017-05-03
copy-my-requirements-migration

benvand:
We need to store whether or not a brief is a clone of another brief, and if so, which brief.
This has been done here with a nullable foreign key to the briefs table indicating 'copied_from_brief_id.

*  Add field `copied_from_brief_id ` to `Briefs` model and migrate


idavidmcdonald:
Your PR message is ouddated (you should update) but when tests green this looks good 👍

PR #569, opened 2017-05-03
copy-my-requirements

benvand:
* Update `apps/models.py::Brief::copy`
 * Add copy string to title
 * Specify fields from data to _not_ copy (see ticket)
* Remove `test_cannot_make_a_draft_copy_of_a_brief_if_the_framework_is_not_live`
 * No longer relevant. Framework does not need to be live
* Update tests

The model method here is triggered by the api end point at `digitalmarketplace-api/app/main/views/briefs.py::copy_brief` which didn't need changing.

Migration:
https://github.com/alphagov/digitalmarketplace-api/pull/568

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1289,15 +1289,11 @@ class Brief(db.Model):
     def copy(self):
         data = self.data.copy()
         title = data.get('title', '')
-        if len(title) <= 95:
+        if 0 < len(title) <= 95:
             data['title'] = u"{} copy".format(title)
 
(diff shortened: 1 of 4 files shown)


idavidmcdonald:
Looks good when tests are green 👍

idavidmcdonald:
Actually, one issue spotted. To use the `copied_from_brief_id` in our scripts repo it must be sent as part of the serialized response. We must add it to the serialize method (at which point camel case also is convention).

PR #571, opened 2017-05-05
Remove `copied_from_brief_id` from model so migration is separate

idavidmcdonald:
The model change will be added to a later PR instead so the code change and migration are separate.


benvand:
:+1: 
lgtm

PR #585, opened 2017-06-01
1. Add Briefs.is_a_copy to replace Briefs.copied_from_brief_id

benvand:


TheDoubleK:
This has the migration and associated model change in the same PR.  Should probably split these into two - and we run the migration first, update the model second.

migrations/versions/900_add_brief_is_a_copy.py
```diff
@@ -0,0 +1,22 @@
...
+# revision identifiers, used by Alembic.
+revision = '900'
+down_revision = '890'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('briefs', sa.Column('is_a_copy', sa.Boolean(), server_default=sa.text(u'false'), nullable=False))
```
TheDoubleK: Why have we got `sa.text(u'false')` here? In e.g. migration #840 we just have `server_default='false'`.

It runs fine as it is (at least looks like it as far as I can tell), but just wondering if there's a reason for the `sa.text()` wrapper?

To be honest both of these seem weird because they're using string/text-types rather than a Boolean, but it clearly works ¯\\_(ツ)_/¯

benvand: This was produced by running the migrate tool to auto create the migration. I was tempted to change it to sa.false() in here but I just thought 'Alembic must know best'. :man_shrugging:

TheDoubleK: Ah OK, yep let's go with it - it works OK.

PR #587, opened 2017-06-01
2. Fix code for new Brief.is_a_copy field replacing Brief.copied_from_brief_id

benvand:


app/models.py
```diff
@@ -1358,8 +1358,8 @@ def serialize(self, with_users=False):
                 'withdrawnAt': self.withdrawn_at.strftime(DATETIME_FORMAT)
             })
 
-        if self.copied_from_brief_id:
-            data.update({'copiedFromBriefId': self.copied_from_brief_id})
+        if self.is_a_copy:
```
TheDoubleK: Do we want this to be conditional?  Seems like having the key always there with a true/false value would be fine?

TheDoubleK:
Yep I think the model change from https://github.com/alphagov/digitalmarketplace-api/pull/585 should be in here instead.

app/models.py
benvand: Good shout

tests/test_models.py
```diff
@@ -437,7 +437,7 @@ def test_copy_brief(self, live_dos_framework):
         assert copy.framework == self.brief.framework
         assert copy.lot == self.brief.lot
         assert copy.users == self.brief.users
-        assert copy.copied_from_brief_id == self.brief.id
+        assert copy.is_a_copy is True
```
TheDoubleK: Do we also want a test somewhere that `is_a_copy is False` for a non-copied brief?  Could probably tag an assertion into an existing test that creates a brief?

benvand: @TheDoubleK Good call, added this to create a brief test

PR #592, opened 2017-06-12
REFACTOR-use-orm-to-change-and-reset-framework-in-test

benvand:
Refactor framework changing test to use ORM and commit successfully

### DIFF
--- a/tests/main/views/test_brief_response.py
+++ b/tests/main/views/test_brief_response.py
@@ -1,15 +1,17 @@
+"""Tests for brief response views in app/views/brief_responses.py."""
+from datetime import datetime, timedelta
+from freezegun import freeze_time
 import json
 import mock
-from datetime import datetime, timedelta
-
-from freezegun import freeze_time
-
-from tests.bases import BaseApplicationTest, JSONUpdateTestMixin
-from tests.helpers import FixtureMixin
-from ... import example_listings
+import pytest
 
 from dmapiclient.audit import AuditTypes
+
 from app.models import db, Lot, Brief, BriefResponse, AuditEvent, Service, Framework
+from tests.bases import BaseApplicationTest, JSONUpdateTestMixin
+from tests.helpers import FixtureMixin
+
+from ... import example_listings
 
 
 class BaseBriefResponseTest(BaseApplicationTest, FixtureMixin):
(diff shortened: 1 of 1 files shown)


tests/main/views/test_brief_response.py
```diff
@@ -550,32 +552,44 @@ def test_can_not_submit_a_brief_response_that_already_been_submitted(self, live_
...
-                    )
-
-                    res = self._submit_brief_response(self.brief_response_id)
-                    data = json.loads(res.get_data(as_text=True))
-                    assert res.status_code == 400
-                    assert data == {'error': 'Brief framework must be live or expired'}
+    @pytest.mark.parametrize('framework_status', ('coming', 'open', 'pending', 'standstill'))
+    def test_can_not_submit_a_brief_response_for_a_framework_that_is_not_live_or_expired(
+            self,
+            live_dos_framework,
```
Wynndow: There's a `live_dos2_framework` fixture available if you wanted to be more current? Although it seems to be exactly the same apart from the `-2` at the end of the slug so probably not worth it.

tests/main/views/test_brief_response.py
```diff
@@ -550,32 +552,44 @@ def test_can_not_submit_a_brief_response_that_already_been_submitted(self, live_
...
+            if existing_brief_response:
+                db.session.delete(existing_brief_response[-1])
+                db.session.commit()
+
+            # Get old framework status so we can reset it later.
+            dos_framework = db.session.query(Framework).filter_by(slug='digital-outcomes-and-specialists').first()
+            original_status = dos_framework.status
+
+            # Make framework live so a brief response can be created.
+            dos_framework.status = 'live'
```
Wynndow: I think you might not need to reset the framework status for each iteration of the test with the code here and on lines 570 and 591-592.
I've dropped a debugger in and the `dos_framework` retrieved on line 569 has a framework status of `live` for each iteration, without having the code to reset it back to its original status.
I'm assuming it's loading the fixture fresh for each iteration of parameters?

tests/main/views/test_brief_response.py
```diff
@@ -550,32 +552,44 @@ def test_can_not_submit_a_brief_response_that_already_been_submitted(self, live_
         data = json.loads(repeat_res.get_data(as_text=True))
         assert data == {'error': 'Brief response must be a draft'}
 
-    def test_can_not_submit_a_brief_response_for_a_framework_that_is_not_live_or_expired(self, live_dos_framework):
-            with self.app.app_context():
-                for framework_status in [status for status in Framework.STATUSES if status not in ('live', 'expired')]:
```
idavidmcdonald: I think we set up this framework_statuses list comprehension so if we ever added a new framework status then we wouldn't need to manually add it to our list which I thought was kind of nice... It's not a game changer though...

tests/main/views/test_brief_response.py
idavidmcdonald: Let's do it. No harm I think...

benvand: So much harm.

Been trying to get it to work for over an hour and this whole file is a quagmire of hard coded ids. I'm deeming it as out of scope as part of this fix. I'll put a ticket on the tech debt Trello Sorry :(

tests/main/views/test_brief_response.py
benvand: Good call

tests/main/views/test_brief_response.py
benvand: Fixed

idavidmcdonald:
👍

PR #593, opened 2017-06-12
Add brief to brief response serialisation

benvand:
Add brief to brief response serialisation and update tests to check for brief in brief response serialisation.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1167,7 +1167,9 @@ class Brief(db.Model):
     lot = db.relationship('Lot', lazy='joined')
     clarification_questions = db.relationship(
         "BriefClarificationQuestion",
-        order_by="BriefClarificationQuestion.published_at")
+        order_by="BriefClarificationQuestion.published_at",
+        lazy='joined'
+    )
 
     @validates('users')
     def validates_users(self, key, user):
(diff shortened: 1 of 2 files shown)


app/models.py
```diff
@@ -1462,10 +1464,12 @@ def validate(self, enforce_required=True, required_fields=None, max_day_rate=Non
 
     def serialize(self):
         data = self.data.copy()
+        parent_brief = self.brief.serialize()
+        parent_brief_fields = ['id', 'title', 'status', 'applicationsClosedAt']
```
idavidmcdonald: Why do we restrict ourselves to these fields? We are already running `brief.serialize` above it and so there will be no further expense to add everything? (Not that saying, yes let's add everything but just wondering if there is a particular reason for it being good to not add more).

Also we won't run in any problems with it for this case, but if we at a later point added brief responses to the `brief.serialize` method then we would run into an infinite loop. Kind of links in to a point that I think George made about the ability to serialize an object without including any of it's related data objects so we can make sure serialisation doesn't propagate out too many other serialises.

TheDoubleK: Out of interest, why are we adding `lazy= joined`? This eager loading has caused some of our performance issues elsewhere, so unless it's essential it might be better to lazy load if possible.

allait: @TheDoubleK because we're always returning clarification questions when brief is serialized, which means that without JOIN we end up with N+1 (101) DB queries for each `/briefs` response from the API.

I guess the question is whether we need clarification questions in the list of briefs response at all.

benvand: > Why do we restrict ourselves to these fields? 
If we want to refactor in the future it'll be easier to remove just a few fields than the whole serialisation :man_shrugging: 


> I guess the question is whether we need clarification questions in the list of briefs response at all
Probably not but that's out of scope of this ticket.

benvand: @TheDoubleK
Actually wrote this up the other day 
https://docs.google.com/document/d/1mm2pEwHC18zkbJ3nec2jz94SjPeu7lv9Mo3o8f62Q5E/edit

idavidmcdonald: Would be nice to add a summary of that as a commit message so that reasoning exists in our git history please

benvand: Done that

PR #594, opened 2017-06-16
Finally purge the ghost of minglis

TheDoubleK:
The API explorer was a nice idea but in reality has never ever really been used by devs to explore the API.  It hasn't been updated in over two years and doesn't actually work any more (page errors with `KeyError: 'BASE_TEMPLATE_DATA'` trying to visit the page locally).

On the day Martyn leaves GDS (for six months at least), let's get rid of this cruft.


pcraig3:
Ha! 👍 

Not related to this pull request, but Martyn is also the one who added the Bower repo nobody knew about.

PR #595, opened 2017-06-19
Remove Flask-Bootstrap

TheDoubleK:
This was used by the API explorer, but that's gone now so this can be removed.


pcraig3:
Ah, good call.

PR #596, opened 2017-06-19
[#147424377]list_audits: add earliest_for_each_object option

risicle:
Story https://www.pivotaltracker.com/story/show/147424377

If we want https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/269 to be safely performant, we're going to have to do the grouping/`DISTINCT ON` server-side.

I've called this mode `earliest_for_each_object`, I'm not totally set on the name, but think the way I've added the feature is sensibly generic.

Also allowed the endpoint to filter by `object_type` without requiring `object_id`. This will be needed in forthcoming stories.

Added `AuditEvent.id` as minor ordering tie-breaker to ensure deterministic ordering and ensure `latest_first=true` returns the exact reverse result of `latest_first=false`.

I haven't split out the migrations as a separate PR as they are both db changes which should be mostly transparent to the app. But if we wanted to be obsessive about that I could.

### DIFF
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -4,6 +4,7 @@ from ...models import AuditEvent
 from sqlalchemy import asc, desc, Date, cast
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.sql.expression import true, false
+from sqlalchemy.orm import class_mapper
 from ...utils import pagination_links, get_valid_page_or_1
 from .. import main
 from ... import db, models
@@ -84,20 +85,23 @@ def list_audits():
     if object_type:
         if object_type not in AUDIT_OBJECT_TYPES:
             abort(400, 'invalid object-type supplied')
-        if not object_id:
-            abort(400, 'object-type cannot be provided without object-id')
+
         model = AUDIT_OBJECT_TYPES[object_type]
         id_field = AUDIT_OBJECT_ID_FIELDS[object_type]
 
-        ref_object = model.query.filter(
-            id_field == object_id
-        ).first()
+        audits = audits.filter(AuditEvent.object.is_type(model))
 
-        if ref_object is None:
-            abort(404, "Object with given object-type and object-id doesn't exist")
+        if object_id:
+            ref_object = model.query.filter(
+                id_field == object_id
+            ).first()
 
-        audits = audits.filter(AuditEvent.object == ref_object)
+            if ref_object is None:
+                abort(404, "Object with given object-type and object-id doesn't exist")
 
+            audits = audits.filter(
+                AuditEvent.object_id == class_mapper(model).identity_key_from_instance(ref_object)[1][0]
+            )
     elif object_id:
         abort(400, 'object-id cannot be provided without object-type')
 
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -43,6 +43,9 @@ def list_audits():
     earliest_for_each_object = convert_to_boolean(request.args.get('earliest_for_each_object'))
 
     if earliest_for_each_object:
+        # the rest of the filters we add will be added against a subquery which we will join back onto the main table
+        # to retrieve the rest of the row. this allows the potentially expensive DISTINCT ON pass to be performed
+        # against an absolutely minimal subset of rows which can probably be pulled straight from an index
         audits = db.session.query(AuditEvent.id)
     else:
         audits = AuditEvent.query
@@ -113,6 +116,7 @@ def list_audits():
         abort(400, 'object-id cannot be provided without object-type')
 
     if earliest_for_each_object:
+        # we need to join the built-up subquery back onto the AuditEvent table to retrieve the rest of the row
         audits_subquery = audits.order_by(
             AuditEvent.object_type,
             AuditEvent.object_id,
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -39,11 +39,12 @@ def list_audits():
     except ValueError:
         abort(400, 'invalid page size supplied')
 
-    audits = AuditEvent.query.order_by(
-        desc(AuditEvent.created_at)
-        if convert_to_boolean(request.args.get('latest_first'))
-        else asc(AuditEvent.created_at)
-    )
+    earliest_for_each_object = convert_to_boolean(request.args.get('earliest_for_each_object'))
+
+    if earliest_for_each_object:
+        audits = db.session.query(AuditEvent.id)
+    else:
+        audits = AuditEvent.query
 
     audit_date = request.args.get('audit-date', None)
     if audit_date:
@@ -99,6 +100,25 @@ def list_audits():
 
     elif object_id:
         abort(400, 'object-id cannot be provided without object-type')
+
+    if earliest_for_each_object:
+        audits_subquery = audits.order_by(
+            AuditEvent.object_type,
+            AuditEvent.object_id,
+            AuditEvent.created_at,
+            AuditEvent.id,
+        ).distinct(
+            AuditEvent.object_type,
+            AuditEvent.object_id,
+        ).subquery()
+
+        audits = AuditEvent.query.join(audits_subquery, audits_subquery.c.id == AuditEvent.id)
+
+    audits = audits.order_by(
+        desc(AuditEvent.created_at)
+        if convert_to_boolean(request.args.get('latest_first'))
+        else asc(AuditEvent.created_at)
+    )
 
     audits = audits.paginate(
         page=page,
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -116,6 +116,17 @@ def list_audits():
         abort(400, 'object-id cannot be provided without object-type')
 
     if earliest_for_each_object:
+        if not (
+                acknowledged and
+                acknowledged != 'all' and
+                audit_type == "update_service" and
+                object_type == "services"
+                ):
+            current_app.logger.warning(
+                "earliest_for_each_object option currently intended for use on acknowledged update_service events. "
+                "If use with any other events is to be regular, the scope of the corresponding partial index "
+                "should be expanded to cover it."
+            )
         # we need to join the built-up subquery back onto the AuditEvent table to retrieve the rest of the row
         audits_subquery = audits.order_by(
             AuditEvent.object_type,
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -86,21 +86,28 @@ def list_audits():
         if object_type not in AUDIT_OBJECT_TYPES:
             abort(400, 'invalid object-type supplied')
 
-        model = AUDIT_OBJECT_TYPES[object_type]
-        id_field = AUDIT_OBJECT_ID_FIELDS[object_type]
+        ref_model = AUDIT_OBJECT_TYPES[object_type]
+        ext_id_field = AUDIT_OBJECT_ID_FIELDS[object_type]
 
-        audits = audits.filter(AuditEvent.object.is_type(model))
+        audits = audits.filter(AuditEvent.object.is_type(ref_model))
 
+        # "object_id" here is the *external* object_id
         if object_id:
-            ref_object = model.query.filter(
-                id_field == object_id
+            ref_object = ref_model.query.filter(
+                ext_id_field == object_id
             ).first()
 
             if ref_object is None:
                 abort(404, "Object with given object-type and object-id doesn't exist")
 
+            # this `.identity_key_from_instance(...)[1][0]` is exactly the method used by sqlalchemy_utils' generic
+            # relationship code to extract an object's pk value, so *should* be relatively stable, API-wise.
+            # the `[1]` is to select the pk's *value* rather than the `Column` object and the `[0]` simply fetches
+            # the first of any pk values - generic relationships are already assuming that compound pks aren't in
+            # use by the target.
+            ref_object_pk = class_mapper(ref_model).identity_key_from_instance(ref_object)[1][0]
             audits = audits.filter(
-                AuditEvent.object_id == class_mapper(model).identity_key_from_instance(ref_object)[1][0]
+                AuditEvent.object_id == ref_object_pk
             )
     elif object_id:
         abort(400, 'object-id cannot be provided without object-type')
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -118,11 +118,8 @@ def list_audits():
 
         audits = AuditEvent.query.join(audits_subquery, audits_subquery.c.id == AuditEvent.id)
 
-    audits = audits.order_by(
-        desc(AuditEvent.created_at)
-        if convert_to_boolean(request.args.get('latest_first'))
-        else asc(AuditEvent.created_at)
-    )
+    sort_order = db.desc if convert_to_boolean(request.args.get('latest_first')) else db.asc
+    audits = audits.order_by(sort_order(AuditEvent.created_at), sort_order(AuditEvent.id))
 
     audits = audits.paginate(
         page=page,
--- a/app/models.py
+++ b/app/models.py
@@ -14,7 +14,14 @@ from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import validates, backref, mapper
 from sqlalchemy.orm.session import Session
-from sqlalchemy.sql.expression import case as sql_case, cast as sql_cast, select as sql_select, false as sql_false
+from sqlalchemy.sql.expression import (
+    case as sql_case,
+    cast as sql_cast,
+    select as sql_select,
+    false as sql_false,
+    true as sql_true,
+    and_ as sql_and,
+)
 from sqlalchemy.types import String
 from sqlalchemy import Sequence
 from sqlalchemy_utils import generic_relationship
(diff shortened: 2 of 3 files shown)


app/main/views/audits.py
```diff
@@ -83,23 +90,43 @@ def list_audits():
...
+            AuditEvent.id,
+        ).distinct(
+            AuditEvent.object_type,
+            AuditEvent.object_id,
+        ).subquery()
+
+        audits = AuditEvent.query.join(audits_subquery, audits_subquery.c.id == AuditEvent.id)
+
+    latest_first = convert_to_boolean(request.args.get('latest_first'))
+    asc_, desc_, nullslast_, nullsfirst_ = reversible_order_by_terms(latest_first)
```
galund: These (target) variables must be wrongly named, as sometimes they receive the opposite thing to what they say they are. If this is a good pattern, then it needs to be clear from the names, somehow, what is actually going on.

risicle: Yes, I couldn't think of better names for them that were still short. I think they're best described as *notionally asc* or *notionally desc*. I mean, ideally what I'd like is to be able to do a regular `order_by` with all its intricacies, but then to be able to `.reverse_order_by()` or something. `pseudo_asc`?

risicle:
Oh hold on I think I might have screwed up the imports of the migrations...

migrations/versions/930_audit_event_data_field_non_nullable.py
```diff
@@ -0,0 +1,32 @@
...
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+def upgrade():
+    op.alter_column(
+        'audit_events',
+        'data',
+        existing_type=postgresql.JSON(astext_type=sa.Text()),
+        nullable=False,
```
allait: So it looks like the `nullable` flag was set on the column 6 months ago but the migration hasn't been generated until now for some reason.

Are we sure that there are no `data=NULL` or `data='null'` events in production DB?

In general, I don't see a reason why data would be required for an audit event, but if the migration will apply without issues then I guess we might as well roll it out and ask everyone to do `data={}` when they don't have anything to store.

tests/main/views/test_audits.py
```diff
@@ -50,8 +54,204 @@ def add_audit_events_with_db_object(self):
...
+        # supplier_audit_event_params events. so if we had 5 service_audit_event_params and 2
+        # supplier_audit_event_params, "5" would refer to the audit event created by the first-listed
+        # supplier_audit_event_params.
+        # similarly, where supplier and service "id"s are referred to, the "id"s we're referring to are normalized
+        # pseudo-ids from 0-4 inclusive
+        chain.from_iterable(
+            ((serv_aeps, supp_aeps, req_params, expected_resp_events) for req_params, expected_resp_events in req_cases)
+            for serv_aeps, supp_aeps, req_cases in (
+                (
+                    (  # service_audit_event_params
```
allait: This would probably be easier to understand if it listed what the parameters are. Or maybe even better, if it used dictionaries instead of tuples.

tests/main/views/test_audits.py
```diff
@@ -50,8 +54,204 @@ def add_audit_events_with_db_object(self):
...
+                    ),
+                    (  # supplier_audit_event_params
+                        (0, AuditTypes.supplier_update, datetime(2010, 6, 6), None,),
+                    ),
+                    (  # and now a series of req_cases - pairs of (req_params, expected_resp_events) to test against
+                       # the above db scenario. these get flattened out into concrete test scenarios by
+                       # chain.from_iterable above before they reach pytest's parametrization
+                        (
+                            {"earliest_for_each_object": "true", "latest_first": "true"},
+                            (3, 0, 2,),
```
allait: This is pretty minor, but I think using tuples and lists semantically here (as opposed to always using tuples) would help visually parse this structure. Eg `(serv_aeps, supp_aeps, req_params, expected_resp_events)` should be a tuple, since each positional element has a different meaning, but `[3, 0, 2]` is a list of ordered ids.

I'm guessing you're doing this to avoid accidental mutation, but the code that's using these values in the `assert` is pretty minimal, and we're only using them once so it doesn't feel like a risk at all.

TheDoubleK: Why do we expect `3` before `0` here - the timestamps are identical?

tests/main/views/test_audits.py
```diff
@@ -50,8 +54,204 @@ def add_audit_events_with_db_object(self):
...
+                        ),
+                        (
+                            {"earliest_for_each_object": "true", "latest_first": "false"},
+                            (2, 0, 3,),
```
TheDoubleK: Similarly, why `0` before `3` - are we just relying on the order of insertion?  That doesn't seem a useful test?

tests/main/views/test_audits.py
```diff
@@ -50,8 +54,204 @@ def add_audit_events_with_db_object(self):
...
+            assert AuditEvent.query.count() == len(service_audit_event_params)+len(supplier_audit_event_params)
+
+            return audit_event_id_lookup
+
 
 class TestAuditEvents(BaseTestAuditEvents):
+    @pytest.mark.parametrize(
+        "service_audit_event_params,supplier_audit_event_params,req_params,expected_resp_events",
+        # where we refer to "id"s in the expected_response_params, because we can't be too sure about the *actual* ids
+        # given to objects, we're using 0-based notional "ids" mased on the order the audit events were inserted into
```
TheDoubleK: Typo: `mased`

tests/main/views/test_audits.py
```diff
@@ -50,8 +54,204 @@ def add_audit_events_with_db_object(self):
...
+                    ),
+                    (  # supplier_audit_event_params
+                        (1, AuditTypes.supplier_update, datetime(2010, 2, 6), None,),
+                        (0, AuditTypes.supplier_update, datetime(2010, 6, 6), None,),
+                        (1, AuditTypes.supplier_update, datetime(2010, 2, 6), None,),
+                    ),
+                    (  # series of req_cases for the above db scenario
+                        (
+                            {"earliest_for_each_object": "true"},
+                            (6, 7, 4, 3, 5,),
```
TheDoubleK: Again, does the appearance of `6` rather than `8` for supplier_updates to supplier #1 just depend on order of insertion?

tests/main/views/test_audits.py
```diff
@@ -50,8 +54,204 @@ def add_audit_events_with_db_object(self):
...
+                        (1, AuditTypes.supplier_update, datetime(2010, 2, 1), None,),
+                        (1, AuditTypes.supplier_update, datetime(2010, 2, 6), None,),
+                        (0, AuditTypes.supplier_update, datetime(2010, 2, 7), None,),
+                        (3, AuditTypes.supplier_update, datetime(2009, 1, 1), None,),
+                        (0, AuditTypes.supplier_update, datetime(2010, 2, 3), None,),
+                        (1, AuditTypes.supplier_update, datetime(2010, 4, 9), None,),
+                        (2, AuditTypes.supplier_update, datetime(2010, 1, 1), None,),
+                        (1, AuditTypes.supplier_update, datetime(2010, 8, 8), None,),
+                    ),
+                    (  # series of req_cases for the above db scenario
```
TheDoubleK: Might be worth having a test for `{"acknowledged": "false", "object-type": "services"}` and check for an empty list in the response?

app/models.py
```diff
@@ -1551,17 +1558,26 @@ def serialize(self):
...
+# partial index for supplier_update-monitoring admin views. specifically this
+# ordering should allow the index to be used for earliest_for_each_object's
+# ordering-followed-by-DISTINCT ON
+db.Index(
+    'idx_audit_events_created_at_per_obj_partial',
+    AuditEvent.object_type,
+    AuditEvent.object_id,
+    AuditEvent.created_at,
+    AuditEvent.id,
+    postgresql_where=sql_and(AuditEvent.acknowledged == sql_false(), AuditEvent.type == "supplier_update"),
```
TheDoubleK: And this should correspondingly be `AuditEvent.type == 'update_service'`.

app/main/views/audits.py
```diff
@@ -83,23 +90,43 @@ def list_audits():
     if object_type:
         if object_type not in AUDIT_OBJECT_TYPES:
             abort(400, 'invalid object-type supplied')
-        if not object_id:
-            abort(400, 'object-type cannot be provided without object-id')
```
risicle: The check isn't being removed - the *ability* is being *added*. There *is* a reason for it. And I should have written down that reason when I did it, because I can't *quite* remember it now...

risicle:
Ok, I think some miscommunication happened here - I was under the impression that a single service being able to effectively disable the listing page would be considered a Bad Thing - limiting the previous approach to 100 updates would effectively do that, making a page that was subjected to that *only* able to see that service. But... acceptable failure modes are always debatable...

Anyway, this is done now and applyable or not applyable.

app/models.py
```diff
@@ -1551,17 +1558,26 @@ def serialize(self):
...
-# dedicated index for objects with multiple events
+
+# DEPRECATED - remove in a migration once supplier update admin app has been updated
 db.Index(
     'idx_audit_events_type_acknowledged',
     AuditEvent.type,
     AuditEvent.acknowledged,
 )
 
+# partial index for supplier_update-monitoring admin views. specifically this
```
risicle: Pas du tout. If the feature gets expanded to another object type, we just add the new update event to the index condition. No new code needs to be written, no new tests. It's a one-liner (apart from the migration...).

And it's not like we don't have existing precedents for adding e.g. specific indexes targeted at specific things "we know" the frontend(s) do. The replaced comment here for the previous index was something about "this is for the supplier updates admin page" and it was set up to use a _specific set_ of filters that that particular frontend view used.

app/main/views/audits.py
```diff
@@ -83,23 +90,43 @@ def list_audits():
...
+            AuditEvent.id,
+        ).distinct(
+            AuditEvent.object_type,
+            AuditEvent.object_id,
+        ).subquery()
+
+        audits = AuditEvent.query.join(audits_subquery, audits_subquery.c.id == AuditEvent.id)
+
+    latest_first = convert_to_boolean(request.args.get('latest_first'))
+    pseudo_asc, pseudo_desc, pseudo_nullslast, pseudo_nullslast = reversible_order_by_terms(latest_first)
```
risicle: Muh enough people don't like this that it'll have to go but I thought it was pretty cool.

(the idea was that it returning all the required tools would prompt any later users to *not forget about nullsfirst!* if applicable and avoid things being overlooked. what I'd really like to do is give sqlalchemy a `.reverse_order_by()` which would cope properly with all cases)

app/main/views/audits.py
```diff
@@ -83,23 +90,43 @@ def list_audits():
...
+            ref_object = model.query.filter(
+                id_field == object_id
+            ).first()
 
-        audits = audits.filter(AuditEvent.object == ref_object)
+            if ref_object is None:
+                abort(404, "Object with given object-type and object-id doesn't exist")
 
+            audits = audits.filter(
+                AuditEvent.object_id == class_mapper(model).identity_key_from_instance(ref_object)[1][0]
```
risicle: Yes, this is the only way I can see of checking the object type separately from the id. `.identity_key_from_instance(ref_object)[1][0]` indeed looks horrible but is exactly what `sqlalchemy_utils` does in its generic relationship code. So, if `sqlalchemy_utils` considers the API stable enough to depend on, I figured we may as well.

There was also a slightly less scary but more verbose way of doing it I discovered at the time which I could try.

migrations/versions/930_audit_event_data_field_non_nullable.py
risicle: Oh poo I'm going to split these out.

tests/main/views/test_audits.py
risicle: dicts tend to get really, really verbose when used like this...

tests/main/views/test_audits.py
risicle: It's mostly just that tuples are my default and sticking with tuples means I don't have to remember whether this or that is a tuple or a list or ...

migrations/versions/920_add_idx_audit_events_created_at_per_obj_.py
```diff
@@ -0,0 +1,28 @@
+"""add_idx_audit_events_created_at_per_obj_partial
```
risicle: *grumble*

migrations/versions/920_add_idx_audit_events_created_at_per_obj_.py
```diff
@@ -0,0 +1,28 @@
...
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+def upgrade():
+    op.create_index(
+        'idx_audit_events_created_at_per_obj_partial',
+        'audit_events',
+        ['object_type', 'object_id', 'created_at', 'id'],
+        unique=False,
+        postgresql_where=sa.text(u"acknowledged = false AND type = 'supplier_update'"),
```
risicle: God damn I'm an idiot.

app/models.py
```diff
@@ -1551,17 +1558,26 @@ def serialize(self):
     AuditEvent.created_at,
 )
 
-# Index for type + acknowledged audit event query count - for admin
-# app "Service updates" requests: while the page of audit events
-# will rely on the created_at index scan and is relatively quick,
-# count(*) query executed by `.paginate` is very slow without a
-# dedicated index for objects with multiple events
+
+# DEPRECATED - remove in a migration once supplier update admin app has been updated
```
risicle: Supplier and service are the same word. Fact.

app/models.py
```diff
@@ -1551,17 +1558,26 @@ def serialize(self):
...
 db.Index(
     'idx_audit_events_type_acknowledged',
     AuditEvent.type,
     AuditEvent.acknowledged,
 )
 
+# partial index for supplier_update-monitoring admin views. specifically this
+# ordering should allow the index to be used for earliest_for_each_object's
```
risicle: Yes I would love to do this.

app/models.py
risicle: I'm going to rebase this out so it's like it never happened and you'll have a hard time proving that it did.

app/main/views/audits.py
risicle: Yup will comment.

app/main/views/audits.py
risicle: * No, it's my damn autocomplete being overzealous
* This is all so gone you won't remember it next week.

app/utils.py
```diff
@@ -154,3 +156,13 @@ def purge_nulls_from_data(data):
 def get_request_page_questions():
     json_payload = get_json_from_request()
     return json_payload.get('page_questions', [])
+
+
+def reversible_order_by_terms(reversed_):
```
risicle: Yeah yeah yeah *I* thought it was kinda cool.

tests/main/views/test_audits.py
risicle: ok

tests/main/views/test_audits.py
risicle: Because we've got a minor `order_by` of `id`, and we're asking for it in reverse (`latest_first=true`).

tests/main/views/test_audits.py
risicle: Well... it's the id. Having a deterministic output becomes particularly useful if you want a query that, when you ask for the reverse ordering, you actually get something that's the reverse ordering, not *mostly* the reverse ordering (Imagine you're wanting to be able to ask for e.g. "the events after number 384827, which I've already seen in the last page I fetched"). Generally people tend to use an API in a way that expects the ordering to be at least stable. If you want I'll dig through the app and find some places where non-deterministic ordering could produce subtly unexpected results.

And if we agree that it's a property that we desire, then it's something that should be tested for surely...?

tests/main/views/test_audits.py
risicle: More specifically `id`, which *most of the time* is based on insertion order.

app/main/views/audits.py
allait: As far as I remember one of the reasons it was disallowed was that these requests ended up generating an expensive query plan due to not using the index. And filtering by audit event type seemed like a reasonable alternative for things you'd look for with a bare object-type filter. Do we need this combination of filters for something now?

tests/main/views/test_audits.py
risicle: Oh - I was avoiding this because I was under the impression that it would 404 in such a case, but I think I must have been temporarily mad, because it doesn't 404 at all.

Maybe something else was doing it.

Yes, I will.

app/models.py
allait: Is there a problem with using a full index for this?

I still feel that we should either make it general right away or make it explicit that it isn't at the moment by returning a 400 when `earliest_for_each_object` is used with a different event type.

app/main/views/audits.py
risicle: It was specifically an index-related thing. This is where I would have to actually check the use of this index but will need to play with my db to do so.

app/models.py
risicle: It's just that it's faster.

> by returning a 400

If you really wanna, I will...

tests/main/views/test_audits.py
TheDoubleK: Yep OK, fair point.

risicle:
Leaving the migrations in this PR for now for ease of mangling, will split at last minute.

I'm going to check this is actually able to use the added index now...

TheDoubleK:
Would be good to split out the migrations into a separate PR and get this up-to date to merge.  All looks good to go I think.

risicle:
Yeah need to do that

TheDoubleK:
If you push up the new version with migrations removed we should be able to get this in now.

risicle:
A new version without the migrations wouldn't pass the tests - I was going to wait for the migrations to get merged then rebase this.

TheDoubleK:
Ah yep, makes sense.

TheDoubleK:
The warning for use without the correct index feels like the right thing to me.

:+1: once Travis is happy.

PR #599, opened 2017-06-20
Speed up get_framework_interest endpoint

TheDoubleK:
The script to countersign G-Cloud 9 agreements is failing with a 502 from production API on `/frameworks/g-cloud-9/interest`.

This patch will hopefully help by speeding up the API request.

Local testing looks promising... timing 600 requests using the API client `get_interested_suppliers` method (100 each for G6, G7, G8, G9, DOS and DOS2) results were as follows:
Unpatched (current master branch): 01:57.958328
Patched (this branch):  0:00:30.681378

i.e. only a quarter of the time taken for the same 600 requests.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -276,6 +276,11 @@ def get_framework_interest(framework_slug):
 
     supplier_frameworks = SupplierFramework.query.filter(
         SupplierFramework.framework_id == framework.id
+    ).options(
+        db.defaultload(SupplierFramework.framework).lazyload("*"),
+        db.defaultload(SupplierFramework.supplier).lazyload("*"),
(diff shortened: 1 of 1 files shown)


risicle:
From a broader pov, it's a pity sqlalchemy doesn't (*seem to*) support an "eager load only for a depth of n" option (for anything other than self/cyclical joins). Because using that could remove the need for any of this per-view join-overriding. As it is, we either make them default lazy and override the views to one-deep eager load or default them to eager and override the views to disable eager loads beyond the first join...

:man_shrugging:

TheDoubleK:
Yep I had a look at doing it the other way round, but these two that we've done seemed to be the main ones that select a huge list of things that we only want a shallow representation of.  There are many more that select just one or a few records (for a specific supplier), where we want all the associated things, so I think this way round is probably marginally better.

PR #600, opened 2017-06-26
Filter brief responses by list of framework slugs

idavidmcdonald:
We need to filter brief responses by framework as part of this trello story
https://trello.com/c/vntb9Xql/396-allow-suppliers-to-view-their-applications

I've also decided to kept the conftest quite light (not following the more indepth pattern of `fixture_inner`) as keen not to over-engineer. 

Potential things for discussion:
- The second test (test_list_brief_responses_by_multiple_framework_slugs) maybe feels a little fragile due to there only being two DOS frameworks
- General conftest and fixture setup

### DIFF
--- a/app/main/views/brief_responses.py
+++ b/app/main/views/brief_responses.py
@@ -6,7 +6,7 @@ from sqlalchemy.exc import IntegrityError, DataError
 from dmapiclient.audit import AuditTypes
 
 from .. import main
-from ...models import db, Brief, BriefResponse, AuditEvent
+from ...models import db, Brief, BriefResponse, AuditEvent, Framework
 from ...utils import (
     get_json_from_request, json_has_required_keys, get_int_or_400,
     pagination_links, get_valid_page_or_1, url_for,
@@ -219,6 +219,13 @@ def list_brief_responses():
         db.defaultload(BriefResponse.brief).defaultload(Brief.lot).lazyload("*"),
         db.defaultload(BriefResponse.supplier).lazyload("*"),
     )
+
+    if request.args.get('framework'):
+        brief_responses = brief_responses.join(BriefResponse.brief).join(Brief.framework).filter(
+            Brief.framework.has(Framework.slug.in_(
+                framework_slug.strip() for framework_slug in request.args["framework"].split(",")
+            ))
+        )
 
     if brief_id or supplier_id:
         return jsonify(
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -175,11 +175,29 @@ def _add_framework(request, app, slug, **kwargs):
 
     def teardown():
         with app.app_context():
+            FrameworkLot.query.filter(FrameworkLot.framework_id == framework.id).delete()
             Framework.query.filter(Framework.slug == slug).delete()
             db.session.commit()
 
     request.addfinalizer(teardown)
     return framework.serialize()
+
+
+def _add_lots_for_framework(request, app, framework_slug, lot_slugs):
+    for lot_slug in lot_slugs:
+        _add_lot_for_framework(request, app, framework_slug, lot_slug)
+
+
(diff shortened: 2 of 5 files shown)


app/main/views/brief_responses.py
```diff
@@ -214,6 +214,13 @@ def list_brief_responses():
     if brief_id is not None:
         brief_responses = brief_responses.filter(BriefResponse.brief_id == brief_id)
 
+    if request.args.get('framework_slug'):
+        brief_responses = brief_responses.join(BriefResponse.brief).join(Brief.framework).filter(
+            Brief.framework.has(Framework.slug.in_(
+                framework_slug.strip() for framework_slug in request.args["framework_slug"].split(",")
```
TheDoubleK: I'm not sure about using `framework_slug` here, I can see the point about clarity, but there's a counterargument about consistency with usage elsewhere, and elsewhere we have `framework` arg for a list of framework slugs.

It's probably OK I guess, not a deal-breaker if you prefer it this way.

tests/conftest.py
```diff
@@ -175,13 +175,24 @@ def _add_framework(request, app, slug, **kwargs):
...
         with app.app_context():
+            FrameworkLot.query.filter(FrameworkLot.framework_id == framework.id).delete()
             Framework.query.filter(Framework.slug == slug).delete()
             db.session.commit()
 
     request.addfinalizer(teardown)
     return framework.serialize()
 
 
+def _add_lot_for_framework(request, app, framework_slug, lot_slug):
```
TheDoubleK: In terms of adding lots for other frameworks... I think probably yes because it's desirable to have consistency in what to expect if you use a framework fixture - i.e. as a developer I shouldn't need to know which frameworks come with lots and which don't (and it makes sense that all frameworks should have lots).  I'd be tempted to have adding lots as part of the base "give me a framework" code, rather than as an optional extra.

tests/main/views/test_brief_response.py
```diff
@@ -729,6 +729,61 @@ def test_list_brief_responses_for_brief_id(self):
         assert all(br['briefId'] == another_brief_id for br in data['briefResponses'])
         assert 'self' in data['links']
 
+    def test_list_brief_responses_by_one_framework_slug(self, live_dos2_framework):
+        with self.app.app_context():
+            dos2_brief = Brief(
```
TheDoubleK: Is it worth shifting this out to be a fixture too, alongside the framework?  Seems like having a mix of ways to set up data in one test is a bit weird (but I appreciate we might be doing it that way all over the place), and I thought we're generally trying to use fixtures these days?

app/main/views/brief_responses.py
idavidmcdonald: So searching the API views for `request.args.get('framework')` I find 4 routes.

Two of them take comma separated framework slugs
One only supports a single framework slug
One filters on `framework.framework` rather than the slug

I'm going to go with the rough consistency therefore of `framework` rather than `framework_slug` but then it feels like we need to fix the `framework.framework` filtering. I will raise in the tech debt meeting

tests/main/views/test_brief_response.py
idavidmcdonald: I'd be tempted to keep it as it is.

The whole file relies on a brief created `BaseBriefResponseTest.setup` so the similar thing to do would be to add the DOS2 brief to that set up. But we only need the DOS2 brief for two of the tests in this file as opposed to all 53 of them so it doesn't feel appropriate to put it there.

Unless we were going to move the whole test set up to be done using fixtures, I'd be tempted not to mix them and have this file responsible for setting itself up.

Do you feel strongly either way?

idavidmcdonald:
I've force pushed with changes that address your comments @TheDoubleK. Let me know what you think :)

benvand:
👍

PR #603, opened 2017-07-04
Improve API test coverage

katstevens:
The coverage stats show some gaps in our test coverage.

This PR covers:
- Adding tests for catching SQLAlchemy integrity/data errors in `app/main/views`
- Reorganising the vast number of tests in `test_drafts` into classes (mostly according to route, though there are some tests that use more than one route)
- Removal of a duplicate test 

Not included:
- Test coverage for areas that would require some refactoring of the code itself first

### DIFF
--- a/tests/main/views/test_drafts.py
+++ b/tests/main/views/test_drafts.py
@@ -361,7 +361,6 @@ class TestFetchAndListDraftServices(DraftServicesTestBase):
         assert res.status_code == 201
 
         data = json.loads(res.get_data())
-
         res = self.client.get('/draft-services/{}'.format(data['services']['id']))
         assert res.status_code == 200
         data = json.loads(res.get_data())
--- a/tests/main/views/test_drafts.py
+++ b/tests/main/views/test_drafts.py
@@ -1336,7 +1336,7 @@ class TestCompleteDraft(BaseApplicationTest, JSONUpdateTestMixin):
 
     def test_complete_draft(self):
         res = self.client.post(
-            '/draft-services/%s/complete' % self.draft_id,
+            '/draft-services/{}/complete'.format(self.draft_id),
             data=json.dumps({'updated_by': 'joeblogs'}),
             content_type='application/json')
 
@@ -1346,7 +1346,7 @@ class TestCompleteDraft(BaseApplicationTest, JSONUpdateTestMixin):
 
     def test_complete_draft_should_create_audit_event(self):
         res = self.client.post(
-            '/draft-services/%s/complete' % self.draft_id,
+            '/draft-services/{}/complete'.format(self.draft_id),
             data=json.dumps({'updated_by': 'joeblogs'}),
             content_type='application/json')
 
@@ -1364,7 +1364,7 @@ class TestCompleteDraft(BaseApplicationTest, JSONUpdateTestMixin):
 
     def test_should_not_complete_draft_without_updated_by(self):
         res = self.client.post(
-            '/draft-services/%s/complete' % self.draft_id,
+            '/draft-services/{}/complete'.format(self.draft_id),
             data=json.dumps({}),
             content_type='application/json')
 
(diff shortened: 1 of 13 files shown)


TheDoubleK:
Caught out by the Python 3 trap... 
Py2: `JSON was not a valid format. u'' is too short`
Py3: `JSON was not a valid format. '' is too short`

Our tests run with both Py2 and Py3, and I don't *think* there's a straightforward way to omit the `u` running under Py2.  One way could be to make two assertions, something like:
```
assert "JSON was not a valid format." in data
assert "'' is too short" in data
```
Unless you have better ideas?

tests/main/views/test_frameworks.py
```diff
@@ -379,6 +381,22 @@ def test_schema_validation_for_framework_agreement_details(self, open_example_fr
...
+            if key in self.attribute_whitelist
+        }
+
+        with self.app.app_context():
+            for key, value in valid_attributes_and_values.items():
+                response = self.post_framework_update({
+                    key: value
+                })
+
+                assert response.status_code == 400
```
TheDoubleK: Should we also check for the "Could not commit" the error message here?  It should be the same each time?

tests/main/views/test_services.py
```diff
@@ -1404,6 +1404,31 @@ def test_supplier_name_in_service_data_is_shadowed(self):
             assert response.status_code == 200
             assert data['services']['supplierName'] == u'Supplier 1'
 
+    @mock.patch('app.search_api_client')
+    def test_cannot_update_existing_service_by_put(self, search_api_client):
+        with self.app.app_context():
+            search_api_client.return_value = "bar"
+            response = self.client.put(
```
TheDoubleK: Probably not worth assigning to `response` here as the value is never used?  (I have a feeling @benvand's Flake8 stuff will complain about this.)

tests/main/views/test_drafts.py
```diff
@@ -1085,6 +1140,17 @@ def test_should_not_copy_draft_documents(self):
         assert "serviceDefinitionDocumentURL" not in data['services']
         assert "sfiaRateDocumentURL" not in data['services']
 
+    @mock.patch('app.db.session.commit')
+    def test_copy_draft_service_should_catch_db_integrity_error(self, db_commit):
+        db_commit.side_effect = IntegrityError("Could not commit", orig=None, params={})
+        res = self.client.post(
+            '/draft-services/%s/copy' % self.draft_id,
```
katstevens: I was wondering about that - one of the existing tests for this endpoint used `%` so I was going for 'consistency'. I can change the old test too 😄

tests/main/views/test_users.py
```diff
@@ -501,22 +503,23 @@ def test_return_400_for_invalid_user_json(self):
...
+                    'phoneNumber': '01234 567890',
+                    'password': '1234567890',
+                    'role': 'buyer',
                     'name': 'joe bloggs'}}),
             content_type='application/json')
-
         assert response.status_code == 400
-        data = json.loads(response.get_data())["error"]
-        assert "JSON was not a valid format" in data
+        assert json.loads(response.get_data())["error"] == "Invalid user role"
```
katstevens: I was wondering about that. That particular `try...except` aborts with that error message in the code. I was a bit loathe to change actual code in this PR rather than just tests, but if the message should be `Database Error:...` I can change it?

tests/main/views/test_drafts.py
TheDoubleK: Yeah that'd be nice.  There are a few places they get used.  We probably don't need to go through them all with a fine toothcomb, but if you spot one no harm in fixing up!

tests/main/views/test_users.py
TheDoubleK: It feels like a (very minor) code bug to me - I think fix it so the message tells the truth about what has gone wrong.

tests/main/views/test_drafts.py
```diff
@@ -1513,3 +1590,14 @@ def test_should_not_update_draft_status_to_invalid_status(self):
 
         assert res.status_code == 400
         assert json.loads(res.get_data()) == {"error": "'INVALID-STATUS' is not a valid status"}
+
+    @mock.patch('app.db.session.commit')
+    def test_update_draft_catches_db_integrity_errors(self, db_commit):
```
katstevens: Good spot! Thanks.

katstevens:
Closing this PR as it's turned into a big rebase mess! See https://github.com/alphagov/digitalmarketplace-api/pull/610 for the new improved version.

PR #604, opened 2017-07-05
Add default.nix

risicle:
See https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/719

Mostly similar to that PR, the only notable addition here is postgres to provide us with the libs for `psycopg2` to link against. And the `pg_config` binary as noted in comments. *In an ideal world* this should be chosen to match your locally running postgres version (most postgres versions are available in nixpkgs as e.g. `pkgs.postgresql94`, `pkgs.postgresql95`, but *in reality* the postgres protocol tends to be mostly cross-minor-version-compatible. So it's probably fine to leave it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ venv*
 __pycache__
 version_label
 .cache
+
+# local environment customizations
+local.nix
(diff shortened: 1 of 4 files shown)


risicle:
See https://github.com/alphagov/digitalmarketplace-manual/pull/6 for proposed corresponding manual page.

TheDoubleK:
:+1:

PR #606, opened 2017-07-06
Optimise queries for brief responses

idavidmcdonald:
We have been seeing memory issues, in particular on the list brief responses endpoint. Running this on similar data on the PaaS, this has seem good memory reduction.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1168,7 +1168,7 @@ class Brief(db.Model):
     clarification_questions = db.relationship(
         "BriefClarificationQuestion",
         order_by="BriefClarificationQuestion.published_at",
-        lazy='joined'
+        lazy='select',
     )
 
     @validates('users')
@@ -1325,7 +1325,7 @@ class Brief(db.Model):
             'requirementsLength': requirements_length
         }
 
-    def serialize(self, with_users=False):
+    def serialize(self, with_users=False, with_clarification_questions=False):
         data = dict(self.data.items())
 
         data.update({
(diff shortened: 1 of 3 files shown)


app/models.py
```diff
@@ -1168,7 +1168,7 @@ class Brief(db.Model):
     clarification_questions = db.relationship(
         "BriefClarificationQuestion",
         order_by="BriefClarificationQuestion.published_at",
-        lazy='joined'
+        lazy='select',
```
allait: Why are we switching to `select` here? We're still loading them for the list briefs endpoint, won't this result in N+1 queries being sent?

idavidmcdonald: If we keep this as joined, it is not a lazy join but is actually eagerly loaded, meaning we will always be joining the clarification questions table which we don't want to be doing. By changing it to select, it is lazily loaded.

By changing `Brief.serialize` to not access `clarification_questions` by default it means we won't trigger a call to the clarification questions table. So at the moment, only the `get_brief` and `add_clarification_questions` will load clarification questions, and will do so lazily, which will do the N+1 queries. We could add db options to those individual routes to override the behaviour and eagerly load them just for those routes?

Alternatively, we could keep the eager loading and add db options to all the routes that we don't want them to load clarification questions (I think this is a larger number than of those that should load clarification questions).

allait: Ah, you're right, I totally misread that as being part of  the `GET /briefs` endpoint

idavidmcdonald: Yep, looking at sql statements produced there will just be one extra query to retrieve the clarification questions.

PR #609, opened 2017-07-07
[#147424377]add migrations for "list_audits: add earliest_for_each_object"

risicle:
Story https://www.pivotaltracker.com/story/show/147424377

Main PR https://github.com/alphagov/digitalmarketplace-api/pull/596


TheDoubleK:
This looks good, but we also have the "view acknowledged" page that queries just for acknowledged changes. Might an index of all "type= service_update" events be better than just unacknowledged?

risicle:
Hmmm I think we're going to have to do some serious consideration of what the acknowledged page *does*...

TheDoubleK:
Yep, hadn't really looked at what we had - just got it wired up and tests fixed as it was.

In that case, let's get this rebased and merged :+1:

PR #611, opened 2017-07-07
[#138095135]acknowledge multiple audit events atomically

risicle:
Firstly, this sits on top of https://github.com/alphagov/digitalmarketplace-api/pull/596 so the "files changed" will be full of chaff. Sorry.

Story https://www.pivotaltracker.com/story/show/138095135

The motivation behind this is to add the ability to mark multiple audit events as acknowledged in a single atomic call which will result in all affected events being given the same `acknowledged_at` timestamp. This will allow for an amount of lightweight grouping to be performed when showing acknowledgement history without the requirement for any new concepts such as addressable "acknowledgement sets".

In the end, the sensible approach for this seemed to be to add a new endpoint, as so little of the code is shared with the `acknowledge_audit` endpoint. It was also difficult to conceive of a sensible way to "mode switch" an endpoint which is designed to operate on one audit event and make it operate on (and also return) multiple events.

Allowing the endpoint to acknowledge an arbitrary set of audit events seemed mostly out of the question:
- when it comes to HTTP interface, there isn't a particularly neat or concise way of supplying these ids to the endpoint
- in cases of unexpectedly extreme numbers of events, it would require a client to first fetch all of the unacknowledged ids to be able to specify them to this endpoint
- minor, but it allows for "bubbles" of unacknowledged events to be created. The rest of this system rather relies on the fact that there aren't any acknowledged update events after the most recent unacknowledged update event.

Being able to specify a range of audit event ids using a `/audit-events/123123...321321/acknowledge` path is an appealing idea but it has a couple of problems:
- while it _presents itself_ as a general interface for blanket updating of all audit events between those ids, we would need it to restrict affected events to just those with the same object and event type. Which would be a bit magic & unexpected. The magic could be removed if we wanted by adding all sorts of filtering abilities to the endpoint, but that seems like a lot of unnecessary complexity. It would also introduce the possibility of the first specified event and last specified event not sharing one of those properties - another category of error cases to check for...
- also it implies `id`-order is the thing we care about, whereas what we actually are ordering these by is `created_at`. Which, yes, *should* generally be monotonic, but... corner cases arising from manual intervention...
- it would also allow bubbles to be formed.

So the most sensible conclusion looked like adding a new endpoint whose specific job is to, given a single event id, acknowledge it and all previously-occurring events for that object for that event type. This prevents bubbles being formed, and prevents any of the property-mismatch issues mentioned above.

### DIFF
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -234,11 +234,21 @@ def acknowledge_including_previous(audit_id):
     result = db.session.execute(AuditEvent.__table__.update().returning(
         AuditEvent.__table__.c.id
     ).where(db.and_(
-        AuditEvent.__table__.c.created_at <= audit_event.created_at,
         AuditEvent.__table__.c.object_id == audit_event.object_id,
         AuditEvent.__table__.c.object_type == audit_event.object_type,
         AuditEvent.__table__.c.type == audit_event.type,
         AuditEvent.__table__.c.acknowledged == db.false(),
+        # ugly, but this just implements the same "id tie breaker" behaviour for created_at-equal events as the
+        # ordering we use in list_audits. this way there is some consistency between the two views as to what events
+        # are considered "previous" in such cases. we could use postgres composite types to express this far more
+        # neatly, but i can't get sqlalchemy to work with anonymous composite types.
+        db.or_(
+            AuditEvent.__table__.c.created_at < audit_event.created_at,
+            db.and_(
+                AuditEvent.__table__.c.created_at == audit_event.created_at,
+                AuditEvent.__table__.c.id <= audit_event.id,
+            ),
+        ),
     )).values(
         acknowledged=db.true(),
         acknowledged_at=datetime.utcnow(),
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -223,6 +223,42 @@ def acknowledge_audit(audit_id):
     return jsonify(auditEvents=audit_event.serialize()), 200
 
 
+@main.route('/audit-events/<int:audit_id>/acknowledge-including-previous', methods=['POST'])
+def acknowledge_including_previous(audit_id):
+    updater_json = validate_and_return_updater_request()
+
+    audit_event = AuditEvent.query.get(audit_id)
+    if audit_event is None:
+        abort(404, "No audit event with this id")
+
+    result = db.session.execute(AuditEvent.__table__.update().returning(
+        AuditEvent.__table__.c.id
+    ).where(db.and_(
+        AuditEvent.__table__.c.created_at <= audit_event.created_at,
+        AuditEvent.__table__.c.object_id == audit_event.object_id,
+        AuditEvent.__table__.c.object_type == audit_event.object_type,
+        AuditEvent.__table__.c.type == audit_event.type,
+        AuditEvent.__table__.c.acknowledged == db.false(),
+    )).values(
+        acknowledged=db.true(),
+        acknowledged_at=datetime.utcnow(),
+        acknowledged_by=updater_json['updated_by'],
+    )).fetchall()
+    db.session.expire_all()
+
+    try:
+        db.session.commit()
+
+    except IntegrityError as e:
+        db.session.rollback()
+        abort(400, e.orig)
+
+    # returning the list of affected ids, each one in its own dict seems the most rest-ful way (but least memory
+    # efficient - well spotted) of returning this information. you would think of these as the most abbreviated
+    # serializations of audit events possible.
+    return jsonify(auditEvents=[{"id": audit_event_id} for (audit_event_id,) in result]), 200
+
+
 @main.route('/audit-events/<int:audit_id>', methods=['GET'])
 def get_audit_event(audit_id):
     audit_event = AuditEvent.query.get(audit_id)
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -223,30 +223,44 @@ def acknowledge_audit(audit_id):
     return jsonify(auditEvents=audit_event.serialize()), 200
 
 
-@main.route('/audit-events/<int:audit_id>/acknowledge-including-previous', methods=['POST'])
-def acknowledge_including_previous(audit_id):
+# this is a "view without a route" for at the moment - it is used as an "inner" view implementation for the service
+# updates acknowledgement view
+def acknowledge_including_previous(
+        audit_id,
+        restrict_object_id=None,
+        restrict_object_type=None,
+        restrict_audit_type=None,
+        ):
     updater_json = validate_and_return_updater_request()
 
-    audit_event = AuditEvent.query.get(audit_id)
+    audit_event_query = db.session.query(AuditEvent)
+    if restrict_object_id is not None:
+        audit_event_query = audit_event_query.filter(AuditEvent.object_id == restrict_object_id)
+    if restrict_object_type is not None:
+        audit_event_query = audit_event_query.filter(AuditEvent.object.is_type(restrict_object_type))
+    if restrict_audit_type is not None:
+        audit_event_query = audit_event_query.filter(AuditEvent.type == restrict_audit_type)
+
+    audit_event = audit_event_query.filter(AuditEvent.id == audit_id).one_or_none()
     if audit_event is None:
-        abort(404, "No audit event with this id")
+        abort(404, "No suitable audit event with this id")
 
     result = db.session.execute(AuditEvent.__table__.update().returning(
-        AuditEvent.__table__.c.id
+        AuditEvent.id
     ).where(db.and_(
-        AuditEvent.__table__.c.object_id == audit_event.object_id,
-        AuditEvent.__table__.c.object_type == audit_event.object_type,
-        AuditEvent.__table__.c.type == audit_event.type,
-        AuditEvent.__table__.c.acknowledged == db.false(),
+        AuditEvent.object_id == audit_event.object_id,
+        AuditEvent.object.is_type(type(audit_event.object)),
+        AuditEvent.type == audit_event.type,
+        AuditEvent.acknowledged == db.false(),
         # ugly, but this just implements the same "id tie breaker" behaviour for created_at-equal events as the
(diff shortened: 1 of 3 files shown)


risicle:
I realize I've got to write a few more tests for this covering the non-happy-path... watch this space.

app/main/views/audits.py
```diff
@@ -180,6 +212,48 @@ def acknowledge_audit(audit_id):
     return jsonify(auditEvents=audit_event.serialize()), 200
 
 
+@main.route('/audit-events/<int:audit_id>/acknowledge-including-previous', methods=['POST'])
```
allait: I think the implementation makes sense, but I'm not sure about the endpoint itself.

We've used "verb" endpoints before, but they're usually tied to a specific object they're operating on (eg `/briefs/<brief_id>/copy`, `/agreeements/<agreement_id>/approve`). This one doesn't though, it affects a collection of objects at once and I don't think a single audit event is a meaningful entry point for this.

First of all, it creates an implicit relationship that hasn't been hardcoded before. The fact that only events with the same type and target object are acknowledged is never expressed by the caller, unlike eg when filtering the events. While that relationship is arguably there, the grouping doesn't necessarily behave the same way for different audit types. Eg I don't think we'd want to acknowledge clarification questions, password resets or failed login events in bulk. But similar to the include earliest flag, an audit-event endpoint defines that behaviour on all event types at once.

From a more practical perspective, by dropping the resource semantics of REST we risk losing some of the benefits it provides. In this case for example, I can imagine it being quite hard to find a log message of the API call that marked a certain audit event as acknowledged. In order to do that we'd need to search the logs for IDs of all subsequent events or find the latest event that was acknowledged at the same time.

These are largely the reasons I tried to push for a separate resource for the service updates. But since we're not doing that one alternative I can see is moving this to something like `POST /services/<id>/updates/acknowledge` with a payload of `{"latest_audit_id": ...}`. This is still a bit awkward, but it immediately limits the audit type to service updates, makes the relationship between affected audit events more explicit and allows us to investigate the logs by only looking up the audit event we're interested in (since it will contain the acknowledged timestamp and the target object id).

allait: > As for ease of log finding, I don't really see 123123...321321 making things much easier. 

That's not what I'm suggesting at all. I'm suggesting `POST /services/<id>/updates/acknowledge`.

I think it's much more likely that we'd need to change the service edits flow than us building an event acknowledgement flow for different types of events. And `/services/<id>/updates/acknowledge` feels like an easier endpoint to evolve for this as well as not forcing us to define operations we don't want to define at the moment (like acknowledging events of different types).

TheDoubleK: The implementation is generic and will work for all audit event types though, so putting it under `/services` seems odd to me. (In that you could inadvertently post to `/services...` and acknowledge a bunch of audit events completely unrelated to services.

I think the frontends should be responsible for when they call this and the API as clear as possible what it does.  Maybe we could have the route return the JSON-equivalent of a NotImplementedError if the type of the audit isn't one we're expecting (i.e. anything other than a service update event)? 

It does feel like this lives better under `/audit-events` than `/services`, but @allait I see your point about the verb referencing a single audit ID.  How about re-jigging the order to something like `/audit-events/acknowledge-including-previous/<audit_event_id>` (similar to what we have for, e.g. `/draft-services/copy-from/<service_id>`)?  Does that make it feel less like it implies operating on a single event?

Or we could go for something like a `POST /audit-events/acknowledge-all-for-object-until` with the `{"latest_audit_id": ...}` payload?

risicle: Hmm yes I see there's a mismatch with the whole `/someobject/<object_id>/someaction` thing manipulating more than one object - it's a pseudo-object-oriented interface and it's like you're addressing something not in context of its collection.

risicle: With `/audit-events/acknowledge-including-previous/<audit_event_id>` there is the whole naughtiness of having `acknowledge-including-previous` occupying the space that is in other cases occupied by an id though...

risicle: It *would* be good to somehow express the "only for this object and type" nature of it too.

allait: Oh yeah, sorry, I meant locking down the event type and object-type if we move it under `/services` to always apply to `update_service`.

I don't really see the benefits of keeping it generic across audit event types to be honest. There's no work on the roadmap that I'm aware of that would require this and as I've said above chances are the flow would be different anyway.

At the same time, my main practical point is about how easy it is to gather information about what's going on. Reordering the URL doesn't really help with that. If you take that draft-services endpoint for example (which I believe is largely unused) - you'd look at the service_id in the draft service and search for log events that contain it. I don't really see an easy way to do this with the proposed audit-events endpoints - you have to find related events until you can search the logs.

app/main/views/audits.py
```diff
@@ -180,6 +212,52 @@ def acknowledge_audit(audit_id):
...
+        AuditEvent.__table__.c.id
+    ).where(db.and_(
+        AuditEvent.__table__.c.object_id == audit_event.object_id,
+        AuditEvent.__table__.c.object_type == audit_event.object_type,
+        AuditEvent.__table__.c.type == audit_event.type,
+        AuditEvent.__table__.c.acknowledged == db.false(),
+        # ugly, but this just implements the same "id tie breaker" behaviour for created_at-equal events as the
+        # ordering we use in list_audits. this way there is some consistency between the two views as to what events
+        # are considered "previous" in such cases. we could use postgres composite types to express this far more
+        # neatly, but i can't get sqlalchemy to work with anonymous composite types.
```
galund: I don't understand, can't the order just be "date then id". Once you've hit the primary / any candidate key, no further ordering is necessary?

app/main/views/audits.py
```diff
@@ -180,6 +212,52 @@ def acknowledge_audit(audit_id):
...
+        acknowledged_by=updater_json['updated_by'],
+    )).fetchall()
+    db.session.expire_all()
+
+    try:
+        db.session.commit()
+
+    except IntegrityError as e:
+        db.session.rollback()
+        abort(400, e.orig)
```
galund: Why bother catching this? There's no particular evidence that it's the client's fault is there (which would motivate the 400 rather than a 500), and if we don't catch it the rollback will happen anyway?

galund:
The discussion here doesn't mention the motivation for adding the "earliest_for_each_object" flag, and it certainly seems to be creating a lot of code :-) (I can't see your pivotal board I'm afraid, if the justification is there.)

app/main/views/audits.py
risicle: > I don't really see the benefits of keeping it generic across audit event types to be honest.

What I'm *trying* to do is keep what we're doing be an API. If we end up special-casing every frontend view with its own special-case, magic API endpoint, are we making an API anymore? No, we're writing an extremely obtuse distributed monolith.

`/services/<service_id>/updates/acknowledge` *itself* wouldn't be ok as not including the specific event id would introduce a race condition - when an admin acknowledges a version they must be acknowledging *that specific version*, not whatever happens to be the latest one at the time the request gets through.

But `/services/<service_id>/updates/acknowledge-including-previous/<event_id>` wouldn't be the craziest thing in the world, and I see how I could implement it with the potential of expansion to other object types.

risicle:
> I don't see why the type can't be in the URL

It would just be yet another error case for me to write code for - a specified id that doesn't match the specified type - which a client couldn't possibly expect to get a sensible result from anyway.

risicle:
And yes, talk about #596 on #596

app/main/views/audits.py
galund: My thought right now would be `/audit-events/services/<service_id>/acknowledge?event-id=<event_id>`?

(I'm finding the 'including-previous' thing a bit hard to parse, and maybe it's easier to just document that if you acknowledge an event, you're also acknowledging all earlier events on the same service?)

app/main/views/audits.py
risicle: This is just cloned from the original `acknowledge_audit`. 🤷‍

app/main/views/audits.py
risicle: Yes, but when you're *filtering* you can't use `ORDER BY` primitives (unless you *actually* want to page through the results to find the starting point). What you *actually* want to say is "audit events where the tuple of `(created_at, id)` is greater than `(2012-10-10, 12345)`, but you can't do that without SQL composite types, which have their own complications.

app/main/views/audits.py
risicle: But people never read documentation (until it's too late...)

If *you're* asked to do some ad-hoc data admin at some odd hour, do *you* stop and find the documentation for an endpoint that otherwise gives no clue that it does anything other than what it says on the tin?

I wouldn't even know where to look for that sort of documentation.

risicle:
I do *love* these "angels on the head of a pin" discussions we get into over our SOA

app/main/views/audits.py
TheDoubleK: There's no race condition if, as @allait originally suggested, the audit event id is passed in as part of the payload rather than the URL. 

A small issue that would arise with something like `/services/<service_id>/updates/acknowledge-including-previous` (with the audit event id in either the payload or the URL) is that there's an extra database lookup required to validate that the `service_id` passed in matches the object referenced by the audit event, because I think the object id stored in the audit table isn't the (string) service ID but rather the numeric id from the services table that we never really expose anywhere else. It's not the end of the world, but a weirdness that would have to be worked with.

@galund I don't know whether the existing "ack single event" endpoint is going away right now or not, but either way can imagine we might want to support both single- and multi-event acknowledgements, and in that case it would be useful to be able to distinguish from the URL which is which without referring to the docs, so I think it needs something other than just `/acknowledge`.  (Not necessarily "including previous" if you have other suggestions... naming things is hard.)

risicle:
Thanks for playing "RESTier than thou" - next time you feel like playing, *don't* invite me.

TheDoubleK:
This needs a rebase but the new setup is neat - I'll have a quick look over after the rebase to see exactly what new is going in, but looks good I think.

TheDoubleK:
Good to go - let's do it :+1:

PR #612, opened 2017-07-10
Update api to Python 3

Wynndow:
### Pull in version 2.0.0 of base-api image

The version 2 series of base images are built with python3. See the docker-base repo for more info.

### Pull in version 27.1.1 of utils and 8.10.2 of apiclient

The new utils version has a fix for header's needing to be strings in
python3. The two major changes, 26 and 27 don't affect the api.

The new apiclient version has a fix to only require enum34 if the python
version being used is 3.4 or below. The major version bump, 8, doesn't
affect the api.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 language: python
 sudo: false
 python:
-  - "2.7"
-  - "3.4"
+  - "3.6"
 addons:
   postgresql: "9.3"
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
   - "3.6"
 addons:
   postgresql: "9.5"
+dist: "trusty"
 env:
   - SQLALCHEMY_DATABASE_URI=postgresql://postgres:@localhost:5432/digitalmarketplace_test
 install:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,8 +1,7 @@
 language: python
 sudo: false
 python:
-  - "2.7"
-  - "3.4"
+  - "3.6"
 addons:
   postgresql: "9.3"
```
allait: In the spirit of updating things, production is now on postgres 9.5

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [pep8]
-exclude = ./migrations,./venv,./venv3
+exclude = ./migrations,./venv,./venv3,./venv3.6
```
allait: `./venv*`?

benvand: Just noticed this, it depends on what venvs are going to be named? If it's just venv we can stick with `/venv,` here and also in the `.gitignore`

benvand: :man_shrugging:

Wynndow: Hey Ben, there's a fix for it in a later commit. Changes, it to `/venv*`

benvand: Nice :+1:

PR #614, opened 2017-07-11
Add option for brief clarification questions to briefs list endpoint

idavidmcdonald:
Our models no longer return them by default (a change to improve
memory usage), so now we must request them from the frontend ourselves
if we wish to get them.

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -118,6 +118,8 @@ def list_briefs():
 
     with_users = request.args.get('with_users', 'false').lower() == 'true'
 
+    with_clarification_questions = request.args.get('with_clarification_questions', 'false').lower() == 'true'
+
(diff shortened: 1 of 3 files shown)


app/main/views/briefs.py
```diff
@@ -118,6 +118,8 @@ def list_briefs():
 
     with_users = request.args.get('with_users', 'false').lower() == 'true'
```
benvand: Not sure I'd bother with the space here because these lines do such a simila job, might be cool to group them.
I _certainly_ don't feel too strongly though.

benvand:
>>> Our models no longer return them by default (a change to improve
memory usage), so now we must request them from the frontend ourselves
if we wish to get them.

Sounds a bit like the clarification questions themselves are coming from the frontend :grin:

PR #615, opened 2017-07-12
Pin all application requirements in requirements.txt

allait:
The list of top-level dependencies is moved to `reqirements-app.txt` which is used to by `make compile-requirements` to generate the full list of requirements in `requirements.txt`.

`requirements_for_test` and releated make step are renamed to `requirements-dev` to reflect the fact that it should contain any dependencies that aren't used by the app itself including dependencies for tests, utility scripts and dev tools.

Rationale
---------

We've had a number of issues caused by unpinned dependencies (eg alphagov/digitalmarketplace-api#607). They can cause things to fail at different stages depending on the root cause, but generally highlight a few problems with our current approach:

* The latest version of unpinned packages are installed, so whenever there's a breaking change in a new package release we're forced to either update our code or pin the package alongside our top-level application dependencies.
* `pip install` is run at different times. Eg Travis CI will install dependencies on each test run but our image builds will only run it if `requirements.txt` have changed. This means that libraries we're running the tests with could be different from libraries that are used by the deployed application.
* Finding out which library release broke the build is hard because there's no easily available list of "known-good" versions. These could be extracted from the previous application container versions, but the process isn't documented.

The common solution to this is to ship a full list of pinned dependencies.
There are currently 3 ways to do this in Python:

1. Generating a `requirements.txt` file with `pip freeze`. This stores the full list of currently installed packages.
2. Using [pip-tools] to generate a full list of required packages from a list of top-level dependencies. pip-tools also provides utilities to upgrade package versions without modifying the files manually.
3. Defining dependencies in a [Pipfile] and using [pipenv] to manage pinned versions and the virtualenv.

[Pipfile] seems to be the planned replacement for requirements files, however the tools appear to have some issues when applied to our repos. For example, [pipenv] doesn't appear to be locking dependencies of the VCS packages (eg utils and apiclient). It also doesn't detect non-semver package versions correctly (eg `functools32==3.2.3.post2`). We could fix the VCS dependencies issue by publishing our packages on PyPI, but my feeling is the tools aren't ready for production just yet.

[pip-tools] appears to be more commonly used at the moment. It works by examining setup.py files of packages listed in `requirements.in` file and generating a full list of dependencies in `requirements.txt`. The main issue with pip-tools when applied to our current applications is that it's stricter than pip in resolving version conflicts. Some of the packages we depend on are locking the libraries we use directly in their own setup.py (eg `mandrill` requires `docopt==0.4`, while we
generally use `0.6.2`). While `pip` would install the latest package version despite the conflict, pip-tools requires us to resolve the conflict, which in this case means relying on a fork of mandrill client.
And pip's behaviour in this case isn't really a problem for us, since docopt isn't used when mandrill is loaded as a module.

`pip freeze` is the simplest approach that doesn't require any additional tools, however it also isn't very usably without additional tooling. The main problems with using `pip freeze` directly is that it
will save whatever is installed in the current virtualenv including the test packages and packages that have been installed manually and that it loses VCS packages, only storing the names. This is solved by wrapping `pip freeze` with a `make compile-requirements` step that:

1. Creates a temporary virtualenv.
2. Installs application requirements.
3. Runs `pip freeze` to get a full list of installed packages.
4. Modifies `pip freeze` output to maintain the list of application dependencies in the original format.

The benefits of this approach is that it keeps the existing process of running `pip install -r requirements.txt` for application builds. It also requires only one command to be run whenever application dependencies change to regenerate `requirements.txt`.

The known downsides and unsolved problems:

* You need to remember to run `make compile-requirements` after changing `requirements-app.txt`
* Since the virtualenv is rebuilt from the application requirements it will still install newer versions of all unpinned packages. The difference is that the process is explicit and changes will be visible in the `requirements.txt` diff. This allows us to either accept the new package version or ignore it by discarding the change. This does mean that if there's a package we want to hold back we'll need to do this repeatedly after each `requirements-app.txt` change. The benefit of this approach is that it makes us aware of new package versions.

Another major point is that development requirements don't go through the same process but are now loading the full list of application dependencies. This in turn means that:

* `requirements-dev.txt` can't list any of the packages present in `requirements.txt` since pip will complain about a duplicate record. We can either rely on the record in `requirements.txt` in this case or move the dependency to `requirements-app.txt`
* While it's not possible for one of the dev requirements to override the application dependencies (pip's resolver prioritises the top-level dependency version) they do install additional packages that aren't present in the application deployment container. This is unchanged from the existing process but could potentially hide issues that are only discovered once the app has been deployed.

[pip-tools]: https://github.com/jazzband/pip-tools
[pipenv]: https://github.com/kennethreitz/pipenv
[Pipfile]: https://github.com/pypa/pipfile


Makefile
```diff
@@ -18,8 +18,19 @@ bootstrap: virtualenv
...
+compile-requirements:
+	virtualenv venv-freeze
+	$$(pwd)/venv-freeze/bin/pip install -r requirements-app.txt
+	echo '# This file is autogenerated. Run `make compile-requirements`' > requirements.txt
+	echo '# to update it with any changes made in requirements-app.txt' >> requirements.txt
+	echo '' >> requirements.txt
+	cat requirements-app.txt >> requirements.txt
+	echo '' >> requirements.txt
+	$$(pwd)/venv-freeze/bin/pip freeze -r requirements-app.txt | sed -n '/The following requirements were added by pip freeze/,$$p' >> requirements.txt
+	rm -rf venv-freeze
```
risicle: It's probably best to do any `rm -rf venv-freeze` at the beginning of the process to clean up the results of any failed runs.

risicle:
1. Well damn, one of us is going to have to rebase
2. I think the choice of the word "compile" here is a slightly confusing collision with the fact that some of the dependencies *do* need to be *compiled* and I initially assumed the new command was for doing that. I think "generate-requirements" would do fine. Or even `freeze-requirements`.

Makefile
allait: Yeah, good point. Will add a clean step at the beginning.

allait:
@risicle I'll keep this open for a bit to give people a chance to give feedback.

2. Yeah, the naming is generally inconsistent across tools. pip-tools calls it "compile", pipenv - "lock" and pip itself uses "freeze". I'm happy with "freeze-requirements" as it's pretty unambiguous and close to what it actually does.
3. Yep. I didn't want to make a big deal out of it by changing everything, but we do tend to use dashes in scripts so it feels like make targets should use a similar convention. But maybe that's the wrong approach here. I'm happy to either rename all steps or keep the new ones with underscores if people think it'd be better.

risicle:
3. Meh please everyone and put aliases in.

risicle:
If I were feeling really crazy I'd hypothesize that this `freeze-requirements` command would be a good place to run `pypi2nix` if people decided they liked that tool.

Makefile
samuelhwilliams: 👍

Makefile
```diff
@@ -43,4 +55,4 @@ docker-push:
 	docker push digitalmarketplace/api:${RELEASE_NAME}
 
 
-.PHONY: virtualenv requirements requirements_for_test test_pep8 test_migrations test_unit test test_all run_migrations run_app run_all docker-build docker-push
+.PHONY: virtualenv requirements requirements-dev freeze-requirements test_pep8 test_migrations test_unit test test_all run_migrations run_app run_all docker-build docker-push
```
samuelhwilliams: I'm totally OK with changing these to dashes. FYI. ;)

Makefile
samuelhwilliams: 👍

Makefile
samuelhwilliams: I'm totally OK with changing all the other ones to dashes as well. FYI. ;)

TheDoubleK:
We could have a comment at the top of `requirements-app.txt` to remind people to run the make freeze thing if they update the file?

Makefile
TheDoubleK: Consistency would be nice if I'm finally going to have to start using the `make` commands :)

Makefile
```diff
@@ -10,16 +10,28 @@ run_migrations: virtualenv
...
+freeze-requirements:
+	rm -rf venv-freeze
+	virtualenv -p python3 venv-freeze
+	$$(pwd)/venv-freeze/bin/pip install -r requirements-app.txt
+	echo '# This file is autogenerated. Run `make freeze-requirements`' > requirements.txt
+	echo '# to update it with any changes made in requirements-app.txt' >> requirements.txt
+	echo '' >> requirements.txt
+	cat requirements-app.txt >> requirements.txt
+	echo '' >> requirements.txt
+	$$(pwd)/venv-freeze/bin/pip freeze -r requirements-app.txt | sed -n '/The following requirements were added by pip freeze/,$$p' >> requirements.txt
```
allait: We pass a requirements file to pip freeze in `pip freeze -r requirements-app.txt`, which makes it keep the packages listed in the file in place and append everything else. The problem is that it also loses git package urls, replacing them with package names. So instead, I copy the original file and then only add the additional packages from the pip freeze output (the sed command skips all lines until it matches the pip freeze comment about added packages).

allait:
I've changed the comments in requirements.txt and requirements-app.txt so that they hopefully make sense when the files are joined. I've also added a `test-requirements` target to check that the files match up and have renamed other make targets to consistently use dashes instead of underscores.

Since there don't seem to be any major objections, I'm going to merge this and will make PRs for other apps. Thanks everyone!

risicle:
Yeah I think this is a positive.

PR #618, opened 2017-07-17
[#138031901]list_audits: add after-event-id filter

risicle:
More from the land of ugly/crazy API endpoints. 🏝 

This allows the results to be limited to those that occurred notionally "after" the specified one. "after" meaning it has a higher created_at value with id used as the tie-breaker, as is the standard way among these endpoints.

This should allow the supplier update acknowledgement pages to perform the "next" functionality without having to page through an arbitrarily large number of events to find its position. 

Another option would have been to have it as "created-after" and accept a timestamp, but that would fail to do the trick in the case of multiple updates sharing the same timestamp - again it would potentially require paging through an "unbounded" number of results to find itself.

Do what you like with this, but, it's there...

### DIFF
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -115,6 +115,27 @@ def list_audits():
     elif object_id:
         abort(400, 'object-id cannot be provided without object-type')
 
+    after_event_id = request.args.get('after-event-id')
+    if after_event_id:
+        ref_audit_event = db.session.query(AuditEvent).get(int(after_event_id))
+        if ref_audit_event is None:
+            abort(400, "No audit event with id {}".format(after_event_id))
+
+        audits = audits.filter(
+            # much like in acknowledge_including_previous, this clause implements the same "id tie breaker" behaviour
+            # ordering we use. this way the filter appears to behave exactly like a regular request truncated. the only
+            # situation where it will behave at all differently is cases where earliest_for_each_object is also used.
+            # again, we could use postgres composite types to express this far more neatly, but i can't get sqlalchemy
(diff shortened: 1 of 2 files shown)


app/main/views/audits.py
```diff
@@ -115,6 +115,27 @@ def list_audits():
     elif object_id:
         abort(400, 'object-id cannot be provided without object-type')
 
+    after_event_id = request.args.get('after-event-id')
```
TheDoubleK: `per_page` and `earliest_for_each_object` are both underscores rather than dashes.  Might be better to go with underscores for this too?  Although `audit-type`, etc. are dashes, so we already have a mix...

risicle: Oh yes so I thought about this: here we go...

I figured `earliest_for_each_object` was more a paging-y type thing so fitted in with `per_page` et al while `after-audit-id` was more of a filtering operation and so fitted better with `object-id`.

The mental gymnastics required for that rationalization...

TheDoubleK: Ha! Fine, I can kind of see that, and as we have an assortment here anyway it probably doesn't matter.

TheDoubleK:
Maybe we shelve this if we're not building the "next" bottom right now?

risicle:
Maybe we should shelve *you*

Sorry

allait:
I'm with Kev on this, I suggest we close the PR for now if we don't have any plans to use this functionality at the moment.

risicle:
(Ssshhhh but I'm working on bits related to this after hours...)

PR #619, opened 2017-07-26
750-Show-size-of-suppliers-on-opportunity-page

benvand:
### What was needed
https://trello.com/c/MUAKZ7G5/750-2-show-size-of-suppliers-on-opportunity-page
We want to be able to show the organisational size of the suppliers that have applied for an opportunity. This is part of the ongoing data transparency work and was born out of a need for suppliers to be able to find out who is applying for work.

### Basic approach
Make organisation_size available on the SupplierFramework model (because it comes from the declaration)
Make supplier_framework available on the BriefResponse model
When serialising BriefResponse return supplier_framework.supplier_organisation_size

### What was done
1. Hybrid property SupplierFramework.supplier_organisation_size
2. Make supplier_framework available on BriefResponse
3. Add `supplierOrganisationSize` to `BriefResponse` serialization and supply it through BriefResponse.supplier_framework.supplier_organisation_size
4. Update tests

### For more info on how this works  and why a normal relationship  wouldn't see:
The docs:
 http://docs.sqlalchemy.org/en/latest/orm/join_conditions.html#relationship-to-non-primary-mapper
Specifically the bit on why a normal relationship would not work in this case.
> since the A->secondary->B pattern does not support any references between A and B directly.

(ie BriefResponse.supplier_id)

The failed attempts branch:
https://github.com/alphagov/digitalmarketplace-api/compare/WIP-sf-on-br?expand=1
😁 

The comments:
https://github.com/alphagov/digitalmarketplace-api/pull/619/files#diff-413e9aa8961126d20e2d90befa13e0f1R1508

### Benchmarks
Does not increase SQL queries for end points. Adds a join.

Endpoint speed:

| With | Without |
| --------- | --------- |
| 0.65  | 0.59  |
| 0.60  | 0.65  |
| 0.62  | 0.59  |
| 0.60  | 0.64  |
| 0.61  | 0.60  |

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -409,6 +409,14 @@ class SupplierFramework(db.Model):
 
     # vvvv current_framework_agreement defined further down (after FrameworkAgreement) vvvv
 
+    @hybrid_property
+    def supplier_organisation_size(self):
+        return self.declaration.get('organisationSize') if self.declaration else None
+
+    @supplier_organisation_size.expression
+    def supplier_organisation_size(cls):
+        return cls.declaration['organisationSize'].astext
+
     @validates('declaration')
     def validates_declaration(self, key, value):
         value = strip_whitespace_from_data(value)
@@ -1481,6 +1489,7 @@ class BriefResponse(db.Model):
             'briefId': self.brief_id,
             'supplierId': self.supplier_id,
             'supplierName': self.supplier.name,
+            'supplierOrganisationSize': self.supplier_framework.supplier_organisation_size,
             'createdAt': self.created_at.strftime(DATETIME_FORMAT),
             'submittedAt': (
                 self.submitted_at and self.submitted_at.strftime(DATETIME_FORMAT)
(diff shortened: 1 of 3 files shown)


galund:
I assume this comes about because there isn't an API service to get the supplier-framework data already? If there is, then this seems like overkill, because there would be existing calls you could make to get this information, if you already have a brief response with its brief id and its supplier id (i.e. look up the brief, then look up the framework, then look up the supplier-framework with the framework id and the supplier id you already have).

The point I'm trying to make, this seems like it might be an optimisation: if so, we should only be putting it in, if the non-optimized version is too slow?

benvand:
Non optimised is N+1 calls to SupplierFrameworks endpoint :(

benvand:
@idavidmcdonald 
On my local I have 3281 SupplierFramework entries for dos2, to get them all requires 2 requests to the api.
Each request takes between 0.58 and 0.63 seconds to complete (low-high over 10 requests).

I set up a brief with 50 responses and the single request required to fetch them all (including the organisation size) takes between 0.062 and 0.071 (same low-high stat) 
For all intents and purposes 10* faster.

Looks like doing it that way will add at least 1.5 seconds to the request time (on local at least).

benvand:
@idavidmcdonald If you can check over the comments and if they need a bit more then I'm happy to go over them tomorrow or something to straighten them out

idavidmcdonald:
Getting all the supplier frameworks for DOS2 takes two API calls? Which are the two calls? Now I think about it, I assumed we would either be doing one API call or we need to use the iter method which will do (number of services divided by 100) calls? 

The other option I can think of is to adapt the list supplier framework endpoint so you can ask for all the supplier frameworks on DOS2 for a given comma separated list of supplier ID's. That might be slightly less resource intensive.

I think the approach we are currently taking is the best solution but would be interested to see if you agree @galund based on our above discussion.

galund:
Sounds like plenty enough reason there for going ahead with this approach, 👍

PR #622, opened 2017-08-09
Collect Awarded supplier and contract details

katstevens:
Ticket: https://trello.com/c/kGVfzXuH/749-3-collect-award-status-on-buyer-dashboard

PR 1 of 4: the API

See also:
PR 2 of 4: API client: https://github.com/alphagov/digitalmarketplace-apiclient/pull/82 (merged)
PR 3 of 4: Question content: https://github.com/alphagov/digitalmarketplace-frameworks/pull/464 (merged)
PR 4 of 4: Briefs front end https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/12

**TL:DR**: Adding the ability to store who won a Brief, and any contract details about the Brief supplied by the briefs-frontend form. This means adding new fields on `BriefResponse`.

**The juicy details:**
- New fields on the `BriefResponse` model: `awarded_at` datestamp and `award_details` JSON blob. This all fits in one handy DB migration (950).
- A new relationship on the `Brief` model, joining any BriefResponse that matches the Brief ID and has an `awarded_at` timestamp
- A unique constraint on the above relationship, so that only one BriefResponse can meet this criteria
- Validation on setting the `awarded_at` value under different status conditions
- 2 new API routes: `award_brief` and `award_brief_details`, to update the BriefResponse object and add the relevant AuditEvents (the corresponding API client changes for these routes are in PR 2 of 4 https://github.com/alphagov/digitalmarketplace-apiclient/pull/82).
- JSON schemas and validation on the routes
- Expose the newly stored data in the BriefResponse serialization.

**Other details:**
- To support a user partially filling in the form (i.e. choosing the winning supplier, but not yet filling out the contract details), we have a `pending-awarded` status on `BriefResponse` if there is a `pending: True` flag set in the award details JSON blob. This will allow us to pre-populate a form if the user returns to complete the form later.
- Huge thanks to @benvand and @allait for helping out with the DB relationship and SQL expressions for filtering by the new statuses.
- Storing the award data on the BriefResponse leaves open the possibility in future of storing more than one winner (for example, if a contract involves assembling a team of people from multiple suppliers). Also keeping the details alongside the supplier makes sense when consuming the data on the front end (see Pea's upcoming PR).
- When we come to do the 'no' path, there will obviously be no BriefResponse associated with the incoming data. It makes sense to me to add an extra JSON blob / datestamp combo on the `Brief` model at that point, depending on exactly what we want to store. There's a separate spike ticket to decide this.

~~NB: we will need to bump the API client version before merging.~~ Now done.


migrations/versions/940_brief_response_award_details.py
```diff
@@ -0,0 +1,34 @@
+"""award_details JSON blob and awarded_at date stamp on BriefResponse
```
Wynndow: Can you let me know when this goes out to the staging environment? I'd like to test the new production to staging data migration stuff when there is an actual difference in alembic versions!

tests/main/views/test_briefs.py
```diff
@@ -1242,3 +1242,243 @@ def test_supplier_is_eligible_even_if_does_not_supply_in_outcome_location(self):
...
+
+    def test_can_change_awarded_brief_response_for_closed_brief_without_awarded_datestamp(self):
+        with self.app.app_context():
+            self.setup_dummy_briefs(1, status="closed")
+            self.setup_dummy_suppliers(1)
+            brief_response1 = BriefResponse(brief_id=1, supplier_id=0, submitted_at=datetime.utcnow(), data={})
+            brief_response2 = BriefResponse(brief_id=1, supplier_id=0, submitted_at=datetime.utcnow(), data={})
+            db.session.add_all([brief_response1, brief_response2])
+            db.session.commit()
+            brief_response1.award_details = {'pending': True}
```
Wynndow: Can you do this before commiting for this first time and save a commit to the db?

tests/main/views/test_briefs.py
```diff
@@ -1242,3 +1242,243 @@ def test_supplier_is_eligible_even_if_does_not_supply_in_outcome_location(self):
...
+            }
+
+    def test_200_if_trying_to_award_a_brief_response_again_even_though_it_has_already_been_awarded(self):
+        with self.app.app_context():
+            self.setup_dummy_briefs(1, status="closed")
+            self.setup_dummy_suppliers(1)
+            brief_response = BriefResponse(brief_id=1, supplier_id=0, submitted_at=datetime.utcnow(), data={})
+            db.session.add(brief_response)
+            db.session.commit()
+            brief_response.award_details = {'pending': True}
```
Wynndow: Same as above, can you save a commit?

tests/test_models.py
```diff
@@ -404,6 +432,92 @@ def test_new_clarification_questions_get_added_to_the_end(self):
...
+        with self.app.app_context():
+            self.setup_dummy_suppliers(1)
+            brief = Brief(data={}, framework=self.framework, lot=self.lot, published_at=datetime(2000, 1, 1))
+            brief_response1 = BriefResponse(brief=brief, supplier_id=0, submitted_at=datetime.utcnow(), data={})
+            brief_response2 = BriefResponse(brief=brief, supplier_id=0, submitted_at=datetime.utcnow(), data={})
+            db.session.add_all([brief, brief_response1, brief_response2])
+            db.session.commit()
+            brief_response2.award_details = {"confirmed": "details"}
+            brief_response2.awarded_at = datetime(2016, 12, 12, 1, 1, 1)
+            db.session.add(brief_response2)
```
Wynndow: Same as above, can this be done in one?

app/main/views/briefs.py
```diff
@@ -204,6 +206,113 @@ def update_brief_status(brief_id, action):
     return jsonify(briefs=brief.serialize()), 200
 
 
+@main.route('/briefs/<int:brief_id>/award', methods=['POST'])
+def award_brief(brief_id):
```
katstevens: Good point. Perhaps something like `award_winning_brief_response` would be more descriptive of what's happening?

Wynndow: maybe something that mentions pending. `award_pending_brief_response` perhaps. Something less final.

tests/main/views/test_briefs.py
katstevens: Aha, yes it does work in a single commit now. I think this was left over from the previous implementation where we were setting a boolean `awarded` flag on the BriefResponse instead, and that required a saved DB object beforehand. I'll change these tests now.

app/models.py
```diff
@@ -1448,6 +1485,8 @@ def status(self):
     def status(cls):
         return sql_case([
             (cls.submitted_at.isnot(None), 'submitted'),
+            (cls.awarded_at.isnot(None), 'awarded'),
```
katstevens: The test for this seems to pass OK? `test_awarded_status_for_brief_response_has_been_awarded`

Wynndow: I think this uses the python version of the hybrid property to get the status. I'm not sure of the best way to test the sql version.

migrations/versions/940_brief_response_award_details.py
```diff
@@ -0,0 +1,34 @@
...
+from sqlalchemy.dialects import postgresql
+
+
+# revision identifiers, used by Alembic.
+revision = '940'
+down_revision = '930'
+
+
+def upgrade():
+    op.add_column('brief_responses', sa.Column('award_details', postgresql.JSON(astext_type=sa.Text()), nullable=True))
```
benvand: Worth defaulting to '{}' here and removing checks like `br.award_status and br.award_status.get('foo') == 'bar'`
To do this you can add a `server_default` to the column declaration to set a default in the database itself.

app/main/views/briefs.py
```diff
@@ -204,6 +206,113 @@ def update_brief_status(brief_id, action):
...
+
+    if not brief.status == 'closed':
+        abort(400, "Brief is not closed")
+
+    brief_responses = BriefResponse.query.filter(
+        BriefResponse.brief_id == brief.id,
+        BriefResponse.status != 'draft'
+    )
+    # Check that BriefResponse in POST data is associated with this brief
+    if not json_payload['brief_response_id'] in [b.id for b in brief_responses]:
```
benvand: JSON keys should be camelCase

app/main/views/briefs.py
```diff
@@ -204,6 +206,113 @@ def update_brief_status(brief_id, action):
...
+        abort(400, "BriefResponse cannot be awarded for this Brief")
+
+    # Find any existing pending awarded BriefResponse
+    audit_events = []
+    pending_awarded_responses = list(filter(lambda x: x.status == 'pending-awarded', brief_responses))
```
benvand: You can use the ORM for this without causing more queries
`brief_responses.filter(BriefResponse.status=='pending-awarded')`

app/main/views/briefs.py
benvand: Can use ORM for this too like:
`breif_responses.filter(BriefResponse.id == json_payload['brief_response_id']).exists()`

benvand: Actually we could define `brief_response` up here like:
```
brief_reponse = brief_responses.get(BriefResponse.id==json_payload['brief_response_id'])
if not brief_response:
    # BAIL
```
Then we don't have to define it on line 251

app/main/views/briefs.py
```diff
@@ -204,6 +206,113 @@ def update_brief_status(brief_id, action):
...
+    if pending_awarded_responses:
```
benvand: Feels like a for loop here would indicate what we're trying to do better.
_If_ there was more than one `pending_awarded_response` we'd want to blank all of them. Also calling the variable `pending_awarded_responses` and then treating it as if there's only going to ever be one (`pending_awarded_responses[0]`) feels a little semantically weird :man_shrugging:

app/main/views/briefs.py
```diff
@@ -204,6 +206,113 @@ def update_brief_status(brief_id, action):
...
+                    'briefId': brief.id,
+                    'briefResponseId': pending_award.id,
+                    'briefResponseAwardedValue': False
+                },
+                db_object=pending_award
+            )
+        )
+
+    # Set new awarded BriefResponse
+    brief_response = list(filter(lambda x: x.id == json_payload['brief_response_id'], brief_responses))[0]
```
benvand: https://github.com/alphagov/digitalmarketplace-api/pull/622/files#r132459384

app/main/views/briefs.py
```diff
@@ -204,6 +206,113 @@ def update_brief_status(brief_id, action):
...
+    updater_json = validate_and_return_updater_request()
+
+    brief = Brief.query.filter(
+        Brief.id == brief_id
+    ).first_or_404()
+
+    brief_response = BriefResponse.query.filter(
+        BriefResponse.id == brief_response_id
+    ).first_or_404()
+    if not (brief_response.award_details and brief_response.award_details.get('pending')):
```
benvand: Re setting the default and not needing the none check
https://github.com/alphagov/digitalmarketplace-api/pull/622/files#r132456726

app/main/views/briefs.py
```diff
@@ -204,6 +206,113 @@ def update_brief_status(brief_id, action):
...
+        abort(400, "Cannot save details for this brief")
+
+    # Drop any null values to ensure correct validation messages
+    award_details = {k: v for k, v in json_payload['award_details'].items() if v is not None}
```
benvand: We've got a helper for this!
```
from utils import purge_nulls_from_data
award_details = purge_nulls_from_data(json_payload['award_details'])```

File: app/models.py [hunk 14]
```diff
@@ -1186,6 +1187,13 @@ class Brief(db.Model):
         lazy='select',
     )
 
+    awarded_brief_response = db.relationship(
+        'BriefResponse',
+        primaryjoin="and_(Brief.id==BriefResponse.brief_id, BriefResponse.awarded_at != None)",
+        uselist=False,
+        lazy='joined'
```
benvand: I'd suggest `viewonly` here.
http://docs.sqlalchemy.org/en/latest/orm/relationship_api.html#sqlalchemy.orm.relationship.params.viewonly

app/models.py
```diff
@@ -1417,10 +1436,13 @@ class BriefResponse(db.Model):
 
     created_at = db.Column(db.DateTime, index=True, nullable=False, default=datetime.utcnow)
     submitted_at = db.Column(db.DateTime, nullable=True)
+    awarded_at = db.Column(db.DateTime, nullable=True)
 
     brief = db.relationship('Brief', lazy='joined')
     supplier = db.relationship('Supplier', lazy='joined')
 
+    award_details = db.Column(JSON, nullable=True)
```
benvand: I'd probably stick this above the relationships, group  everything that's an actual column in the db

app/models.py
```diff
@@ -1431,6 +1453,17 @@ def validates_data(self, key, data):
 
         return data
 
+    @validates('awarded_at')
+    def validates_awarded_at(self, key, awarded_at):
+        if self.status == "awarded":
+            raise ValidationError('Brief response award cannot be changed as the brief has already been awarded.')
+        if awarded_at:
```
benvand: It's probably worth changing this to break out of the function early like:
```
if not awarded_at:
    return None
if self.brief.status != "closed":
    ...
```
Saves you a level of indentation at the cost of another line and some people find it more readable:  https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

I don't care about indentation level personally but if some people find it easier to read then fine.

app/models.py
benvand: Yeah I tested this in the shell
`python application.py shell`
```
In [1]: from app.models import BriefResponse
In [2]: from app import db
In [3]: db.session.query(BriefResponse).filter(BriefResponse.status=='awarded').first() is None
Out[3]: True
```

tests/main/views/test_briefs.py
```diff
@@ -1242,3 +1242,239 @@ def test_supplier_is_eligible_even_if_does_not_supply_in_outcome_location(self):
...
+    award_url = "/briefs/1/award/{}/contract-details"
+    valid_payload = {
+        "award_details": {
+            "awardedContractStartDate": "2020-12-31",
+            "awardedContractValue": "99.95",
+        },
+        "updated_by": "user@email.com"
+    }
+
+    def _post_to_award_details_endpoint(self, payload, brief_response_id):
```
benvand: Could create a base class for `_get_brief_response_audit_events` and `_post_to_award_details_endpoint`. Not a dealbreaker.

tests/test_models.py
```diff
@@ -404,6 +432,90 @@ def test_new_clarification_questions_get_added_to_the_end(self):
...
+    def test_no_awarded_brief_response_if_no_brief_responses(self):
+        with self.app.app_context():
+            brief = Brief(data={}, framework=self.framework, lot=self.lot, published_at=datetime(2000, 1, 1))
+            db.session.add(brief)
+            db.session.commit()
+
+            brief = Brief.query.get(brief.id)
+            assert brief.awarded_brief_response is None
+
+    def _setup_brief_and_awarded_brief_response(self, context, awarded_at=True, pending=True):
```
benvand: Nice to put helper methods at the top of test classes

tests/test_models.py
```diff
@@ -550,6 +666,18 @@ def test_draft_status_for_brief_response_with_no_submitted_at(self):
...
+    def test_awarded_status_for_brief_response_has_been_awarded(self):
+        with self.app.app_context():
+            brief = Brief.query.get(self.brief_id)
+            brief_response = BriefResponse(
+                brief=brief,
+                submitted_at=datetime.utcnow(),
+                award_details={'confirmed': 'details'},
+                awarded_at=datetime(2016, 1, 1)
+            )
+
```
benvand: To check for the sql `hybrid_property.expression` case we can add, commit and reload the brief_response before checking the status

katstevens: Cool - have pushed a commit to fix (re-ordering the case statement), which also includes a test for the `pending-awarded` status.

tests/test_models.py
katstevens: Whereas this test class is HUGE... Perhaps I'll split out the awards-related tests.

tests/main/views/test_briefs.py
katstevens: I've moved the audit events helper to `helpers.py` as we can reuse that for other `AuditEvent` assertions. The `_post_to_award_details_endpoint` is only used in this class, I think?

app/models.py
katstevens: Moving this case to the top would mean that we could potentially un-award a BriefResponse by setting the `awarded_at` datestamp to None, which we're currently disallowing. Should we be this strict? (My gut says no...)

app/models.py
```diff
@@ -1431,6 +1457,18 @@ def validates_data(self, key, data):
 
         return data
 
+    @validates('awarded_at')
+    def validates_awarded_at(self, key, awarded_at):
+        if self.awarded_at is not None:
+            raise ValidationError('Brief response award cannot be changed as the brief has already been awarded.')
```
idavidmcdonald: This error message could be a little clearer either to say that you can not award this brief response more than once or that this brief response has already been awarded

app/main/views/briefs.py
```diff
@@ -204,6 +207,113 @@ def update_brief_status(brief_id, action):
...
+
+    brief = Brief.query.filter(
+        Brief.id == brief_id
+    ).first_or_404()
+
+    brief_response = BriefResponse.query.filter(
+        BriefResponse.id == brief_response_id
+    ).first_or_404()
+    if not brief_response.award_details.get('pending'):
+        abort(400, "Cannot save details for this brief")
```
galund: As Ben I think mentioned elsewhere, it would be good to say _why_ in the message, even if we don't think it's likely / possible to actually come up in real life. (Because developers are users too :) )

galund: Actually it was David. I'm a little unclear why here we are choosing abort(400), whereas elsewhere we're throwing ValidationErrors. Presumably the effect is much the same - do those errors get converted into 4xx responses by something?

app/models.py
```diff
@@ -1286,10 +1297,17 @@ def status(self, value):
...
     def status(cls):
+        # To filter by 'awarded' status, we need an explicit EXISTS query on BriefResponse.
+        # This mirrors the awarded_brief_response relationship query above.
         return sql_case([
             (cls.withdrawn_at.isnot(None), 'withdrawn'),
             (cls.published_at.is_(None), 'draft'),
-            (cls.applications_closed_at > datetime.utcnow(), 'live')
+            (
+                exists([BriefResponse.id]).where(
+                    sql_and(cls.id == BriefResponse.brief_id, BriefResponse.awarded_at != None)  # noqa
```
galund: Can we used the `isnot(None)` construction per discussion @ https://stackoverflow.com/questions/21784851/sqlalchemy-is-not-null-select

The `!= None` thing is super-ugly :)

katstevens: We can indeed - I'd tried `is not None` and got an error, but `isnot(None)` works fine. Thanks George!

PR #624, opened 2017-08-15
Direct Award Procurement Projects

samuelhwilliams:
## Summary
Adds basic API endpoints to allow creating+listings projects, saving+listing searches, and adding searches to projects.

## Blocked by
~~https://github.com/alphagov/digitalmarketplace-apiclient/pull/84 needs to go in first (for audit events)~~
~~https://github.com/alphagov/digitalmarketplace-api/pull/630 (must be merged first; contains the separate migrations for direct awards that should sit underneath the logic changes)~~

## Associated PRs/branches
~~APIClient: https://github.com/alphagov/digitalmarketplace-apiclient/pull/84~~
Buyer FE: https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/581 (for trying out the flow only; not for review yet)

## Ticket
https://trello.com/c/pbJiKaVl/625-api-for-direct-award-projects


TheDoubleK:
The migration should go in a PR on its own, separate from the model changes. And I'm not mad keen on the `da_` prefix on the new table names. I'd rather go with something longer and more immediately meaningful.

samuelhwilliams:
Fair point re:`da`. We agreed it as a team but you're probably right that it may as well be more explicit, so I'll update it to `directaward` unless anyone else vehemently disagrees.

Might need you to explain the migration thing to me tomorrow, don't immediately understand why it should be separate or where the line should be drawn.

tests/test_models.py
```diff
@@ -1358,3 +1361,170 @@ def test_current_framework_agreement_with_countersigned_and_new_signed(self):
...
+
+            project.active = False
+            db.session.add(project)
+            db.session.commit()
+            assert project.status == 'abandoned'
+
+            project.completed_at = datetime.utcnow()
+            db.session.add(project)
+            db.session.commit()
+            assert project.status == 'completed'
```
galund: So I'm thinking in terms of "creating a list" being the "completed" step.

But asking for a big refactoring now to have "list" as a first-class concept seems like a big ask at this point in the sprint.

So maybe the middle-ground approach is to split the "status" field into two, I'd suggest:

* active (boolean, not null)
* list_created_at (datetime, nullable)

Explanation: we don't need a separate time field, as we can derived whether or not the list was created from whether the datetime was set.

samuelhwilliams: I guess I'm understanding you wrong.

We currently have an `active` boolean, not null and a `completed_at` datetime nullable. Are you just saying that `completed_at` should be renamed? If so that's fine.

Status is derived from those two properties already; it doesn't exist in the DB itself..

tests/main/views/test_direct_award.py
```diff
@@ -0,0 +1,499 @@
...
+        res = self.client.get('/direct-award/projects?user_id={}'.format(self.user_id))
+        data = json.loads(res.get_data(as_text=True))
+
+        assert res.status_code == 200
+        assert len(data['projects']) == 2
+
+        with self.app.app_context():
+            projects = DirectAwardProject.query.filter(DirectAwardProject.users.any(User.id == self.user_id))
+
+            assert all([x.serialize() in data['projects'] for x in projects.all()])
```
MuriloDalRi: I like x, it's one of my favourite chars, but maybe we could have something a bit more descriptive? :P

app/main/views/direct_award.py
```diff
@@ -0,0 +1,192 @@
...
+        per_page=current_app.config['DM_API_PROJECTS_PAGE_SIZE'],
+    )
+
+    pagination_params = request.args.to_dict()
+    pagination_params['project_id'] = project_id
+
+    return jsonify(
+        searches=[search.serialize() for search in searches.items],
+        meta={
+            "total": searches.total,
```
galund: Assuming we already call it 'total' elsewhere then this is fine, but I find the name confusing. I'd prefer `size` or `count` or `total_count` or something?

tests/main/views/test_direct_award.py
samuelhwilliams: Good catch.

app/main/views/direct_award.py
samuelhwilliams: Already called `total` elsewhere.

README.md
```diff
@@ -52,6 +52,13 @@ apt-get install gcc virtualenv python-dev libffi-dev libpq-dev
 source ./venv/bin/activate
 ```
 
+### Creating a new database migration
```
galund: I think this probably belongs further down - it's not part of the "Full setup" process that's being described here. Maybe next to "Updating application dependencies".

app/models.py
```diff
@@ -1614,6 +1614,75 @@ def serialize(self):
         }
 
 
+class DirectAwardProject(db.Model):
```
galund: So I was wanting to put these into an `app.models.direct_award` module. Then this model would just be called `Project`.

Which would mean moving all the existing code in `app.models` into a new `__init__.py`, which is a pain... @allait if you don't agree then we can just move on?

samuelhwilliams:
@galund added the `locked_at` property and added validation on editing ``DirectAwardSearch`` to reject if its associated project is locked. Added a test for it.

app/main/views/direct_award.py
```diff
@@ -0,0 +1,193 @@
...
+@main.route('/direct-award/projects', methods=['GET'])
+def list_projects():
+    user_id = get_int_or_400(request.args, 'user_id')
+    if user_id is None:
+        abort(400, "User ID not supplied")
+
+    page = get_valid_page_or_1()
+
+    projects = DirectAwardProject.query.filter(DirectAwardProject.users.any(id=user_id))
+    projects = projects.order_by(desc(DirectAwardProject.created_at), desc(DirectAwardProject.id))
```
allait: Usually `asc` order by id is preferred as the default sort since it allows us to safely iterate through API pages without catching duplicates (since for desc order someone creating a new project between page=1 request and page=2 request will shift 100th record to 101, making it appear twice).

Are we planning a paginated display of all projects on the frontend? If we do, this should be a "reverse order", similar to "human" flag in briefs or "earliest_first" in audit events.

app/main/views/direct_award.py
```diff
@@ -0,0 +1,193 @@
...
+    db.session.commit()
+
+    return jsonify(project=project.serialize()), 201
+
+
+@main.route('/direct-award/projects/<int:project_id>', methods=['GET'])
+def get_project(project_id):
+    user_id = get_int_or_400(request.args, 'user_id')
+    if user_id is None:
+        abort(400, "User ID not supplied")
```
allait: Same argument as the list searches check below

app/main/views/direct_award.py
```diff
@@ -0,0 +1,193 @@
...
+from ...utils import (
+    get_json_from_request, get_int_or_400, json_has_required_keys, pagination_links,
+    get_valid_page_or_1, validate_and_return_updater_request, validate_user_can_access_direct_award_project_or_403)
+
+
+@main.route('/direct-award/projects', methods=['GET'])
+def list_projects():
+    user_id = get_int_or_400(request.args, 'user_id')
+    if user_id is None:
+        abort(400, "User ID not supplied")
```
MuriloDalRi: It might be something for our tech debt, but this bit of code is repeated everywhere, it would be nice to move it somewhere else and use it as a decorator like we do in other apps.

app/main/views/direct_award.py
```diff
@@ -0,0 +1,193 @@
...
+
+    searches = DirectAwardSearch.query.filter(DirectAwardSearch.project_id == project_id)\
+        .order_by(desc(DirectAwardSearch.created_at), desc(DirectAwardSearch.id))
+
+    searches = searches.paginate(
+        page=page,
+        per_page=current_app.config['DM_API_PROJECTS_PAGE_SIZE'],
+    )
+
+    pagination_params = request.args.to_dict()
```
allait: I don't think we filter any unrecognised parameters from the request, so I'm not entirely sure how safe this is to use. Is it possible to explicitly pick things we're interested and have checked are fine here?

app/models.py
```diff
@@ -1614,6 +1614,99 @@ def serialize(self):
...
+    id = db.Column(db.Integer, primary_key=True)
+    project_id = db.Column(db.Integer, db.ForeignKey('direct_award_projects.id'), index=True, nullable=False)
+    user_id = db.Column(db.Integer, db.ForeignKey('users.id'), index=True, nullable=False)
+
+
+class DirectAwardSearch(db.Model):
+    __tablename__ = 'direct_award_searches'
+
+    id = db.Column(db.Integer, primary_key=True)
+    created_by = db.Column(db.Integer, db.ForeignKey('users.id'), nullable=False)
```
allait: Why do we have both `created_by` and `user`?

app/main/views/direct_award.py
```diff
@@ -0,0 +1,193 @@
...
+    except IntegrityError as e:
+        db.session.rollback()
+        abort(400, e.orig)
+
+    audit = AuditEvent(
+        audit_type=AuditTypes.create_project,
+        user=updater_json['updated_by'],
+        data={
+            'project_id': project.id,
+            'project_json': project_json,
```
samuelhwilliams: Yeah so I get this is the standard, but I started using that and it ended up just becoming really unclear where to draw the line between python vars, json keys, and query params and I was spending loads of (wasted) time mixing up between them all. I opted to be internally consistent with this section of code. Hope you don't mind... :)

samuelhwilliams: I'd love to know the reason behind wanting to use camelCase for the json keys though. Is there a good reason or just a general style choice?

app/main/views/direct_award.py
samuelhwilliams: We pass `request.args` to `pagination_links` everywhere else - is this notably different?

app/models.py
samuelhwilliams: `created_by` is the column in the db, `user` is just sqlalchemy niceness to easily access the corresponding row in the User db.

app/models.py
```diff
@@ -1614,6 +1614,99 @@ def serialize(self):
...
+    project_id = db.Column(db.Integer, db.ForeignKey('direct_award_projects.id'), nullable=False, index=True)
+    created_at = db.Column(db.DateTime, default=datetime.utcnow, nullable=False)  # When this search was first saved.
+    searched_at = db.Column(db.DateTime, nullable=True)  # When this search was run for the project services download.
```
samuelhwilliams: So the search is 'created' when the user saves it. The search is 'run'/'searched' (as far as Direct Award goes) when they download their list of services and start the shortlisting process.

A user can save multiple searches while they're refining their criteria, but then they will select one (or possibly more) searches to re-run, and the resulting services form their long/shortlist.

app/models.py
```diff
@@ -1614,6 +1614,99 @@ def serialize(self):
...
+    search_url = db.Column(db.Text, nullable=False)  # Search API URL to hit when retrieving services.
```
samuelhwilliams: Yes, we considered this (and I was a proponent of the idea) - but given the search URL is the more canonical representation of what the user has done and the result set it describes, we opted to store that instead.

app/models.py
```diff
@@ -1614,6 +1614,99 @@ def serialize(self):
...
+
+    # Partial index on project_id,active==1. Enforces only one active search per project at a time.
+    __table_args__ = (db.Index('idx_project_id_active', project_id, active, unique=True,
+                               postgresql_where=db.Column('active')),)
+
+    def serialize(self):
+        return {
+            "id": self.id,
+            "created_by": self.created_by,
+            "project_id": self.project_id,
```
samuelhwilliams: try_to_make_me

app/main/views/direct_award.py
samuelhwilliams: Removed

samuelhwilliams: Ok, I'll remove the user_id checks. I suppose I thought it added some modicum of extra precaution against displaying data to the wrong user (in case something went weird in the frontend), but maybe in reality that's unlikely to be an issue or one that we don't want to account for.

app/main/views/direct_award.py
```diff
@@ -0,0 +1,193 @@
...
+    validate_user_can_access_direct_award_project_or_403(user_id, project_id)
+
+    return jsonify(project=DirectAwardProject.query.get(project_id).serialize())
+
+
+@main.route('/direct-award/projects/<int:project_id>/searches', methods=['GET'])
+def list_project_searches(project_id):
+    user_id = get_int_or_400(request.args, 'user_id')
+    if user_id is None:
+        abort(400, "User ID not supplied")
```
samuelhwilliams: Ok, I'll remove the user_id checks. I suppose I thought it added some modicum of extra precaution against displaying data to the wrong user (in case something went weird in the frontend), but maybe in reality that's unlikely to be an issue or one that we don't want to account for.

app/main/views/direct_award.py
```diff
@@ -0,0 +1,193 @@
...
+
+@main.route('/direct-award/projects/<int:project_id>/searches', methods=['POST'])
+def create_project_search(project_id):
+    updater_json = validate_and_return_updater_request()
+
+    json_payload = get_json_from_request()
+    json_has_required_keys(json_payload, ['search'])
+
+    search_json = json_payload['search']
+    json_has_required_keys(search_json, ['user_id', 'search_url'])
```
samuelhwilliams: As above... curious to know why we do this.

app/main/views/direct_award.py
galund: hmm, I think these might be important. If we want to have a model where multiple users (but not all users) can have access to certain resources, then I'd want this logic to be enforced by the back-end. Now of course the frontend has to do the authentication, but once we have a verified user id then the right place for this logic is definitely the API I think?

app/main/views/direct_award.py
samuelhwilliams: Nice, thanks.

app/main/views/direct_award.py
allait: @galund I know what you mean and I definitely have a feeling that the checks in the frontend can be unreliable since they're easy to forget without noticing anything's wrong.

The main problem with GET is that it goes against the read behaviours we expect from the API:
1. It's not obvious what's the best way to list project searches with a script for example since we'd need to attach a user_id for each request, which isn't actually what's happening in that case.
2. In pretty much every other list endpoint I'd read the `user_id` GET argument as "only return objects from this user". This goes against that and makes it hard to combine with that filter (or any other filter) in the same endpoint.
3. GET parameter just feels like a wrong place for authentication details in terms of caching, logging access etc.

I think if we extend this to "how would this look in a public API" it becomes clear that API would have to do both authentication and authorization for this and other read and write endpoints that are currently doing this in the frontend apps (eg draft briefs, draft services). At that point, we'd need to find a way to securely transfer authentication details to the API and authorize everywhere based on that. Since we haven't thought about what that would look like and whether we'd want to use it in our own frontends as well I think we should stick with the current common approach for now and do the checks in the frontends.

galund: Fair points.

Re 1 I think that scripts should be subject to the same access-control restrictions and (more importantly) audit logging, in which case there's a strong argument for giving them role accounts in the system anyway?

But I take your other points about the user id - in this case - not actually affecting the resource we're describing at all, so the authentication should be in another band.

app/utils.py
```diff
@@ -154,3 +154,14 @@ def purge_nulls_from_data(data):
 def get_request_page_questions():
     json_payload = get_json_from_request()
     return json_payload.get('page_questions', [])
+
+
+def validate_user_can_access_direct_award_project_or_403(user_id, project_id):
```
galund: This is named wrong - it sometimes throws a 404 :)

But also I'm not sure I like this being in 'utils' - can we just have a private method in `views/direct_award.py`?

In light of the other conversation, are we actually going to leave it in? Do we really want to do access control on POST/PUT, but _not_ on GET? That seems weird too. Hrmmm. (cc @allait )

allait: It's worth checking what other write endpoints do, but I think if the user_id is not stored or used for integrity checks and is only present for authorization then we follow the same rule of checking in the frontend.

galund: We do store it, so we definitely need it. But I'm wondering whether we should be validating auth/access or not.

I really definitely still do think we need to have a way to start putting that logic into the API, but it can be a separate PR once we start on the multi-user thing. So we should try and have the conversation about the right way to do that soon.

app/main/views/direct_award.py
allait: Yep, good point, I should've checked first. It looks like they're only used to build up the URL for next page so it will keep whatever unused params were present in the original URL.

app/utils.py
samuelhwilliams: @galund So rip this out for now?

galund: Yes I suppose so.

It seems like we also have the "validate_and_return_updater_request", and that also has a user id - the one that gets written to the audit log. ISTM there's therefore already some ambiguity about which of the two user id's is actually "doing" the operation - and therefore which would even be the "right" one to use to do any access control against.

But I think please do add a TODO because we do need to come back to this after we discuss with more generally...

app/main/views/direct_award.py
```diff
@@ -0,0 +1,190 @@
+from flask import jsonify, abort, current_app, request
+from sqlalchemy.exc import IntegrityError
+from sqlalchemy import asc, desc
+
+from dmapiclient.audit import AuditTypes
+from dmutils.config import convert_to_boolean
+from .. import main
+from ... import db
+from ...models import User, AuditEvent
+from app.models.direct_award import Project, Search
```
allait: Why isn't this a relative import?

tests/main/views/test_direct_award.py
```diff
@@ -0,0 +1,497 @@
...
+
+class DirectAwardSetupAndTeardown(BaseApplicationTest, FixtureMixin):
+    def setup(self):
+        super(DirectAwardSetupAndTeardown, self).setup()
+        self.user_id = self.setup_dummy_user(id=1, role='buyer')
+        self.direct_award_project_name = DIRECT_AWARD_PROJECT_NAME
+        self.direct_award_search_url = DIRECT_AWARD_SEARCH_URL
+
+    def teardown(self):
+        super(DirectAwardSetupAndTeardown, self).teardown()
```
galund: Does this actually do anything? If we don't override the teardown function, it's still available to get called, right? Or is there something about the multiple inheritance thing that I'm forgetting...

samuelhwilliams: It's not necessary, yeah.

app/main/views/direct_award.py
galund: My bad - it got put in by PyCharm's refactoring tool rather than me, but I didn't spot that :)

app/main/views/direct_award.py
```diff
@@ -0,0 +1,190 @@
...
+    get_json_from_request, get_int_or_400, json_has_required_keys, pagination_links,
+    get_valid_page_or_1, validate_and_return_updater_request)
+
+
+@main.route('/direct-award/projects', methods=['GET'])
+def list_projects():
+    page = get_valid_page_or_1()
+
+    projects = Project.query
+    user_id = get_int_or_400(request.args, 'user-id')
```
galund: `app.main.views.briefs.create_brief has `userId`. I'm just really confused now about what standard we are supposed to be following :(

samuelhwilliams: Posts (i.e. json) will use camel case, gets (i.e. query params) use kebab case.

samuelhwilliams: I think we all find it confusing but it's 'consistent' with what we already have.

app/main/views/direct_award.py
samuelhwilliams: It is, what do you mean? 👍

app/main/views/direct_award.py
galund: Hmm so the comparison should have been with `app.main.views.briefs.list_briefs` which uses underscores!!

Arrrggghh :-D

PR #626, opened 2017-08-18
Model update add supplier details columns

TheDoubleK:
The first commit here (the DB migration file) is from #625, so will disappear from this PR once that is reviewed and merged.

The following commits have the corresponding model changes.  There were no tests of the suppliers model at all so I've added some.  In order to get them to work I had to add a couple of things: the `SERVER_NAME` variable in the test config, which allows `url_for` to work outside of a request context (which we need for directly testing model serialization functions that include links, e.g. to "self").  This also, for some reason, seems to need the `main` blueprint specified for finding the right route methods so I've added that and it lets the tests work while apparently making no difference to the running app :crossed_fingers:

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -826,9 +826,7 @@ class TestSuppliers(BaseApplicationTest, FixtureMixin):
             assert self.supplier.organisation_size == "medium"
             assert self.supplier.trading_status == "Sticky"
 
-    def test_serialization_of_supplier_with_all_details_added(self):
-        with self.app.app_context():
-            self._update_supplier_from_json_with_all_details()
+            # Check that serialization of a supplier with all details added looks as it should
             assert self.supplier.serialize() == {
                 'clients': ['Parcel Wrappers Ltd'],
                 'companiesHouseNumber': '98765432',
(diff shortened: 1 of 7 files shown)


tests/test_models.py
```diff
@@ -755,6 +755,108 @@ def test_answers_can_be_100_words(self):
...
+                'links': {'self': 'http://127.0.0.1:5000/suppliers/0'},
+                'name': 'String and Sticky Tape Inc.',
+                'organisationSize': 'medium',
+                'otherCompanyRegistrationNumber': '',
+                'registeredName': 'Tape and String Inc.',
+                'registrationCountry': 'Wales',
+                'registrationDate': '1973-08-10',
+                'tradingStatus': 'Sticky',
+                'vatNumber': '321321321',
+            }
```
risicle: The two above tests could effectively be merged into one - they have the same prologue and function call, just different assertions. Doing so just makes the tests execute that bit faster and style-wise it makes it clear that only a single path is actually being tested here. Of course, none of this is *important* important.

TheDoubleK: Hmm, my thinking when I wrote this was that the first tests the update_from_json method and the second ensures things are serialized as we expect, but in fact the serialization test depends on the update_from_json working properly, so yep maybe they should be assertions in the same test.  And like you say, those milliseconds are precious.

TheDoubleK:
I'm going to hold off merging this as I have a weird error updating contact details locally against this API branch. It's fine with current master so think I've broken something here :/

risicle:
Again, good (even better)

PR #627, opened 2017-08-21
flake8 config: ignore all venv* directories

risicle:
Certainly the nix configuration at least uses `venvx.y` `x.y` being the python version


benvand:
👍 Good shout

PR #629, opened 2017-08-22
Update apiclient

samuelhwilliams:
## Summary
Pull in apiclient==9.0.0 which removes the default index parameter from the Search API Client.

Blocked until https://github.com/alphagov/digitalmarketplace-apiclient/pull/86 is merged. Will need to regen requirements after that goes in.

## Ticket
Required work for https://trello.com/c/Hlq5mzAy/619-save-a-search-and-create-a-project

### DIFF
--- a/app/service_utils.py
+++ b/app/service_utils.py
@@ -135,7 +135,9 @@ def index_service(service):
         service.status == 'published'
     ):
         try:
-            search_api_client.index(service.service_id, service.serialize())
+            search_api_client.index(index=service.framework.slug,
+                                    service_id=service.service_id,
+                                    service=service.serialize())
         except dmapiclient.HTTPError as e:
             current_app.logger.warning(
                 'Failed to add {} to search index: {}'.format(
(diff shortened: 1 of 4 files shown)


app/service_utils.py
```diff
@@ -135,7 +135,9 @@ def index_service(service):
         service.status == 'published'
     ):
         try:
-            search_api_client.index(service.service_id, service.serialize())
+            search_api_client.index(index=service.framework.slug,
```
galund: Did not realise this function already existed :)

galund:
So the problem we need to figure out here, is how (if at all) to cover the scenario of multiple frameworks in one index.

The problem being e.g. in the case of editing a G-Cloud 8 service, where the correct index to use would probably have been called `g-cloud-6`.

Of course, one option will be for us to maintain an alias for every framework slug. And maybe that's a good idea.

samuelhwilliams:
We could do that, or we could use the inverse to `get_latest_live_framework` (e.g. a new `get_earliest_live_framework`). If they're live at the same time, they should be in the same index, and that index is the earliest still-live one, right?

galund:
Yeah, although get_earliest_live_framework would have to be available from here (i.e. the API) as well as from the frontends, and potentially the implementation in the two cases would look a little different.

It might be easier to maintain if we just say we will always create the alias whenever a framework goes live. We'll be doing that anyway if the new framework is going in a new index, so this is just about whether we _skip_ that step in the case of the new services going into an existing index?

samuelhwilliams:
Does sound easier. Let's go for that.

PR #633, opened 2017-08-25
supplier json schema: allow null companiesHouseNumber

risicle:
Broadly related to https://trello.com/c/v0ouuEMF/61-migrate-the-data-from-the-g9-dos2-declaration-to-account-level-2

We're supposed to be allowing a null `companiesHouseNumber` for suppliers but it would appear to be impossible to *set* it as such - clearly we were relying on the initial state of suppliers for those without CHIDs.

Also adds a test for this.

### DIFF
--- a/json_schemas/suppliers.json
+++ b/json_schemas/suppliers.json
@@ -11,9 +11,16 @@
       "type": "array"
     },
     "companiesHouseNumber": {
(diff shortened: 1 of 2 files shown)


TheDoubleK:
:+1: after a rebase.

PR #637, opened 2017-09-01
New cancelled and unsuccessful statuses for Brief

katstevens:
Trello ticket: https://trello.com/c/ggDGUY5d/823-2-collect-status-when-a-buyer-is-not-proceeding-with-an-award-no-journey-backend

The migration branch for these fields should be merged first (this branch contains the same commit to make sure tests past): https://github.com/alphagov/digitalmarketplace-api/pull/634

- new fields and status property
- status setting, filtering and ordering
- include the new statuses as actions in the `update_brief_status` route
- reorganise Brief tests

### DIFF
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -1210,6 +1210,8 @@ class Brief(db.Model):
                            default=datetime.utcnow, onupdate=datetime.utcnow)
     published_at = db.Column(db.DateTime, index=True, nullable=True)
     withdrawn_at = db.Column(db.DateTime, index=True, nullable=True)
+    cancelled_at = db.Column(db.DateTime, index=True, nullable=True)
+    unsuccessful_at = db.Column(db.DateTime, index=True, nullable=True)
 
     __table_args__ = (db.ForeignKeyConstraint([framework_id, _lot_id],
                                               ['framework_lots.framework_id', 'framework_lots.lot_id']),
@@ -1315,6 +1317,10 @@ class Brief(db.Model):
             return 'draft'
         elif self.applications_closed_at > datetime.utcnow():
             return 'live'
+        elif self.cancelled_at:
+            return 'cancelled'
+        elif self.unsuccessful_at:
+            return 'unsuccessful'
         elif self.awarded_brief_response:
             return 'awarded'
         else:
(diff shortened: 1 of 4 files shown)


migrations/versions/970_add_brief_unsuccessful_and_closed_statuses.py
```diff
@@ -0,0 +1,29 @@
+"""Add 2 new status columns; one for 'unsuccessful', one for 'cancelled'.
+
+Revision ID: 970
+Revises: 960
+Create Date: 2017-08-25 11:38:57.947569
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
```
benvand: Too many blank lines after imports.
PEP8 doesn't run here 😁

tests/main/views/test_briefs.py
```diff
@@ -744,6 +744,34 @@ def test_withdraw_a_brief(self):
...
+
+        assert res.status_code == 200
+        assert data['briefs']['status'] == 'cancelled'
+
+    def test_update_a_brief_as_unsuccessful(self):
+        self.setup_dummy_briefs(1, title='The Title', status='closed')
+
+        res = self.client.post(
+            '/briefs/1/unsuccessful',
+            data=json.dumps({
```
benvand: Just had to test this but we don't actually have to `json.dumps` the data. Although given it is consistent with the rest of the file it's the right thing to do here.

migrations/versions/970_add_brief_unsuccessful_and_closed_statuses.py
katstevens: Cool - I'll fix this in the migration PR (I have to update that anyway as someone else pushed a migration and the version numbers are out of date...)

katstevens:
Thanks Ben - I'll rebase this once the migration branch is merged as the first commit will give a conflict that'll need resolving.

app/models/main.py
```diff
@@ -1351,6 +1363,8 @@ def status(cls):
         "live": 0,
         "closed": 1,
         "awarded": 1,
+        "cancelled": 1,
+        "unsuccessful": 1,
```
Wynndow: What does this ordering do? I'm just curious as I've not used it before.

katstevens: On the opportunity listings, the Briefs were previously sorted by status/publish date/ID, where 'status' was descending alphabetically (because there were just `live` and `closed` in the list). Now we want `awarded`, `cancelled` and `unsuccessful` to have the same sort level as `closed` Briefs, so we have this mapping.

Wynndow: Nice.

PR #638, opened 2017-09-04
Add `downloaded_at`

samuelhwilliams:
## Summary
Add `downloaded_at` column to the DirectAwardProject table. This is required for the project/search overview page to identity that the "download shortlist" step has been completed.

## Migration PR
https://github.com/alphagov/digitalmarketplace-api/pull/639

## Ticket
https://trello.com/c/9YtT8DmX/675-task-list-saved-search-overview-page


app/models/direct_award.py
```diff
@@ -23,6 +24,7 @@ def serialize(self, with_users=False):
             "id": self.id,
             "name": self.name,
             "createdAt": self.created_at.strftime(DATETIME_FORMAT),
+            "downloadedAt": self.downloaded_at.strftime(DATETIME_FORMAT) if self.downloaded_at is not None else None,
```
galund: We could have a function in dmutils that knows how to handle the null case, if this gets very tedious, perhaps.

samuelhwilliams:
This PR will be picked up as part of a later ticket when the implementation for the downloaded_at parts of the API is required.

PR #642, opened 2017-09-07
Allow cancelled and unsuccessful updates to brief status from expired or closed framework.

benvand:
https://trello.com/c/Umt24W6d/816-3-collect-status-when-buyer-has-cancelled-procurement-no-journey-frontend

Briefs: https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/26

As a brief can never be published without the framework being in a live or expired state we don't need to perform checks here when changing the brief status to either `unsuccessful` or `cancelled`

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -174,9 +174,7 @@ def update_brief_status(brief_id, action):
     brief = Brief.query.filter(
         Brief.id == brief_id
     ).first_or_404()
-
-    if brief.framework.status != 'live':
-        abort(400, "Framework is not live")
+    framework_status = brief.framework.status
 
     action_to_status = {
         'publish': 'live',
@@ -184,9 +182,14 @@ def update_brief_status(brief_id, action):
         'cancel': 'cancelled',
         'unsuccessful': 'unsuccessful'
     }
-    if brief.status != action_to_status[action]:
-        previousStatus = brief.status
-        brief.status = action_to_status[action]
+    new_brief_status = action_to_status[action]
+
+    if new_brief_status in ['live', 'withdrawn'] and framework_status != 'live':
+        abort(400, "Framework is not live")
+
+    if brief.status != new_brief_status:
+        old_brief_status = brief.status
+        brief.status = new_brief_status
 
         if action == 'publish':
             validate_brief_data(brief, enforce_required=True)
@@ -196,7 +199,7 @@ def update_brief_status(brief_id, action):
             user=updater_json['updated_by'],
             data={
                 'briefId': brief.id,
-                'briefPreviousStatus': previousStatus,
+                'briefPreviousStatus': old_brief_status,
                 'briefStatus': brief.status,
             },
             db_object=brief,
--- a/tests/main/views/test_briefs.py
+++ b/tests/main/views/test_briefs.py
@@ -744,8 +744,14 @@ class TestUpdateBriefStatus(FrameworkSetupAndTeardown):
         assert data['briefs']['status'] == 'withdrawn'
         assert data['briefs']['withdrawnAt'] is not None
 
-    def test_cancel_a_brief(self):
+    @pytest.mark.parametrize('framework_status', ('pending', 'expired'))
+    def test_cancel_a_brief(self, framework_status):
         self.setup_dummy_briefs(1, title='The Title', status='closed')
+        with self.app.app_context():
+            framework = Framework.query.filter(Framework.slug == 'digital-outcomes-and-specialists').first()
+            framework.status = framework_status
+            db.session.add(framework)
+            db.session.commit()
 
         res = self.client.post(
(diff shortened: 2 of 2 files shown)


app/main/views/briefs.py
```diff
@@ -174,19 +174,24 @@ def update_brief_status(brief_id, action):
...
     }
-    if brief.status != action_to_status[action]:
-        previousStatus = brief.status
-        brief.status = action_to_status[action]
+    new_status = action_to_status[action]
+
+    if new_status in ['live', 'withdrawn'] and framework_status != 'live':
+        abort(400, "Framework is not live")
+    if new_status in ['cancelled', 'unsuccessful'] and framework_status not in ['live', 'expired']:
+        abort(400, "Framework is not live or expired")
```
katstevens: Nice catch, I hadn't thought about the need for cancelling briefs on older frameworks. We should check that the award path allows this to?

app/main/views/briefs.py
```diff
@@ -174,19 +174,24 @@ def update_brief_status(brief_id, action):
...
     action_to_status = {
         'publish': 'live',
         'withdraw': 'withdrawn',
         'cancel': 'cancelled',
         'unsuccessful': 'unsuccessful'
     }
-    if brief.status != action_to_status[action]:
-        previousStatus = brief.status
-        brief.status = action_to_status[action]
+    new_status = action_to_status[action]
```
katstevens: Much more readable 👍 But should we call it `new_brief_status` to differentiate between it and the `framework_status` (I'm easy either way)?

tests/main/views/test_briefs.py
```diff
@@ -836,6 +836,30 @@ def test_withdrawn_at_is_not_updated_if_withdrawn_brief_is_withdrawn(self):
         withdrawn_at = json.loads(res.get_data(as_text=True))['briefs']['withdrawnAt']
         assert withdrawn_at == original_withdrawn_at
 
+    @pytest.mark.parametrize('url_arg', ('cancel', 'unsuccessful'))
+    @pytest.mark.parametrize('framework_status', filter(lambda i: i not in ('live', 'expired'), Framework.STATUSES))
+    def test_cannot_make_brief_cancelled_or_unnsuccessfull_if_framework_is_not_live_or_expired(
```
katstevens: Typo in test name - `unsuccessful`

app/main/views/briefs.py
katstevens: (Just checked - the award path is fine)

app/main/views/briefs.py
```diff
@@ -174,19 +174,24 @@ def update_brief_status(brief_id, action):
...
+
+    if new_status in ['live', 'withdrawn'] and framework_status != 'live':
+        abort(400, "Framework is not live")
+    if new_status in ['cancelled', 'unsuccessful'] and framework_status not in ['live', 'expired']:
```
idavidmcdonald: I don't think we need this second check. If the framework is not live or expired then it will not be in a state where you could ever create a brief, let alone publish one

benvand: Yeah it will never be `closed` if the framework isn't `live` or `expired`

app/main/views/briefs.py
samuelhwilliams: Yeah, `old_brief_status`/`new_brief_status` would be a bit nicer.

app/main/views/briefs.py
```diff
@@ -174,19 +174,22 @@ def update_brief_status(brief_id, action):
...
         'cancel': 'cancelled',
         'unsuccessful': 'unsuccessful'
     }
-    if brief.status != action_to_status[action]:
-        previousStatus = brief.status
-        brief.status = action_to_status[action]
+    new_status = action_to_status[action]
+
+    if new_status in ['live', 'withdrawn'] and framework_status != 'live':
+        abort(400, "Framework is not live")
```
samuelhwilliams: Would be good to have a test around this case as it's a new bit of logic. Previously if the framework was not live, you couldn't use this endpoint at all - now you can if the update is to either `cancelled` or `unsuccessful`.

app/main/views/briefs.py
benvand: Done 👍

app/main/views/briefs.py
benvand: Updated existing test for this

benvand: The point that David made is that a brief can't be 'closed' or 'expired' unless its corresponding framework is 'open' or 'expired'. This means that all the validation we need is done in the status setter on the brief model when it checks that we cannot move to 'cancelled' or 'unsuccessful' for any status other than 'closed'.

benvand: @samuelhwilliams added a test to ensure we change when on the expected framework statuses

PR #644, opened 2017-09-08
Reset failed login count and unlock user on password change

allait:
This is an alternative to alphagov/digitalmarketplace-apiclient#93

We don't have a good use case for resetting the password without unlocking the user, so we could combine the two operations and unlock the user each time the password is updated.


TheDoubleK:
I prefer this way, but we can wait and see what consensus is if you like... 
:+1:

PR #648, opened 2017-09-12
Bump postgres version for Travis

samuelhwilliams:
## Summary
We use PostgreSQL 9.6 in production, so let's do the same in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 python:
   - "3.6"
 addons:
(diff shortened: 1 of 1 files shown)


allait:
We're on postgres 9.5 on all PaaS environments. Why did you think we were on 9.6?

samuelhwilliams:
Huh. I thought I had a conversation with someone a while ago where we decided we were on 9.6 I guess I was mistaken!

PR #650, opened 2017-09-15
Add a parameter to allow only returning active searches

samuelhwilliams:
## Summary
In the buyer-frontend we always ask for a list of searches associated to a Direct Award project, but we are currently only ever interested in the active search which leads to a lot of repeated filtering. This adds a query param to the endpoint that will cause only the active search(es) to be returned.

Bonus: Also adds 404s to direct award endpoints if the project doesn't exist, and tests to boot.

## Ticket
https://trello.com/c/igRhA3Bt/711-shortlist-file-download

### DIFF
--- a/app/main/views/direct_award.py
+++ b/app/main/views/direct_award.py
@@ -110,6 +110,9 @@ def list_project_searches(project_id):
     else:
         searches = searches.order_by(asc(DirectAwardSearch.id))
 
+    if convert_to_boolean(request.args.get('only-active', False)):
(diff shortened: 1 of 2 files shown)


app/main/views/direct_award.py
```diff
@@ -110,6 +113,9 @@ def list_project_searches(project_id):
     else:
         searches = searches.order_by(asc(DirectAwardSearch.id))
 
+    if convert_to_boolean(request.args.get('only-active', False)):
```
galund: I think I'd have called it `active-only`?

PR #652, opened 2017-09-18
Fix brief awarding tests, must go from-pending to awarded

benvand:
If we are instantiating an awarded `brief_response` the `awarded_at` validator (`validates_awarded_at`) is run against the kwargs that are passed in, and runs against them in order. Although the validator passes when we order the kwargs correctly this isn't representative of what happens in the app. 

What we really want the validator to do is prevent us going to ‘awarded’ from any state but ‘pending-awarded’. That is not what the validator is doing in this case.

We should instantiate all our brief responses in tests as pending-awarded then award them after. As would happen in the actual app.

### DIFF
--- a/tests/models/test_main.py
+++ b/tests/models/test_main.py
@@ -815,9 +815,10 @@ class TestBriefResponses(BaseApplicationTest, FixtureMixin):
                 data={},
                 supplier_id=0,
                 submitted_at=datetime.utcnow(),
-                award_details={'confirmed': 'details'},
-                awarded_at=datetime(2016, 1, 1)
+                award_details={'pending': True},
             )
+            brief_response.award_details = {'confirmed': 'details'},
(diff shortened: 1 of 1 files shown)


katstevens:
Yeah, this is a good shout Ben. I think we were just being 'economical' with these tests when we wrote them, taking advantage of the award_details being set first. Reflecting reality is definitely preferable!

Will the setup of awarded Briefs/BriefResponses can be standardised with FactoryBoy at all?

benvand:
It will indeed. Without going into too much detail the I'm going to make an `AwardedBriefFactory` that subclasses the `PendingBriefFactory`. The `AwardedBriefFactory` will then have a post generation hook that does the actual awarding 💥

benvand:
Happy to give this an approve?

katstevens:
Yep! Mergemergemerge!

PR #655, opened 2017-09-18
Store Search-API-relative URL for all saved searches.

galund:
As well as saving some bytes, and making our saved search data portable between environments, this removes a theoretical attack vector i.e. where a client sends a malicious URL into the API as a 'saved search'.

https://trello.com/c/jjVWnK1e/722-security-fixes-for-save-search-feature

### DIFF
--- a/app/url_utils.py
+++ b/app/url_utils.py
@@ -1,0 +1,24 @@
+from urllib.parse import urlsplit, urlunsplit
+
+
+def force_relative_url(base_url, url):
+    """
+    Forcibly removes the hostname and port from a URL, returning a new URL relative to base_url.
+    :param base_url: the path segment of base_url will be used to shorten the return url if possible
+    :param url: the URL to process into a relative URL
+    :return: sanitised URL relative to base_url
+    """
+    # convert args to str; urllib can accept bytes but our string ops can't
+    base_url, url = str(base_url), str(url)
+
+    # remove the netloc and port from the url...
+    parsed_url = urlsplit(url)
+    sanitised_url = urlunsplit(('', '', parsed_url.path, parsed_url.query, parsed_url.fragment))
+
+    # ...and remove any common path prefix up to and including the final slash
+    base_path = urlsplit(base_url).path or '/'
+    base_path = base_path[:base_path.rfind('/') + 1]
+    if sanitised_url.startswith(base_path):
+        sanitised_url = sanitised_url[len(base_path):]
+
+    return sanitised_url
--- a/app/models/direct_award.py
+++ b/app/models/direct_award.py
@@ -1,8 +1,11 @@
 from datetime import datetime
+from urllib.parse import urljoin
 
 from sqlalchemy.orm import validates
+from flask import current_app
 
 from app import db
+from app.url_utils import force_relative_url
 from app.models import User, ValidationError, ArchivedService
 from dmutils.formats import DATETIME_FORMAT
 
@@ -81,13 +84,14 @@ class DirectAwardSearch(db.Model):
                                postgresql_where=db.Column('active')),)
 
     def serialize(self):
+        resolved_search_url = urljoin(current_app.config['DM_SEARCH_API_URL'], self.search_url)
         return {
             "id": self.id,
             "createdBy": self.created_by,
             "projectId": self.project_id,
             "createdAt": self.created_at.strftime(DATETIME_FORMAT),
             "searchedAt": self.searched_at.strftime(DATETIME_FORMAT) if self.searched_at is not None else None,
-            "searchUrl": self.search_url,
+            "searchUrl": resolved_search_url,
             "active": self.active
         }
 
@@ -96,8 +100,12 @@ class DirectAwardSearch(db.Model):
         if self.project and self.project.locked_at:
             raise ValidationError('Cannot change attributes of a search under a project ({}) that '
                                   'has been locked.'.format(self.project_id))
+        if key == 'search_url' and value:
+            # remove hostname etc so that stored URI is always relative to the base Search API URI in our config
+            return force_relative_url(current_app.config['DM_SEARCH_API_URL'], value)
 
-        return value
+        else:
+            return value
 
 
 class DirectAwardSearchResultEntry(db.Model):
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -38,7 +38,9 @@ COMPLETE_DIGITAL_SPECIALISTS_BRIEF = {
 
 DIRECT_AWARD_PROJECT_NAME = 'My Direct Award Project'
 DIRECT_AWARD_FROZEN_TIME = '2017-01-01T00:00:00.000000Z'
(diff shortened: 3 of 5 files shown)


app/url_utils.py
```diff
@@ -0,0 +1,24 @@
...
+    # convert args to str; urllib can accept bytes but our string ops can't
+    base_url, url = str(base_url), str(url)
+
+    # remove the netloc and port from the url...
+    parsed_url = urlsplit(url)
+    sanitised_url = urlunsplit(('', '', parsed_url.path, parsed_url.query, parsed_url.fragment))
+
+    # ...and remove any common path prefix up to and including the final slash
+    base_path = urlsplit(base_url).path or '/'
+    base_path = base_path[:base_path.rfind('/') + 1]
```
benvand: Why do we care about the preceding slash? why not always have it or always don't have it?
Also why do we care about the matching bits? Can't we just store this as 
```
url = urlsplit(str(url))
urlunsplit(['', '', url.path, url.query, url.fragment])
```

app/url_utils.py
```diff
@@ -0,0 +1,24 @@
+from urllib.parse import urlsplit, urlunsplit
```
benvand: I'd drop this in the multi purpose utils file. Seems overkill to have another file for a single function.

app/url_utils.py
galund: So it seemed weird to have a field where the first character was _always_ a slash.

So then I wanted a consistent/logical process for getting rid of that first slash, and I wanted something that could be expressed in terms of an explainable function. So this is really to cover some edge cases that shouldn't ever be seen in real life, but are (somewhat?) explained by the tests.

I'll admit that those lines only got added after I'd factored the code into a separate function and given it a name, and then thought about the implications of the edge cases. They aren't really essential to what we're doing, but making this a function that will behave consistently with its 'reverse' i.e. `urljoin` seemed like a nice idea.

app/url_utils.py
galund: Personally I think multi-purpose utils files are the spawn of the devil. They only ever get longer and longer and longer and before you know it you have a 4000 line pile of spew. I've seen it (in a previous employment that had existed for longer than GDS, admittedly).

galund:
For added clarity: the weird prefix-removal is necessary in order for the last two test cases to pass, neither of which we expect to happen in real life in our application, but I think are necessary if the function is to have well-defined behaviour in all the edge cases. (Including the case of our search API being mounted somewhere other than `/`, which seems like a terribly bad idea but not entirely impossible.)

app/url_utils.py
benvand: Unless we're planning on splitting out that other file into different files, ie `json_utils`, `request_utils`, `data_utils` then an extra file is inconsistent. :man_shrugging:

app/url_utils.py
benvand: It only removes the preceding slash in some cases though. That's the bit that I don't get.

app/url_utils.py
samuelhwilliams: Got to start somewhere.

app/url_utils.py
galund: Yup, this is quite hard to explain. The case where we don't do anything at all is where there isn't any kind of common prefix: this means that *if* there is some non-empty path in the base URL, that is not a prefix of the URL we are trying to shorten, then we are going to have to eliminate that by making sure there's a slash on the frond.

The "is not a prefix" thing in that previous sentence then needs clarification, because in general you don't have to consider anything after the final slash when deciding if something is a prefix, because of how relative paths are interpreted.

Anyway, this is what the last test case is trying to cover, and it shouldn't ever happen with the kinds of URL we will be passing around.

app/url_utils.py
galund: Quite: councils of perfection that result in complete inaction aren't the incremental agile change we're signed up to ;-) In this context, inconsistency is inevitable and good, as long as we're agreed about the direction of travel.

galund: Incidentally I wanted to put this into dmutils, but I won't be doing that cos supporting python 2 is a PITA :-p

app/url_utils.py
benvand: but saying `c:my/weird/path` is the same as saying `c:/my/weird/path` so it doesn't matter if we always have a preceeding slash.
So we could just do this:
```
url = urlsplit(str(url))
path_string = urlunsplit(['', '', url.path, url.query, url.fragment])
return path_string if path_string.startswith('/') else '/' + path_string
```

benvand: Grim. Tested this on my windows box at home
`c:my/weird/path is the same as saying c:/my/weird/path`
It's not. I stand corrected. I thought `c:` pointed to the mount point as an absolute.



Here's what I'm unsure of:
OK so the point of this story is to remove the hostname and port to prevent people injecting malicious urls into our db?

What do you mean by 'mounted somewhere other than `/`'?

If all the paths are going to be relative then why does it matter if they begin with a slash or not?

app/url_utils.py
benvand: Creating inconsistencies in the code base can make you feel like you're doing the _right_ thing, and you might well be doing the right thing, in this case you probably are, butmy problem is that it inconsistency _reduces confidence_ in the code. People will be confused by the multiple conventions and will wonder why that code had to be kept separately, or be tacked on, or not be properly integrated.

It doesn't reduce confidence in all the other code that follows the same patterns (no matter how wrong they might be); It reduces confidence in the different code, the inconsistent code.

I personally thing that we should be using class based views for all our views but if I started writing class based views it would look weird and out of place. And besides, that wouldn't achieve my goal of all views being class based. It would just mean we had 2 paradigms until I classed up all the views.

It's all about the easiest way for you to achieve your goal of not having a big utils file:
Say that it falls outside the scope of this ticket. (If this conversation was the only thing holding up this PR how annoying would that be?)
Highlight it to other members of the team to make sure they agree.
Then either:
 * Create a tech debt ticket and get someone else to do it in a separate PR all at once 🎉 
 * Do it yourself in a separate PR all at once
That way the PR will be accepted quickly because people will see it's just arefacotr and not have to go digging through a refactor for functional changes

Half staring a refactor and tagging it onto a bug fix PR doesn't help you achieve your goal of not having a massive utils file. It makes the PR harder to read, introduces confusion and reduces confidence.

I'm not trying to 💩  on the idea or anything like that. I think breaking the code into other files is a good idea. But I think that if we do it, we should do the whole thing. Every code base I've ever worked with has had a legacy of accepting PRs that introduce new styles and paradigms and conventions without evaluating closely enough the effort it will take to implement them fully. In this case the effort is low, but the cost of failing to complete the refactor is high (reduced confidence)

That's why I think consistency is important, even if what you're being consistent with isn't quite perfect.

That's the last I'll say on it, it's up to you if you want to change it.

app/url_utils.py
samuelhwilliams: I _think_ George means in the case that we change our search-api to be mounted at, say, www.digitalmarketplace.service.gov.uk/search-api/

PR #656, opened 2017-09-19
implement list services API endpoint

samuelhwilliams:
## Summary
Adds the endpoint required for constructing a download file containing services matched by a Direct Award search. This endpoint paginates normally but we are going to limit the number of services that can be saved into a locked project to 100, which will effectively mean no pagination with our current configuration. We've had to limit this to 100 to remain responsive and minimise memory usage; ideally we'd have some kind of Digital Marketplace task runner that manage creating documents, combining lots of disparate data, etc, as opposed to doing it all on the frontends.

The list services endpoint accepts a `fields` parameter that allows the frontend to specify a subset of the Service model's `data` field to be returned, limiting the size of the response further. Ideally we would have a way of creating jobs for long-running tasks.

Tests for the list services endpoint requires archived services in the database, so I've slightly generalised the `setup_dummy_service` method to accept a model to use when setting up services. For Direct Award, this model is `ArchivedService`, as all locked searches store archived service instances which are (theoretically) immutable.

Also fixes a bug with the `lock_project` endpoint that wasn't checking for the currently `active` search. This could have been nasty. ☹️  Also stubbed out a test for this.

A second endpoint has also been added to record the downloading of a shortlist so that the tasklist page can correctly update to unlock the final steps. I have implemented this so that an AuditEvent is created every time the tasklist is downloaded, but the timestamp is only recorded for the first call.

## Depends upon
~~https://github.com/alphagov/digitalmarketplace-apiclient/pull/98~~

## Required by
https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/615
https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/32

## Other associated PRs
https://github.com/alphagov/digitalmarketplace-utils/pull/332
~~https://github.com/alphagov/digitalmarketplace-frameworks/pull/473~~

## Ticket
https://trello.com/c/igRhA3Bt/711-shortlist-file-download

### DIFF
--- a/app/main/views/direct_award.py
+++ b/app/main/views/direct_award.py
@@ -12,7 +12,7 @@ from ...models import User, AuditEvent, ArchivedService
 from ...models.direct_award import DirectAwardProject, DirectAwardSearch
 from ...utils import (
     get_json_from_request, get_int_or_400, json_has_required_keys, pagination_links,
-    get_valid_page_or_1, validate_and_return_updater_request)
+    get_valid_page_or_1, validate_and_return_updater_request, drop_all_other_fields)
 
 
 @main.route('/direct-award/projects', methods=['GET'])
@@ -201,7 +201,59 @@ def get_project_search(project_id, search_id):
 
 @main.route('/direct-award/projects/<int:project_id>/services', methods=['GET'])
 def list_project_services(project_id):
-    raise NotImplementedError()
+    """This endpoint returns all the services associated with a particular (locked) Direct Award Project. It returns
+    some fairly arbitrary data which is obviously not ideal, but as we don't have a task runner that we can utilise to
+    manage jobs like this and we need it to be responsive on-click, we're having to jerryrig in some data extraction and
+    transformation in here. Specifically we are returning supplier name and contact information, and returning from
+    the service data JSON keys as specified in the request (which will be loaded via a framework-specific manifest)."""
+    page = get_valid_page_or_1()
+    requested_fields = request.args.get('fields', '').split(',')
+    project = DirectAwardProject.query.filter(DirectAwardProject.id == project_id).first_or_404()
+
+    if not project.locked_at:
+        abort(400, 'Project has not been locked: {}'.format(project_id))
+
+    # TODO: This should work for _all_ active searches, not just the first (although we currently enforce only one
+    # TODO: active search) - SW 21/09/2017
+    search = DirectAwardSearch.query.filter(
+        DirectAwardSearch.project_id == project_id,
+        DirectAwardSearch.active == True  # noqa
+    ).first()
+    if not search:
+        abort(400, 'Project does not have a saved search: {}'.format(project_id))
+
+    paginated_archived_services = search.archived_services.paginate(
+        page=page,
+        per_page=current_app.config['DM_API_PROJECTS_PAGE_SIZE'],
+    )
+
+    project_archived_services = list(map(lambda service: {
+        'id': service.service_id,
+        'supplier': {
+            'name': service.supplier.name,
+            'contact': {
+                'name': service.supplier.contact_information[0].contact_name,
+                'phone': service.supplier.contact_information[0].phone_number,
+                'email': service.supplier.contact_information[0].email,
+            },
+        },
+        'data': drop_all_other_fields(service.data, requested_fields),
+    }, paginated_archived_services.items))
+
+    pagination_params = request.args.to_dict()
+    pagination_params['project_id'] = project_id
+
+    return jsonify(
+        services=project_archived_services,
+        meta={
+            "total": paginated_archived_services.total,
+        },
+        links=pagination_links(
+            paginated_archived_services,
+            '.list_project_services',
+            pagination_params
+        ),
+    )
 
 
 @main.route('/direct-award/projects/<int:project_id>/lock', methods=['POST'])
@@ -214,7 +266,8 @@ def lock_project(project_id):
         abort(400, 'Project has already been locked: {}'.format(project_id))
 
     search = DirectAwardSearch.query.filter(
-        DirectAwardSearch.project_id == project_id
+        DirectAwardSearch.project_id == project_id,
+        DirectAwardSearch.active == True,  # noqa
     ).first_or_404()
 
     now = datetime.datetime.utcnow()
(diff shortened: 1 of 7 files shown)


galund:
Make sure that filing out that test is captured in a ticket, either this one or another one :)

samuelhwilliams:
Filling in those tests is part of the end search ticket, but I'll drop a comment in there just to make sure Dilwoar is aware.

app/main/views/direct_award.py
```diff
@@ -201,7 +201,60 @@ def get_project_search(project_id, search_id):
...
+    requested_fields = request.args.get('fields', '').split(',')
+    project = DirectAwardProject.query.filter(DirectAwardProject.id == project_id).first_or_404()
+
+    if not project.locked_at:
+        abort(400, 'Project has not been locked: {}'.format(project_id))
+
+    search = DirectAwardSearch.query.filter(
+        DirectAwardSearch.project_id == project_id,
+        DirectAwardSearch.active == True  # noqa
+    ).first()
```
galund: FWIW I think this should be the combination of results from all active searches. The fact that only one search can be active at a time would be better managed in just one place? (i.e. no assumption made here)

samuelhwilliams: I was going to do that initially and decided against it... 🙈

app/main/views/direct_award.py
```diff
@@ -201,7 +201,60 @@ def get_project_search(project_id, search_id):
...
+        'id': service.service_id,
+        'supplier': {
+            'name': service.supplier.name,
+            'contact': {
+                'name': service.supplier.contact_information[0].contact_name,
+                'phone': service.supplier.contact_information[0].phone_number,
+                'email': service.supplier.contact_information[0].email,
+            },
+        },
+        'data': drop_all_other_fields(service.data, requested_fields),
```
galund: I'm not sure I entirely hold with this "recursively wipe fields named/not-named X regardless of where they are in the structure" thing. Feels like we should know the structure, and wipe the specific things from where they are. Otherwise in a world of properly-nested related objects, random bits of other objects get wiped if the names clash. But I get that this is what we already do, so eh.

tests/main/views/test_direct_award.py
```diff
@@ -534,6 +538,107 @@ def test_get_search_returns_serialized_search(self):
...
+                                                                                     self.user_id))
+        data = json.loads(res.get_data(as_text=True))
+
+        assert res.status_code == 200
+
+        assert len(data['services']) == self.app.config['DM_API_PROJECTS_PAGE_SIZE']
+        assert data['meta']['total'] == project_services_count
+
+        next_url = 'http://127.0.0.1:5000/direct-award/projects/{}/services?user-id={}&page=2'.format(
+            self.project_id, self.search_id, self.user_id
```
DilwoarH: where is the search_id being added in?

app/main/views/direct_award.py
galund: okay maybe we can just add a to-do in then

app/main/views/direct_award.py
samuelhwilliams: This call doesn't recurse, otherwise yes that wouldn't be good.

tests/main/views/test_direct_award.py
samuelhwilliams: Good spot.

app/main/views/direct_award.py
```diff
@@ -251,3 +304,31 @@ def lock_project(project_id):
...
+
+
+@main.route('/direct-award/projects/<int:project_id>/record-download', methods=['POST'])
+def record_project_download(project_id):
+    updater_json = validate_and_return_updater_request()
+
+    project = DirectAwardProject.query.filter(DirectAwardProject.id == project_id).first_or_404()
+
+    # Up for discussion: should this endpoint abort if downloaded_at is already set, or should it continue (so that
+    # and audit event is created for every download)?
```
galund: FWIW I think if we are having a separate endpoint, then the expectation is that you can call it separately to getting the download? In which case it would really make sense to audit every download.

In terms of the language on the frontend, I think (but @karlchillmaid may differ) that if we are talking about "when was this thing downloaded", then the _assumption_ would be "when _last_ did that happen". And if you meant "when was the project _first_ downloaded" then we'd have needed to call that out.

But I'm not quite sure why we're recording this anyway, except so that we can have one extra step that can get ticked :-)

tests/main/views/test_direct_award.py
```diff
@@ -24,6 +29,18 @@ def setup(self):
...
+        return {
+            'project': {
+                'userId': self.user_id,
+                'name': self.direct_award_project_name,
+            },
+            'updated_by': str(self.user_id)
+        }.copy()
+
+    def _updated_by_data(self):
+        return {'updated_by': str(self.user_id)}.copy()
```
galund: Why do we need `copy` here? A new dict will get created on every call anyway... won't it?

samuelhwilliams: Looks like you're right.

app/main/views/direct_award.py
samuelhwilliams: That's really the only reason - so that the last state of the tasklist is unlocked. We don't ever show on the frontend the datetime of the download, so it doesn't matter overly much if we record the first or last timestamp. Happy to change if you'd prefer.

galund: I guess I'd marginally prefer to just keep recording the last date, yeah? But makes little odds if we aren't showing it for sure.

samuelhwilliams: I guess we'll have the audit events if we ever need to go back and see, so sure.

app/models/direct_award.py
```diff
@@ -17,6 +17,7 @@ class DirectAwardProject(db.Model):
     name = db.Column(db.String(255), nullable=False)
     created_at = db.Column(db.DateTime, default=datetime.utcnow, nullable=False)
     locked_at = db.Column(db.DateTime, nullable=True)  # When the project's active search/es are final and cannot change
+    downloaded_at = db.Column(db.DateTime, nullable=True)  # When the project's shortlist was first downloaded
```
galund: The comment here is now wrong

PR #663, opened 2017-10-04
Return live at the end of the Brief.status hybrid property checks

samuelhwilliams:
## Summary
Our preview smoke tests are currently failing due to some manually mutated data on a few briefs. However, this has highlighted that the ordering of checks in the Brief.status field are wrong - it should only return live if it doesn't meet the criteria for other statuses that should take precedence (cancelled, withdrawn).

Also added cancelledAt and unsuccessfulAt to the serialization, as if these keys were present in the response it would have helped us debug this faster.

### DIFF
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -1349,6 +1349,7 @@ class Brief(db.Model):
         return sql_case([
             (cls.withdrawn_at.isnot(None), 'withdrawn'),
             (cls.published_at.is_(None), 'draft'),
+            (cls.applications_closed_at > datetime.utcnow(), 'live'),
             (cls.cancelled_at.isnot(None), 'cancelled'),
             (cls.unsuccessful_at.isnot(None), 'unsuccessful'),
             (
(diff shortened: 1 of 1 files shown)


app/models/main.py
```diff
@@ -1479,6 +1479,16 @@ def serialize(self, with_users=False, with_clarification_questions=False):
...
+        if self.unsuccessful_at:
+            data.update({
+                'unsuccessfulAt': self.unsuccessful_at.strftime(DATETIME_FORMAT)
+            })
+
+        if self.cancelled_at:
+            data.update({
+                'cancelledAt': self.cancelled_at.strftime(DATETIME_FORMAT)
+            })
+
```
katstevens: Good catch on these! I honestly thought they had already been included. I'll have a quick check on the Briefs FE to see if we were expecting them in the response already...

app/models/main.py
```diff
@@ -1315,14 +1315,14 @@ def status(self):
...
-        elif self.applications_closed_at > datetime.utcnow():
-            return 'live'
         elif self.cancelled_at:
             return 'cancelled'
         elif self.unsuccessful_at:
             return 'unsuccessful'
         elif self.awarded_brief_response:
             return 'awarded'
+        elif self.applications_closed_at > datetime.utcnow():
+            return 'live'
```
katstevens: 👍 for making the ordering consistent - this was the intended priority!

app/models/main.py
katstevens: Confirmed that Briefs FE is only checking the status, not the datestamp.

PR #664, opened 2017-10-04
60 Remove unused fields from model and drop constraint

MuriloDalRi:
## Summary
In the supplier information rework we have several data fields now completely unused.  We should remove them from our API data model and drop them from the database.

## Ticket
https://trello.com/c/hAY4ccVY/60-tidy-up-unused-supplier-information-fields

This has to go in before #665


migrations/versions/1000_nullable_clients_field.py
```diff
@@ -0,0 +1,32 @@
...
+revision = '1000'
+down_revision = '990'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('suppliers', 'clients',
+               existing_type=postgresql.JSON(astext_type=sa.Text()),
+               nullable=True)
+    # ### end Alembic commands ###
```
katstevens: I think we've been removing the Alembic-autogenerated comments on upgrade and downgrade - definitely not a blocker though!

MuriloDalRi: Oh I was planning to blame Alembic if things broke! :trollface:

PR #672, opened 2017-10-11
DB migration to add new buyer email domain table

katstevens:

PR 1: schema migration to add a simple DB table to store the email domain (currently held in a text file).

Related:
PR 2: Add API client method and AuditType https://github.com/alphagov/digitalmarketplace-apiclient/pull/102
PR 3: API model/add domain view/backwards compatibility for existing domain checks https://github.com/alphagov/digitalmarketplace-api/pull/674
PR 4: data migration to transfer domains from text file to DB https://github.com/alphagov/digitalmarketplace-api/pull/675

### DIFF
--- a/migrations/versions/1020_add_buyer_email_domain_table.py
+++ b/migrations/versions/1020_add_buyer_email_domain_table.py
@@ -1,0 +1,33 @@
+"""Add buyer email domain table
+
+Revision ID: 1020
+Revises: 1010
+Create Date: 2017-10-10 15:18:22.683693
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '1020'
+down_revision = '1010'
+
+
+def upgrade():
+    op.create_table(
+        'buyer_email_domains',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('domain_name', sa.String(), nullable=False),
+        sa.PrimaryKeyConstraint('id', name=op.f('buyer_email_domains_pkey'))
+    )
+    op.create_unique_constraint(
+        op.f('uq_buyer_email_domains_domain_name'), 'buyer_email_domains', ['domain_name']
+    )
+
+
+def downgrade():
+    op.drop_constraint(
+        op.f('uq_buyer_email_domains_domain_name'), 'buyer_email_domains', type_='unique'
+    )
+    op.drop_table('buyer_email_domains')


migrations/versions/1020_add_buyer_email_domain_table.py
```diff
@@ -0,0 +1,27 @@
...
+# revision identifiers, used by Alembic.
+revision = '1020'
+down_revision = '1010'
+
+
+def upgrade():
+    op.create_table(
+        'buyer_email_domains',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('domain_name', sa.String(length=255), nullable=False),
```
allait: Should we add a uniqueness constraint on the domain?

katstevens: We could do - the validation around what counts as an 'existing' domain is a bit more complicated so I thought it might be easier to keep it in the code, but there'd be no harm having `unique=True` as well I suppose.

TheDoubleK: I don't think we ever set a length for our String fields anywhere else - we just have `sa.String()` in our migrations.  I don't see any reason to be explicit here?  Varchar FTW!

katstevens: That was auto-generated by Alembic because it was on my model, which I copied the structure of from the most recent model i.e. `DirectAwardProject` 😸 We seem to specify `String(255)` for a lot of our `name` columns but I agree that it doesn't seem necessary here. Happy to remove!

katstevens:
Uniqueness constraint now added (and the string length removed).

TheDoubleK:
One of us is going to have to renumber our migration...

migrations/versions/1020_add_buyer_email_domain_table.py
```diff
@@ -0,0 +1,33 @@
...
+
+def upgrade():
+    op.create_table(
+        'buyer_email_domains',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('domain_name', sa.String(), nullable=False),
+        sa.PrimaryKeyConstraint('id', name=op.f('buyer_email_domains_pkey'))
+    )
+    op.create_unique_constraint(
+        op.f('uq_buyer_email_domains_domain_name'), 'buyer_email_domains', ['domain_name']
```
TheDoubleK: Not sure about the naming of the constraint here, or about Postgres implementation of unique constraints, but running the migration we get  a listing under "Indexes" - all our other indexes are prefixed with `ix_` rather than `uq_`.  I have no idea if this has any broader implications than consistency/tidiness... @allait might know more/have opinions on this.  There was something sometime about a pattern that gets used to generate these names, but I can't really remember the details.

```
Indexes:
    "buyer_email_domains_pkey" PRIMARY KEY, btree (id)
    "uq_buyer_email_domains_domain_name" UNIQUE CONSTRAINT, btree (domain_name)
```

TheDoubleK: I'm not saying there's anything wrong, just I'd like another opinion before approving.

katstevens: The name was autogenerated by Alembic, if that makes any difference?

TheDoubleK: Yeah, I think that means it's probably the right name :)

PR #673, opened 2017-10-11
End your search test

DilwoarH:
As a G-Cloud buyer
I need a fixed list of services
so that I can start the shortlisting process

Ticket: https://trello.com/c/gs2I3Cj2/676-end-your-search

### DIFF
--- a/tests/main/views/test_direct_award.py
+++ b/tests/main/views/test_direct_award.py
@@ -5,18 +5,25 @@ import math
 import random
 import sys
 
+import mock
+from app import db
+
 import pytest
-from tests.helpers import FixtureMixin
 from tests.bases import BaseApplicationTest
 
+from sqlalchemy import desc
 from dmapiclient.audit import AuditTypes
 
-from app import db
 from app.models import DATETIME_FORMAT, AuditEvent, User, ArchivedService
 from app.models.direct_award import (
-    DirectAwardProjectUser, DirectAwardSearch, DirectAwardProject, DirectAwardSearchResultEntry
+    DirectAwardProjectUser,
+    DirectAwardSearch,
+    DirectAwardProject,
+    DirectAwardSearchResultEntry
 )
-from ...helpers import DIRECT_AWARD_SEARCH_URL, DIRECT_AWARD_PROJECT_NAME, DIRECT_AWARD_FROZEN_TIME
+from ...helpers import (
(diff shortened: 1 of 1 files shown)


tests/main/views/test_direct_award.py
```diff
@@ -659,30 +663,71 @@ def _lock_project_data(self):
...
+            }),
+            content_type='application/json')
+
+        assert res.status_code == 404
 
-    @pytest.mark.skip
-    def test_lock_project_saved_result_entries_are_for_the_active_search_in_the_project(self):
-        pass
+    @mock.patch('app.main.views.direct_award.search_api_client')
+    def test_lock_project_success(self, search_api_client):
```
samuelhwilliams: The name of this test ('success') is pretty generic. It doesn't help someone reading it understand what, specifically, is being tested.

galund: I think this is ok - there's a single path which is "all the things that happen when the user successfully locks a project". We don't want to rename the function whenever we find more side-effects to test, and we didn't want separate test functions because we'd end up testing the same code over and over.

If there were more pathways that were 'successful' then yes we'd need to distinguish. And particular failure modes need to be in separate tests too, for sure.

DilwoarH:
@samuelhwilliams - is this fine to merge?

tests/main/views/test_direct_award.py
```diff
@@ -659,29 +666,98 @@ def _lock_project_data(self):
...
+                filter(ArchivedService.service_id.in_([service_id])).\
+                group_by(ArchivedService.id).\
+                order_by(ArchivedService.service_id, desc(ArchivedService.id)).\
+                distinct(ArchivedService.service_id).all()
+
+            assert len(archived_services) > 0
+            assert search_result_entry.count() == 1
+            assert search_result_entry.all()[0].archived_service_id == archived_services[0].id
+
+    def _create_service(self):
```
samuelhwilliams: This should be moved higher up (i.e probably after setup). Also it's creating and updating a service, so probably call it `_create_service_and_update`.

tests/main/views/test_direct_award.py
```diff
@@ -659,29 +666,98 @@ def _lock_project_data(self):
...
+        service['createdAt'] = "2017-12-23T14:51:19Z"
+
+        res2 = self.client.post(
+            '/services/{}'.format(service['id']),
+            data=json.dumps(
+                {
+                    'updated_by': 'example',
+                    'services': service,
+                    'update_details': {
+                        'updated_by': 'example'
```
samuelhwilliams: Does this json blob really need updated_by _and_ update_details->updated_by?

PR #674, opened 2017-10-13
Create buyer email domains via the API

katstevens:

- Add the `BuyerEmailDomain` model (used to generate the schema migration in PR 1) and a `serialize` method
- New route to create new `BuyerEmailDomain` objects
- Validate requests using JSON schema
- Ensure buyers with a domain already approved (either in the DB or the .txt file) can still create an account

NB This branch is based on the schema migration branch in order to make the tests pass, so the migration will be included in the diff until that gets merged. Travis will fail on this PR until the APIClient PR is merged. 

This is PR 3 out of 4.
PR 1: Schema migration alphagov/digitalmarketplace-api#672
PR 2: API Client https://github.com/alphagov/digitalmarketplace-apiclient/pull/102
PR 4: Data migration https://github.com/alphagov/digitalmarketplace-api/pull/675

### DIFF
--- a/app/main/views/buyer_domains.py
+++ b/app/main/views/buyer_domains.py
@@ -5,7 +5,7 @@ from flask import jsonify, abort
 from .. import main
 from ... import db
 from ...models import BuyerEmailDomain, AuditEvent
-from ...validation import validate_buyer_email_domain_json_or_400, buyer_email_already_approved
+from ...validation import validate_buyer_email_domain_json_or_400, buyer_email_domain_approved
 from ...utils import get_json_from_request, json_has_required_keys, validate_and_return_updater_request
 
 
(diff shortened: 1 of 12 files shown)


allait:
I feel like even for MVP we should have at least a list domains endpoint, otherwise there's no way to review the list.

If we're planning to add admin features for this I think delete could be pretty important as well, because typos will happen sooner or later. We'd need to do something about buyer accounts that only exist thanks to that domain record though, which could be an argument for making this check  affect user `active` flag on each read for example.

katstevens:
We did think about whether list/delete endpoints were necessary just yet - I think we agreed they're probably out of scope for this particular story. Happy to discuss further on Monday with the team?

app/main/views/buyer_domains.py
```diff
@@ -0,0 +1,71 @@
...
+from ... import db
+from ...models import BuyerEmailDomain, AuditEvent
+from ...validation import validate_buyer_email_domain_json_or_400, buyer_email_domain_approved
+from ...utils import (
+    get_json_from_request, json_has_required_keys, validate_and_return_updater_request,
+    get_valid_page_or_1, pagination_links
+)
+
+
+@main.route('/buyer-email-domain', methods=['POST'])
```
allait: This and the GET one should probably be plural `/buyer-email-domains`, since that's what we do with every other endpoint

katstevens: Yikes, good catch 👍

katstevens: Now done - @allait could you sanity check the corresponding API client PR? https://github.com/alphagov/digitalmarketplace-apiclient/pull/106

PR #675, opened 2017-10-13
 Data migration to move buyer domains from .txt file to DB

katstevens:

Once the migration has been applied, we can delete the .txt file in a later branch and this migration will exit early.

NB this branch is based off the `buyer-email-domains-to-db` which contains the schema migration and the main API changes. (Travis will fail until the API Client is merged.) This should be merged last of all.

The text file and its associated functions/tests can be safely deleted once all four PRs for this ticket are deployed.

Related:
PR 1: Schema migration alphagov/digitalmarketplace-api#672
PR 2: API Client alphagov/digitalmarketplace-apiclient#102
PR 3: API model and route alphagov/digitalmarketplace-api#674

NB I have run the data migration locally and it seems to work fine.

### DIFF
--- a/migrations/versions/1030_buyer_email_domain_data_migration.py
+++ b/migrations/versions/1030_buyer_email_domain_data_migration.py
@@ -1,0 +1,45 @@
+"""Migrate buyer email domains from .txt file to DB
+
+Revision ID: 1030
+Revises: 1020
+Create Date: 2017-10-13 11:18:22.683693
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '1030'
+down_revision = '1020'
+
+buyer_domains_table = sa.Table(
+    'buyer_email_domains',
+    sa.MetaData(),
+    sa.Column('id', sa.Integer, primary_key=True),
+    sa.Column('domain_name', sa.String, nullable=True),
+)
+
+
+def upgrade():
+    try:
+        with open('./data/buyer-email-domains.txt') as f:
+            legacy_buyer_domains = [line.strip() for line in f]
+    except FileNotFoundError as e:
+        # Once we have deleted the file, we can exit the migration early
+        return
+
+    conn = op.get_bind()
+    for legacy_domain in legacy_buyer_domains:
+        # SELECT domain_name FROM buyer_email_domains WHERE domain_name == 'legacy-domain.org'
+        query = buyer_domains_table.select(buyer_domains_table.c.domain_name == legacy_domain)
+
+        result = conn.execute(query).first()
+        if not result:
+            # INSERT INTO buyer_email_domains (domain_name) VALUES ("legacy-domain.org");
+            query = buyer_domains_table.insert().values(domain_name=legacy_domain)
+            conn.execute(query)
+
+
+def downgrade():
+    pass


app/main/views/buyer_domains.py
```diff
@@ -0,0 +1,46 @@
...
+    updater_json = validate_and_return_updater_request()
+
+    json_payload = get_json_from_request()
+    json_has_required_keys(json_payload, ["buyerEmailDomains"])
+    validate_buyer_email_domain_json_or_400(json_payload["buyerEmailDomains"])
+
+    new_domain = json_payload["buyerEmailDomains"]['domainName'].lower()
+
+    if buyer_email_domain_approved(BuyerEmailDomain.query.all(), new_domain):
+        abort(409, "Domain name {} has already been approved".format(new_domain))
```
TheDoubleK: 409, fancy!

app/main/views/users.py
```diff
@@ -293,7 +294,12 @@ def email_has_valid_buyer_domain():
     if not email_address:
         abort(400, "'email_address' is a required parameter")
 
-    return jsonify(valid=is_valid_buyer_email(email_address))
+    domain_ok = buyer_email_domain_approved(BuyerEmailDomain.query.all(), email_address.split('@')[-1])
+    if not domain_ok:
+        # If the data migration hasn't happened yet, check the .txt file
```
TheDoubleK: The migration runs before new code gets deployed, so I don't think we should need this? i.e. the reading of the text file will definitely have happened by the time this code is live.

And if not I think we can take the risk of a buyer user trying to sign up in the few seconds it takes to run the migration, failing once, and then succeeding the next time they try it.

app/models/buyer_domains.py
```diff
@@ -0,0 +1,14 @@
+from app import db
+
+
+class BuyerEmailDomain(db.Model):
+    __tablename__ = 'buyer_email_domains'
+
+    id = db.Column(db.Integer, primary_key=True)
+    domain_name = db.Column(db.String(255), nullable=False)
```
TheDoubleK: We dropped the 255 from the table definition, should also drop here.

json_schemas/buyer-email-domains.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "additionalProperties": false,
+  "properties": {
+    "domainName": {
+      "pattern": "^(?!:\/\/)([a-zA-Z0-9-_]+\\.)*[a-zA-Z0-9][a-zA-Z0-9-_]+\\.[a-zA-Z]{2,64}?$",
```
TheDoubleK: While I'm pretty sure that actual government buyer domains will be ascii and digits, I'm not sure we should be so restrictive here - internationalised domain names are a real thing, and I don't see a good reason to exclude them here.

I think the main things we should check are that there is no `@` and at least one `.` - beyond that all bets are off. (We could have a length restriction - probably sensible... 255?)

json_schemas/buyer-email-domains.json
```diff
@@ -0,0 +1,13 @@
...
+      "type": "string"
+    }
+  },
+  "required": ["domainName"],
+  "title": "Create Buyer Domain Validation Schema",
```
TheDoubleK: Could just be `Buyer Domain Schema` - we don't mention the action ("Create") or validation in any of our other schema titles.

migrations/versions/1030_buyer_email_domain_data_migration.py
```diff
@@ -0,0 +1,47 @@
...
+
+        result = conn.execute(query).first()
+        if not result:
+            # INSERT INTO buyer_email_domains (domain_name) VALUES ("legacy-domain.org");
+            query = buyer_domains_table.insert().values(domain_name=legacy_domain)
+            conn.execute(query)
+
+
+def downgrade():
+    pass
```
TheDoubleK: Could do a thing to delete all domains that don't have a "create_buyer_domain" audit event associated with them, but... nah 😛

migrations/versions/1030_buyer_email_domain_data_migration.py
```diff
@@ -0,0 +1,47 @@
...
+    sa.Column('domain_name', sa.String, nullable=True),
+)
+
+
+def upgrade():
+    try:
+        with open('./data/buyer-email-domains.txt') as f:
+            legacy_buyer_domains = [line.strip() for line in f]
+    except FileNotFoundError as e:
+        # When we eventually delete the file, we can exit the migration early
```
TheDoubleK: "Eventually"?! We should do it right away.

migrations/versions/1030_buyer_email_domain_data_migration.py
```diff
@@ -0,0 +1,47 @@
...
+        return
+
+    conn = op.get_bind()
+    for legacy_domain in legacy_buyer_domains:
+        # SELECT domain_name FROM buyer_email_domains WHERE domain_name == 'legacy-domain.org'
+        query = buyer_domains_table.select(
+            buyer_domains_table.c.domain_name == legacy_domain
+        ).with_only_columns([buyer_domains_table.c.domain_name])
```
benvand: Don't think the `with_only_columns` gains us anything here. An `exists` would but only if we had tens of thousands of columns. Interesting though. Not a dealbreaker.

app/models/buyer_domains.py
katstevens: Yep - that should disappear on the rebase.

migrations/versions/1030_buyer_email_domain_data_migration.py
katstevens: Yep, good shout. I stuck `with_only_columns` in there as the last data migration (I think on the `Briefs` model) used it. Will remove so as to keep things as simple as possible.

app/main/views/users.py
katstevens: Sorry for the confusion - this branch with the data migration is just 1 commit, the code changes are in https://github.com/alphagov/digitalmarketplace-api/pull/674 and will be merged separately.

migrations/versions/1030_buyer_email_domain_data_migration.py
katstevens: I promise to do another PR as soon as this one has deployed 😸

tests/main/views/test_buyer_domains.py
```diff
@@ -0,0 +1,186 @@
...
+
+            assert audit.type == "create_buyer_email_domain"
+            assert audit.user == "example user"
+            assert audit.data == {
+                'buyerEmailDomainId': buyer_email_domain.id,
+                'buyerEmailDomainJson': {'domainName': 'example.com'}
+            }
+
+    @pytest.mark.parametrize(
+        'new_domain', ['no-dots', 'at@symbol.org', 'contains space.org', 'bad!char.org', 'tld-too-short.x']
```
TheDoubleK: Would also be nice to have a test case for a plain TLD like `.com`, which shouldn't be allowed.  And maybe we don't allow dots at the start at all?  e.g. should `.gov.uk` fail but `gov.uk` be allowed?

tests/main/views/test_users.py
```diff
@@ -1475,13 +1475,23 @@ def test_response_agreed_contract_variation(self):
...
+        assert response.status_code == 200
+        assert json.loads(response.get_data())['valid'] is True
+
+    def test_valid_email_is_ok_if_domain_found_in_text_file(self):
         response = self.client.get('/users/check-buyer-email', query_string={'email_address': 'buyer@gov.uk'})
         assert response.status_code == 200
         assert json.loads(response.get_data())['valid'] is True
 
     def test_invalid_email_is_not_ok(self):
-        response = self.client.get('/users/check-buyer-email', query_string={'email_address': 'someone@notgov.uk'})
```
TheDoubleK: I quite liked `notgov.uk` here.

tests/models/test_buyer_domains.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def test_create_new_buyer_email_domain(self):
+        with self.app.app_context():
+
+            buyer_domain = BuyerEmailDomain(domain_name="superkalifrajilisticexpialidocious.org.uk")
+            db.session.add(buyer_domain)
+            db.session.commit()
+
+            assert buyer_domain.id is not None
+            assert buyer_domain.domain_name == "superkalifrajilisticexpialidocious.org.uk"
```
TheDoubleK: Typo => should be `supercalifragilisticexpialidocious` :trollface:

app/models/buyer_domains.py
TheDoubleK: OK, sorry, haven't figured out what belongs to what.

tests/models/test_buyer_domains.py
katstevens: 🙀 Will rectify immediately!

tests/main/views/test_users.py
katstevens: Happy to put it back.

json_schemas/buyer-email-domains.json
katstevens: https://blog.farrant.me/getting-an-emoji-domain-in-2015/ 😹 

Happy to change this. I think I'd just nabbed the regex above from an example online. I think the TLD part is ok (apparently the max length for TLDs is 64 otherwise the DNS can't resolve!), so how about something like 

```
"^([^@\s])+\.[a-zA-Z]{2,64}$"
```

This should skip `@` and whitespace, and require at least one dot before a sane TLD. My regex powers are pretty dire so any advice gratefully received!

json_schemas/buyer-email-domains.json
katstevens: Cool - now done.

tests/main/views/test_buyer_domains.py
```diff
@@ -0,0 +1,143 @@
...
+    def test_create_buyer_email_domain_with_no_data(self):
+        res = self.client.post(
+            '/buyer-email-domain',
+            content_type='application/json'
+        )
+
+        assert res.status_code == 400
+
+    @pytest.mark.parametrize('new_domain', ["fine.org", "also-fine.org", "also.fine.org"])
+    @pytest.mark.parametrize('existing_domain', ["ine.org", ".org", "o-fine.org"])
```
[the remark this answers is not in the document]
katstevens: Agreed. I think the current regex wouldn't allow `.org` on the request, however as this test is creating it directly in the DB we're bypassing validation, so this is misleading. I'll remove it.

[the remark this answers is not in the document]
katstevens: Now done - have added a TLD as an invalid case.

tests/main/views/test_buyer_domains.py
```diff
@@ -0,0 +1,186 @@
...
+            '/buyer-email-domain',
+            data=json.dumps({
+                'buyerEmailDomains': {'domainName': 'EXAMPLE.org'},
+                'updated_by': 'example'
+            }),
+            content_type='application/json')
+        data = json.loads(res.get_data(as_text=True))
+
+        assert res.status_code == 201
+        assert data['buyerEmailDomains']['domainName'] == "example.org"
```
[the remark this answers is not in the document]
katstevens: Yeah, good point. Following Alexey's comment on the other PR, I've added a list endpoint, but no get by ID just yet.

tests/main/views/test_buyer_domains.py
katstevens: Done, as per the comment above

migrations/versions/1030_buyer_email_domain_data_migration.py
katstevens: These wouldn't have an audit event as we're creating them directly! Should they? Who would be the 'updated_by' user?

TheDoubleK: Exactly - if they have no audit event then that means we could roll back this migration without deleting stuff that had been added subsequently (which does have an audit event).

But no, I don't think a rollback is really necessary.

And no, I don't think this migration should create audit events.

app/main/views/users.py
TheDoubleK: So the plan is to remove this bit once this final migration is run?  Could we do the migration before merging the code changes and avoid the interim bit?  Maybe too late to suggest that at this point?

migrations/versions/1030_buyer_email_domain_data_migration.py
katstevens: Ahhh I misread your comment. Yes that makes sense, on all points 😸

app/main/views/users.py
katstevens: Yeah, I guess I'm quite cautious when it comes to data migrations that we can't easily undo. I guess the steps are currently: 
1) release schema migration 
2) release new feature code 
3) run data migration 
4) use DB for adding/checking new domains but use text file for checking against old domains 
5) release (as yet unwritten) code change to remove old functionality and check against DB for everything. 

Then 3) and 4) could potentially happen in either order. This way if the release pipeline gets blocked for whatever reason we're still ok.

json_schemas/buyer-email-domains.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "additionalProperties": false,
+  "properties": {
+    "domainName": {
+      "pattern": "^(?!\\.)([^:\/@!&=\\s])+\\.[a-zA-Z]{2,64}?$",
```
TheDoubleK: This looks like it will work as far as I can tell (and we have tests to reassure us).  

I'm not sure we need so many brackets in it, and I'm not sure the nonmatching lookahead group at the start adds anything more than a straight "not a dot" to start?

So I think this does the same thing, perhaps marginally more readable, but it is marginal:
```
^[^\.][^:/@!&=\s]+\.[a-zA-Z]{2,64}?$
```
(Probably with added escapes for the slashes)

TheDoubleK: And I don't think we need the final `?` - if it's matching the start and end of the line then greedy/non-greedy makes no sense - it's going to match it all whatever.
So I think `^[^\.][^:/@!&=\s]+\.[a-zA-Z]{2,64}$` is equivalent.

TheDoubleK: Aha!  OK so your negative capturing group means yours will match single-character before the domain e.g. `a.com`, but mine requires at least two characters before the dot (so `a.com` will fail but `ab.com` will match).

Which of these we want, I'm not 100% sure... probably stick with yours?

katstevens: You're right - I'm used to grouping the different bits to match and use later, but we don't need that here. Tests pass without the lookahead and the `?`, pushing a commit now 😸

katstevens: I'm not keen on `a.com`! They aren't going to come up very much for government buyers: https://en.wikipedia.org/wiki/Single-letter_second-level_domain

katstevens: Though I suppose something like `a.cymru` might. Hmm, will fiddle a bit and see what I can do.

katstevens: Have reinstated the lookahead but simplified the rest of the regex.

PR #676, opened 2017-10-13
Migration to mark service edits before 19-10-2017 as acknowledged

TheDoubleK:
Before we turn on supplier editing their own services (see https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/752) we need to mark edits up to now (performed by admin users) as acknowledged, so that CCS admin users start from a "clean slate" of changes to review.

Ticket: https://trello.com/c/Avb6xJNr/78-accept-all-old-g-cloud-service-edits

Upgrade and downgrade tested locally and looks good as far as I can see.

### DIFF
--- a/migrations/versions/1040_mark_service_edits_as_acknowledged.py
+++ b/migrations/versions/1040_mark_service_edits_as_acknowledged.py
@@ -1,0 +1,30 @@
+"""Mark all audit events of type "update_service" up to 19th October 2017 as acknowledged
+
+Revision ID: 1040
+Revises: 1030
+Create Date: 2017-10-13 13:25:01.000000
+
+"""
+
+revision = '1040'
+down_revision = '1030'
+
+from alembic import op
+
+
+def upgrade():
+    op.execute("""
+        UPDATE audit_events
+           SET acknowledged = TRUE, acknowledged_by = 'pre-supplier-editing-migration-1040', acknowledged_at = NOW()
+         WHERE type = 'update_service'
+           AND acknowledged is FALSE
+           AND created_at < '2017-10-19 00:00:00'
+    """)
+
+
+def downgrade():
+    op.execute("""
+        UPDATE audit_events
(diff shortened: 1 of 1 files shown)


benvand:
Looks good. Didn't realise datetime comparisons are automatically cast to a datetime. 
`created_at < '2017-10-16 00:00:00'`
Thought you might have to 
`created_at < '2017-10-16 00:00:00::TIMESTAMP'

I've done similar migrations with the ORM but this looks fine.
https://github.com/alphagov/digitalmarketplace-api/blob/master/migrations/versions/910_delete_copied_from_brief_id_column.py

TheDoubleK:
I never really got on with ORMs.  Is there a benefit?  In the one you linked to you've written the SQL anyway and then translated into ORM - I don't see the point, but might be missing something.

katstevens:
Hmm. While this looks correct and is perfectly readable, I'm not super keen on having raw SQL queries in the migrations if we can avoid them. 

Could we use SQLAlchemy instead, with something like:
```
audit_events_table = sa.Table('audit_events')  # some other meta data etc needed here I think
query = audit_events_table.update().values({'acknowledged': True, ''acknowledged_by": ...})

conn = op.get_bind()
result = conn.execute(query)
```
Or am I overthinking this?

katstevens:
Isn't the benefit with the ORM that it's platform agnostic? (Say there's a new shiny DB platform we want to upgrade to - or *have* to upgrade to for compliance or similar). For a simple `UPDATE` it's unlikely to be a problem but we should probably be consistent. Of course saying that I'm sure there are a bunch of migrations with raw SQL lying around already!

TheDoubleK:
I always thought SQL is platform-agnostic 😛

katstevens:
I'm thinking more of the scenario where a developer has previously been used to coding raw SQL for MySQL (or some other non-Postgres DB backend), is less likely to make a mistake, and more likely to get a helpful error message, if they use the ORM.

PR #677, opened 2017-10-16
fetch_draft_service: handle case where draft service has no audit history, add test

risicle:
This can sometimes happen e.g. on an instance that's been restored from a partial backup

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -170,7 +170,7 @@ def fetch_draft_service(draft_id):
 
     return jsonify(
         services=draft.serialize(),
-        auditEvents=last_audit_event.serialize(include_user=True),
+        auditEvents=(last_audit_event.serialize(include_user=True) if last_audit_event else None),
         validationErrors=get_service_validation_errors(draft)
     )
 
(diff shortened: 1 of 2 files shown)


TheDoubleK:
This feels more like an issue with our data copying (not copying relevant audit event types over) than with the way our API currently works. Fixing 500s here might result in surfacing errors at the front-end where a timestamp is expected.

So I'm not sure this is the right fix. If we're seeing errors on Staging because of this we should fix the data copying instead.

risicle:
Right but this is part of the "robust by default" thing. Reality is, sometimes it *does* happen, and breaking in unexpected ways when it doesn't *have* to turns out to prove annoying when you're tinkering with stuff.

TheDoubleK:
It might make API tinkering more robust, but if you're running frontends that rely on a value here the same pages will break with or without this fix in, so the gains are marginal at best I think.

This is in reality an invalid database state - if this happened in production it should set alarm bells ringing, not silently swallow the errors.

We should fix our data copying to not create invalid states, rather than fix the code to hide shoddy data and potentially prevent errors from being detected.

But hey, George has approved the PR already, and this is just my opinion, do as you will...

risicle:
Things that are invalid database states should have constraints and the database should not allow them to exist, they shouldn't randomly break at GET time. This just wasted my time chasing a phantom breakage which turned out not to be a breakage, just an artefact of how someone's been tinkering with the database.

PR #679, opened 2017-10-16
Update project award endpoints for data tracking

Wynndow:
Part of this ticket: https://trello.com/c/eo8gJE4C

### Allow list projects with `with_users`
The `DirectAwardProject` model's serialize method has a `with_users`
parameter to include a projects users.

There was no way for this to be used on the `list_projects` view.

### Return projectId in list_project_services services
The services being returned by the `list_project_services` endpoint did
not include the project id associated with the request.

This made dealing with the data returned difficult, especially if making
multiple calls for different projects, such as when extracting data to
analyse.

This also updates the test of the response to assert that the data is
correct, and not just that the keys are in place.

### Test list_project_services returns requested fields
There was no test to ensure that the data specified in the fields param
was being returned in the service data.

### Remove unnecessary string params
The string formatting wasn't using one of the params, and the other was
requesting data that doesn't exist.

### DIFF
--- a/app/main/views/direct_award.py
+++ b/app/main/views/direct_award.py
@@ -21,11 +21,8 @@ def list_projects():
 
     projects = DirectAwardProject.query
 
-    with_users = convert_to_boolean(request.args.get('with-users', False))
-    if not isinstance(with_users, bool):
-        abort(400, "with-users param must be True of False")
-
-    if with_users:
+    includes = request.args.get('include', '').split(',')
+    if 'users' in includes:
         projects = projects.options(db.joinedload('users').lazyload('supplier'))
 
     user_id = get_int_or_400(request.args, 'user-id')
@@ -46,7 +43,7 @@ def list_projects():
     )
 
     return jsonify(
-        projects=[project.serialize(with_users=with_users) for project in projects.items],
+        projects=[project.serialize(with_users='users' in includes) for project in projects.items],
         meta={
             "total": projects.total,
         },
--- a/app/main/views/direct_award.py
+++ b/app/main/views/direct_award.py
@@ -20,6 +20,11 @@ def list_projects():
     page = get_valid_page_or_1()
 
     projects = DirectAwardProject.query
+
+    with_users = convert_to_boolean(request.args.get('with-users', False))
+    if not isinstance(with_users, bool):
+        abort(400, "with-users param must be True of False")
+
     user_id = get_int_or_400(request.args, 'user-id')
     if user_id:
         projects = projects.filter(DirectAwardProject.users.any(id=user_id))
@@ -38,7 +43,7 @@ def list_projects():
     )
 
     return jsonify(
-        projects=[project.serialize() for project in projects.items],
+        projects=[project.serialize(with_users=with_users) for project in projects.items],
         meta={
             "total": projects.total,
         },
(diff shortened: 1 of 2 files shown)


app/main/views/direct_award.py
```diff
@@ -38,7 +43,7 @@ def list_projects():
     )
 
     return jsonify(
-        projects=[project.serialize() for project in projects.items],
+        projects=[project.serialize(with_users=with_users) for project in projects.items],
```
allait: I think project.users is not `lazy=joined`, so this will generate 100 additional queries to the database per page (assuming each project has only one user).

Even if this is meant to be used in scripts, we've seen similar requests time out in the past and a slow response will show up in our monitoring, which would make any alerting we build on top of it less useful.

I think we could use the query option setter here to set the join option on the initial query, the same way we disable it when we don't need it in other endpoints.

Wynndow: What's the query option setter? Can you link me to an example?

app/main/views/direct_award.py
```diff
@@ -20,6 +20,11 @@ def list_projects():
     page = get_valid_page_or_1()
 
     projects = DirectAwardProject.query
+
+    with_users = convert_to_boolean(request.args.get('with-users', False))
+    if not isinstance(with_users, bool):
+        abort(400, "with-users param must be True of False")
```
galund: Are we really saying that `convert_to_boolean` doesn't guarantee a bool output??

app/main/views/direct_award.py
```diff
@@ -20,6 +20,11 @@ def list_projects():
     page = get_valid_page_or_1()
 
     projects = DirectAwardProject.query
+
+    with_users = convert_to_boolean(request.args.get('with-users', False))
```
galund: Can we use an `include=users` style argument instead, per <http://jsonapi.org/format/#fetching-includes>?

app/main/views/direct_award.py
Wynndow: Yep. If it gets a thing that it can't convert to boolean (like a random string) it just returns it.

app/main/views/direct_award.py
Wynndow: Nice! I've not seen that before.

app/main/views/direct_award.py
allait: https://github.com/alphagov/digitalmarketplace-api/blob/d8a7ee8267ce1331d45b12e151f4b8c2f6307348/app/main/views/brief_responses.py#L219-L223

Wynndow: Thanks, have updated to a joinedload for users. Is much quicker now!

app/main/views/direct_award.py
```diff
@@ -38,7 +43,7 @@ def list_projects():
     )
 
     return jsonify(
-        projects=[project.serialize() for project in projects.items],
+        projects=[project.serialize(with_users='users' in includes) for project in projects.items],
```
samuelhwilliams: We could maybe factor `'users' in includes` out to a `with_users` variable at the first point you use it, just to make this read more easily.

Wynndow: I actually went the other way, and changed from your suggest to what it is above. But thinking about it again, for readability, I think you're right. Will update.

PR #680, opened 2017-10-18
Filtering Briefs by datestamp fields

katstevens:
A lot of our scripts involve getting Briefs that closed on a certain day. This involves querying the API for Briefs by 'closed' status, then applying a date filter to the query result in Python. 

I thought it would be nice to be able to filter Briefs by their date fields with an API call. We rarely need to query for a specific datestamp, it's usually 'Briefs closed/published on day X'.

This PR introduces some new `query_class` methods:
- ~~`data_key_contains_value` to filter by any top level key list in `Brief.data` (copied wholesale from the Service model)~~ we're not using this, let's get rid of it
- `has_date_field_between` which will accept either date strings or datetime objects, and filter the Brief's date attributes (currently `created_at`, `updated_at`, `published_at`, `cancelled_at`, `unsuccessful_at`, `withdrawn_at`). 
- equivalent `has_date_field_before` and `has_date_field_after` which are hopefully self explanatory. 

The date filter methods have an `inclusive` flag (defaults to False).

(NB there is no `awarded_at` attribute on the Brief model - this is on the BriefResponse model.)

The `list_briefs` view makes use of these query class methods by accepting the following GET parameters: 'created_on', 'published_on', 'withdrawn_on', 'cancelled_on', 'unsuccessful_on'. Note the `_on` suffix, to differentiate from `_at` - the value we're supplying is a single day in `YYYY-MM-DD` format, not a datestamp or date range (I am not about to write a new API syntax for that!). 

**UPDATE: ** The view now accepts `foo_after` and `foo_before` args! 🎉 

For example: `GET /briefs?closed_on=2017-01-01` will return all Briefs with an `applications_closed_at` attribute between 2017-01-01 00:00:00 and 2017-01-01 23:59:59.

There are some TODOs:
- ~~We can't filter explicitly by `closed_on` yet, because it could be 1 or 2 weeks after the published date. But we'd know it would be one of two `published_on` dates, either 1 or 2 weeks previously, which might help us avoid making 117 paginated API calls in order to get 1 or 2 briefs.~~ DONE
- Missing test cases for the `created_at` and `updated_at` queries, for some reason `freeze_time` isn't working
- ~~Filtering on non-list types in `Brief.data` is untested (and probably won't work yet)~~ Not going to do this

Thoughts welcome.

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -154,12 +154,13 @@ def list_briefs():
         status = "applications_closed" if status == "closed" else status  # Hack for 'closed' status
         day_value = datetime.strptime(request.args.get(status_date_key), "%Y-%m-%d")
         if temporal_filter == 'on':
-            day_end = datetime(day_value.year, day_value.month, day_value.day, 23, 59, 59)
-            briefs = briefs.has_date_field_between("{}_at".format(status), day_value, day_end, inclusive=True)
+            day_end = datetime(day_value.year, day_value.month, day_value.day, 23, 59, 59, 999999)
+            briefs = briefs.has_datetime_field_between("{}_at".format(status), day_value, day_end, inclusive=True)
         if temporal_filter == 'before':
-            briefs = briefs.has_date_field_before("{}_at".format(status), day_value)
+            briefs = briefs.has_datetime_field_before("{}_at".format(status), day_value)
         if temporal_filter == 'after':
-            briefs = briefs.has_date_field_after("{}_at".format(status), day_value)
+            day_end = datetime(day_value.year, day_value.month, day_value.day, 23, 59, 59, 999999)
+            briefs = briefs.has_datetime_field_after("{}_at".format(status), day_end)
 
     if user_id:
         return jsonify(
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -143,6 +143,24 @@ def list_briefs():
             *(status.strip() for status in request.args['status'].split(','))
         )
 
+    status_filters = ['created', 'published', 'withdrawn', 'cancelled', 'unsuccessful', 'closed']
+    temporal_filters = ['on', 'before', 'after']
+    status_date_filters = [
+        '{}_{}'.format(status, temporal) for status in status_filters for temporal in temporal_filters
+    ]
+
+    for status_date_key in [arg for arg in request.args.keys() if arg in status_date_filters]:
+        status, temporal_filter = status_date_key.split('_')
+        status = "applications_closed" if status == "closed" else status  # Hack for 'closed' status
+        day_value = datetime.strptime(request.args.get(status_date_key), "%Y-%m-%d")
+        if temporal_filter == 'on':
+            day_end = datetime(day_value.year, day_value.month, day_value.day, 23, 59, 59)
+            briefs = briefs.has_date_field_between("{}_at".format(status), day_value, day_end, inclusive=True)
+        if temporal_filter == 'before':
+            briefs = briefs.has_date_field_before("{}_at".format(status), day_value)
+        if temporal_filter == 'after':
+            briefs = briefs.has_date_field_after("{}_at".format(status), day_value)
+
     if user_id:
         return jsonify(
             briefs=[brief.serialize(with_users, with_clarification_questions) for brief in briefs.all()],
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -1371,6 +1371,35 @@ class Brief(db.Model):
         def has_statuses(self, *statuses):
             return self.filter(Brief.status.in_(statuses))
 
+        def data_key_contains_value(self, k, v):
+            return self.filter(Brief.data[k].astext.contains(u'"{}"'.format(v)))  # Postgres 9.3: use string matching
+
+        def has_date_field_after(self, attr, start_date, inclusive=None):
+            """Date filter values can be either strings or datetime objects."""
+            if inclusive:
+                return self.filter(getattr(Brief, attr) >= str(start_date))
+            return self.filter(getattr(Brief, attr) > str(start_date))
+
+        def has_date_field_before(self, attr, start_date, inclusive=None):
+            """Date filter values can be either strings or datetime objects."""
+            if inclusive:
+                return self.filter(getattr(Brief, attr) <= str(start_date))
+            return self.filter(getattr(Brief, attr) < str(start_date))
+
+        def has_date_field_between(self, attr, start_date, end_date, inclusive=None):
+            """
+            Date filter values can be either strings or datetime objects.
+            e.g. Brief.query.has_date_field_between('published_at', '2017-01-01', datetime(2017, 1, 2), inclusive=True)
+            would return briefs published between 2017-01-01T00:00:00.000000Z and 2017-01-02T00:00:00.000000Z.
+            """
+            if inclusive:
+                return self.filter(
+                    sql_and(getattr(Brief, attr) >= str(start_date), getattr(Brief, attr) <= str(end_date))
+                )
+            return self.filter(
+                sql_and(getattr(Brief, attr) > str(start_date), getattr(Brief, attr) < str(end_date))
+            )
+
     def add_clarification_question(self, question, answer):
         clarification_question = BriefClarificationQuestion(
             brief=self,
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -1371,9 +1371,6 @@ class Brief(db.Model):
         def has_statuses(self, *statuses):
             return self.filter(Brief.status.in_(statuses))
 
-        def data_key_contains_value(self, k, v):
-            return self.filter(Brief.data[k].astext.contains(u'"{}"'.format(v)))  # Postgres 9.3: use string matching
-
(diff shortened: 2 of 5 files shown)


app/models/main.py
```diff
@@ -1371,6 +1371,28 @@ class query_class(BaseQuery):
         def has_statuses(self, *statuses):
             return self.filter(Brief.status.in_(statuses))
 
+        def data_key_contains_value(self, k, v):
+            return self.filter(Brief.data[k].astext.contains(u'"{}"'.format(v)))  # Postgres 9.3: use string matching
+
+        def has_inclusive_date_range(self, attr, start_date=None, end_date=None):
```
benvand: As a model method this actually does 3 things. I'd rather see it as 3 methods on the model. before, after and between. That way they can be tested separately and processing the args we get from the get request could be done in the view. Ie (if not (start_date or end_date), if not end_date, if not start_date) then call the respective model methods in the view

katstevens:
I've now split that query_class method into three, which also means we can have `_before` and `_after` request args on the view. Hurray!

katstevens:
An update on this:

- I removed the JSON field filtering, we're not really going to use it.
- Thanks to @benvand's keen eyes we can now use the `Brief.applications_closed_at` attribute to filter by `closed_on`, `closed_before`, `closed_after` 🎉 
- Setting up closed Briefs in tests remains painful 😿 but at least it's possible 😸 
- I squashed up the commits for tidyness.

I think it's ready for merging. @idavidmcdonald @benvand ?

tests/models/test_main.py
```diff
@@ -565,6 +565,55 @@ def test_query_brief_applications_closed_at_date_for_mix_of_brief_lengths(self):
...
+
+    @pytest.mark.parametrize('inclusive,expected_count', [(True, 2), (False, 1)])
+    @pytest.mark.parametrize('date_format', ["2017-01-01", datetime(2017, 1, 1, 0, 0, 0)])
+    @pytest.mark.parametrize('datestamp_attr', ['published_at', 'withdrawn_at', 'cancelled_at', 'unsuccessful_at'])
+    def test_query_brief_has_date_field_before_and_after(self, datestamp_attr, date_format, inclusive, expected_count):
+        with self.app.app_context():
+            new_briefs = [Brief(data={}, framework=self.framework, lot=self.lot) for i in range(3)]
+            setattr(new_briefs[0], datestamp_attr, datetime(2016, 12, 31, 23, 59, 59))  # < date
+            setattr(new_briefs[1], datestamp_attr, datetime(2017, 1, 1, 0, 0, 0))  # <= and >= date
+            setattr(new_briefs[2], datestamp_attr, datetime(2017, 1, 1, 0, 0, 1))  # > date
```
idavidmcdonald: `setattr(new_briefs[2], datestamp_attr, datetime(2017, 1, 1, 0, 0, 1))  # > date` should this instead be
`setattr(new_briefs[2], datestamp_attr, datetime(2017, 1, 2, 0, 0, 0))  # > date`?

What you currently have is 1 minute into the search date, which is still therefore on the same date rather than being > than the search date.

tests/models/test_main.py
```diff
@@ -565,6 +565,55 @@ def test_query_brief_applications_closed_at_date_for_mix_of_brief_lengths(self):
...
+        ]
+    )
+    @pytest.mark.parametrize('datestamp_attr', ['published_at', 'withdrawn_at', 'cancelled_at', 'unsuccessful_at'])
+    def test_query_brief_has_date_field_between(self, datestamp_attr, start_date, end_date, inclusive, expected_count):
+        with self.app.app_context():
+            new_briefs = [Brief(data={}, framework=self.framework, lot=self.lot) for i in range(4)]
+            setattr(new_briefs[0], datestamp_attr, datetime(2016, 12, 31, 23, 59, 59))  # < date range
+            setattr(new_briefs[1], datestamp_attr, datetime(2017, 1, 1, 0, 0, 0))  # <= date range
+            setattr(new_briefs[2], datestamp_attr, datetime(2017, 1, 3, 0, 0, 0))  # >= date range
+            setattr(new_briefs[3], datestamp_attr, datetime(2017, 1, 3, 0, 0, 1))  # > date range
```
idavidmcdonald: Again, I think this last date is still inside an inclusive date range of 1/1/2017 to 3/1/2017.
A non inclusive range would just be 2/1/17.

It feels like there may be slight confusion/misunderstanding around filtering based on dates rather than datetimes. Feel free to grab me to chat this through if you'd prefer.

katstevens: Yikes, you're right. I had written my docstring wrong on the method itself, and the tests were based on that! Carelessness on my part, sorry.

idavidmcdonald: No problem :) Let me know when you want me to take a look after you've made relevant changes.

katstevens: As discussed, the query methods now only accept `datetime` objects and raise a ValueError if a string is supplied. Hopefully this should remove the ambiguity.

app/models/main.py
```diff
@@ -1371,6 +1371,40 @@ class query_class(BaseQuery):
         def has_statuses(self, *statuses):
             return self.filter(Brief.status.in_(statuses))
 
+        def has_date_field_after(self, attr, start_datetime, inclusive=None):
```
idavidmcdonald: Could we name these methods `datetime_field` so it's really clear by looking at the method name what they take?

app/main/views/briefs.py
```diff
@@ -143,6 +143,24 @@ def list_briefs():
...
+        '{}_{}'.format(status, temporal) for status in status_filters for temporal in temporal_filters
+    ]
+
+    for status_date_key in [arg for arg in request.args.keys() if arg in status_date_filters]:
+        status, temporal_filter = status_date_key.split('_')
+        status = "applications_closed" if status == "closed" else status  # Hack for 'closed' status
+        day_value = datetime.strptime(request.args.get(status_date_key), "%Y-%m-%d")
+        if temporal_filter == 'on':
+            day_end = datetime(day_value.year, day_value.month, day_value.day, 23, 59, 59)
+            briefs = briefs.has_date_field_between("{}_at".format(status), day_value, day_end, inclusive=True)
```
idavidmcdonald: Feels slightly weird that we are saying to inclusively include `23:59:59` given that it on the same day and not in a different time period but our method does require us to do this.

Not entirely sure if this is a good suggestion but you could instead pass the next day at `00:00:00` non inclusively in as the end value.

app/models/main.py
katstevens: We can! Done.

tests/main/views/test_briefs.py
```diff
@@ -713,6 +714,88 @@ def test_list_briefs_no_pagination_if_user_id_supplied(self):
...
+    )
+    def test_list_briefs_filter_after_date(self, date_arg):
+        for i, brief_status in enumerate(["draft", "withdrawn", "published", "cancelled", "unsuccessful"]):
+            with freeze_time('2017-01-01 00:00:01'):
+                self.setup_dummy_briefs(1, lot='digital-outcomes', status=brief_status, brief_start=i + 1)
+            # The following briefs should be filtered out ('after' is not inclusive)
+            with freeze_time('2017-01-01 00:00:00'):
+                self.setup_dummy_briefs(1, lot='digital-outcomes', status=brief_status, brief_start=i + 6)
+
+        res = self.client.get('/briefs?{}=2017-01-01'.format(date_arg))
```
idavidmcdonald: This test feels weird. If you said to someone find me anything published after Thursday, you would either include both briefs published at 00:00 and 00:01 on Thursday because you are being inclusive of Thursday. Or you would include neither because you would only find briefs published on Friday.

katstevens: I'd prefer to include neither ('after' and 'before' imply non-inclusive to me). I'll change so that the `after_` case will hack the date in the same way as for the upper bound on `on_`, i.e. replace the param with 'param + 23:59:59'.

app/main/views/briefs.py
katstevens: I suppose this would cover an edge case of a `23:59:59.999999` timestamp! However our `has_datetime_field_between` doesn't currently allow separate inclusive flags for the lower and upper bounds. It's possible to chain the before/after query methods and use them separately, but I think this would be slightly less efficient than using the `sql_and` in the between method.

I'll update the range to include microseconds as a compromise.

tests/main/views/test_briefs.py
```diff
@@ -713,6 +714,88 @@ def test_list_briefs_no_pagination_if_user_id_supplied(self):
...
+
+    @pytest.mark.parametrize(
+        'date_arg', ['published_after', 'withdrawn_after', 'cancelled_after', 'unsuccessful_after']
+    )
+    def test_list_briefs_filter_after_date(self, date_arg):
+        for i, brief_status in enumerate(["draft", "withdrawn", "published", "cancelled", "unsuccessful"]):
+            with freeze_time('2017-01-02 00:00:00'):
+                self.setup_dummy_briefs(1, lot='digital-outcomes', status=brief_status, brief_start=i + 1)
+            # The following briefs should be filtered out ('after' is not inclusive)
+            with freeze_time('2017-01-01 00:00:00'):
```
idavidmcdonald: For continuity with your other tests, is it worth having this as `23:59:59`?

PR #682, opened 2017-10-25
Add buyer domains: wandle, iwm, bl

benvand:


idavidmcdonald:
Do we not do these in the database now?

risicle:
Should this file still exist then if that's the case? Otherwise it's just confusing.

katstevens:
Ah that's my fault for not tidying up after myself! Now the DB table etc has been deployed to prod, we can get rid of the file (the data migration should happily skip the missing file if we have to roll forward/backward for whatever reason). There's a little bit of untangling to do as well as just deleting the file, I'm happy to do a quick PR now.

benvand:
Done these using the api! Thanks @katstevens !

katstevens:
That PR: https://github.com/alphagov/digitalmarketplace-api/pull/683

PR #683, opened 2017-10-25
Removing the old buyer domain text file

katstevens:
The final bit of cleanup following the successful (woo!) deployment of this Trello story: https://trello.com/c/JxwDhnDh/82-migrate-existing-domain-check-to-use-database

- Updated the User validation (and the `/check-buyer-email` route) to use the shiny new database buyer domain validation 
- Updated a bunch of tests which relied on a variation of `.gov.uk` being present in the old file. The `BaseApplicationTest` class now creates `digital.gov.uk` as a default fixture, so any test that creates a buyer account should be able to use this for validation. For testing the actual buyer domain bits, this fixture is removed beforehand.
- There were a few tests which needed some TLC to work with the new validation (e.g. for some of the User tests). 
- Finally, removed the old validation code and the file.


tests/models/test_main.py
```diff
@@ -20,26 +20,26 @@
...
+                updated_at=now,
+                password_changed_at=now
+            )
 
-    with app.app_context():
-        assert_equal(user.serialize()['emailAddress'], "email@digital.gov.uk")
-        assert_equal(user.serialize()['name'], "name")
-        assert_equal(user.serialize()['role'], "buyer")
-        assert_equal('password' in user.serialize(), False)
+            assert_equal(user.serialize()['emailAddress'], "email@digital.gov.uk")
```
idavidmcdonald: I believe we are trying to remove nose functions from all our tests suites so we just use `assert user.serialize()['emailAddress'] ==  "email@digital.gov.uk"` instead. I think we had removed these from nearly everywhere but these might be some final ones if you'd like to do that whilst your at it please?

tests/main/views/test_buyer_domains.py
```diff
@@ -158,6 +158,13 @@ def test_create_buyer_email_domain_fails_if_higher_level_domain_already_approved
 
 
 class TestListBuyerEmailDomains(BaseApplicationTest):
+    def setup(self):
+        super(TestListBuyerEmailDomains, self).setup()
+        with self.app.app_context():
+            # Remove default fixture so we can test pagination
```
idavidmcdonald: Looking at the below tests, all of them require there to be no domains in the table, so the comment should be more general about wanting to start from an empty state. With that being the case, is it worth deleting everything in the table rather than specifically mentioning the one domain which we assume will be there (and be the only domain there)?

tests/bases.py
```diff
@@ -26,6 +26,11 @@ def setup(self):
             HTTP_AUTHORIZATION='Bearer {}'.format(self.app.config['DM_API_AUTH_TOKENS'])
         )
         self.client = self.app.test_client()
+        with self.app.app_context():
+            # Ensure there is at least one pre-approved buyer domain
+            if BuyerEmailDomain.query.filter(BuyerEmailDomain.domain_name == 'digital.gov.uk').count() == 0:
```
idavidmcdonald: Do we need this on our base class meaning it will run on every single test we run? There probably are some knock on effects that I haven't thought through but it feels like this could be something we want to avoid adding?

app/main/views/users.py
```diff
@@ -295,10 +294,6 @@ def email_has_valid_buyer_domain():
         abort(400, "'email_address' is a required parameter")
 
     domain_ok = buyer_email_domain_approved(BuyerEmailDomain.query.all(), email_address.split('@')[-1])
-    if not domain_ok:
-        # If the data migration hasn't happened yet, check the .txt file
```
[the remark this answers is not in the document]
katstevens: The migration is definitely there! 😸  https://github.com/alphagov/digitalmarketplace-api/blob/master/migrations/versions/1030_buyer_email_domain_data_migration.py

tests/main/views/test_buyer_domains.py
katstevens: Yep, good call. Will fix

tests/models/test_main.py
katstevens: Yep, happy to change. This test was clearly a bit of an outlier!

app/main/views/users.py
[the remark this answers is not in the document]
idavidmcdonald: Oh thanks, the ordering of numbers alphabetically meant I didn't see it. Although just to confirm, if any developer or new joiner ran the migration after this PR they would not end up with the domains in their DB because we are about to delete the file?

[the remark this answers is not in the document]
katstevens: Yes, I think so. We'd have to do a fresh DB dump 😸

app/validation.py
```diff
@@ -400,4 +387,10 @@ def validate_buyer_email_domain_json_or_400(submitted_json):
 
 
 def buyer_email_domain_approved(existing_buyer_domains, new_domain):
```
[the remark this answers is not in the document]
katstevens: I think so. I thought I could get away with having one function, but it makes more sense to have the two functions for the two different usecases: validating a new domain added by the admin, and validating an email address for a  new buyer user. Have added a commit to fix.

tests/bases.py
katstevens: It turns out that putting it as a method in `FixtureMixin` works well enough. Have added a commit to fix.

tests/main/views/test_buyer_domains.py
```diff
@@ -158,6 +158,13 @@ def test_create_buyer_email_domain_fails_if_higher_level_domain_already_approved
 
 
 class TestListBuyerEmailDomains(BaseApplicationTest):
+    def setup(self):
+        super(TestListBuyerEmailDomains, self).setup()
+        with self.app.app_context():
+            # Remove any default fixtures so we can test pagination
```
idavidmcdonald: comment still specifically references pagination - more general would be good please as it's applicable for all tests.

tests/test_validation.py
```diff
@@ -790,3 +779,21 @@ def test_translate_unknown_oneoff_eerror():
             {'message': "Unknown type", 'validator': 'type'}
         ],
     }]) == {'example': [{'error': 'failed', 'index': 0}]}
+
+
+@pytest.mark.parametrize('email, expected_result', [('hurray@cool.gov', True), ('hurray@notcool.gov', False)])
```
idavidmcdonald: Do we need to cover subdomains here? like hurray@digital.cool.gov?

tests/main/views/test_buyer_domains.py
katstevens: Fixed!

tests/test_validation.py
katstevens: Done!

PR #686, opened 2017-10-30
Push context in tests

benvand:
If we add this in the setup we can later remove all the instances of `with self.app.context():` and access objects that are created within tests throughout the entire test instead of needing to enter and exit the context to access the objects we create. This has broad implications for setups, fixtures and factories. Making them accessible as part of a 'test-wide' context.

This will make writing tests cleaner and easier hopefully faster by:
* Making it possible to attach objects to a class in a `setup` method and access them in tests
* Reducing database calls
* Reducing unneeded lines of code and indentation
* Easing the implementation of factory boy factories

See documentation here for naming convention and how this works:
http://flask.pocoo.org/docs/0.12/reqcontext/#diving-into-context-locals

### DIFF
--- a/tests/bases.py
+++ b/tests/bases.py
@@ -25,18 +25,20 @@ class BaseApplicationTest(object):
             self.app.wsgi_app,
             HTTP_AUTHORIZATION='Bearer {}'.format(self.app.config['DM_API_AUTH_TOKENS'])
         )
+        self.ctx = self.app.app_context()
+        self.ctx.push()
         self.client = self.app.test_client()
 
     def teardown(self):
-        with self.app.app_context():
-            db.session.remove()
-            for table in reversed(db.metadata.sorted_tables):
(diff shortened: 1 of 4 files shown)


tests/bases.py
```diff
@@ -25,18 +25,20 @@ def setup(self):
             self.app.wsgi_app,
             HTTP_AUTHORIZATION='Bearer {}'.format(self.app.config['DM_API_AUTH_TOKENS'])
         )
+        self.ctx = self.app.app_context()
```
samuelhwilliams: I guess I'd rather it was called `self.app_context`, but :man_shrugging:

samuelhwilliams:
(After you fix tests obviously :trollface:)

benvand:
This work is being continued in https://github.com/alphagov/digitalmarketplace-api/compare/TESTS-bv?expand=1

PR #687, opened 2017-11-07
Use new edit by admin audit event

TheDoubleK:
Tests will fail until API client https://github.com/alphagov/digitalmarketplace-apiclient/pull/110 is merged.

Part of this ticket: https://trello.com/c/DbYlqSxv/121-exclude-changes-made-by-admin-in-the-approve-edits-section

This makes the `update_service` endpoint take an option GET parameter `user-role`. If the value of this is 'admin' then the view should use the new `update_service_admin` audit type, rather than the plain `update_service` type.

This will allow the admin frontend to list only services edited by suppliers, thus excluding edits to Digital Outcomes and Specialists services that only admin users can/will be able to edit.


requirements-app.txt
```diff
@@ -10,7 +10,7 @@ SQLAlchemy==1.1.4
 SQLAlchemy-Utils==0.30.5
 
 git+https://github.com/alphagov/digitalmarketplace-utils.git@29.0.0#egg=digitalmarketplace-utils==29.0.0
-git+https://github.com/alphagov/digitalmarketplace-apiclient.git@10.8.0#egg=digitalmarketplace-apiclient==10.8.0
+git+https://github.com/alphagov/digitalmarketplace-apiclient.git@11.4.0#egg=digitalmarketplace-apiclient==11.4.0
```
katstevens: Is there a breaking change here?

TheDoubleK: Yes but it doesn't affect API (which just needs the new audit event type).

tests/main/views/test_services.py
```diff
@@ -463,9 +463,11 @@ def setup(self):
                 **payload
             )
 
-    def _post_service_update(self, service_data, service_id=None):
+    def _post_service_update(self, service_data, service_id=None, user_role=''):
```
katstevens: Nitpicking considering this is a test, but could we have `user_role=None` as the default instead?

TheDoubleK: Sure, if it makes you happy 😄

katstevens: Lovely. Giving the thumbs up now 👍

PR #688, opened 2017-11-07
[WIP] Obfuscate saved search IDs

samuelhwilliams:
## Summary
Option #1 for implementing this ticket.

We want to obscure saved search/direct award project IDs when they leave the API so we aren't leaking information and to minimise the attack surface for malicious users.

Tell me if this is too much and we should just do something simpler... :)

We now have an `IdentifierObscuringMixin` (name up for debate) that you can add to a class to easily obscure data in that table. In the serialization method, you call `self.obfuscate` on whichever fields you want to hide (i.e. the PK). The field is encrypted with a key derived from a DM secret and the name of the table, meaning that identical keys (e.g. 1) will encode to different external identifiers for each table.

When a request comes back into the API for a given resource that utilises this encoding, we have a new route converter that can be used to seamlessly decode it before handing off to the view. Whereas you might normally have `<int:project_id>` or `<string:project_id>` or whatever in the route decorater, you can now have e.g. `<obscured(DirectAwardProject):project_id>`. When `project_id` comes into the view method, you'll be dealing with the internal ID again even though the URL used included the obscured ID (e.g. `/direct-award/projects/ehk7pl6pjidta`).

[WIP]: I had hoped to be able to store the external ID in a new column but I haven't yet found a way to do this given that it needs to be generated at the same time as the PK, which isn't determined until the row is committed to the DB. This would mean we don't have to generate it on the fly every time and would mean there's no risk that a change in the future could alter the encoding algorithm and change the encoded ID, effectively breaking all user routes.

I also need to write tests.

## Corresponding PRs
https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/662
https://github.com/alphagov/digitalmarketplace-credentials/pull/41
https://github.com/alphagov/digitalmarketplace-aws/pull/354

## Ticket


risicle:
Neat!

Incidentally, I remember why I originally needed to use an encrypted id scheme and not just random ids - I needed each object to have multiple different ids dependent on whether our user was using the trial or full version.

risicle:
Idea: rather than storing the external id on the row, you could add it to the audit log...

risicle:
(or just do that as well)

risicle:
Especially neat is the `DataUnobscuringConverter` - how does this interoperate with `to_url`? are you just expected to obfuscate it yourself before the call (which is a perfectly fine approach)?

samuelhwilliams:
@risicle Yeah, currently I think the approach would be that you need to obfuscate the ID before building the URL. I haven't actually looked at that yet, so maybe it should do the conversion automatically as well

samuelhwilliams:
@TheDoubleK raises a good point in that this approach (with obscured IDs) makes it hard for developers to manually poke resources through the API, especially if the external ID isn't stored in the database. Will definitely need to have a solution for this.

risicle:
But clearly `url_for` doesn't freak out when trying to generate a url section for `obscured(DirectAwardProject):...` so it seeeeeeeems fine.

risicle:
I don't understand how this makes it difficult to poke resources through the API, though I do have some ideas about how we could store an external id using ORM events

risicle:
Oh and past experience makes me suggest you add an option to override the apparent "table name" used for the key.

app/converters.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def to_python(self, value):
+        """Takes an obscured ID (as passed to Flask's routing system) and attempts to unobscure it, transforming
+        it back to the associated internal ID (e.g. a Direct Award Project/Saved Search PK/ID)."""
+        # TODO: Remove this after buyer-fe re-direct for internal IDs has been removed.
+        if value.isdigit() and len(value) <= 13:
+            return int(value)
+
+        try:
+            unobscured_value = self.model.unobscure(value, raise_errors=True)
```
galund: This is the only place we use self.model. And it's a class method, on the base/mixin class, that seems unlikely ever to need to be overridden. So I'd question whether we actually need to pass the model in at all - which would make this all seem a lot more simple?

app/models/mixins.py
```diff
@@ -0,0 +1,76 @@
+import base64
+import binascii
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+from cryptography.hazmat.backends import default_backend
+from flask import abort
+import hashlib
+import math
+
+
+class DataObscuringMixin:
```
galund: Maybe rename see https://github.com/alphagov/digitalmarketplace-credentials/pull/41/files#r149741874

app/converters.py
samuelhwilliams: The obfuscation is based in part on the table name of the model, so without a reference to the model (or, at least, the table name), it's impossible to get back to the 'real' ID.

app/models/mixins.py
samuelhwilliams: IdentifierObscuringMixin? IDObscuringMixing?

risicle: Well strictly what you've created *here* isn't id specific - could be used with any integer right?

samuelhwilliams: Yep, technically true.

samuelhwilliams:
Yeah, I agree that we need to think about AuditEvents. I'd plan to include the obscured ID in any audit events we raise, and we'd need to run a migration to update historical ones. My priority is definitely to keep is as invisible as possible, and the potential burden on developers is something I'm thinking about and won't feel comfortable landing this until it's no harder than what we have currently.

The plan was to make it as easy as possible to roll it out to other IDs if we decide to go ahead with that, so hopefully it shouldn't be too bad. But yes, it'll be interesting to think about briefs/responses.

risicle:
When it comes to audit events, I don't think this situation is significantly different from that of services & service ids

samuelhwilliams:
Closing due to decision to go ahead with random public/external IDs as opposed to derived ones. See https://github.com/alphagov/digitalmarketplace-api/pull/690

samuelhwilliams:
Closing due to decision to go ahead with random public/external IDs as opposed to derived ones. See https://github.com/alphagov/digitalmarketplace-api/pull/690

PR #689, opened 2017-11-09
 Change brief_response editing validation to use brief status

CrystalPea:
Earlier the brief response could only be edited if it had draft status. Now we are introducing a new feature that allows suppliers to edit submitted brief responses while the brief is live, so the validation system needed updating.

Trello ticket for the new feature: https://trello.com/c/kQpbazwt/85-allow-suppliers-edit-submitted-application-if-opportunity-is-still-open


katstevens:
Agreed - hopefully no-one with a theoretical `withdrawn` application would try and edit it but you never know 😸 Thanks Ben!

PR #690, opened 2017-11-10
Add an external ID to the Direct Award Project/'saved search' table

samuelhwilliams:
## Summary
We will generate random external integer IDs for our Direct Award saved searches in line with best practice to mitigate risks around flawed implementation of authorization. For a transitional period, we will allow URLs on the frontend using a saved search's internal ID to go ahead (although we will redirect to the external ID).

The randomly generated IDs are consistently 15 digits in length, fitting into a PostgreSQL BigInteger and (hopefully) below the threshold for scientific notation in spreadsheet programs. For the sake of preserving our existing analytics setup, these won't be (eg base32) encoded when they leave the API - they will remain as raw integers. I don't envision this causing a problem or looking too awful, although they are 15 characters, which could be considered long.

There is no collision detection/avoidance; on the slim chance that a randomly generated ID has already been submitted, a unique constraint on the external ID will cause the API to error if a duplicate entry is committed. After speaking to @galund and @risicle it seems we are happy to accept this.

Existing AuditEvents will be updated to inject the new external ID of existing direct award projects so that we retain the ability to filter AuditEvents on the ID we are likely to be given (e.g. by the support team).

## Ticket

### DIFF
--- a/app/main/views/direct_award.py
+++ b/app/main/views/direct_award.py
@@ -11,8 +11,23 @@ from ... import db
 from ...models import User, AuditEvent, ArchivedService
 from ...models.direct_award import DirectAwardProject, DirectAwardSearch
 from ...utils import (
-    get_json_from_request, get_int_or_400, json_has_required_keys, pagination_links,
-    get_valid_page_or_1, validate_and_return_updater_request, drop_all_other_fields)
+    drop_all_other_fields,
+    get_int_or_400,
+    get_json_from_request,
+    get_valid_page_or_1,
+    json_has_required_keys,
+    pagination_links,
+    validate_and_return_updater_request,
+)
+
+
+def get_project_by_id_or_404(project_id: int):
+    """During the transitional period, project_id may be either the internal or the public ID of the project."""
+    project = DirectAwardProject.query.filter(DirectAwardProject.id == project_id).first()
+    if project:
+        return project
+
+    return DirectAwardProject.query.filter(DirectAwardProject.external_id == project_id).first_or_404()
 
 
 @main.route('/direct-award/projects', methods=['GET'])
@@ -71,20 +86,28 @@ def create_project():
     if user is None:
         abort(400, "User ID not supplied")
 
-    project = DirectAwardProject(name=project_json['name'], users=[user])
-    db.session.add(project)
+    retries = 0
+    while True:
+        project = DirectAwardProject(name=project_json['name'], users=[user])
+        db.session.add(project)
 
-    try:
-        db.session.flush()
-    except IntegrityError as e:
-        db.session.rollback()
-        abort(400, e.orig)
+        try:
+            db.session.flush()
+            break
+
+        except IntegrityError as e:
+            current_app.logger.warning("Project ID collision on {}".format(project.external_id))
+            db.session.rollback()
+
+            retries += 1
+            if retries >= 5:
+                abort(400, str(e.orig))
 
     audit = AuditEvent(
         audit_type=AuditTypes.create_project,
         user=updater_json['updated_by'],
         data={
-            'projectId': project.id,
+            'projectExternalId': project.external_id,
             'projectJson': project_json,
         },
         db_object=project,
@@ -96,21 +119,20 @@ def create_project():
     return jsonify(project=project.serialize()), 201
 
 
-@main.route('/direct-award/projects/<int:project_id>', methods=['GET'])
-def get_project(project_id):
-    project = DirectAwardProject.query.filter(DirectAwardProject.id == project_id).first_or_404()
+@main.route('/direct-award/projects/<int:project_external_id>', methods=['GET'])
+def get_project(project_external_id):
+    project = get_project_by_id_or_404(project_external_id)
 
     return jsonify(project=project.serialize(with_users=True))
 
 
-@main.route('/direct-award/projects/<int:project_id>/searches', methods=['GET'])
-def list_project_searches(project_id):
-    # If the project doesn't exist, we should 404 early.
-    DirectAwardProject.query.filter(DirectAwardProject.id == project_id).first_or_404()
+@main.route('/direct-award/projects/<int:project_external_id>/searches', methods=['GET'])
+def list_project_searches(project_external_id):
+    project = get_project_by_id_or_404(project_external_id)
 
     page = get_valid_page_or_1()
 
-    searches = DirectAwardSearch.query.filter(DirectAwardSearch.project_id == project_id)
+    searches = DirectAwardSearch.query.filter(DirectAwardSearch.project_id == project.id)
 
     if 'latest-first' in request.args:
         if convert_to_boolean(request.args.get('latest-first')):
@@ -129,7 +151,7 @@ def list_project_searches(project_id):
     )
 
     pagination_params = request.args.to_dict()
-    pagination_params['project_id'] = project_id
+    pagination_params['project_external_id'] = project.external_id
 
     return jsonify(
         searches=[search.serialize() for search in searches.items],
@@ -144,12 +166,11 @@ def list_project_searches(project_id):
     )
 
 
-@main.route('/direct-award/projects/<int:project_id>/searches', methods=['POST'])
-def create_project_search(project_id):
+@main.route('/direct-award/projects/<int:project_external_id>/searches', methods=['POST'])
+def create_project_search(project_external_id):
     updater_json = validate_and_return_updater_request()
 
-    # If the project doesn't exist, we should 404 early.
-    DirectAwardProject.query.filter(DirectAwardProject.id == project_id).first_or_404()
+    project = get_project_by_id_or_404(project_external_id)
 
     json_payload = get_json_from_request()
     json_has_required_keys(json_payload, ['search'])
@@ -163,9 +184,9 @@ def create_project_search(project_id):
 
     # TODO: Validate user has authorisation to access resource.
 
-    db.session.query(DirectAwardSearch).filter(DirectAwardSearch.project_id == project_id).\
+    db.session.query(DirectAwardSearch).filter(DirectAwardSearch.project_id == project.id).\
         update({DirectAwardSearch.active: False})
-    search = DirectAwardSearch(created_by=user.id, project_id=project_id,
+    search = DirectAwardSearch(created_by=user.id, project_id=project.id,
                                search_url=search_json['searchUrl'], active=True)
 
     db.session.add(search)
@@ -181,7 +202,8 @@ def create_project_search(project_id):
         audit_type=AuditTypes.create_project_search,
         user=updater_json['updated_by'],
         data={
-            'projectId': project_id,
+            'projectId': project.id,
+            'projectExternalId': project.external_id,
             'searchJson': search_json,
         },
         db_object=search,
@@ -193,21 +215,20 @@ def create_project_search(project_id):
     return jsonify(search=search.serialize()), 201
 
 
-@main.route('/direct-award/projects/<int:project_id>/searches/<int:search_id>', methods=['GET'])
-def get_project_search(project_id, search_id):
-    # If the project doesn't exist, we should 404 early.
-    DirectAwardProject.query.filter(DirectAwardProject.id == project_id).first_or_404()
+@main.route('/direct-award/projects/<int:project_external_id>/searches/<int:search_id>', methods=['GET'])
+def get_project_search(project_external_id, search_id):
+    project = get_project_by_id_or_404(project_external_id)
 
     search = DirectAwardSearch.query.filter(
         DirectAwardSearch.id == search_id,
-        DirectAwardSearch.project_id == project_id
+        DirectAwardSearch.project_id == project.id
     ).first_or_404()
 
     return jsonify(search=search.serialize())
 
 
-@main.route('/direct-award/projects/<int:project_id>/services', methods=['GET'])
-def list_project_services(project_id):
+@main.route('/direct-award/projects/<int:project_external_id>/services', methods=['GET'])
+def list_project_services(project_external_id):
     """This endpoint returns all the services associated with a particular (locked) Direct Award Project. It returns
     some fairly arbitrary data which is obviously not ideal, but as we don't have a task runner that we can utilise to
     manage jobs like this and we need it to be responsive on-click, we're having to jerryrig in some data extraction and
@@ -215,19 +236,19 @@ def list_project_services(project_id):
     the service data JSON keys as specified in the request (which will be loaded via a framework-specific manifest)."""
     page = get_valid_page_or_1()
     requested_fields = request.args.get('fields', '').split(',')
-    project = DirectAwardProject.query.filter(DirectAwardProject.id == project_id).first_or_404()
+    project = get_project_by_id_or_404(project_external_id)
 
     if not project.locked_at:
-        abort(400, 'Project has not been locked: {}'.format(project_id))
+        abort(400, 'Project has not been locked: {}'.format(project.id))
 
     # TODO: This should work for _all_ active searches, not just the first (although we currently enforce only one
     # TODO: active search) - SW 21/09/2017
     search = DirectAwardSearch.query.filter(
-        DirectAwardSearch.project_id == project_id,
+        DirectAwardSearch.project_id == project.id,
         DirectAwardSearch.active == True  # noqa
     ).first()
     if not search:
-        abort(400, 'Project does not have a saved search: {}'.format(project_id))
+        abort(400, 'Project does not have a saved search: {}'.format(project.id))
 
     paginated_archived_services = search.archived_services.paginate(
         page=page,
@@ -236,7 +257,7 @@ def list_project_services(project_id):
 
     project_archived_services = list(map(lambda service: {
         'id': service.service_id,
-        'projectId': project_id,
+        'projectId': project.external_id,
         'supplier': {
             'name': service.supplier.name,
             'contact': {
@@ -249,7 +270,7 @@ def list_project_services(project_id):
     }, paginated_archived_services.items))
 
     pagination_params = request.args.to_dict()
-    pagination_params['project_id'] = project_id
+    pagination_params['project_external_id'] = project.external_id
 
     return jsonify(
         services=project_archived_services,
@@ -264,17 +285,17 @@ def list_project_services(project_id):
     )
 
 
-@main.route('/direct-award/projects/<int:project_id>/lock', methods=['POST'])
-def lock_project(project_id):
+@main.route('/direct-award/projects/<int:project_external_id>/lock', methods=['POST'])
+def lock_project(project_external_id):
     updater_json = validate_and_return_updater_request()
 
-    project = DirectAwardProject.query.filter(DirectAwardProject.id == project_id).first_or_404()
+    project = get_project_by_id_or_404(project_external_id)
 
     if project.locked_at:
-        abort(400, 'Project has already been locked: {}'.format(project_id))
+        abort(400, 'Project has already been locked: {}'.format(project.id))
 
     search = DirectAwardSearch.query.filter(
-        DirectAwardSearch.project_id == project_id,
+        DirectAwardSearch.project_id == project.id,
         DirectAwardSearch.active == True,  # noqa
     ).first_or_404()
 
@@ -303,7 +324,8 @@ def lock_project(project_id):
         audit_type=AuditTypes.lock_project,
         user=updater_json['updated_by'],
         data={
-            'projectId': project.id
+            'projectId': project.id,
+            'projectExternalId': project.external_id,
         },
         db_object=search,
     )
@@ -314,11 +336,11 @@ def lock_project(project_id):
     return jsonify(project=project.serialize())
 
 
-@main.route('/direct-award/projects/<int:project_id>/record-download', methods=['POST'])
-def record_project_download(project_id):
+@main.route('/direct-award/projects/<int:project_external_id>/record-download', methods=['POST'])
+def record_project_download(project_external_id):
     updater_json = validate_and_return_updater_request()
 
-    project = DirectAwardProject.query.filter(DirectAwardProject.id == project_id).first_or_404()
+    project = get_project_by_id_or_404(project_external_id)
 
     # We want to track the latest datetime the results were downloaded, overriding previous (although they're audited).
     project.downloaded_at = datetime.datetime.utcnow()
@@ -329,7 +351,7 @@ def record_project_download(project_id):
         audit_type=AuditTypes.downloaded_project,
         user=updater_json['updated_by'],
         data={
-            'projectId': project.id
+            'projectExternalId': project.external_id,
         },
         db_object=project,
     )
--- a/app/models/direct_award.py
+++ b/app/models/direct_award.py
@@ -5,8 +5,10 @@ from sqlalchemy.orm import validates
 from flask import current_app
 
 from app import db
+from app.utils import random_positive_external_id
 from app.url_utils import force_relative_url
 from app.models import User, ValidationError, ArchivedService
+
 from dmutils.formats import DATETIME_FORMAT
 
 
@@ -14,6 +16,7 @@ class DirectAwardProject(db.Model):
     __tablename__ = 'direct_award_projects'
 
     id = db.Column(db.Integer, primary_key=True)
+    external_id = db.Column(db.BigInteger, default=random_positive_external_id, nullable=False, index=True, unique=True)
     name = db.Column(db.String(255), nullable=False)
     created_at = db.Column(db.DateTime, default=datetime.utcnow, nullable=False)
     locked_at = db.Column(db.DateTime, nullable=True)  # When the project's active search/es are final and cannot change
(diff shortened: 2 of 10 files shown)


app/main/views/direct_award.py
```diff
@@ -15,6 +16,22 @@
...
+
+    try:
+        project_public_id = PublicIdMixin.decode(project_id)
+
+    except ValueError:
+        abort(404)
+
+    return DirectAwardProject.query.filter(DirectAwardProject.public_id == project_public_id).first_or_404()
+
+
```
risicle: Could probably abstract this into a general "get transitional id" function, which could be `partial`-ized for actual use and revived any time we needed to do another transition, but that's a future concern

app/utils.py
```diff
@@ -1,11 +1,16 @@
...
 from flask import abort, request
 from six import iteritems, string_types
+import os
 from werkzeug.exceptions import BadRequest
 
 from .validation import validate_updater_json_or_400
 
 
+def random_positive_public_id():
+    return int.from_bytes(os.urandom(8), 'big') >> 1
```
risicle: You can just use `random.SystemRandom` y'know...

app/main/views/direct_award.py
```diff
@@ -181,7 +198,8 @@ def create_project_search(project_id):
         audit_type=AuditTypes.create_project_search,
         user=updater_json['updated_by'],
         data={
-            'projectId': project_id,
+            'projectId': project.id,
+            'projectPublicId': project.public_id,
```
risicle: Would the idea be for the id key always be referred to externally as the "public id" or for this to just be a transitional key?

It'd be 'orrible if we ended up with all of our `id` fields being `publicId`.

app/utils.py
samuelhwilliams: /shrug

app/main/views/direct_award.py
samuelhwilliams: Yeah, I'd probably leave that for next time.

app/main/views/direct_award.py
samuelhwilliams: External of the API, the public ID should just be 'the' ID of the resource. Internal to the API, including AuditEvents, we probably need to keep a distinction between the ID and the public/external ID.

app/models/mixins.py
```diff
@@ -0,0 +1,31 @@
...
+    def __pad_as_bytes(val):
+        """Converts an int to its equivalent in bytes and prepends with \x00 bytes until it's a multiple of 8 bytes
+        long to create ciphertexts of the same length."""
+        return val.to_bytes(8, 'big')
+
+    @staticmethod
+    def __append_equals(text):
+        """Appends '=' to the given text until it's a multiple of 8 characters long and so suitable for base32
+        decoding."""
+        return text + ''.join('=' * (4 - (len(text) % 4)))
```
galund: I don't really get why we need this - if someone passes something that isn't 8 characters long, then surely it isn't one of our identifiers, so we shouldn't be trying to decode it?

samuelhwilliams: Our encoding function strips any extra '=' on the end, because I think it looks ... nicer and it's then just an alphanumeric string, which I expect users would feel more comfortable with. It's arguable whether we want to do this or not, but if we do, then we need to put them back on when the IDs come back.

app/main/views/direct_award.py
risicle: Oh yes duh I wasn't looking at the context of this

app/models/mixins.py
```diff
@@ -0,0 +1,31 @@
...
+
+    @staticmethod
```
galund: prefer `@classmethod` cos then you can use `cls. __pad_as_bytes` rather than having to name the class again

app/models/mixins.py
galund: Ah right yeah I see what you mean, they are ugly.

On the assumption that they're always 16 characters long I'd be tempted to write this as it's more clear? `text.ljust(16, '=')`

louzoid-gds:
Hi Sam,
As discussed yesterday, out of the two options, this one is my (slight) preference.  Reasons being:
- it's consistent with how we're already doing other public id's in that it's randomly generated
- it's one less key to look after

risicle:
*boring...*

samuelhwilliams:
We'll be going ahead with this implementation as opposed to encrypted/obscured IDs. Have added the migration and updated tests. Please look over again when you have the chance @risicle @galund @TheDoubleK @idavidmcdonald

tests/main/views/test_direct_award.py
```diff
@@ -134,9 +148,11 @@ def test_list_projects_orders_by_id(self):
             res = self.client.get('/direct-award/projects?user-id={}&page={}'.format(self.user_id, i + 1))
             data = json.loads(res.get_data(as_text=True))
 
-            # Check that each project has a created_at date older than the last, i.e. reverse chronological order.
-            for project in data['projects']:
-                assert last_seen < project['id']
```
samuelhwilliams: Yep. This was broken previously. :man_facepalming:

app/main/views/direct_award.py
```diff
@@ -12,7 +12,27 @@
 from ...models.direct_award import DirectAwardProject, DirectAwardSearch
 from ...utils import (
     get_json_from_request, get_int_or_400, json_has_required_keys, pagination_links,
-    get_valid_page_or_1, validate_and_return_updater_request, drop_all_other_fields)
+    get_valid_page_or_1, validate_and_return_updater_request, drop_all_other_fields, RANDOM_ID_BIT_LENGTH)
+
+
+def get_project_by_id(project_id: int):
+    """During the transitional period, project_id may be either the internal or the public ID of the project."""
```
galund: We think we'll probably maintain this functionality going forwards, so the comment needs a little adjustment.

Arguably the function name could also be more clear, that it is intended to do something with multiple kinds of id. Somehow :)

migrations/versions/1060_add_external_id_column.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def upgrade():
+    op.add_column('direct_award_projects', sa.Column('external_id', sa.BIGINT()))
+    op.create_index(op.f('ix_direct_award_projects_external_id'), 'direct_award_projects', ['external_id'], unique=True)
+
+    # Manually generate random IDs for all existing saved searches
+    project_connection = op.get_bind()
+    project_id_internal_to_external = {}
+    for row_id, *_ in project_connection.execute('select id from direct_award_projects'):
```
idavidmcdonald: I'm not aware of the `*_` syntax. Easy explanation or a link to explain?

app/main/views/direct_award.py
```diff
@@ -193,47 +207,50 @@ def create_project_search(project_id):
...
-        abort(400, 'Project does not have a saved search: {}'.format(project_id))
+        abort(400, 'Project does not have a saved search: {}'.format(project.id))
 
     paginated_archived_services = search.archived_services.paginate(
         page=page,
         per_page=current_app.config['DM_API_PROJECTS_PAGE_SIZE'],
     )
 
+    # Serializing gives us the ID that is intended to leave the API.
+    project_id = project.serialize()['id']
```
idavidmcdonald: Do we need to do this? This is the external ID right? Can't we just reference the external ID in the archived service instead?

app/utils.py
```diff
@@ -1,11 +1,20 @@
...
 from six import iteritems, string_types
+import random
 from werkzeug.exceptions import BadRequest
 
 from .validation import validate_updater_json_or_400
 
 
+def random_positive_external_id():
+    """
+    Generate a random ID in the range 2^60 to 2^63-1. This will generate integers that are a consistent 19 digits long.
```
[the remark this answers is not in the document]
galund: I was wondering about whether it would all be nicer if we could base64 encode the new identifiers. Apparently that may cause problems with our analytics, which assume the id's are int's... but it'd both make our URLs shorter and also make it [even] more obvious which kind of id is in play.

app/main/views/direct_award.py
```diff
@@ -15,6 +15,18 @@
     get_valid_page_or_1, validate_and_return_updater_request, drop_all_other_fields)
 
 
+def get_project_by_id(project_id: int):
+    """During the transitional period, project_id may be either the internal or the public ID of the project."""
+    max_project_id = db.session.query(func.max(DirectAwardProject.id)).first()[0]
```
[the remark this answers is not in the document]
samuelhwilliams: I thought about checking ID length and decided I didn't like that. I don't think it matters too much either way to be honest. It's transitional only.

That's a better name. I'll update.

app/main/views/direct_award.py
samuelhwilliams: This will return the external ID, yes. It's not strictly necessary, and I'm open to being convinced otherwise. My justification was simply that serialize() then becomes the only place where you have a direct reference to 'external_id', but maybe hiding this fact just makes it more confusing.

I think it probably does in reality. I'll change it.

migrations/versions/1060_add_external_id_column.py
```diff
@@ -0,0 +1,77 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '1060'
+down_revision = '1050'
+
+DIRECT_AWARD_AUDIT_TYPES = ('create_project', 'create_project_search', 'lock_project', 'downloaded_project')
+
+
+def random_positive_external_id():
+    return random.randint(2 ** 60, 2 ** 63 - 1)
```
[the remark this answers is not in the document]
samuelhwilliams: 1) I didn't find a way to import it cleanly.
2) We shouldn't import it any way, as should the function ever change in code, this migration would be unable to generate IDs of the kind that this migration wanted at the time it went in.

migrations/versions/1060_add_external_id_column.py
samuelhwilliams: https://www.python.org/dev/peps/pep-3132/

idavidmcdonald: Cool. Makes sense.

What extra values are we expecting to get back from `execute`? Although as you say, not that we will need them

app/utils.py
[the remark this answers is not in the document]
samuelhwilliams: @idavidmcdonald 10^19-1 is 64-bits long, which would roll over into a negative integer ID when stored in postgres. We could do do 10**18 to 2**63-1, but I think that's probably just as opaque.

I'm open to base32/64 encoding them, but it would delay this further as we'd need to get analytics fixed up before deploying this, or accept the loss of data. Currently we can only handle integers in our analytics funnels.

I agree it would make the ID type more immediately obviously, but also the ID length is a fairly good indicator.

Open to being convinced/encouraged to go that way if people feel strongly.

migrations/versions/1060_add_external_id_column.py
samuelhwilliams: None, but it returns a sequence of one-item tuples, which would otherwise go into row_id as a tuple. This way, row_id contains the actual id (integer), which feels nicer.

tests/main/views/test_direct_award.py
```diff
@@ -48,12 +48,24 @@ def _create_project_data(self):
     def _updated_by_data(self):
         return {'updated_by': str(self.user_id)}
 
+    @staticmethod
+    def _assert_one_audit_event_created_for_only_project(audit_type):
+        projects = DirectAwardProject.query.all()
+        assert len(projects) == 1
+
+        audit_events = AuditEvent.query.filter(AuditEvent.type == audit_type
```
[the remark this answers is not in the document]
samuelhwilliams: Yeah, it just felt nice to tighten it up and get the id/external_id assertions in there as well.

app/utils.py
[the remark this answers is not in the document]
risicle: base64 is horrible to work with. base32 nice.

migrations/versions/1060_add_external_id_column.py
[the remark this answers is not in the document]
risicle: Is this not a place to use `SystemRandom`?

migrations/versions/1060_add_external_id_column.py
```diff
@@ -0,0 +1,77 @@
...
+        new_audit_data = json.dumps({'projectExternalId': project_id_internal_to_external[project_id], **audit_data})
+        audit_connection.execute(text('update audit_events set data = (:new_audit_data)::json where id = :audit_id'),
+                                 {'new_audit_data': new_audit_data, 'audit_id': audit_id})
+
+
+def downgrade():
+    op.drop_index(op.f('ix_direct_award_projects_external_id'), table_name='direct_award_projects')
+    op.drop_column('direct_award_projects', 'external_id')
+
+    # I don't think we ought to delete the external ID from the Audit Event on a downgrade... But if we do, here's
```
TheDoubleK: Why not?  The external IDs would be meaningless once they no longer exist in the projects table?

app/main/views/direct_award.py
```diff
@@ -85,6 +97,7 @@ def create_project():
         user=updater_json['updated_by'],
         data={
             'projectId': project.id,
+            'projectExternalId': project.external_id,
```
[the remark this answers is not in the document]
samuelhwilliams: This is just creating an AuditEvent, which is still fairly internal to the API. It's probably good to have both references because you could end up with either one (internal from some error in the API, external from a user), so allows you to search audit events with either reference.

This implementation isn't really about having transparent external IDs any more. It's now something developers need to know about and understand as part of how we reference resources - internal IDs for eg database joins, but external IDs as soon as it leaves the API/DB.

`project.id` should never be emitted in a response by the API, but it will still get used internally.

app/main/views/direct_award.py
```diff
@@ -303,7 +320,8 @@ def lock_project(project_id):
         audit_type=AuditTypes.lock_project,
         user=updater_json['updated_by'],
         data={
-            'projectId': project.id
+            'projectId': project.id,
```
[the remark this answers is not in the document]
samuelhwilliams: Again, this is making an Audit Event rather than being sent out of the API in a response to the request. I think we'd still want both.

migrations/versions/1060_add_external_id_column.py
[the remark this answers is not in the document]
samuelhwilliams: Is it? Why is that better than just doing this directly?

migrations/versions/1060_add_external_id_column.py
```diff
@@ -0,0 +1,77 @@
...
+    op.add_column('direct_award_projects', sa.Column('external_id', sa.BIGINT()))
+    op.create_index(op.f('ix_direct_award_projects_external_id'), 'direct_award_projects', ['external_id'], unique=True)
+
+    # Manually generate random IDs for all existing saved searches
+    project_connection = op.get_bind()
+    project_id_internal_to_external = {}
+    for row_id, *_ in project_connection.execute('select id from direct_award_projects'):
+        # Some logic to ensure all IDs generated are unique; keep track so we can update AuditEvents.
+        random_id = random_positive_external_id()
+        while random_id in project_id_internal_to_external:
```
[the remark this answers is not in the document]
samuelhwilliams: Good spot. Had it as a set originally.

migrations/versions/1060_add_external_id_column.py
samuelhwilliams: In theory, if we ever did a downgrade migration on production, the IDs _would have_ existed and users would have seen them. They might start sending in support requests saying "Where's my project? It was reference `xyz` but that seems to not work anymore."

Maybe. (of course there are other ways they could find the search they're after, so maybe that's a bad example...)

But if we keep them, we would always be able to look up by the external ID.

Anyway I wasn't sure what we _should_ do, hence the comment and writing the SQL. If we should get rid of them and make it as if it never happened, I'll uncomment it.

migrations/versions/1060_add_external_id_column.py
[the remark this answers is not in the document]
risicle: The basic version will be based on the mersenne twister, which is quite predictable, and especially so when doing things in batch, all in sequence in the same run derived from the same seed.

migrations/versions/1060_add_external_id_column.py
idavidmcdonald: Cool, thanks

app/utils.py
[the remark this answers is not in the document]
idavidmcdonald: If we don't go down the alphanumeric route (which I would suggest we don't or are at least somewhat blocked on this given that Ash is away), then I think we should go for 15 digits to match the service IDs in length.

app/main/views/direct_award.py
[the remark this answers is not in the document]
TheDoubleK: Ah OK, sorry yeah missed that. But in that case isn't the `projectId` the same as the `objectId` on the audit event, in which case we shouldn't need to capture it in the `data` field? And then we can look up the external idfrom the object id  where necessary, as we do for service IDs I think?  Although maybe it is nicer not to have to do that lookup...

app/utils.py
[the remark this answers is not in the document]
samuelhwilliams: @TheDoubleK Switched to generating 15-digit IDs. Good spot. Didn't think about how they look in spreadsheets.

@idavidmcdonald Given they're now fully containable within a BIGINT, I've switched to using `.randint(10 ** 14, (10 ** 15) - 1)`

@risicle I've updated to use `random.SystemRandom`; thanks for the explanation.

migrations/versions/1060_add_external_id_column.py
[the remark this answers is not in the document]
samuelhwilliams: Will use SystemRandom now. 👍

.gitignore
```diff
@@ -4,6 +4,8 @@ venv*
 __pycache__
 version_label
 .cache
+.idea/*
```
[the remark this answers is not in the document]
samuelhwilliams: 👍

app/main/views/direct_award.py
```diff
@@ -15,6 +15,18 @@
     get_valid_page_or_1, validate_and_return_updater_request, drop_all_other_fields)
 
 
+def get_project_by_id(project_id: int):
+    """During the transitional period, project_id may be either the internal or the public ID of the project."""
+    max_project_id = db.session.query(func.max(DirectAwardProject.id)).first()[0]
+
+    if 0 < project_id <= max_project_id:
```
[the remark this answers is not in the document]
samuelhwilliams: It's enforced only by the fact that PKs are predominantly tiny (<1000) and the randomly-generated IDs are all 15 digits long.

It's an assumption... but doesn't feel like much of a risk. I'll comment.

app/main/views/direct_award.py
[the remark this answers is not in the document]
samuelhwilliams: Hmm, looks like you're right, we have object_id directly and that is indeed the project's internal ID. So you think we should retain the `projectId` key in the AuditEvent's data and populate it with the project's external ID?

tests/main/views/test_drafts.py
```diff
@@ -994,14 +994,14 @@ def test_drafts_made_from_services_are_deleted_when_published(self):
...
-        generate_new_service_id.side_effect = (
-            '1234567890123457',
-            '1234567890123457',
-            '1234567890123457',
-            '2222222222222222',
+        random_positive_external_id.side_effect = (
+            '123456789012345',
+            '123456789012345',
+            '123456789012345',
+            '222222222222222',
```
samuelhwilliams: I've changed these IDs only so that they match the length of randomly generated IDs.

migrations/versions/1060_add_external_id_column.py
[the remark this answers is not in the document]
risicle:
```
try:
    random_id = next(new_id for new_id in (
        random_positive_external_id()
        for _ in range(1024)
    ) if new_id not in project_id_internal_to_external.values())
except StopIteration:
    raise SomeAppropriateError("This is getting silly")
```

app/utils.py
```diff
@@ -1,11 +1,20 @@
...
 from six import iteritems, string_types
+import random
 from werkzeug.exceptions import BadRequest
 
 from .validation import validate_updater_json_or_400
 
 
+def random_positive_external_id() -> int:
+    """
+    Generate a random ID in the range 2^60 to 2^63-1. This will generate integers that are a consistent 19 digits long.
```
idavidmcdonald: Docblock needs updating

migrations/versions/1060_add_external_id_column.py
samuelhwilliams: Be interested in your thoughts @idavidmcdonald

app/utils.py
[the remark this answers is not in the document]
risicle: Y'know if you wanted to restrict the range to a fixed-decimal length, you could use a range up to `2**49` and stuff the most significant bits with 1s to ensure you never have the "leading zeros" ambiguity.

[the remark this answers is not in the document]
risicle: (so, most significant *bit* i guess)

samuelhwilliams:
Migration split out to https://github.com/alphagov/digitalmarketplace-api/pull/703

app/main/views/direct_award.py
```diff
@@ -15,6 +15,20 @@
     get_valid_page_or_1, validate_and_return_updater_request, drop_all_other_fields)
 
 
+def get_project_by_id_or_404(project_id: int):
+    """During the transitional period, project_id may be either the internal or the public ID of the project."""
+    max_project_id = db.session.query(func.max(DirectAwardProject.id)).first()[0]
+
+    # We assume that our internal project IDs/PKs will never be as long as our randomly-generated external IDs (which
+    # are 15-digits long.
```
TheDoubleK: Closing bracket?

app/main/views/direct_award.py
```diff
@@ -15,6 +15,20 @@
     get_valid_page_or_1, validate_and_return_updater_request, drop_all_other_fields)
 
 
+def get_project_by_id_or_404(project_id: int):
+    """During the transitional period, project_id may be either the internal or the public ID of the project."""
+    max_project_id = db.session.query(func.max(DirectAwardProject.id)).first()[0]
```
TheDoubleK: Do we even need this check?  Could just try to fetch by ID, if no project found then try for external ID or 404?

app/main/views/direct_award.py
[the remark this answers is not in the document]
TheDoubleK: No, I think the object ID for audit events should probably always reference the internal ID.  At least, this is what we do for audit events relating to services.  Which might not be the best thing, but now we've got it let's stay consistent.

For services we don't even capture the external service id in the data at all.

But feels like it could be convenient to do so, so maybe keep the `external` in data here?

app/models/direct_award.py
```diff
@@ -1,19 +1,23 @@
...
 from app.models import User, ValidationError, ArchivedService
+
 from dmutils.formats import DATETIME_FORMAT
 
 
 class DirectAwardProject(db.Model):
     __tablename__ = 'direct_award_projects'
 
     id = db.Column(db.Integer, primary_key=True)
+    external_id = db.Column(BIGINT, default=random_positive_external_id, nullable=False, index=True, unique=True)
```
TheDoubleK: For supplier ID we use `db.BigInteger` - I'm pretty sure these will be the same thing under the hood, but is it worth using that here for consistency?
```
    supplier_id = db.Column(db.BigInteger,
                            db.ForeignKey('suppliers.supplier_id'),
                            index=True, unique=False, nullable=True)
```

migrations/versions/1060_add_external_id_column.py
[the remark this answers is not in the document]
TheDoubleK: I don't think this issue is resolved yet?

migrations/versions/1060_add_external_id_column.py
TheDoubleK: It's an unlikely scenario, but I guess there's no harm in keeping them. And would save this scary looking update of all the audit JSON from existing.

tests/main/views/test_drafts.py
TheDoubleK: Yeah this must have been from the good-old-days of 16 digit IDs that everyone hated.

app/main/views/direct_award.py
samuelhwilliams: Guess not. I'll add a try/catch around the query instead.

app/main/views/direct_award.py
samuelhwilliams: :trollface:

app/main/views/direct_award.py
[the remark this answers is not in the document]
samuelhwilliams: Sorry, I wasn't clear. `object_id` should definitely stay as the internal ID. I was asking if we should change projectId to reference the external ID or if we should add a new key to the data for projectExternalId.... as this change is.

Basically do you want me to change anything here?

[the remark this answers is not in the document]
TheDoubleK: The change here puts two Id's in the JSON data field of the audit event. We need at most one in there (the external ID). The internal project ID is already stored as the object ID on the audit event, so definitely doesn't need to go in the data.

I'm in two minds about having the external one in there, as we don't do that for services AFAIK (although come to think of it maybe we do in the service creation event? Definitely not for other service events). But it feels like it could be useful and no harm to having it in here, so maybe keep it.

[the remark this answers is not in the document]
TheDoubleK: The change here puts two Id's in the JSON data field of the audit event. We need at most one in there (the external ID). The internal project ID is already stored as the object ID on the audit event, so definitely doesn't need to go in the data.

I'm in two minds about having the external one in there, as we don't do that for services AFAIK (although come to think of it maybe we do in the service creation event? Definitely not for other service events). But it feels like it could be useful and no harm to having it in here, so maybe keep it.

samuelhwilliams:
All comments addressed @TheDoubleK.

Also added tests around service ID collisions.

PR #691, opened 2017-11-15
Index briefs on update

Wynndow:
For this card on trello: [https://trello.com/c/JR2Ch0ml](https://trello.com/c/JR2Ch0ml)

There will be a PR on the apiclient too, which will need to be merged and pulled in here. Will update when it's done.

### Create `index_object` and use in `index_services`
We were only ever indexing services, now we want to index briefs too.
This moves the call to the api client to a shareable function which can
be used by different object types.

### Add framework/doc_type/index mapping to config
We've agreed that this isn't the best place for this to live. It would
be better in the frameworks table in th db. However, for now, this will
work and we'll look to moving it out at the end of the mission.

### Index briefs on status update
The only time we'll need to re-index a brief is when it's status is
changed.

When a brief response is marked with an `awarded-at` timestamp, this has
the knock on effect of changing the briefs status, hence why we reindex.

### Remove yield tests to remove deprecation warnings
Yield tests are being removed in pytest 4.0, and as a result we were
getting a lot of annoying warnings every test run.

Also a warning about using [pytest] instead of [tool:pytest] in
setup.cfg has been fixed

### Fix tests to work with new indexing function
The new index_object function takes different arguments, and different
objects needed mocking to ensure that existing tests are kept relevant.

More tests need adding to test the new functionality.

### Handle key errors when getting mapping
If an attempt was made to index an object which there wasn't a mapping
for in the config, we'd get a KeyError which would cause a 500. We
probably don't want to do this and instead log a warning

### Constrain `index_briefs` to non draft DOS
For a brief, the framework family should always be
digital-outcomes-and-specialists at the moment.

We also don't want to be indexing briefs before a buyer has published
them, so no drafts.

### DIFF
--- a/app/brief_utils.py
+++ b/app/brief_utils.py
@@ -40,9 +40,14 @@ def get_supplier_service_eligible_for_brief(supplier, brief):
 
 
 def index_brief(brief):
-    index_object(
-        framework=brief.framework.slug,
-        object_type='briefs',
-        object_id=brief.id,
-        serialized_object=brief.serialize(),
-    )
+    if (
+        brief.framework.framework == 'digital-outcomes-and-specialists' and
+        brief.status != 'draft'
+    ):
+
+        index_object(
+            framework=brief.framework.slug,
+            object_type='briefs',
+            object_id=brief.id,
+            serialized_object=brief.serialize(),
+        )
--- a/app/brief_utils.py
+++ b/app/brief_utils.py
@@ -47,7 +47,7 @@ def index_brief(brief):
 
         index_object(
             framework=brief.framework.slug,
-            object_type='briefs',
+            doc_type='briefs',
             object_id=brief.id,
             serialized_object=brief.serialize(),
         )
--- a/app/brief_utils.py
+++ b/app/brief_utils.py
@@ -3,6 +3,7 @@ from flask import abort
 from .models import Service
 from .validation import get_validation_errors
 from .service_utils import filter_services
+from .utils import index_object
 
 
 def validate_brief_data(brief, enforce_required=True, required_fields=None):
@@ -36,3 +37,12 @@ def get_supplier_service_eligible_for_brief(supplier, brief):
     services = services.filter(Service.supplier_id == supplier.supplier_id)
 
     return services.first()
+
+
+def index_brief(brief):
+    index_object(
+        framework=brief.framework.slug,
+        object_type='briefs',
+        object_id=brief.id,
+        serialized_object=brief.serialize(),
+    )
--- a/app/brief_utils.py
+++ b/app/brief_utils.py
@@ -40,11 +40,7 @@ def get_supplier_service_eligible_for_brief(supplier, brief):
 
 
 def index_brief(brief):
-    if (
-        brief.framework.framework == 'digital-outcomes-and-specialists' and
-        brief.status != 'draft'
-    ):
-
+    if brief.status != 'draft':
         index_object(
             framework=brief.framework.slug,
             doc_type='briefs',
--- a/app/utils.py
+++ b/app/utils.py
@@ -164,21 +164,23 @@ def get_request_page_questions():
     return json_payload.get('page_questions', [])
 
 
-def index_object(framework, object_type, object_id, serialized_object):
+def index_object(framework, doc_type, object_id, serialized_object):
     try:
-        index_name = current_app.config['DM_FRAMEWORK_TO_ES_INDEX_MAPPING'][framework][object_type]
+        index_name = current_app.config['DM_FRAMEWORK_TO_ES_INDEX_MAPPING'][framework][doc_type]
 
         try:
-            search_api_client.index(index_name=index_name,
-                                    object_type=object_type,
-                                    object_id=object_id,
-                                    serialized_object=serialized_object)
+            search_api_client.index(
+                index_name=index_name,
+                object_id=object_id,
+                serialized_object=serialized_object,
+                doc_type=doc_type,
+            )
         except dmapiclient.HTTPError as e:
             current_app.logger.warning(
                 'Failed to add {} object with id {} to {} index: {}'.format(
-                    object_type, object_id, index_name, e.message))
+                    doc_type, object_id, index_name, e.message))
 
     except KeyError as e:
         current_app.logger.warning(
-            "Failed to find index name for framework '{}' with object type '{}'".format(framework, object_type)
+            "Failed to find index name for framework '{}' with object type '{}'".format(framework, doc_type)
         )
--- a/app/utils.py
+++ b/app/utils.py
@@ -181,6 +181,6 @@ def index_object(framework, doc_type, object_id, serialized_object):
                     doc_type, object_id, index_name, e.message))
 
     except KeyError as e:
-        current_app.logger.warning(
+        current_app.logger.error(
             "Failed to find index name for framework '{}' with object type '{}'".format(framework, doc_type)
         )
--- a/app/utils.py
+++ b/app/utils.py
@@ -166,7 +166,7 @@ def get_request_page_questions():
 
 def index_object(framework, doc_type, object_id, serialized_object):
     try:
-        index_name = current_app.config['DM_FRAMEWORK_TO_ES_INDEX_MAPPING'][framework][doc_type]
+        index_name = current_app.config['DM_FRAMEWORK_TO_ES_INDEX'][framework][doc_type]
 
         try:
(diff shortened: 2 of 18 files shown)


config.py
```diff
@@ -42,6 +42,15 @@ class Config:
 
     VCAP_SERVICES = None
 
+    DM_FRAMEWORK_TO_ES_INDEX_MAPPING = {
+        'g-cloud-9': {
+            'services': 'g-cloud-9'
+        },
+        'digital-outcomes-and-specialists-2': {
```
idavidmcdonald: What about our DOS1 briefs which are able to change status still? Also, I assume both DOS1 and DOS2 briefs are being indexed together in the same index rather than in two different indexes or will the FE be searching through two indexes combined?

risicle:
> The only time we'll need to re-index a brief is when it's status is changed.

Disagree - the brief is going to go into the index when it's first created (as a draft) and the draft's updates are going to need to be synchronized the whole way.

Life is just much easier if we just try and keep objects in sync *everywhere*. (hence my original desire to use a more low level ORM trigger)

Wynndow:
@galund Can me and Rob grab a hang out with you to talk about the drafts thing? ^^^

app/brief_utils.py
```diff
@@ -36,3 +37,17 @@ def get_supplier_service_eligible_for_brief(supplier, brief):
     services = services.filter(Service.supplier_id == supplier.supplier_id)
 
     return services.first()
+
+
+def index_brief(brief):
+    if (
+        brief.framework.framework == 'digital-outcomes-and-specialists' and
```
galund: I'm not sure I see why we need this constraint. We only have DOS briefs anyway, and if there was another framework, then we'd need to do a whole heap of work to decide where those briefs lived _anyway_. They might well want to end up in the same index, I don't think we know?

app/utils.py
```diff
@@ -161,3 +162,25 @@ def purge_nulls_from_data(data):
...
+                serialized_object=serialized_object,
+                doc_type=doc_type,
+            )
+        except dmapiclient.HTTPError as e:
+            current_app.logger.warning(
+                'Failed to add {} object with id {} to {} index: {}'.format(
+                    doc_type, object_id, index_name, e.message))
+
+    except KeyError as e:
+        current_app.logger.warning(
```
galund: I reckon that should be an error, not a warning.

app/brief_utils.py
Wynndow: I was trying to stay in keeping with what was being done when indexing g-cloud services. I guess the check is there to prevent something like a g-cloud service being indexed as a brief by mistake. Happy to remove if you don't think it's necessary?

app/utils.py
Wynndow: `logger.error()` ?

galund: `logger.error` I think should be fine; I don't know how anyone else is searching Kibana, but I tend to do `loglevel: ERROR OR status:500` when doing the 2nd line thang?

We don't want to abandon the entire transaction just because indexing failed.

app/brief_utils.py
galund: So hopefully no-one will call this with a service, I assume that's fairly not-possible.

For services, perhaps this ought now to be controlled purely by `DM_FRAMEWORK_TO_ES_INDEX_MAPPING`?

Of course that's a bit of a contradiction with the other conversation, where I'm claiming a missing mapping should be an error - this implies we should just ignore it. Ho hum...

Wynndow: Yeah, it's not a particularly helpful check. I'll get rid of it.

config.py
Wynndow: So actually yes, dos1 and dos2 briefs will be going into the same index as they are both compatible with the same mapping.

galund: - [ ] (And we should therefore add DOS1 into this mapping?)

tests/test_brief_utils.py
```diff
@@ -0,0 +1,32 @@
...
+            dos2 = Framework.query.filter(Framework.slug == 'digital-outcomes-and-specialists-2').first()
+
+            with mock.patch.object(Brief, "serialize", return_value={'serialized': 'object'}):
+                brief = Brief(status='live', framework=dos2, data={'requirementsLength': '1 week'})
+                index_brief(brief)
+
+            index_object.assert_called_once_with(
+                framework='digital-outcomes-and-specialists-2',
+                doc_type='briefs',
+                object_id=None,
```
idavidmcdonald: Should we change the object_id to not be `None` so it is more realistic?

config.py
```diff
@@ -42,6 +42,18 @@ class Config:
 
     VCAP_SERVICES = None
 
+    DM_FRAMEWORK_TO_ES_INDEX_MAPPING = {
+        'g-cloud-9': {
+            'services': 'g-cloud-9'
+        },
+        'digital-outcomes-and-specialists': {
+            'briefs': 'briefs-digital-outcomes-and-specialists-2'
```
idavidmcdonald: Please do correct me as I have very little idea about the difference between the search API mapping and the search API indexes but in terms of longevity, what is the plan for when DOS3 comes along - assuming nothing changes with how we display briefs on the brief listing page. Will we then add a new mapping for `briefs-digital-outcomes-and-specialists-3` and swap everything to it or are we intending to keep everything on the same mapping (in which case a more generic name like `briefs-digital-outcomes-and-specialists` would need no changes).

Wynndow: It really depends on what happens with DOS3. If nothing much changes then it may be possible to reuse the mapping. If briefs do change, it may mean that they need a different mapping which would require a new index. If we can reuse the mapping, it's a pretty minor change to update the index name if that's what we decide to do.

galund: If we were to end up with something non-compatible in the data structure of DOS3 briefs, we might potentially have to define some custom process to transform the old structure into the new structure, as part of indexing. Potentially the Search API would be in charge of doing that (somehow!). It's likely very desirable to be able to continue to get all the briefs into one index, at least all the DOS ones.

galund: Oh, but yes probably the name of the index should just be `digital-outcomes-and-specialists` - the 'briefs' part will generally in the URLs anyway as it's the mapping type (of which as Chris said, there can only be one).

Wynndow: I think the idea behind the index name, is that it's a pattern we can re-use if we have multiple objects indexed for the same framework. So if dos had services we could have `services-digital-outcomes-and-specialists-2` as well as `briefs-digital-outcomes-and-specialists-2`

tests/test_brief_utils.py
Wynndow: Yep, makes sense. Have updated the test to commit the brief so it has an ID.

config.py
galund: Yeah, fair enough, but I don't propose we change the g-cloud one, and it seems fairly unlikely we'll be indexing DOS services any time soon. But as it will contain DOS1 stuff, we'd still better drop the `-2`?

Wynndow: I had a chat with @risicle about the `-2`. The mapping for the index is generated from the dos2 set up in the frameworks repo. If we drop the `-2` it could look like it's referring to dos1, as it's not actually called `digital-outcomes-and-specialists-1`, rather just `digital-outcomes-and-specialists`.

galund: Hmm, I think if the index contains both, then it would be weird for it's name to just mention one of them. Same goes for the name of the mapping, we can discuss Wednesday.

Wynndow: Have had a chat with Rob, and he has no strong opinion about it, and I'm happy to change. `digital-outcomes-and-specialists` it is!

config.py
```diff
@@ -44,6 +44,18 @@ class Config:
 
     VCAP_SERVICES = None
 
+    DM_FRAMEWORK_TO_ES_INDEX_MAPPING = {
```
galund: hmm I just spotted that this name is potentially confusing, because it uses the word 'mapping', which is - uh - overloaded now somewhat!

Perhaps `DM_INDEX_NAME_BY_FRAMEWORK_AND_DOC_TYPE`? (eww long I know)

I was going to suggest we come up with some more self-documenting structure like

```
DM_INDEXES = [
    { 'framework-slug': 'g-cloud-9', 'doc-type': 'services', 'index-name': 'g-cloud-9', }
    # etc
]

But then you'd have to have a function somewhere that parsed it out, and I've no idea where that would go, so let's not do that.

Wynndow: How about just `DM_FRAMEWORK_TO_ES_INDEX`?

galund: Fine by me

PR #692, opened 2017-11-16
straw man for standardising responses

idavidmcdonald:
I've just done the brief-responses view to see what this might look like.

https://trello.com/c/HzdillME/50-api-response-counts

### DIFF
--- a/app/main/views/brief_responses.py
+++ b/app/main/views/brief_responses.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 
-from flask import jsonify, abort, request, current_app
+from flask import abort, request, current_app
 from sqlalchemy.exc import IntegrityError, DataError
 
 from dmapiclient.audit import AuditTypes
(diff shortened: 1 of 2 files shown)


app/main/views/brief_responses.py
```diff
@@ -230,21 +231,22 @@ def list_brief_responses():
         )
 
     if brief_id or supplier_id:
-        return jsonify(
-            briefResponses=[brief_response.serialize() for brief_response in brief_responses.all()],
+        brief_responses = brief_responses.all()
+        return list_result_response(
+            total_count=len(brief_responses),
+            briefResponses=[brief_response.serialize() for brief_response in brief_responses],
             links={'self': url_for('.list_brief_responses', supplier_id=supplier_id, brief_id=brief_id)}
```
idavidmcdonald: Will likely remove this individual link as I highly doubt it is being used. Either we can be consistent that all endpoints contain a link to themselves or none should.

samuelhwilliams: I think i's generally nice to include `self` link on our responses, but as we're the only consumer it doesn't matter too much.

TheDoubleK: +1 to having a self link everywhere.
http://blog.ploeh.dk/2013/05/03/rest-lesson-learned-consider-a-self-link-on-all-resources/

idavidmcdonald: Interesting read. We could easily add this to all our resources using this PR. Not that it would probably have any immediate effects for us but there are things that we could do to produce more canonical URLs e.g. ordering query params.

idavidmcdonald:
Closing as will not be implementing this way.

PR #693, opened 2017-11-16
Standardise serialized json responses

idavidmcdonald:
https://trello.com/c/HzdillME/50-api-response-counts

We very often return JSON responses which are
- single serialized objects from our models
- lists of serialized objects from our models
- paginated lists of serialized objects from our models

However, when doing this in the past we have not had consistency.
For example, some lists of objects may include meta such as the
total number of results. Other responses may include links to
their own resources.

This introduces the wrappers for those common types of responses
so we can introduce consistency. This means all lists of results
will contain meta totals for example. If we want to add links to
all our results then that can easily be done in one change.

To implement this around the codebase there were a few exceptions/
things that needed changing. Some times we are not serializing
objects but instead jsonifying dictionarys of data - this case has
been ignored. A handful of our list responses were including either
an empty links dict or a dict containing a link to the resource it
self. These have been removed for consistency as it was rarely done
and our front end does not use them. We can choose to implement
this for all responses if we so choose at a later point.

Two smaller changes also include fixing an incorrect status code
which should have been 201 rather 200 when creating an object, and
moving the number of paginated results for buyer domains into the
config.

### DIFF
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -1,20 +1,23 @@
 from datetime import datetime
+
 from flask import jsonify, abort
-from .. import main
 from sqlalchemy.exc import IntegrityError
 from dmapiclient.audit import AuditTypes
+
+from .. import main
 from ...models import (
     AuditEvent, db, Framework, FrameworkAgreement, SupplierFramework, User
 )
-
 from ...utils import (
     get_json_from_request,
     json_has_required_keys,
     json_has_keys,
+    single_result_response,
     validate_and_return_updater_request,
 )
-
 from ...supplier_utils import validate_agreement_details_data
+
+RESOURCE_NAME = "agreement"
 
 
 @main.route('/agreements', methods=['POST'])
@@ -66,14 +69,13 @@ def create_framework_agreement():
     db.session.add(audit_event)
     db.session.commit()
 
-    return jsonify(agreement=framework_agreement.serialize()), 201
+    return single_result_response(RESOURCE_NAME, framework_agreement), 201
 
 
 @main.route('/agreements/<int:agreement_id>', methods=['GET'])
 def get_framework_agreement(agreement_id):
     framework_agreement = FrameworkAgreement.query.filter(FrameworkAgreement.id == agreement_id).first_or_404()
-
-    return jsonify(agreement=framework_agreement.serialize())
+    return single_result_response(RESOURCE_NAME, framework_agreement), 200
 
 
 @main.route('/agreements/<int:agreement_id>', methods=['POST'])
@@ -137,7 +139,7 @@ def update_framework_agreement(agreement_id):
         db.session.rollback()
         return jsonify(message="Database Error: {0}".format(e)), 400
 
-    return jsonify(agreement=framework_agreement.serialize())
+    return single_result_response(RESOURCE_NAME, framework_agreement), 200
 
 
 @main.route('/agreements/<int:agreement_id>/sign', methods=['POST'])
@@ -187,7 +189,7 @@ def sign_framework_agreement(agreement_id):
         db.session.rollback()
         return jsonify(message="Database Error: {0}".format(e)), 400
 
-    return jsonify(agreement=framework_agreement.serialize())
+    return single_result_response(RESOURCE_NAME, framework_agreement), 200
 
 
 @main.route('/agreements/<int:agreement_id>/on-hold', methods=['POST'])
@@ -224,7 +226,7 @@ def put_signed_framework_agreement_on_hold(agreement_id):
         db.session.rollback()
         return jsonify(message="Database Error: {0}".format(e)), 400
 
-    return jsonify(agreement=framework_agreement.serialize())
+    return single_result_response(RESOURCE_NAME, framework_agreement), 200
 
 
 @main.route('/agreements/<int:agreement_id>/approve', methods=['POST'])
@@ -288,4 +290,4 @@ def approve_for_countersignature(agreement_id):
         db.session.rollback()
         return jsonify(message="Database Error: {0}".format(e)), 400
 
-    return jsonify(agreement=framework_agreement.serialize())
+    return single_result_response(RESOURCE_NAME, framework_agreement), 200
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -1,18 +1,27 @@
-from flask import jsonify, abort, request, current_app
 from datetime import datetime, timedelta
-from ...models import AuditEvent
+
+from flask import jsonify, abort, request, current_app
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.sql.expression import true, false
 from sqlalchemy.orm import class_mapper
-from ...utils import pagination_links, get_valid_page_or_1
-from .. import main
-from ... import db, models
 from dmapiclient.audit import AuditTypes
 from dmutils.config import convert_to_boolean
 from dmutils.formats import DATE_FORMAT
-from ...validation import is_valid_date, is_valid_acknowledged_state
-from ...utils import get_json_from_request, json_has_required_keys, validate_and_return_updater_request
 
+from .. import main
+from ... import db, models
+from ...models import AuditEvent
+from ...validation import is_valid_date, is_valid_acknowledged_state
+from ...utils import (
+    get_json_from_request,
+    get_valid_page_or_1,
+    json_has_required_keys,
+    paginated_result_response,
+    single_result_response,
+    validate_and_return_updater_request,
+)
+
+RESOURCE_NAME = "auditEvents"
 
 AUDIT_OBJECT_TYPES = {
     "suppliers": models.Supplier,
@@ -148,19 +157,14 @@ def list_audits():
     sort_order = db.desc if convert_to_boolean(request.args.get('latest_first')) else db.asc
     audits = audits.order_by(sort_order(AuditEvent.created_at), sort_order(AuditEvent.id))
 
-    audits = audits.paginate(
+    return paginated_result_response(
+        result_name=RESOURCE_NAME,
+        results_query=audits,
         page=page,
-        per_page=per_page
-    )
-
-    return jsonify(
-        auditEvents=[audit.serialize() for audit in audits.items],
-        links=pagination_links(
-            audits,
-            '.list_audits',
-            request.args
-        )
-    )
+        per_page=per_page,
+        endpoint='.list_audits',
+        request_args=request.args
+    ), 200
 
 
 @main.route('/audit-events', methods=['POST'])
@@ -201,7 +205,7 @@ def create_audit_event():
     db.session.add(audit_event)
     db.session.commit()
 
-    return jsonify(auditEvents=audit_event.serialize()), 201
+    return single_result_response(RESOURCE_NAME, audit_event), 201
 
 
 @main.route('/audit-events/<int:audit_id>/acknowledge', methods=['POST'])
@@ -225,7 +229,7 @@ def acknowledge_audit(audit_id):
         db.session.rollback()
         abort(400, e.orig)
 
-    return jsonify(auditEvents=audit_event.serialize()), 200
+    return single_result_response(RESOURCE_NAME, audit_event), 200
 
 
 # this is a "view without a route" for at the moment - it is used as an "inner" view implementation for the service
@@ -294,4 +298,4 @@ def get_audit_event(audit_id):
     if audit_event is None:
         abort(404, "No audit event with this id")
 
-    return jsonify(auditEvents=audit_event.serialize()), 200
+    return single_result_response(RESOURCE_NAME, audit_event), 200
--- a/app/main/views/brief_responses.py
+++ b/app/main/views/brief_responses.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 
-from flask import jsonify, abort, request, current_app
+from flask import abort, request, current_app
 from sqlalchemy.exc import IntegrityError, DataError
 
 from dmapiclient.audit import AuditTypes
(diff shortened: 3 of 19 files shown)


samuelhwilliams:
This version feels a lot cleaner to me. Nice one. 👍 

Would still be good to keep those self links as discussed on the other PR.

app/main/views/agreements.py
```diff
@@ -8,10 +8,8 @@
 )
 
 from ...utils import (
-    get_json_from_request,
-    json_has_required_keys,
-    json_has_keys,
-    validate_and_return_updater_request,
+    get_json_from_request, json_has_required_keys, json_has_keys,
+    validate_and_return_updater_request, single_result_response
```
risicle: Formatting wise, much *much* prefer the former. Far *far* more copy-pastable and easier to see what's present and what's absent from a quick glance.

app/main/views/audits.py
```diff
@@ -4,14 +4,16 @@
...
 from ... import db, models
 from dmapiclient.audit import AuditTypes
 from dmutils.config import convert_to_boolean
 from dmutils.formats import DATE_FORMAT
 from ...validation import is_valid_date, is_valid_acknowledged_state
-from ...utils import get_json_from_request, json_has_required_keys, validate_and_return_updater_request
+from ...utils import (
+    get_json_from_request, json_has_required_keys, validate_and_return_updater_request,
+    single_result_response, paginated_result_response, get_valid_page_or_1
+)
```
risicle: Again, _please_ can we formalize on single import per line, trailing comma, preferably alphabetized.

risicle: when i say import i obviously mean imported object

app/main/views/briefs.py
```diff
@@ -107,7 +107,11 @@ def get_brief(brief_id):
         Brief.id == brief_id
     ).first_or_404()
 
-    return jsonify(briefs=brief.serialize(with_users=True, with_clarification_questions=True))
+    return single_result_response(
+        "briefs",
+        brief,
+        serialize_kwargs={"with_users": True, "with_clarification_questions": True}
```
risicle: nice

risicle:
Definitely the direction things should be moving.

app/utils.py
```diff
@@ -58,6 +58,28 @@ def pagination_links(pagination, endpoint, args):
...
+    return jsonify(**{result_name: result.serialize(**serialize_kwargs)})
+
+
+def list_result_response(result_name, results_query, serialize_kwargs={}):
+    meta = result_meta(results_query.count())
+    serialized_results = [result.serialize(**serialize_kwargs) for result in results_query]
+    return jsonify(meta=meta, **{result_name: serialized_results})
+
+
+def paginated_result_response(result_name, results_query, page, per_page, endpoint, request_args, serialize_kwargs={}):
```
risicle: This should probably have a docstring. In fact, they all probably should.

app/utils.py
```diff
@@ -58,6 +58,28 @@ def pagination_links(pagination, endpoint, args):
...
+    meta = result_meta(results_query.count())
```
risicle: I suspect this will cause the query to get executed for a *third* time. The pagination routines are bad enough in that they make a separate "count" call to calculate the page count, which is often ~90% of the effort for the db to perform as the actual query.

I would advise trying to dig in to the pagination routines to get its count out. If it makes this difficult or impossible, that probably means it needs to be hit with a hammer itself.

app/main/views/audits.py
risicle: lovely thankyou

risicle:
So basically neat, thumbs up except for the additional db hit for the total.

app/main/views/frameworks.py
```diff
@@ -1,26 +1,29 @@
...
 
 @main.route('/frameworks', methods=['GET'])
 def list_frameworks():
-    frameworks = Framework.query.all()
+    frameworks = Framework.query
 
-    return jsonify(
-        frameworks=[f.serialize() for f in frameworks]
-    )
+    return list_result_response("frameworks", frameworks)
```
samuelhwilliams: Probably don't need a separate variable at this point, but :man_shrugging:

app/utils.py
idavidmcdonald: Good idea. Have now implemented.

app/main/views/frameworks.py
idavidmcdonald: Will tidy

app/utils.py
idavidmcdonald: Added, let me know what you think

app/utils.py
```diff
@@ -58,6 +58,40 @@ def pagination_links(pagination, endpoint, args):
...
+    return jsonify(**{result_name: result.serialize(**serialize_kwargs)})
+
+
+def list_result_response(result_name, results_query, serialize_kwargs={}):
+    """
+    Return a standardised JSON response for a SQLAlchemy result query e.g. a query that will retrieve closed briefs.
+    The query should not be executed before being passed in as a argument so we can manipulate the query object (i.e. to
+    get the count). Results will be returned in a list and use the results `serialize` method for presentation.
+    """
+    meta = result_meta(results_query.count())
```
idavidmcdonald: @risicle Was just checking if you had any thoughts about my use of `count` here? It will trigger another call to the DB. 

Two alternatives:
1. Pass in a result query as we are currently doing but use use python to count the results after they have been retrieved.
2. Pass in a list of results instead of a result query (similar to how we do for a single result rather than how we do for pagination) and do the counting in python again

I think I prefer the current solution but that opinion is lightly held.

risicle: Why do we need to give a count in this case at all? I thought the purpose of a count was to allow you to see the number of items when most of them are paged away. If we're getting a list of all of them, well, the consumer can just count them themselves, surely? If it's just for completeness/api symmetry then yeah, count them in python.

idavidmcdonald: We are generally doing it on this type of response already (it also feels slightly helpful when doing dev work) so I'd be keen to keep it.

I'll swap and use python to count.

risicle: If we're doing this already and incurring an unnecessary db hit just to get the count then my mouth is agape.

idavidmcdonald:
@samuelhwilliams I've added some tests. I'm not particularly happy with them though so would be interested in you taking a look and maybe doing some quick pairing tomorrow to see if we can improve them a bit. There was no unit test coverage for `pagination_links` itself and I started looking at adding it but it's a bit of a mess to add and you end up essentially just mocking most of the things so I'm not sure if it's worth trying further on.

Apart from that I think this will be good to go (after a squash of my WIP test commits).

app/main/views/agreements.py
```diff
@@ -66,14 +67,13 @@ def create_framework_agreement():
     db.session.add(audit_event)
     db.session.commit()
 
-    return jsonify(agreement=framework_agreement.serialize()), 201
+    return single_result_response("agreement", framework_agreement), 201
```
samuelhwilliams: It would be good to have `"agreement"` as a constant (same with other places where we pass in the name of the resource).

app/main/views/brief_responses.py
```diff
@@ -230,21 +237,13 @@ def list_brief_responses():
         )
 
     if brief_id or supplier_id:
-        return jsonify(
-            briefResponses=[brief_response.serialize() for brief_response in brief_responses.all()],
-            links={'self': url_for('.list_brief_responses', supplier_id=supplier_id, brief_id=brief_id)}
-        )
+        return list_result_response(RESOURCE_NAME, brief_responses)
```
samuelhwilliams: Hmm, in the change above you've explicitly added a return code of 200. Do we want to do this explicitly everywhere? If so, you're missing a few places (eg here and below, possibly more).

app/main/views/briefs.py
```diff
@@ -166,26 +179,20 @@ def list_briefs():
...
-            },
-            links=pagination_links(
-                briefs,
-                '.list_briefs',
-                request.args
-            ),
+            endpoint='.list_briefs',
+            request_args=request.args,
+            serialize_kwargs={"with_users": with_users, "with_clarification_questions": with_clarification_questions}
         )
```
samuelhwilliams: Another bit without a return code.

app/main/views/buyer_domains.py
```diff
@@ -46,26 +52,19 @@ def create_buyer_email_domain():
...
-        },
-        links=pagination_links(
-            buyer_email_domains,
-            '.list_buyer_email_domains',
-            request.args
-        ),
+        per_page=current_app.config['DM_API_BUYER_DOMAINS_PAGE_SIZE'],
+        endpoint='.list_buyer_email_domains',
+        request_args=request.args
     )
```
samuelhwilliams: ☝️

app/main/views/direct_award.py
```diff
@@ -116,14 +112,14 @@ def create_project():
...
-    return jsonify(project=project.serialize()), 201
+    return single_result_response("project", project), 201
 
 
 @main.route('/direct-award/projects/<int:project_external_id>', methods=['GET'])
 def get_project(project_external_id):
     project = get_project_by_id_or_404(project_external_id)
 
-    return jsonify(project=project.serialize(with_users=True))
+    return single_result_response("project", project, serialize_kwargs={"with_users": True})
```
samuelhwilliams: ☝️

app/main/views/services.py
```diff
@@ -69,26 +77,18 @@ def list_services():
         if not supplier:
             abort(404, "supplier_id '%d' not found" % supplier_id)
 
-        items = services.default_order().filter(Service.supplier_id == supplier_id).all()
-        return jsonify(
-            services=[service.serialize() for service in items],
-            links=dict()
-        )
+        services = services.default_order().filter(Service.supplier_id == supplier_id)
+        return list_result_response("services", services)
```
samuelhwilliams: This should use `RESOURCE_NAME`.

app/main/views/suppliers.py
```diff
@@ -72,18 +73,14 @@ def list_suppliers():
     suppliers = suppliers.distinct(Supplier.name, Supplier.supplier_id)
 
     try:
-        suppliers = suppliers.paginate(
+        return paginated_result_response(
+            result_name="suppliers",
```
samuelhwilliams: Given this set of views is primarily about suppliers, I think we _could_ define the resource name as suppliers and use that here. Anything else could continue to use string literals (eg contact information).

Can take it or leave it with this one, but just a thought.

app/main/views/users.py
```diff
@@ -30,7 +44,7 @@ def auth_user():
         db.session.add(user)
         db.session.commit()
 
-        return jsonify(users=user.serialize()), 200
+        return single_result_response("users", user), 200
```
samuelhwilliams: Another one that should be `RESOURCE_NAME`.

app/main/views/users.py
```diff
@@ -58,10 +72,8 @@ def list_users():
         single_user = user_query.filter(
             User.email_address == email_address.lower()
         ).first_or_404()
-
         return jsonify(
-            users=[single_user.serialize()],
-            links={}
+            users=[single_user.serialize()]
```
samuelhwilliams: Is this one we've missed or are we not doing it for a reason, maybe compatibility with the old one that sends an empty links reference? Think it's probably fine to update.

app/utils.py
```diff
@@ -67,6 +68,44 @@ def pagination_links(pagination, endpoint, args):
...
+    in as a argument so we can manipulate the query object (i.e. to do the pagination). Results will be returned in a
+    list and use the results `serialize` method for presentation.
+    """
+    pagination = results_query.paginate(page=page, per_page=per_page)
+    meta = result_meta(pagination.total)
+    serialized_results = [
+        result.serialize(**(serialize_kwargs if serialize_kwargs else {})) for result in pagination.items
+    ]
+    links = pagination_links(pagination, endpoint, request_args)
+    return jsonify(meta=meta, links=links, **{result_name: serialized_results})
```
samuelhwilliams: Yeah this (the new functions in this file) is good. Definitely a step in the right direction, and when we get around to having the name of the resource singular/plural as appropriate, it could be quite easy to do that here by passing in the singular name and using (eg) `inflect.plural` in the list/paginated functions.

app/main/views/services.py
idavidmcdonald: Good spot

app/main/views/users.py
idavidmcdonald: Nice

app/main/views/users.py
idavidmcdonald: This one is a little strange. If it doesn't find a user then it will 404. Whereas all our other list routes would just return an empty list. It's kind of like a mix between a single resource and a list resource. That's why I left it. Does that seem reasonable?

app/main/views/suppliers.py
idavidmcdonald: I've done as you suggested

idavidmcdonald:
Think I've addressed all your comments @samuelhwilliams. Let me know what you think and if you are happy I'll squash it all down.

app/main/views/briefs.py
```diff
@@ -472,5 +483,4 @@ def list_brief_services(brief_id):
     )
 
     services = services.filter(Service.supplier_id == supplier.supplier_id)
-
-    return jsonify(services=[service.serialize() for service in services])
+    return list_result_response("services", services), 200
```
samuelhwilliams: We could do `from app.main.views.services import RESOURCE_NAME as SERVICE_RESOURCE_NAME` here?

samuelhwilliams: We could do this?

```
from app.main.views.services import RESOURCE_NAME as SERVICE_RESOURCE_NAME
return list_result_response(SERVICE_RESOURCE_NAME, services), 200
```

app/main/views/users.py
samuelhwilliams: Hmm. It's non-traditional, that's for sure, but I think it still qualifies as a `list_result_response`, and I think making it that (probably with a comment) would be better than leaving it.

PR #695, opened 2017-11-21
New admin-manager user role

TheDoubleK:
These are the minimal changes necessary to allow creation of users with "admin-manager" role.

I think it's OK to put the migration with model changes here as it's only adding a value to an enum.  But will split it if you, the reviewer, feels strongly it should be split.

There will be more changes to come later when we enforce only certain email domains can have this role, but I think it's worth getting this in on its own now as a starter for ten.

### DIFF
--- a/migrations/versions/1050_add_admin_manager_role.py
+++ b/migrations/versions/1050_add_admin_manager_role.py
@@ -1,0 +1,22 @@
+"""Add admin-manager role
+
+Revision ID: 1050
+Revises: 1040
+Create Date: 2017-11-21 11:20:00.000000
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '1050'
+down_revision = '1040'
+
+from alembic import op
+
+
+def upgrade():
+    op.execute("COMMIT")  # See: http://stackoverflow.com/a/30910417/15720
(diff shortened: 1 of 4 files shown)


migrations/versions/1050_add_admin_manager_role.py
```diff
@@ -0,0 +1,22 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '1050'
+down_revision = '1040'
+
+from alembic import op
+
+
+def upgrade():
+    op.execute("COMMIT")  # See: http://stackoverflow.com/a/30910417/15720
```
katstevens: I can't decide whether this `COMMIT` statement is a neat workaround, or something potentially dangerous, waiting to bite us... Previously in this scenario I've had to swap round two explicit lists of enum types, similar to the accepted answer on that stackoverflow link. It's long winded but it's clear what's going on. This may well be a better solution, but I don't have a good understanding of why committing the transaction allows the alter statement to 'work', or what else that commit might affect.

TheDoubleK: It seems to be recommended in other places on the web too, and I haven't seen any way to alter enums using SQLAlchemy ORM stuff.  We have it in an earlier migration already so if it's going to bite us then maybe we're already too late to avoid that!

The accepted answer there is horrible :/

PR #696, opened 2017-11-21
REFACTOR-digitalmarketplace-api/app/main/views/users.py

benvand:
* Standardise imports
* Move helper function to live in utils file

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -2,12 +2,13 @@ from datetime import datetime
 
 from dmapiclient.audit import AuditTypes
 from sqlalchemy.orm import lazyload
-from sqlalchemy.exc import IntegrityError, DataError
-from flask import jsonify, abort, request, current_app
+from sqlalchemy.exc import DataError, IntegrityError
+from flask import abort, current_app, jsonify, request
 
 from .. import main
 from ... import db, encryption
-from ...models import User, AuditEvent, Supplier, Framework, SupplierFramework, BuyerEmailDomain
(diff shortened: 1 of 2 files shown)


TheDoubleK:
Looks like you made Flake 8 sad :joy:

benvand:
@TheDoubleK Whaaaaa?? It's flake8ing the git refs 🤔

benvand:
@TheDoubleK https://github.com/alphagov/digitalmarketplace-api/pull/697

app/main/views/users.py
```diff
@@ -1,16 +1,28 @@
...
+from ...utils import (
+    get_json_from_request, json_has_required_keys, json_has_matching_id, pagination_links, get_valid_page_or_1,
+    validate_and_return_updater_request
+)
+from ...validation import (
+    buyer_email_address_has_approved_domain, validate_user_auth_json_or_400, validate_user_json_or_400
+)
+
+
+def check_supplier_role(role, supplier_id):
```
TheDoubleK: If we're moving this round then maybe `supplier_utils.py` would be a better home for it?

benvand:
Yeah good call, those utils files are a bit weird but there is another view helper in there. I think those files would make more sense in `app/main/views/utils/supplier_utils.py`. And then stuff like `app/service_utils.py::commit_and_archive_service` would live in `app/main/models/utils/service_utils.py`. Maybe that's overkill but feels like there's a logical split there. Not exactly the most urgent thing to do.

benvand:
No probs @TheDoubleK :) this was just a bugbear

PR #699, opened 2017-11-24
Bump pycodestyle to support Python 3 syntax

samuelhwilliams:
Remove flake8-putty dependency and install flake8-per-file-ignores, which allows us to bump flake8/pycodestyle, which allows us to use Python3 syntax like static typing and f-strings.


samuelhwilliams:
@katstevens That's the plan - I'll probably run through and do them all (or at least the easy ones :trollface:) after this has gone in.

PR #701, opened 2017-11-29
User model validates admin email domain

katstevens:
Trello: https://trello.com/c/D2QFAPod/158-api-should-only-allow-creation-of-admin-accounts-with-allowed-email-domains

We already have an API route to check that an admin user's email is valid. This PR adds this validation at the User model level.

- For DRYness (and consistency with the buyer email checks), I moved the shared config check to the `validation` helper functions.
- Some test data needed updating too.


app/models/main.py
```diff
@@ -771,9 +774,13 @@ def validate_email_address(self, key, value):
     @validates('role')
```
benvand: So in the method above we do the role check for buyers on email change too if we're changing an email address. do we want to do this for admins too? (Should we even be doing it for buyers?)

katstevens: Oops I didn't even see that one. I think checking both is a good idea - will fix.

PR #702, opened 2017-11-29
Add framework manager user role

katstevens:
Trello: https://trello.com/c/iUS9D0ll/163-add-new-framework-manager-role

Almost entirely copied from the recent branch to add the `admin-manager` role: https://github.com/alphagov/digitalmarketplace-api/pull/695

- DB Migration to allow the new role to be added
- Update JSON schema
- Add new role to User model
- Add a basic test for creating a user with the new role

### DIFF
--- a/migrations/versions/1060_add_framework_manager_role.py
+++ b/migrations/versions/1060_add_framework_manager_role.py
@@ -19,4 +19,5 @@ def upgrade():
 
 
 def downgrade():
-    raise NotImplemented("Cannot remove user role value")
+    # Cannot remove user role value
+    pass
--- a/migrations/versions/1060_add_framework_manager_role.py
+++ b/migrations/versions/1060_add_framework_manager_role.py
@@ -1,0 +1,22 @@
+"""Add admin-framework-manager role
+
+Revision ID: 1060
+Revises: 1050
+Create Date: 2017-11-29 15:20:00.000000
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '1060'
+down_revision = '1050'
+
+from alembic import op
+
+
+def upgrade():
+    op.execute("COMMIT")  # See: http://stackoverflow.com/a/30910417/15720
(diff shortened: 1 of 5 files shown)


migrations/versions/1060_add_framework_manager_role.py
```diff
@@ -0,0 +1,22 @@
...
+# revision identifiers, used by Alembic.
+revision = '1060'
+down_revision = '1050'
+
+from alembic import op
+
+
+def upgrade():
+    op.execute("COMMIT")  # See: http://stackoverflow.com/a/30910417/15720
+    op.execute("ALTER TYPE user_roles_enum ADD VALUE 'admin-framework-manager' AFTER 'admin-manager';")
```
TheDoubleK: After merging the last one I regretted not adding an `IF NOT EXISTS` to this.

I don't think we really need it because the migration really should only run once - that's the whole point of Alembic.  But see what you think.

katstevens: I'd rather keep it consistent with the previous one, to be honest. Plus the `NotImplementedError` on the `downgrade` means we can't go back and forth without an error anyway....

Erk! That says `NotImplemented` rather than `NotImplementedError`, as does the previous migration. I missed that before. Did you try downgrading at all?

TheDoubleK: Damn, not sure I did TBH. And thinking about it this should maybe even just skip rather than raise - it's not really an error that should prevent anything earlier being rolled back - it's just something that can't be undone.

TheDoubleK: Damn my copy-pasting from ages ago :disappointed:

katstevens: OK - I'm going to update this migration (and the last one, if that's OK with you) to `pass` instead.

katstevens: And of course to test the upgrade/downgrade I had to add in the `IF NOT EXISTS` after all 😸

PR #706, opened 2017-12-04
Restrict email domains for admin category/sourcing/managers

katstevens:
Trello: https://trello.com/c/D2QFAPod/158-api-should-only-allow-creation-of-admin-accounts-with-allowed-email-domains

Follow up PR to https://github.com/alphagov/digitalmarketplace-api/pull/701. Email restrictions are now applied at API level for `admin-ccs-sourcing`, `admin-ccs-category`, `admin-manager` and `admin-framework-manager` users.

### DIFF
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -724,14 +724,17 @@ SupplierFramework.current_framework_agreement = db.relationship(
 class User(db.Model):
     __tablename__ = 'users'
 
-    ROLES = [
-        'buyer',
-        'supplier',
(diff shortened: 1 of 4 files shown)


TheDoubleK:
"Prefer", not "order"  stupid phone, sorry.

katstevens:
I prefer precision, and not just because it makes my test parametrization easier 😸 Anyway, this allows us to be more flexible with our role naming - we could add `admin-foo` or `foo-admin` to `User.ADMIN_ROLES` then the validation would still pick it up.

PR #711, opened 2017-12-06
Start using external IDs in responses from the API

samuelhwilliams:
## Summary
Now that we've got a database properly populated with external IDs, we can start replying with them from the API. Adds a test to ensure both internal and external IDs are accepted.

### DIFF
--- a/tests/main/views/test_direct_award.py
+++ b/tests/main/views/test_direct_award.py
@@ -132,7 +132,7 @@ class TestDirectAwardListProjects(DirectAwardSetupAndTeardown):
 
         assert len(set(projects_seen)) == self.app.config['DM_API_PROJECTS_PAGE_SIZE'] * pages
 
-    def test_list_projects_orders_by_id(self):
+    def test_list_projects_orders_by_internal_id(self):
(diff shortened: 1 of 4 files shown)


tests/main/views/test_direct_award.py
```diff
@@ -375,18 +375,20 @@ def test_list_searches_returns_only_for_project_requested(self):
         self.create_direct_award_project(user_id=self.user_id + 1, project_id=self.project_id + 1)
         self.create_direct_award_project_search(created_by=self.user_id, project_id=self.project_id + 1)
 
-        res = self.client.get('/direct-award/projects/{}/searches'.format(self.project_id))
+        res = self.client.get('/direct-award/projects/{}/searches'.format(self.project_external_id))
         data = json.loads(res.get_data(as_text=True))
 
         with self.app.app_context():
-            assert all([search['projectId'] == self.project_id for search in data['searches']])
+            print(data['searches'])
```
idavidmcdonald: Leftover print statement

samuelhwilliams: Removed.

PR #712, opened 2017-12-06
Upgrade API client to v13.1.0

katstevens:
Addresses a vulnerability in the Python `requests` library.

The API should not be affected by the v13.0.0 breaking change - the `update_user` method is called with keyword args rather than positional args.


katstevens:
Nope, I grepped thoroughly - it's calling it as part of a script! But I don't know if we actually run that script anywhere.

PR #713, opened 2017-12-08
Bump api client version to 14.0.0

Wynndow:
It allows `make_iter_method` accept more than one model name. We need
this so we can upgrade the search-api to return `documents` instead of
just `services`.

The direct award `lock_project` view uses the search-api to fetch
service ids. We need to bump the api-client so this view doesn't break
when the search-api starts returning documents.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ SQLAlchemy==1.1.4
 SQLAlchemy-Utils==0.30.5
 
 git+https://github.com/alphagov/digitalmarketplace-utils.git@29.0.0#egg=digitalmarketplace-utils==29.0.0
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -60,6 +60,6 @@ s3transfer==0.1.12
 six==1.11.0
 unicodecsv==0.14.1
 urllib3==1.22
-Werkzeug==0.12.2
+Werkzeug==0.13
```
risicle: You'll enjoy this @galund - this got bumped by the pip freezing script. Seems our version of werkzeug is being dictated to us by our version of the apiclient :D

risicle:
Cool I'll verify my functional tests pass against this.

PR #715, opened 2017-12-12
Migration to correct external IDs

samuelhwilliams:
## Summary
After deploying code that starts correctly storing external IDs in audit events for direct award projects, we need to fix up existing audit events that have stored the internal ID under the 'projectExternalId' key.

Sits on top of the commit in https://github.com/alphagov/digitalmarketplace-api/pull/714 which needs to go out as a separate release before this migration.

## Ticket

### DIFF
--- a/migrations/versions/1090_fix_external_ids.py
+++ b/migrations/versions/1090_fix_external_ids.py
@@ -1,0 +1,46 @@
+"""Fix external IDs that were populated into 'create_project' audit events incorrectly
+
+Revision ID: 1090
+Revises: 1080
+Create Date: 2017-12-12 09:10:00.00000
+
+"""
+from alembic import op
+import json
+import sqlalchemy as sa
+from sqlalchemy.sql import text
+
+# revision identifiers, used by Alembic.
+revision = '1090'
+down_revision = '1080'
+
+
+def upgrade():
+    # Grab existing external IDs and populate the translation table
+    project_id_internal_to_external = {}
+    project_prefill_connection = op.get_bind()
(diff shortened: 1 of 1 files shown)


migrations/versions/1090_fix_external_ids.py
```diff
@@ -0,0 +1,47 @@
+"""Fix external IDs that were populated into audit events incorrectly
```
idavidmcdonald: Real small but might be worth putting in the top line that is audit events regarding direct award projects

idavidmcdonald: Sorry, in particular `create_project` audit events

migrations/versions/1090_fix_external_ids.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # For audit events that have stored an internal project ID (represented by IDs less than 1 million), update the data
+    # using the translation table above to convert it to its external ID.
+    audit_connection = op.get_bind()
+    for audit_id, audit_data, project_id in audit_connection.execute(
+            "SELECT id, data, object_id "
+            "FROM audit_events "
+            "WHERE type='create_project' AND (data->>'projectExternalId')::bigint < 1000000"
+    ):
+        if project_id:
```
idavidmcdonald: Would we ever expect there to be no project_id?

samuelhwilliams: No.

migrations/versions/1090_fix_external_ids.py
samuelhwilliams: Updated.

PR #718, opened 2017-12-19
Firebreak remove nose style assertions and replace with pytest

benvand:
Replace nose style assertions with pytest style assertions
Remove nose from requirements


benvand:
Thanks for the review @CrystalPea !

PR #719, opened 2017-12-19
Firebreak give test requests isolated context

benvand:
This PR gives our test requests their own isolated `app_context` to attach db operations to.

The test changes reflect the fact that it is no longer possible, in a test, to:

Create an object in the DB
POST to a view to change that object
Then check the result of that POST against the original object without reloading it from the db

### DIFF
--- a/tests/bases.py
+++ b/tests/bases.py
@@ -1,5 +1,34 @@
+from flask.testing import FlaskClient
+
 from app import create_app, db
 from app.models import Framework, FrameworkLot
+
+
+class TestClient(FlaskClient):
+    """This is a custom Test Client for handling the creation of a dedicated application context on a per request basis.
+
+    Flask-SQLAlchemy attaches its db operations to the top application context on the context stack.
+    Requests use the top application context on the context stack or create a new one if none exists.
+
+    Normally this isn't an issue. Each new request in production will use its own thread and
+    on finding that there is no existing application context will create a new one.
+
+    In tests however we require an application context to create/ update the database with the data required for the
+    test. We can then end up using this polluted application context in the view we're testing if we don't pop it. In
+    the open method of this class we create a fresh application context for the request/ view to use and remove it
+    after so it doesn't leak back to the test.
+    """
+
+    def open(self, *args, **kwargs):
+        app_context = self.application.app_context()
+        app_context.push()
+
+        res = super(TestClient, self).open(*args, **kwargs)
+
+        db.session.expire_all()
+        app_context.pop()
+
+        return res
 
 
 class WSGIApplicationWithEnvironment(object):
@@ -23,6 +52,7 @@ class BaseApplicationTest(object):
             self.app.wsgi_app,
             HTTP_AUTHORIZATION='Bearer {}'.format(self.app.config['DM_API_AUTH_TOKENS'])
         )
+        self.app.test_client_class = TestClient
         self.client = self.app.test_client()
 
     def teardown(self):
--- a/tests/main/views/test_brief_response.py
+++ b/tests/main/views/test_brief_response.py
@@ -605,17 +605,11 @@ class TestSubmitBriefResponse(BaseBriefResponseTest):
                 db.session.delete(existing_brief_response[-1])
                 db.session.commit()
 
-            # Get dos framework.
-            dos_framework = db.session.query(Framework).filter_by(slug='digital-outcomes-and-specialists').first()
-
-            # Make framework live so a brief response can be created.
-            dos_framework.status = 'live'
-            db.session.commit()
-
             # Create a brief response while the framework is live.
             self._setup_existing_brief_response()
 
             # Set framework status to the invalid status currently under test.
+            dos_framework = db.session.query(Framework).filter_by(slug='digital-outcomes-and-specialists').first()
             dos_framework.status = framework_status
             db.session.commit()
 
--- a/tests/main/views/test_briefs.py
+++ b/tests/main/views/test_briefs.py
@@ -1376,6 +1376,7 @@ class TestAwardPendingBriefResponse(FrameworkSetupAndTeardown):
             brief_response = BriefResponse(brief_id=1, supplier_id=0, submitted_at=datetime.utcnow(), data={})
             db.session.add(brief_response)
             db.session.commit()
+            brief_response_id = brief_response.id
 
             res = self._post_to_award_endpoint({'briefResponseId': brief_response.id})
             assert res.status_code == 200
@@ -1384,7 +1385,7 @@ class TestAwardPendingBriefResponse(FrameworkSetupAndTeardown):
             assert len(brief_response_audits) == 1
             assert brief_response_audits[0]['data'] == {
                 'briefId': 1,
-                'briefResponseId': brief_response.id,
+                'briefResponseId': brief_response_id,
                 'briefResponseAwardedValue': True
             }
 
@@ -1397,7 +1398,8 @@ class TestAwardPendingBriefResponse(FrameworkSetupAndTeardown):
             brief_response1.award_details = {'pending': True}
             db.session.add_all([brief_response1, brief_response2])
             db.session.commit()
-
+            brief_response1_id = brief_response1.id
+            brief_response2_id = brief_response2.id
             # Update to be awarded to brief response 2 rather than brief response 1
             res = self._post_to_award_endpoint({'briefResponseId': brief_response2.id})
             assert res.status_code == 200
@@ -1406,12 +1408,12 @@ class TestAwardPendingBriefResponse(FrameworkSetupAndTeardown):
             assert len(brief_response_audits) == 2
             assert brief_response_audits[0]['data'] == {
                 'briefId': 1,
-                'briefResponseId': brief_response1.id,
+                'briefResponseId': brief_response1_id,
                 'briefResponseAwardedValue': False
             }
             assert brief_response_audits[1]['data'] == {
                 'briefId': 1,
-                'briefResponseId': brief_response2.id,
+                'briefResponseId': brief_response2_id,
                 'briefResponseAwardedValue': True
             }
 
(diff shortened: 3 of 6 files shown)


tests/main/views/test_brief_response.py
```diff
@@ -605,19 +605,15 @@ def test_can_not_submit_a_brief_response_for_a_framework_that_is_not_live_or_exp
...
 
-            # Get dos framework.
-            dos_framework = db.session.query(Framework).filter_by(slug='digital-outcomes-and-specialists').first()
-
-            # Make framework live so a brief response can be created.
-            dos_framework.status = 'live'
-            db.session.commit()
-
+            # Ensure dos framework is live.
+            db.session.query(Framework).filter_by(slug='digital-outcomes-and-specialists').update({'status': 'live'})
```
galund: I don't understand why an 'update' is possible but a fetch-change-commit is not possible?

tests/main/views/test_briefs.py
```diff
@@ -1504,18 +1506,19 @@ def test_can_supply_award_details_for_closed_brief_with_awarded_brief_response(s
             db.session.add(brief_response)
             db.session.commit()
             assert brief_response.status == 'pending-awarded'
+            brief_response_id = brief_response.id
```
galund: I'm not sure I see the benefit of introducing a new variable?

tests/main/views/test_brief_response.py
benvand: So this line here:
https://github.com/alphagov/digitalmarketplace-api/pull/719/files#diff-62b9141c1c9c33809d1460a992905216L619
`dos_framework.status = framework_status`
Was the line operating on a stale object.

To successfully update that object I would have had to repeat the lines above, first getting, then setting the attr, then committing the action. 

Then I remembered 'there's a shortcut for that'.

So I used update. But then I thought I might as well standardise both operations.

galund: Ah hmmm but you're kinda doing the wrong thing then, aren't you: the intention of the code is presumably to _reset_ the entire object back how it _was_ - including the status. So using a 'stale' object was _correct_, no? 

Just updating the one field might not put things back exactly how they used to be?

tests/main/views/test_briefs.py
benvand: `brief_response.id` becomes stale once the request is made to the view. You can't access it any more.

tests/main/views/test_brief_response.py
benvand: That _may_ have been the intention :man_shrugging: 
I don;t think we can assume so because:
1. The `_setup_existing_brief_response` shouldn't _change_ the framework object
2. We're only updating the framework so we _can_ create the objects for this test therefore we only care about slug, status and maybe family.

Regardless of the intention:
1. SQLAlchemy doesn't commit the whole object on a single attribute change just the changed attributes with an `UPDATE  SET WHERE id =` command
2. That line isn't actually doing anything at all. The live_dos_framework fixture is being used here too. And SQLAlchemy won't perform a useless operation (ie issuing an update that doesn't change anything, updating status to 'live' when it's already 'live').
3. Also rather than updating the whole object back to how it was issuing an update on a stale object doesn't do _anything_

benvand: Removed the first one 👍

tests/bases.py
```diff
@@ -1,7 +1,23 @@
+from flask.testing import FlaskClient
+
 from app import create_app, db
 from app.models import Framework, FrameworkLot
 
 
+class TestClient(FlaskClient):
+
```
idavidmcdonald: Please can we inline document what this is doing, why it is doing and what a impact it has for a developer when writing tests

tests/main/views/test_briefs.py
idavidmcdonald: What makes the object become stale?
Is it the new context being used? Or is it within the new context, a change is committed to the DB?

idavidmcdonald:
Oh and you say this is a required step for us to remove app context from all our tests? Could you just give a quick run through of how that will be done please?

tests/main/views/test_briefs.py
benvand: > Is it the new context being used
Yeah, as soon as a new context is introduced the old one is stale

tests/bases.py
benvand: Will do

tests/main/views/test_brief_response.py
```diff
@@ -605,19 +605,13 @@ def test_can_not_submit_a_brief_response_for_a_framework_that_is_not_live_or_exp
                 db.session.delete(existing_brief_response[-1])
                 db.session.commit()
 
-            # Get dos framework.
-            dos_framework = db.session.query(Framework).filter_by(slug='digital-outcomes-and-specialists').first()
-
-            # Make framework live so a brief response can be created.
-            dos_framework.status = 'live'
-            db.session.commit()
```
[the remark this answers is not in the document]
benvand: Yep :)

tests/main/views/test_brief_response.py
```diff
@@ -605,19 +605,13 @@ def test_can_not_submit_a_brief_response_for_a_framework_that_is_not_live_or_exp
...
-
             # Create a brief response while the framework is live.
             self._setup_existing_brief_response()
 
             # Set framework status to the invalid status currently under test.
-            dos_framework.status = framework_status
-            db.session.commit()
+            db.session.query(Framework).filter_by(slug='digital-outcomes-and-specialists').update(
+                {'status': framework_status}
+            )
```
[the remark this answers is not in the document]
benvand: > I can't quite work out why we don't need to commit this?

Update has an implicit commit

> any knock on effects?

It _could_ but tests should take care of setting up the data they need in their own setup. So knock on effects would only effect poorly written tests. It looks like we get away with it here anyway seeing as tests pass.

tests/main/views/test_services.py
```diff
@@ -999,20 +1004,17 @@ def setup(self):
...
             self.setup_dummy_service(
                 service_id=self.service_id,
                 **self.service
             )
-            service = Service.query.filter(
-                Service.service_id == self.service_id
-            ).one()
             for status in valid_statuses:
-                index_service.reset_mock()
+                # index_service.reset_mock()
```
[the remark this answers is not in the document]
benvand: Will remove

tests/main/views/test_brief_response.py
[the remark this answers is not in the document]
idavidmcdonald: I was trying to find some docs about update having an implicit commit but had no luck. Do you know where I might find them?

tests/main/views/test_services.py
```diff
@@ -545,16 +545,21 @@ def test_invalid_json_causes_failure(self, index_service):
             assert b'Invalid JSON' in response.get_data()
             assert index_service.called is False
 
-    @mock.patch('app.main.views.services.index_service', autospec=True)
-    def test_can_post_a_valid_service_update(self, index_service):
+    @mock.patch('app.service_utils.index_object', autospec=True)
```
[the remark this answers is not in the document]
benvand: Sort of. 

When a mock object is called it attaches the arguments to itself in `call_args_list`. When we then perform `mock.assert_called_once_with(object)` it checks the object in the `call_args_list`.

The problem with this is that `index_service` is called within the view with a db model instance. This instance is appended to `call_args_list`. We then drop out of the view and try to compare a new object against the object in `call_args_list`. By this point the object in `call_args_list` is stale. 

The only way around this was to mock a method that doesn't accept a db model instance.

tests/main/views/test_services.py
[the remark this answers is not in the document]
benvand: Done 👍

tests/bases.py
benvand: Done 👍

tests/bases.py
```diff
@@ -1,7 +1,36 @@
...
+    Flask-SQLAlchemy attaches its db operations to the top application context on the context stack.
+    Requests use the top application context on the context stack or create a new one if none exists.
+
+    Normally this isn't an issue. Each new request in production will use its own thread and
+    on finding that there is no existing application context will create a new one.
+
+    In tests however we require an application context to create/ update the database with the data required for the
+    test. We can then end up using this polluted application context in the view we're testing if we don't pop it. In
+    the open method of this class we create a fresh application context for the request/ view to use and remove it
+    after so it doesn't leak back to the test.
```
idavidmcdonald: Nice description

tests/main/views/test_services.py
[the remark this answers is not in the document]
idavidmcdonald: Very interesting

tests/main/views/test_services.py
```diff
@@ -1022,14 +1023,21 @@ def test_should_update_service_with_valid_statuses(self, index_service):
...
-
+                service = Service.query.filter(
+                    Service.service_id == self.service_id
+                ).one()
                 assert response.status_code == 200
                 data = json.loads(response.get_data())
                 assert status == data['services']['status']
-                if status == 'disabled':
-                    assert index_service.called is False
+                if status in ('disabled', 'enabled'):
```
idavidmcdonald: How come we are adding the `enabled` case here?

benvand: Ah that method only only gets called on publish. 
https://github.com/alphagov/digitalmarketplace-api/blob/master/app/service_utils.py#L134

benvand: I think that test should probably be split out into 2 but that's a bit out of scope of this PR

benvand: I don't really like the logic in that view tbf.
It should probably be more like:
```
    if prior_status != status:

        # If it's being unpublished, delete it from the search api.
        if prior_status == 'published':
            delete_service_from_index(service)
        elif status == 'published':
            # If it's being published, index in the search api.
            index_service(service)
        else:
            pass
```

benvand: Explicit is better than implicit sort of thing

benvand: https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/services.py#L329

benvand: Saying that, it hasn't been changed for 2 and a half years, so... If it ain't broke.

tests/main/views/test_brief_response.py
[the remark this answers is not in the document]
benvand: In the Django documentation :man_facepalming:
Django ORMs update is applied instantly without saving the object. SQLAlchemys is not. I've changed this back
https://docs.djangoproject.com/en/1.9/topics/db/queries/#updating-multiple-objects-at-once

benvand:
Awesome. @idavidmcdonald I've updated that commit message to indicate why the changes are being made. Going to take this out to prod 👍

PR #720, opened 2017-12-21
Provide app context in setup

benvand:
Instead of using 
`with self.app.app_context():`
in every test we can provide the app context in the setup now the work to isolate requests from said context is done.

This will have the benefit of making objects created in the setup available in the tests before a request is made.

Requires https://github.com/alphagov/digitalmarketplace-api/pull/719


benvand:
Awesome, thanks @samuelhwilliams

PR #723, opened 2018-01-04
Remove _all_ extraneous calls to `app.app_context()` within tests

benvand:
Remove extraneous `with self.app.app_context()` calls  …
The app context is now provided by the tests.bases.BaseApplicationTest.setup.
This makes these calls, within tests, which previously allowed access to the db, unneeded.
See information here on the app context:
http://flask.pocoo.org/docs/0.12/appcontext/

And see the commit here:
cfcbdea
(introduces a app_context push and pop in the BaseApplicationTest.setup/teardown respectively)

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -51,38 +51,36 @@ def fixture_params(fixture_name, params):
 class FixtureMixin(object):
 
     def setup_default_buyer_domain(self):
-        with self.app.app_context():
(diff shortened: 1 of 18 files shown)


benvand:
@samuelhwilliams Legend 🏆 Thanks 🙌

samuelhwilliams:
Nice work Ben. I really hope this wasn't done manually. 👍

PR #724, opened 2018-01-04
Remove requirement for app_context to be passed to method.

benvand:
No longer required as context is pushed in setup and popped in teardown of BaseApplicationTest


benvand:
Haha 😅  Thanks!

PR #725, opened 2018-01-08
Allow IntegrityErrors to be serialized

idavidmcdonald:
https://trello.com/c/vzTF64Q7/160-sqlalchemy-integrityerrors-in-the-api-cant-be-serialized-so-the-actual-error-message-is-hidden

We had seen 500s in the app if an IntegrityError, `e`, was caught and
we passed `e.orig` into the `abort` function. This is because
`e.orig` refers to the original DB-API IntegrityError that is
raised by the database adapater (psycopg2 in our case). The pyscopg2
IntegrityError is not JSON serializable and we ran into problems.

We can instead use `format` to allow our exception to be
displayed as a string. This is a pattern we are already using in
the many other catching of IntegrityErrors across the app.

### DIFF
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 
-from flask import jsonify, abort
+from flask import abort
 from sqlalchemy.exc import IntegrityError
 from dmapiclient.audit import AuditTypes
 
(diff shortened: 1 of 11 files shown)


idavidmcdonald:
Happy to do that.

I plan on replacing everything with
`abort(400, format(e))`

Sound good?

samuelhwilliams:
Yep, I think that's fine by me. Worth double-checking nothing relies on getting json responses back from the endpoints that did return that. But I doubt they do.

idavidmcdonald:
Ah I didn't realise that just using `abort` won't return you a json response.

It looks like you either need not use abort and `return jsonify(message='error'), 400` or `abort(make_response(jsonify(message="Message goes here"), 400))`.

A quick grep shows about 175 uses off `abort` in our codebase so I assume we aren't currently feeling too tied to it being JSON (or like me people didn't realise).

I will go ahead with using abort and will see if we have any unit tests relying on it (I think it's fair to assume our FE wouldn't be, right?).

PR #728, opened 2018-01-10
Load schemas for G10 and DOS3 service validation

TheDoubleK:


app/validation.py
```diff
@@ -13,69 +14,14 @@
...
 FORMAT_CHECKER = FormatChecker()
 
 
-def load_schemas(schemas_path, schema_names):
+def load_schemas(schema_paths):
     loaded_schemas = {}
-    for schema_name in schema_names:
-        schema_path = os.path.join(schemas_path, '{}.json'.format(schema_name))
+    for schema_path in schema_paths:
+        schema_name = os.path.splitext(os.path.basename(schema_path))[0]
```
TheDoubleK: This will work as long as we don't end up with `some.idiot-did-this.json` in the folder?

samuelhwilliams: Nope, `os.path.splitext` splits would split that into two parts: `some.idiot-did-this`, and `.json`, so it would still be fine.

risicle: Could do just a simple assertion that the `schema_name` isn't already in `loaded schemas` before you stick it in there...

risicle: ok well thats fine then

PR #730, opened 2018-01-15
Pre-fetch awarded_brief_response relationship for /brief-responses

katstevens:
https://trello.com/c/SptNj93b/229-slow-requests-on-brief-responses-endpoint

The `/brief-responses` endpoint was very slow, triggering an alert on our metrics during the data export job.

Adding the prefetch for the award relationship improves the endpoint response time by a factor of 10! Bonus: the tests also run about ~20% faster.

Not sure if this will affect the memory usage @idavidmcdonald (just looking at the old PR you linked to on the Trello ticket)?

### DIFF
--- a/app/main/views/brief_responses.py
+++ b/app/main/views/brief_responses.py
@@ -226,6 +226,7 @@ def list_brief_responses():
     brief_responses = brief_responses.options(
         db.defaultload(BriefResponse.brief).defaultload(Brief.framework).lazyload("*"),
         db.defaultload(BriefResponse.brief).defaultload(Brief.lot).lazyload("*"),
+        db.defaultload(BriefResponse.brief).defaultload(Brief.awarded_brief_response).lazyload("*"),
         db.defaultload(BriefResponse.supplier).lazyload("*"),
     )
 


samuelhwilliams:
I wonder if we should use `.raiseload('*')` instead of `lazyload` 🤔 

It would make the endpoint raise an error on attribute access rather than slow down (by emitting more db queries). If we never _expect_ awarded_brief_response to be used by that endpoint, it would alert us sooner to a problem rather than just suddenly degrade the response speed (which would be harder to catch in development/review). If we go with raiseload, probably need to comment the line to explain as I don't think we use it elesewhere.

If the endpoint does use awarded_brief_response, then `lazyload` is probably the right thing to do. But I can't see anywhere it gets accessed. What do you think?

katstevens:
We need it to evaluate the Brief status (included in the serialization).

katstevens:
I tried out `raiseload()` and it still worked fine - I think because we don't use any attributes on the object loaded by the `Brief.awarded_brief_response` relationship (so for example, using `raiseload(Brief.awarded_brief_response)` earlier in the chain does raise the error). But there's no noticeable performance improvement using `raiseload('*')` over `lazyload('*')`.

samuelhwilliams:
Ah, makes sense. In which case this is perfect. Great performance boost, that's for sure. 👍 

I wonder if we could test our endpoints in some automated way to check they all respond (with a variety of arguments) within a reasonable time. 🤔

PR #734, opened 2018-01-31
Copy supplier account info to declaration

samuelhwilliams:
## Summary
Paired PR with @CrystalPea.

Now that we no longer ask supplier account questions as part of a framework application's declaration, we need to patch supplier declarations with the information we already have. When a framework is set to `pending` (i.e. applications close), we will inject the relevant data we hold in the Supplier/ContactInformation tables into the SupplierFramework.declaration for every supplier who is applying to get onto the framework. We will do this even if the supplier's application is incomplete. Where this is the case, any empty fields will be stored under their corresponding key in the JSON blob as nulls/Nones.

We hardcode a number of keys to inject into the `SupplierFramework.declaration` JSON blob. These aren't currently needed by any of the frontends, so we won't centralise these keys yet.

After much effort and many attempts (thanks @katstevens & @benvand for your time), I couldn't coerce SQLAlchemy into running the query we need through the ORM mechanism, so I've resorted to using a raw SQL statement. Given that our G-Cloud framework has thousands of suppliers, the statement needs to be executed fully on the DB-side as the performance hit from shunting data back and forth between the DB and the API probably would not be acceptable.

## Ticket
https://trello.com/c/lUQOcuSS/16-api-populate-declaration-with-supplier-data-when-framework-becomes-pending

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -1,13 +1,23 @@
 import datetime
 
 from flask import jsonify, abort, request
+from sqlalchemy import func, orm, case, text
 from sqlalchemy.exc import IntegrityError, DataError
-from sqlalchemy import func, orm, case
 from dmapiclient.audit import AuditTypes
 from dmutils.config import convert_to_boolean
 
 from .. import main
-from ...models import db, Framework, DraftService, User, Supplier, SupplierFramework, AuditEvent, Lot
+from ...models import (
+    AuditEvent,
+    db,
+    DraftService,
+    Framework,
+    Lot,
+    Supplier,
+    SupplierFramework,
+    User,
+)
+from ... import supplier_constants
 from ...utils import (
     get_json_from_request,
     json_has_required_keys,
@@ -108,6 +118,55 @@ def update_framework(framework_slug):
             json_payload['frameworks']['frameworkAgreementDetails'],
             enforce_required=True
         )
+
+    if json_payload['frameworks'].get('status', None) == 'pending' and framework.status != 'pending':
+        # When we set a framework to pending, we want to polyfill all supplier declarations (regardless of whether
+        # their application is 'complete' or not) with information from their supplier account. We have a set of
+        # hardcoded keys for these values. Currently these keys don't need to mean anything outside of the API, so
+        # they are defined in dm-api/app/supplier_constants. At the time this was written, it didn't seem feasible to
+        # manage this query through SQLAlchemy in an efficient (i.e. 100% server-side) manner, so we are using raw SQL.
+
+        # We pump this data into the declaration before updating the framework state, so that if for some reason the
+        # framework state fails to update, when a second attempt is made the data is overwritten. We deal with the
+        # theoretical scenario where a supplier has multiple ContactInformation entries by taking the first associated
+        # entry and using the details from that. If we add more advanced logic in the future, this will need updating.
+        statement = text(f"""
+UPDATE
+    supplier_frameworks
+SET
+    declaration = supplier_frameworks.declaration::jsonb || (
+        SELECT
+            row_to_json(data)
+        FROM
+        (
+            SELECT
+                suppliers.name AS "{supplier_constants.KEY_TRADING_NAME}",
+                suppliers.registered_name AS "{supplier_constants.KEY_REGISTERED_NAME}",
+                COALESCE(suppliers.companies_house_number,
+                         suppliers.other_company_registration_number) AS "{supplier_constants.KEY_REGISTRATION_NUMBER}",
+                suppliers.registration_date AS "{supplier_constants.KEY_REGISTRATION_DATE}",
+                suppliers.duns_number AS "{supplier_constants.KEY_DUNS_NUMBER}",
+                suppliers.vat_number AS "{supplier_constants.KEY_VAT_NUMBER}",
+                suppliers.trading_status AS "{supplier_constants.KEY_TRADING_STATUS}",
+                suppliers.organisation_size AS "{supplier_constants.KEY_ORGANISATION_SIZE}",
+                suppliers.registration_country AS "{supplier_constants.KEY_REGISTRATION_COUNTRY}",
+                contact_information.address1 AS "{supplier_constants.KEY_REGISTRATION_BUILDING}",
+                contact_information.city AS "{supplier_constants.KEY_REGISTRATION_TOWN}",
+                contact_information.postcode AS "{supplier_constants.KEY_REGISTRATION_POSTCODE}"
+            FROM
+                suppliers
+            JOIN
+                contact_information ON suppliers.supplier_id = contact_information.supplier_id
+            WHERE
+                suppliers.supplier_id = supplier_frameworks.supplier_id
+            ORDER BY contact_information.id ASC
+            LIMIT 1
+        ) AS data
+    )::jsonb
+WHERE
+    supplier_frameworks.framework_id = :framework_id""")
+
+        db.session.execute(statement, {"framework_id": framework.id})
 
     for whitelisted_key, value in attribute_whitelist.items():
         if whitelisted_key in json_payload['frameworks']:
--- a/app/supplier_constants.py
+++ b/app/supplier_constants.py
@@ -1,0 +1,15 @@
+# Here we define a set of hardcoded keys that we use when denormalizing data from Supplier/ContactInformation tables
+# into the SupplierFramework.declaration field. These are only used internally by the API - the frontends currently
+# do not need knowledge of them, so there is no reason to centralise them e.g. in utils at this time.
+KEY_DUNS_NUMBER = 'supplierDunsNumber'
+KEY_ORGANISATION_SIZE = 'supplierOrganisationSize'
+KEY_REGISTERED_NAME = 'supplierRegisteredName'
+KEY_REGISTRATION_BUILDING = 'supplierRegisteredBuilding'
+KEY_REGISTRATION_COUNTRY = 'supplierRegisteredCountry'
+KEY_REGISTRATION_DATE = 'supplierRegistrationDate'
+KEY_REGISTRATION_NUMBER = 'supplierCompanyRegistrationNumber'
+KEY_REGISTRATION_POSTCODE = 'supplierRegisteredPostcode'
+KEY_REGISTRATION_TOWN = 'supplierRegisteredTown'
+KEY_TRADING_NAME = 'supplierTradingName'
+KEY_TRADING_STATUS = 'supplierTradingStatus'
+KEY_VAT_NUMBER = 'supplierVatNumber'
--- a/tests/main/views/test_frameworks.py
+++ b/tests/main/views/test_frameworks.py
@@ -3,11 +3,15 @@ import datetime
 from flask import json
 from freezegun import freeze_time
 import mock
+import pytest
 from sqlalchemy.exc import IntegrityError
 
 from tests.bases import BaseApplicationTest, JSONUpdateTestMixin
 from app.models import db, Framework, SupplierFramework, DraftService, User, FrameworkLot
+from app.models.main import ContactInformation, Supplier
 from tests.helpers import FixtureMixin
+
+from app import supplier_constants
 
 
 class TestListFrameworks(BaseApplicationTest):
@@ -185,7 +189,7 @@ class TestGetFramework(BaseApplicationTest):
         assert response.status_code == 404
 
 
-class TestUpdateFramework(BaseApplicationTest, JSONUpdateTestMixin):
+class TestUpdateFramework(BaseApplicationTest, JSONUpdateTestMixin, FixtureMixin):
     endpoint = '/frameworks/example'
     method = 'post'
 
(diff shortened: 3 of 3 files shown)


app/constants.py
```diff
@@ -0,0 +1,15 @@
...
+KEY_REGISTERED_NAME = 'supplierRegisteredName'
+KEY_REGISTRATION_BUILDING = 'supplierRegisteredBuilding'
+KEY_REGISTRATION_COUNTRY = 'supplierRegisteredCountry'
+KEY_REGISTRATION_DATE = 'supplierRegistrationDate'
+KEY_REGISTRATION_NUMBER = 'supplierCompanyRegistrationNumber'
+KEY_REGISTRATION_POSTCODE = 'supplierRegisteredPostcode'
+KEY_REGISTRATION_TOWN = 'supplierRegisteredTown'
+KEY_TRADING_NAME = 'supplierTradingName'
+KEY_TRADING_STATUS = 'supplierTradingStatus'
+KEY_VAT_NUMBER = 'supplierVatNumber'
```
katstevens: I can foresee this file getting misused/messy/enormous in the future. Can we organise it by model somehow - `supplier_constants.py` maybe?

samuelhwilliams: Seems like a reasonable idea! 🤔

tests/main/views/test_frameworks.py
```diff
@@ -378,6 +395,158 @@ def test_update_framework_catches_db_errors(self, db_commit, open_example_framew
...
+        KEY_REGISTRATION_BUILDING: "1 Sweater Lane",
+        KEY_REGISTRATION_TOWN: "Milton Keynes",
+        KEY_REGISTRATION_POSTCODE: "SW3 8T",
+        KEY_REGISTRATION_NUMBER: "000123000",
+    }
+
+    def setup(self):
+        super(TestUpdateFrameworkPending, self).setup()
+
+        self.framework = Framework(slug='g-cloud-10', name='G-Cloud 10', framework='g-cloud', status='open',
```
benvand: IIRC: If you give this framework and `id` of over 100 it will be implicitly handled by the teardown in `BaseApplicationTest` and you can get rid of the teardown below.

app/supplier_constants.py
```diff
@@ -0,0 +1,15 @@
+# Here we define a set of hardcoded keys that we use when denormalizing data from Supplier/ContactInformation tables
```
benvand: I'd be tempted to get rid of this :man_shrugging: 

It's not a pattern we're using elsewhere.
We only us them in the application code once (In the SQL)
They should probably be hard coded in the tests so that people reading the tests can see exactly what the result looks like without having to chase the variables about. Verbose tests are not a bad thing :man_shrugging:

tests/main/views/test_frameworks.py
```diff
@@ -378,6 +395,158 @@ def test_update_framework_catches_db_errors(self, db_commit, open_example_framew
...
+        if framework:
+            FrameworkLot.query.filter(FrameworkLot.framework_id == framework.id).delete()
+            Framework.query.filter(Framework.id == framework.id).delete()
+            db.session.commit()
+
+        super(TestUpdateFrameworkPending, self).teardown()
+
+    def _assert_declaration_matches(self, expected_declaration_data):
+        supplier_framework = SupplierFramework.query.filter(SupplierFramework.supplier_id == 1).first()
+        assert supplier_framework.declaration == expected_declaration_data
```
katstevens: I'm not very familiar with how the ContactInformation relates to the SupplierFramework model - do we need to query and assert the ContactInformation separately or does this cover it?

tests/main/views/test_frameworks.py
```diff
@@ -378,6 +395,158 @@ def test_update_framework_catches_db_errors(self, db_commit, open_example_framew
...
+
+        self._assert_declaration_matches(self.EXPECTED_FRAMEWORK_DECLARATION_SUPPLIER_INFO)
+
+    def test_updating_framework_includes_keys_with_none_value_for_null_supplier_account_information(self):
+        supplier = Supplier.query.filter(Supplier.supplier_id == 1).first()
+        supplier.registration_date = None
+        supplier.vat_number = None
+        supplier.registered_name = None
+
+        ci = ContactInformation.query.filter(ContactInformation.supplier_id == supplier.supplier_id).first()
```
CrystalPea: would it be ok to change `ci` to `contact_information`? Just to increase readability.

app/main/views/frameworks.py
```diff
@@ -1,13 +1,36 @@
...
 from flask import jsonify, abort, request
+from sqlalchemy import func, orm, case, text
 from sqlalchemy.exc import IntegrityError, DataError
-from sqlalchemy import func, orm, case
 from dmapiclient.audit import AuditTypes
 from dmutils.config import convert_to_boolean
 
 from .. import main
-from ...models import db, Framework, DraftService, User, Supplier, SupplierFramework, AuditEvent, Lot
+from ...constants import (
```
CrystalPea: is this still `constants`, or `supplier_constants` now?

tests/main/views/test_frameworks.py
```diff
@@ -3,12 +3,29 @@
...
 import mock
+import pytest
 from sqlalchemy.exc import IntegrityError
 
 from tests.bases import BaseApplicationTest, JSONUpdateTestMixin
 from app.models import db, Framework, SupplierFramework, DraftService, User, FrameworkLot
+from app.models.main import ContactInformation, Supplier
 from tests.helpers import FixtureMixin
 
+from app.constants import (
```
CrystalPea: same here, is this still `constants`, or `supplier_constants` now?

CrystalPea: Also, why do we have to import these keys if we define them at class level anyway? Am I missing something? :)

app/supplier_constants.py
katstevens: It's a good point about the tests, however I think keeping the hardcoded keys out of the SQL statement is reasonable. (If we do keep this as a separate file remember to update the imports elsewhere!)

app/main/views/frameworks.py
```diff
@@ -109,6 +132,53 @@ def update_framework(framework_slug):
             enforce_required=True
         )
 
+    if json_payload['frameworks'].get('status', None) == 'pending' and framework.status != 'pending':
+        # When we set a framework to pending, we want to polyfill all supplier declarations (regardless of whether
+        # their application is 'complete' or not) with information from their supplier account. We have a set of
+        # hardcoded keys for these values. Currently these keys don't need to mean anything outside of the API, so
+        # they are defined in dm-api/app/constants. At the time this was written, it didn't seem feasible to manage
```
CrystalPea: same here re: `constants` name :)

app/main/views/frameworks.py
```diff
@@ -109,6 +132,53 @@ def update_framework(framework_slug):
...
+            JOIN
+                contact_information AS ci ON s.supplier_id = ci.supplier_id
+            WHERE
+                s.supplier_id = sf.supplier_id
+            ORDER BY ci.id ASC
+            LIMIT 1
+        ) AS data
+    )::jsonb
+WHERE
+    sf.framework_id = :framework_id""")
```
katstevens: I would ❤️ love ❤️  for this big old statement (and the excellent comment above it) to be pulled into a helper function for readability. 

It's a real shame there's no ORM equivalent (firebreak project anyone?) of the JSON update, but given the expensiveness of the operation, raw SQL seems the sensible way to go.

app/main/views/frameworks.py
```diff
@@ -109,6 +132,53 @@ def update_framework(framework_slug):
...
+                s.duns_number AS "{KEY_DUNS_NUMBER}",
+                s.vat_number AS "{KEY_VAT_NUMBER}",
+                s.trading_status AS "{KEY_TRADING_STATUS}",
+                s.organisation_size AS "{KEY_ORGANISATION_SIZE}",
+                s.registration_country AS "{KEY_REGISTRATION_COUNTRY}",
+                ci.address1 AS "{KEY_REGISTRATION_BUILDING}",
+                ci.city AS "{KEY_REGISTRATION_TOWN}",
+                ci.postcode AS "{KEY_REGISTRATION_POSTCODE}"
+            FROM
+                suppliers AS s
```
CrystalPea: Is there a reason for abbreviating stuff in this query?

app/main/views/frameworks.py
samuelhwilliams: I'm going to keep trying to work out how this could be ORM-ified. If I still haven't done it by next firebreak, then yeah, I'll use that time to have another serious crack at it.

app/main/views/frameworks.py
samuelhwilliams: Not really, it's just a shorthand I've become accustomed to when writing SQL. I could get rid of it if you find it hard to follow?

tests/main/views/test_frameworks.py
samuelhwilliams: Magic! Seems to work. Will change.

app/supplier_constants.py
samuelhwilliams: Yeah my reasoning is as Kat says here. I think it's good to keep them separate from the statement itself. I also don't think it's that valuable to require devs to update the test code when we change the keys in the core application code. The value of the variable doesn't really matter - devs just need to know that it's a "magic" string we're using as names.

tests/main/views/test_frameworks.py
samuelhwilliams: This should cover it. SupplierFramework pulls in three fields from ContactInformation (address1, city, postcode), so when we assert a match on the declaration we're also asserting the appropriate contact information has been pulled in.

tests/main/views/test_frameworks.py
samuelhwilliams: Sure 👍

app/main/views/frameworks.py
samuelhwilliams: Good spot. That's why tests are failing. :man_facepalming:

tests/main/views/test_frameworks.py
samuelhwilliams: We don't need to import them explicitly, however if we just `from app.supplier_constants` then we'd need to reference them as `supplier_constants.KEY_X`, which is a bit verbose.

CrystalPea: I think you are right 👍

TheDoubleK:
Should probably squash the "WIP" commit.

app/supplier_constants.py
TheDoubleK: I think it's good to have these "magic keys" clearly listed outside of the SQL, even if they are only used here in this one place at the moment.  

As to whether or not we should be using them in tests... I'm kind-of undecided.  There's always an element of marking your own homework if tests refer back to things in the main codebase, and best practice says don't do it.  But I also see Sam's point about the values not actually mattering in terms of code correctness.  The only thing you're going to detect is a typo, but maybe that's still useful... If we end up with a load of declarations with a `suplpierRegisterdCounty`instead of `supplierRegisteredCountry` it could matter.

tests/main/views/test_frameworks.py
```diff
@@ -378,6 +395,152 @@ def test_update_framework_catches_db_errors(self, db_commit, open_example_framew
...
+        )
+
+        expected_declaration_data = self.EXPECTED_FRAMEWORK_DECLARATION_SUPPLIER_INFO.copy()
+        expected_declaration_data[KEY_REGISTRATION_DATE] = None
+        expected_declaration_data[KEY_VAT_NUMBER] = None
+        expected_declaration_data[KEY_REGISTERED_NAME] = None
+        expected_declaration_data[KEY_REGISTRATION_TOWN] = None
+
+        self._assert_declaration_matches(expected_declaration_data)
+
```
TheDoubleK: It might be nice to have one further test, where we see a  declaration with pre-existing values for some of the "magic" fields have these values overwritten when the framework status changes to pending again, and all the other "non-magic" fields remain unchanged.

There's no test here with existing declaration data - it would definitely be good to have at least one test where we check that we don't inadvertently wipe all the existing declaration data when we do the copy over.

app/main/views/frameworks.py
```diff
@@ -1,13 +1,36 @@
...
 from flask import jsonify, abort, request
+from sqlalchemy import func, orm, case, text
 from sqlalchemy.exc import IntegrityError, DataError
-from sqlalchemy import func, orm, case
 from dmapiclient.audit import AuditTypes
 from dmutils.config import convert_to_boolean
 
 from .. import main
-from ...models import db, Framework, DraftService, User, Supplier, SupplierFramework, AuditEvent, Lot
+from ...supplier_constants import (
```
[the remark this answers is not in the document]
samuelhwilliams: Done.

app/main/views/frameworks.py
```diff
@@ -109,6 +132,53 @@ def update_framework(framework_slug):
...
+                contact_information.address1 AS "{KEY_REGISTRATION_BUILDING}",
+                contact_information.city AS "{KEY_REGISTRATION_TOWN}",
+                contact_information.postcode AS "{KEY_REGISTRATION_POSTCODE}"
+            FROM
+                suppliers
+            JOIN
+                contact_information ON suppliers.supplier_id = contact_information.supplier_id
+            WHERE
+                suppliers.supplier_id = supplier_frameworks.supplier_id
+            ORDER BY contact_information.id ASC
```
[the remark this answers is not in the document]
samuelhwilliams: Sure.

tests/main/views/test_frameworks.py
samuelhwilliams: Done. 👍

TheDoubleK:
The WIP commit is still there...

samuelhwilliams:
Nah it isn't. :trollface:

tests/main/views/test_frameworks.py
```diff
@@ -378,6 +382,151 @@ def test_update_framework_catches_db_errors(self, db_commit, open_example_framew
...
+        self._assert_declaration_matches(self.EXPECTED_FRAMEWORK_DECLARATION_SUPPLIER_INFO)
+
+    def test_updating_framework_includes_keys_with_none_value_for_null_supplier_account_information(self):
+        self.supplier.registration_date = None
+        self.supplier.vat_number = None
+        self.supplier.registered_name = None
+        self.contact_info.city = None
+
+        # db.session.add(self.supplier)
```
TheDoubleK: Commented code can go?

samuelhwilliams: Yep.

app/main/views/frameworks.py
```diff
@@ -109,6 +119,55 @@ def update_framework(framework_slug):
             enforce_required=True
         )
 
+    if json_payload['frameworks'].get('status', None) == 'pending' and framework.status != 'pending':
```
TheDoubleK: Minor thing, but I think `.get()` will return `None` anyway, so no need to set it as a default return value here.

samuelhwilliams: You're right, I just preferred to be explicit. It is a tad redundant though :man_shrugging:

TheDoubleK: Yeah, fine either way.

PR #735, opened 2018-02-07
Filter brief responses by awarded_at date

brenetic:
The framework family is used when building the url in email alerts to suppliers. We need to surface this when notifying them of briefs being awarded. `awarded_at` can now be passed into the api as a date in the form of `YYY-MM-DD`.

We also now nest framework related fields together within the serialised payload. This is unfortunately not consistent with how the brief serialisation works. It would be good to make this consistent in the _near_ future

Added `has_datetime_field_between` query in order to be able to filter within a date range. This should potentially be pulled out into a 😱helper 😱. Thoughts?

https://trello.com/c/LsEbL3jP

Paired with 😺@katstevens 😺


app/main/views/brief_responses.py
```diff
@@ -223,6 +224,13 @@ def list_brief_responses():
     if brief_id is not None:
         brief_responses = brief_responses.filter(BriefResponse.brief_id == brief_id)
 
+    if awarded_at is not None:
+        day_start = datetime.strptime(awarded_at, "%Y-%m-%d")
```
risicle: Do we not already have a standard date-argument-parsing function in the api? (the answer may well be no)

app/models/main.py
```diff
@@ -1653,13 +1656,37 @@ def validate(self, enforce_required=True, required_fields=None, max_day_rate=Non
...
+                        getattr(BriefResponse, attr) <= str(end_datetime)
+                    )
+                )
+            return self.filter(
+                sql_and(
+                    getattr(BriefResponse, attr) > str(start_datetime),
+                    getattr(BriefResponse, attr) < str(end_datetime)
+                )
+            )
+
```
risicle: This feels really unnecessary and special-casey to me. I mean - it's just a range test, two comparisons - why make the way we handle range tests be special an different for dates on `BriefResponse`s?

app/models/main.py
```diff
@@ -1653,13 +1656,37 @@ def validate(self, enforce_required=True, required_fields=None, max_day_rate=Non
...
+        def has_datetime_field_between(self, attr, start_datetime, end_datetime, inclusive=None):
+            """
+            Date filter values should be datetime objects. e.g.
+                BriefResponse.query.has_date_field_between(
+                    'published_at', datetime(2017, 1, 1), datetime(2017, 1, 2, 23, 59, 59, 999999), inclusive=True
+                )
+            would return brief responses published between 2017-01-01T00:00:00.000000Z and 2017-01-02T23:59:59.999999Z.
+            """
+            if not (isinstance(start_datetime, datetime) and isinstance(end_datetime, datetime)):
+                raise ValueError('Datetime object required')
```
risicle: We often try not to do special type-testing in python - and in this particular case I think it's a test that a) would be the responsibility of SQLAlchemy and b) I suspect SQLAlchemy does it anyway.

app/main/views/brief_responses.py
brenetic: Quite right! Have updated 👍

app/models/main.py
risicle: and c) it would perhaps unnecessarily prevent us from being able to compare `date`s with `datetime` fields.

app/models/main.py
risicle: While it's true that our models do have some of these weird special-casey comparison methods in their `query_class`es, but they were generally written a long time ago and it's not a pattern I think we want to continue following.

app/models/main.py
```diff
@@ -1653,13 +1656,37 @@ def validate(self, enforce_required=True, required_fields=None, max_day_rate=Non
         if errs:
             raise ValidationError(errs)
 
+    class query_class(BaseQuery):
+        def has_datetime_field_between(self, attr, start_datetime, end_datetime, inclusive=None):
+            """
+            Date filter values should be datetime objects. e.g.
+                BriefResponse.query.has_date_field_between(
```
galund: I guess you changed the name of the function, so this comment needs updating?

app/models/main.py
```diff
@@ -1653,13 +1656,37 @@ def validate(self, enforce_required=True, required_fields=None, max_day_rate=Non
         if errs:
             raise ValidationError(errs)
 
+    class query_class(BaseQuery):
+        def has_datetime_field_between(self, attr, start_datetime, end_datetime, inclusive=None):
```
galund: If `inclusive` is a bool, the default should be `False` rather than `None`? That said, if you are only ever calling with True, maybe the False path should just go away? (Actually, seems like what you really want in both cases is a half-closed range: it'd be better to add one to the day, rather than have to catch the last millisecond of the day in your criteria?

I'm not sure I quite like this abstraction (and if it does need to be a thing, it should be unit tested really). I wonder if it might be more clear to just add the filters you need in the two places.

app/models/main.py
galund: oh look we said the same thing, sorry, not intending to pile on, but FWIW I agree with this

app/models/main.py
```diff
@@ -1687,6 +1714,16 @@ def serialize(self):
 
         return purge_nulls_from_data(data)
 
+    def _parent_brief_fields(self, parent_brief):
+        return {
+            **self._unpack_fields(self.PARENT_BRIEF_FIELDS, parent_brief),
+            **{'framework': self._unpack_fields(self.FRAMEWORK_FIELDS, parent_brief)}
```
galund: Can we go to the underlying model instances rather than go via the existing serialization.

I don't want "frameworkFramework" nonsense to leak into our new more-perfect thing: we need to have a place to rename it to 'family'.

app/models/main.py
```diff
@@ -1687,6 +1714,16 @@ def serialize(self):
 
         return purge_nulls_from_data(data)
 
+    def _parent_brief_fields(self, parent_brief):
```
risicle: It might be nicer to name this in a way that makes it clear it's related to serialization. And also it's often nice to name "getter" methods with "get", so maybe prefixed with `_get_...`

app/models/main.py
```diff
@@ -1687,6 +1714,16 @@ def serialize(self):
...
+        return {
+            **self._unpack_fields(self.PARENT_BRIEF_FIELDS, parent_brief),
+            **{'framework': self._unpack_fields(self.FRAMEWORK_FIELDS, parent_brief)}
+        }
+
+    @staticmethod
+    def _unpack_fields(required_fields, collection):
```
risicle: _Personally_ probably wouldn't bother splitting this out and just write it inline instead, even if it's repeated. Means reader doesn't have to go & find the implementation of the function to see what it's doing, and doesn't pollute any namespaces...

tests/main/views/test_brief_response.py
```diff
@@ -866,3 +869,16 @@ def test_filter_brief_response_list_for_all_statuses(self):
 
         assert res.status_code == 200
         assert len(data['briefResponses']) == 4
+
+    def test_filter_responses_awarded_yesterday(self):
```
katstevens: I'd rename the test to `test_filter_brief_responses_awarded_yesterday` for consistency with the others.

tests/models/test_main.py
```diff
@@ -1176,6 +1182,30 @@ def test_cannot_remove_award_from_brief_response_if_brief_awarded(self):
 
         assert 'Cannot remove or change award datestamp on previously awarded Brief Response' in e.value.message
 
+    @pytest.mark.parametrize('awarded_at, expected_result_count', [
+        (datetime(2016, 3, 16, 23, 59, 59), 0),
+        (datetime(2016, 3, 17, 0, 0, 0), 1),
+        (datetime(2016, 3, 17, 23, 59, 59), 1),
+        (datetime(2016, 3, 18, 0, 0, 0), 0),
+    ])
+    def test_query_brief_response_returns_awarded_at_between_two_datetimes(self, awarded_at, expected_result_count):
```
katstevens: And this one I think we should rename `test_query_brief_responses_awarded_within_date_range`.

tests/main/views/test_brief_response.py
```diff
@@ -866,3 +869,16 @@ def test_filter_brief_response_list_for_all_statuses(self):
...
+        yesterday = datetime.utcnow() - timedelta(days=1)
+        self.setup_dummy_brief_response(submitted_at=None)
+        self.setup_dummy_awarded_brief_response(brief_id=111, awarded_at=yesterday)
+        self.setup_dummy_awarded_brief_response(brief_id=222, awarded_at=yesterday - timedelta(days=5))
+        self.setup_dummy_brief_response(award_details={"pending": True})
+
+        res = self.list_brief_responses(awarded_at=yesterday.strftime("%Y-%m-%d"))
```
katstevens: We should use the `DATE_FORMAT` instead of `"%Y-%m-%d"` for consistency.

app/models/main.py
katstevens: Yep, maybe `_get_serialized_parent_brief_fields` is better in this case.

app/models/main.py
katstevens: Some background: we introduced something similiar for the `Brief` model with the flexibility to filter by different date fields (`withdrawn_at`, `published_at` etc). This code is copied from there with a bit of tweaking. We were thinking of potentially DRYing up the shared stuff into a Mixin (I know how much we all love Mixins)?

risicle: I just don't really see it as a particularly useful abstraction - it's inventing a new set of syntax & semantics for these particular cases, when in fact the task it's doing is not really anything special.

app/models/main.py
brenetic: Not sure I follow about the underlying model instances?

Kept `frameworkFramework` for consistency's sake however if we have decided to name it `family` it should start somewhere

app/models/main.py
galund: Put it another way: it feels like if the function wasn't there, it'd be the same total volume of code? At which point I'm not clear there's a benefit to having it.

And yes having to pass in a field name feels rather like going against the way the ORM wants you to work, it's not the usual syntax/style as Rob says.

app/models/main.py
galund: Hmm okay I've thought some more and we can fix several things at the same time here, I _think_:
 - go to the Brief serialization and add the `framework` key (i.e. do the duplicating thing, I genuinely don't think the slight increase in data volume is significant - we'll remove the old keys later)
 - here, then you can just add `framework` to `parent_brief_fields` and boom the data you need is here too?

katstevens: I like that idea @galund - hopefully it won't make too many more tests fail 😸 I don't think any of the consumers of serialized Briefs should care about an extra key.

brenetic:
@galund @risicle Have updated. The use of `has_datetime_field_between` perhaps should be decided by  🤘📜✂️?

app/models/main.py
katstevens: I'll have a look at this - if we can avoid it here we should take it out of the equivalent `Brief.query.has_datetime_field_between` too.

galund: If you're looking anyway then it really might be worth considering whether having this kind of function adds much value ;-)

katstevens: Good news everyone: the built-in `Foo.query.mydatefield.between(start_date, end_date)` works just fine if we only want inclusive ranges, which we do here 👍

app/models/main.py
katstevens: Now done. I included all the `Brief.frameworkFoo` fields in the `framework` sub dict so we can be consistent and use it for everything.

katstevens:
Ugh I don't know what's happened to this branch but I can't rebase it locally properly or squash any commits.

katstevens:
Closing this PR as it's descended into merge conflict hell. Will cherry pick the relevant commits into a new branch + PR.

PR #737, opened 2018-02-09
Ensure db transaction is closed

benvand:
This warning on our tests:
```SAWarning: Connection is already established for the given bind; execution_options ignored```

Highlighted that we were allowing open db transactions to bleed into our view code when testing.
In this PR the open session transactions are closed before the request to the view is made causing the view to create its own transaction.

This broke some of our tests because they were relying on this incorrect behaviour, adding objects to the transaction but not committing them, leaving the transaction open, then operating on them in the view. These have been updated to issue commits before the request to the view.

We also remove the feature flag for using `'SERIALIZED'` transaction isolation level allowing this to be used in production. Specifically, and only, on the `edit_draft_service` view.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -59,8 +59,7 @@ def isolation_level(level):
     def decorator(view):
         @wraps(view)
         def view_wrapper(*args, **kwargs):
-            if flask_featureflags.is_active('TRANSACTION_ISOLATION'):
-                db.session.connection(execution_options={'isolation_level': level})
+            db.session.connection(execution_options={'isolation_level': level})
             return view(*args, **kwargs)
         return view_wrapper
     return decorator
--- a/config.py
+++ b/config.py
@@ -1,5 +1,5 @@
 import os
-from dmutils.status import enabled_since, get_version_label
+from dmutils.status import get_version_label
 
 
 class Config:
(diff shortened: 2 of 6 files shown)


risicle:
I think it's useful to *have* a (working) transaction-isolation-setting decorator (bit sad that flask-sqlalchemy doesn't supply one), but it just so happens that the single place we use it it is total overkill.

risicle:
(I will probably thumbs-up this once I get a chance to test it locally)

galund:
Do we think it's worth replacing the existing thing with the `with_for_update` thing?

I think it's probably okay to remove the transaction isolation level thing if we aren't likely to need it: @risicle are there any views where you think we should add it?

galund:
https://stackoverflow.com/questions/10081121/sqlalchemy-select-for-update-example has some helpful stuff I thought BTW

benvand:
@ris awesome thanks. Yeah it looks like we should consider changing the isolation level at some point.

This is pretty interesting stuff so hopefully it won't get left behind but the sort of playing with this I'd like to do... I'd probably not get around to it until next firebreak

risicle:
Yeah 99% of the time the default transaction isolation level does exactly the right thing. Increasing the isolation level is for cases where you're doing complex interwoven operations. We do very very few of them, but there's always the possibility.

risicle:
:fireworks:

PR #740, opened 2018-02-12
Add validation for country-picker values

TheDoubleK:
The accessible country picker select widget we're now using stores vales as "country" or "territory" followed by a colon followed by a two-capital-letters country code.

We want to enforce this format in the database, for consistency.

Adding the validation here now will enforce it for all future updates. A migration will be needed to update the existing "gb" values to "country:GB".

## NOTE
I thought this would be OK to just go in whenever, but actually we'll need the migration to update existing "gb" values to "country:GB" first, because if a supplier with `registration_country` already set to "gb" tries to update their details (e.g. contact name, etc.) then the update will fail because of the now-invalid country data.

Still, be good to get this reviewed and ready to go for when the time comes.

### DIFF
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -293,6 +293,9 @@ class Supplier(db.Model):
     # Companies House numbers consist of 8 numbers, or 2 letters followed by 6 numbers
     COMPANIES_HOUSE_NUMBER_REGEX = re.compile('^([0-9]{2}|[A-Za-z]{2})[0-9]{6}$')
 
+    # The registration country values that come from our country picker select are in this format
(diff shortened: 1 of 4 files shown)


Wynndow:
I'm gonna throw a quick spanner in the works.
Here are some valid codes from the country picker:

`territory:AE-AZ`
`territory:XXD`
`territory:UM-89`

I'm _fairly_ sure that updating the regex to cover these should cover the lot though.

TheDoubleK:
@Wynndow you told me it was all two-letter codes!

Will fix up.  On Monday now.

TheDoubleK:
Fixed @Wynndow

PR #741, opened 2018-02-12
Fall back to created_at if no _logged_in_at timestamp

TheDoubleK:
Quick fix for this tech-debt ticket: 
https://trello.com/c/kTZYf8QY/38-newly-created-user-is-not-logged-in-so-has-null-lastloggedin-in-admin

When a user is created they are automatically authenticated when they first set a password.  This means that the "lastLoggedIn" timestamp is null, even if they have done work (e.g. created a brief) that you would think needs a login.  This is confusing for admin users who see users listed who have apparently never logged in.

By falling back to the created_at timestamp we accurately reflect that they were logged in when they first created the account, even if they have never been through the normal log-in flow.

### DIFF
--- a/tests/models/test_main.py
+++ b/tests/models/test_main.py
@@ -2,6 +2,7 @@ from datetime import datetime, timedelta
 
 import mock
 import pytest
+from freezegun import freeze_time
 from sqlalchemy.exc import IntegrityError
 
 from app import db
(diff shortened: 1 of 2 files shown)


tests/models/test_main.py
```diff
@@ -36,10 +37,30 @@ def test_should_not_return_password_on_user(self):
...
+                role='buyer',
+                password='password',
+                active=True,
+                failed_login_count=0,
+                created_at=then,
+                updated_at=later,
+                password_changed_at=later
+            )
+
+        assert user.serialize()['loggedInAt'] == "2018-01-09T00:00:00.000000Z"
```
katstevens: Very minor quibbling but should this test be explicit and set `logged_in_at=None`? Do we need to set `failed_login_count` if it defaults to 0?

TheDoubleK: Yeah that might make it more obvious exactly what's going on, will update.

TheDoubleK: Fixed.  Turns out we do need to set `failed_login_count` or it complains - it has a not null constraint in the table.

PR #742, opened 2018-02-20
Update validation on tradingStatus

samuelhwilliams:
## Summary
We now have a more limited set of options that are suitable for suppliers to set as their trading status, and we no longer allow free text entries, so we will allow only an enum of choices as valid entries.

This PR should be merged before the other two (data migration in scripts and new page in suppler-fe).

## Ticket
https://trello.com/c/Wvvtb1yW/52-new-editable-page-trading-status

### DIFF
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -281,6 +281,14 @@ class Supplier(db.Model):
     __tablename__ = 'suppliers'
 
     ORGANISATION_SIZES = (None, 'micro', 'small', 'medium', 'large')
+    TRADING_STATUSES = (None,
+                        "limited company (LTD)",
+                        "limited liability company (LLC)",
+                        "public limited company (PLC)",
+                        "limited liability partnership (LLP)",
+                        "sole trader",
+                        "public body",
+                        "other")
 
     # Companies House numbers consist of 8 numbers, or 2 letters followed by 6 numbers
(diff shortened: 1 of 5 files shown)


TheDoubleK:
We have other supplier validators (for `organisation_size` and `companies_house_number` on the model itself rather than in the JSON schema (also adding validation for country codes here: https://github.com/alphagov/digitalmarketplace-api/pull/740)

It might be nice, for consistency, to do the same with the values here?  It feels like the way we're using the JSON schema here is as a first pass "have I roughly got all the things I need" and then doing the tighter validation at the model level.

I'm prepared to be convinced that it should be in the JSON rather than the model, but I think that (1) it should all be done in the same place and (2) checks shouldn't be duplicated in both places.

Wynndow:
My understanding was that we use the json schemas so we can validate data that lives inside json data blobs in our tables. If we have properties like these that are defined on the model, shouldn't we just validate there? Or is that not how we use the json schemas?

samuelhwilliams:
Yeah, you're both right, will update.

samuelhwilliams:
@TheDoubleK @Wynndow updated

samuelhwilliams:
@CrystalPea added a parametrized test for updating with all valid trading statuses.

PR #744, opened 2018-02-22
Bump Dockerfile to 3.1.1

TheDoubleK:
🙏 this should fix the "no logs" issue

PR #746, opened 2018-02-22
Build psycopg2 from source

samuelhwilliams:
## Summary
When `pip install`ing psycopg2, we need to build it from source rather than use a pre-compiled binary/wheel so that it gets linked against the correct libssl.

https://github.com/unbit/uwsgi/issues/1569

We also need to update the freeze-requirements step to filter out any pip flags during the `pip freeze` step, as these apparently can't be parsed.

## A bit more detail
Today we merged a [seemingly innocuous change](https://github.com/alphagov/digitalmarketplace-api/pull/743) in the API to Preview . This was successfully deployed and appeared to be working well, until Hosted Graphite sent us an alert declaring that logs were no longer being shipped from the API to CloudWatch. This loss of logging could be tied back to around the time of the deploy, so we had a likely candidate. Investigating the container, we saw that the awslogs agent couldn't find `botocore.sessions`. @benvand had previously encountered this issue and identified the remedy: upgrade our out-of-date base image for the API from 2.1.1 to 3.x.x to pull in the changes introduced [here](https://github.com/alphagov/digitalmarketplace-docker-base/pull/27) that would separate the environments we use for installing python dependencies for some of our core libs (like awslogs, uwsgi, etc).

We went about doing this and updated the base image to 3.1.1 [here](https://github.com/alphagov/digitalmarketplace-api/pull/744). This changes our base image from debian jessie to debian stretch, and installs some extra libs. After deploying this and checking Grafana/HostedGraphite, we saw that logs had begun to ship correctly again. Hoorah.

Then our smoke tests failed. The API was now unable to connect to our DB, essentially taking the entire Preview DM offline. We had logs again, but they were all 5xx errors. 😨 After some googling we landed on [this issue logged with uWSGI](https://github.com/unbit/uwsgi/issues/1569). It presented the same symptoms and the setup described by a number of commenters mirrors ours very closely. After trying a number of the recommended solutions (falling back to uWSGI==2.0.14 - unable to install against our openssl, installing libssl-dev after uwsgi - unable to locate libssl-dev 😕 ) , I tried disabling binary installs of psycopg2 on our API and finally appeared to have a viable solution. Logs continued to be shipped and the API could access the DB again. The last step was to update our requirements.txt/requirements-app.txt files to include the new pip flag to exclude psycopg2 binaries. Now our `freeze-requirements` didn't want to play ball as it seems `pip freeze`'s `-r` flag doesn't like pip flags being listed inside them. So I've updated our `make freeze-requirements` step to strip out any lines beginning with `--` (i.e. pip flags). They get pumped into the output file requirements.txt in an earlier step, so they are still preserved and accounted for. With this in place, I think we now have an API that can be deployed and function correctly. (In fact, I've manually deployed a new API app instance to Preview built from this PR and it seems to be working correctly - we have both db connectivity and logs).

There's more to all of this and I'll try to fill in some of the gaps tomorrow (Fri 23 Feb)


galund:
(Just to say thanks for the comprehensive explanation here that got me up to speed with all the chatter on dm-2ndline, much appreciated.)

risicle:
Well diagnosed & handled.

But at some point we've got to realize that we're a frog and the water's boiling.

Makefile
```diff
@@ -36,8 +36,7 @@ freeze-requirements:
 	echo '# This file is autogenerated. Do not edit it manually.' > requirements.txt
 	cat requirements-app.txt >> requirements.txt
 	echo '' >> requirements.txt
-	$$(pwd)/venv-freeze/bin/pip freeze -r requirements-app.txt | sed -n '/The following requirements were added by pip freeze/,$$p' >> requirements.txt
-	rm -rf venv-freeze
```
idavidmcdonald: The removal of `rm -rf venv-freeze` is intentional?

benvand:
Epic write up.

One suggestion:
`$$(pwd)/venv-freeze/bin/pip freeze -r <(sed '/^--/d' requirements-app.txt) | sed -n '/The following requirements were added by pip freeze/,$$p' >> requirements.txt`
we can lose the `-r` argument all together if it's broken/ not fit for what we need it for 🤷‍♂️ 

It only affects the ordering.

Makefile
TheDoubleK: As it also does this as the first step of `freeze-requirements` I guess there's no harm in leaving it hanging around at the end?

benvand:
Ooooo or we could use the existing file 🤔 I wonder what that would do...

Makefile
samuelhwilliams: Ah that should be put back please sorry.

PR #747, opened 2018-02-26
Add service count to supplier export CSV

benvand:
As part of the work for the contract variations.
CCS only want to email suppliers with an active account and who have not been removed from the framework.

This can be determined as any supplier on framework with a service that is published.

Adding the published service count allows them to determine which suppliers are active on the framework.


DilwoarH:
LGTM (but I'm not an expert on this repo)

PR #748, opened 2018-02-27
2. Improvements to the `export_users_for_framework` view

benvand:
* Remove unnecessary `get_supplier_ids_for_completed_service` method. Only used in one place.
  * Remove tests for the above.
* Refactor 'suppliers with a published service' to be a boolean. Count is not required and is less efficient.
  * Update tests to expect boolean and to use `is` to check for boolean equality.
* Refactor sql to be more efficient
  * Select suppliers with a completed service and suppliers with a published service as subquerys
  * Join subqueries to main query
* Reorder imports
* Add new test

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -1,14 +1,13 @@
 from datetime import datetime
 
 from dmapiclient.audit import AuditTypes
-from sqlalchemy import func
 from sqlalchemy.orm import lazyload
 from sqlalchemy.exc import DataError, IntegrityError
 from flask import abort, current_app, jsonify, request
 
 from .. import main
 from ... import db, encryption
-from ...models import AuditEvent, BuyerEmailDomain, Framework, Service, Supplier, SupplierFramework, User
+from ...models import AuditEvent, BuyerEmailDomain, DraftService, Framework, Service, Supplier, SupplierFramework, User
 from ...supplier_utils import check_supplier_role
 from ...utils import (
     get_json_from_request,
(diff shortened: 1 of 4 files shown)


benvand:
So I added in the `published_service_count` field last week because we needed a way to see who is on a framework. 

We were in a bit of a rush and I decided to come back and refactor later. While refactoring I realised that the original need is satisfied by a bool.

To release I'd probably just pick a quiet time, remove the old `published_service_count` field from the admin, release the api, then add the new `published_service` field

benvand:
Admin FE PR here:
https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/380

benvand:
@samuelhwilliams good to go if you/ someone can review this and the admin one

app/main/views/users.py
```diff
@@ -313,7 +315,7 @@ def export_users_for_framework(framework_slug):
             'framework_agreement': framework_agreement,
             'application_result': application_result,
             'variations_agreed': variations_agreed,
-            'published_service_count': supplier_id_published_service_count.get(sf.supplier_id, 0)
+            'published_service': bool(has_a_published_service)
```
samuelhwilliams: I know this isn't really anything to do with this PR, but can we add a note to this return that this is a non-standard format? For JSON responses from the API we generally use camelCase rather than underscoring. But would be good to call it out so that we know this is an odd case that (should but probably won't) get fixed one day.

benvand: Done 👍

PR #749, opened 2018-02-28
Add column to user table

MuriloDalRi:
* Add user_research_preference column to users table

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -139,7 +139,8 @@ def create_user():
         active=True,
         created_at=now,
         updated_at=now,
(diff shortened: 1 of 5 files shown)


samuelhwilliams:
Not sure about the name of this. `user_research_preference` as True doesn't really make instant sense to me. Could it be clearer as something like `user_researched_opted_in` (which I think is what it's trying to say)?

DilwoarH:
lgtm in terms of code however sam does have a point about the variable name.

MuriloDalRi:
Changed the variable name. I think this is ready to go now.

PR #750, opened 2018-03-02
Add notifications.service.gov.uk to allowed domains for non-Production envs

TheDoubleK:
The test email address for Notify, used by our functional tests to invite an admin user, is at this domain.

### DIFF
--- a/config.py
+++ b/config.py
@@ -31,7 +31,7 @@ class Config:
     DM_API_BUYER_DOMAINS_PAGE_SIZE = 100
     DM_API_PROJECTS_PAGE_SIZE = 100
 
-    DM_ALLOWED_ADMIN_DOMAINS = ['digital.cabinet-office.gov.uk', 'crowncommercial.gov.uk', 'user.marketplace.team']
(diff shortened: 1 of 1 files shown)


config.py
```diff
@@ -31,7 +31,7 @@ class Config:
     DM_API_BUYER_DOMAINS_PAGE_SIZE = 100
     DM_API_PROJECTS_PAGE_SIZE = 100
 
-    DM_ALLOWED_ADMIN_DOMAINS = ['digital.cabinet-office.gov.uk', 'crowncommercial.gov.uk', 'user.marketplace.team']
+    DM_ALLOWED_ADMIN_DOMAINS = ['digital.cabinet-office.gov.uk', 'crowncommercial.gov.uk', 'user.marketplace.team', 'notifications.service.gov.uk']
```
samuelhwilliams: Might be good to only add this to the Preview config rather than all configs.

TheDoubleK: But also runs on Staging.  And devs might want to run locally.  The "user.marketplace.team" here is also only for functional tests, so this seems like the best place to me.  Production is the only environment that overrides this default.

samuelhwilliams: Oh so it does. Didn't scroll down far enough.

PR #752, opened 2018-03-05
Remove registration date

samuelhwilliams:
## Summary
Registration date is a field we are no longer interested in, so we should remove all references from it in order to later drop the column from the database.

## Note
This PR must go in **before**, and be released through a separate pipeline to, https://github.com/alphagov/digitalmarketplace-api/pull/753

## Ticket
https://trello.com/c/M2jPJ3kZ/58-purge-registration-date-from-logic-and-data


MuriloDalRi:
@CrystalPea I think that will come in a different PR, this one is just dropping all the references before actually dropping the column.

PR #755, opened 2018-03-06
Add company details complete column

samuelhwilliams:
## Summary
Add a new column to the Suppliers table that tracks whether or not a
supplier has confirmed that the company details that have been submitted
are complete and correct. Defaults to False in the migration so that all
existing suppliers will get a one-off chance to update their existing
details. After this, the frontend will insist that suppliers go through
support to get further corrections (but the API will remain editable
regardless of the flag's value).

 ## Ticket
https://trello.com/c/7UIX3xPs/59


tests/main/views/test_suppliers.py
```diff
@@ -461,7 +461,9 @@ def test_update_response_matches_payload(self):
         supplier.pop('links')
         supplier.pop('id')
 
-        assert supplier == payload
+        payload_with_defaults = {**payload, 'companyDetailsConfirmed': False}
+
+        assert supplier == payload_with_defaults
```
Wynndow: how does `'companyDetailsConfirmed': False` get into the supplier returned in the response in this test? I'm assuming it's the supplier that's POSTed on line 403. But that's loaded from the example listing from `example_listings/supplier_creation.json` which doesn't have the new `companyDetailsConfirmed ` key in it?

TheDoubleK: It's a default value - everyone get's False to start with.

Wynndow: Ah, because the migrations get run before the tests.

PR #759, opened 2018-03-07
Add audit event for creation of supplier

idavidmcdonald:
https://trello.com/c/0Q9fOz46/234-creating-a-supplier-is-not-audited-there-is-no-createsupplier-audit-event

We were not doing this previously and wanted to add coverage
of it to match that we store audit events for all other DB
actions.

Note, unlike most other actions, we do not have a logged in user
at this point so who did the action is unknown.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -143,11 +143,11 @@ def import_supplier(supplier_id):
 
 @main.route('/suppliers', methods=['POST'])
 def create_supplier():
-    supplier_data = validate_and_return_supplier_request()
+    request_data = validate_and_return_supplier_request()
 
-    contact_informations_data = supplier_data['contactInformation']
+    contact_informations_data = request_data['contactInformation']
     supplier_data = drop_foreign_fields(
-        supplier_data,
+        request_data.copy(),
         ['contactInformation']
     )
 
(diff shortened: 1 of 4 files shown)


TheDoubleK:
For this PR it would just mean taking an optional field in the request, and setting it for the user on the audit event if it exists.  We can then patch up the frontend call at our leisure...

idavidmcdonald:
Could do, although if it's patching up the frontend call at our leisure (i.e. never) then I'd rather do it all or not do it at all as we wouldn't have delivered any actual value if it's just half done. At the moment I lean towards to not doing it at all but I think that's because it's the end of the day... I'll rethink tomorrow morning :)

Small question which I don't think is a stopper. Should we store an email address as the updated by value if we haven't yet confirmed that the person who filled in the form owns that email address? I don't think we would care too much as if they don't validate it then they won't ever log in and the account won't become active and if they do activate it then everyone is happy.

TheDoubleK:
I think it's OK to grab that email address and save it - if they never sign up it doesn't matter to the audit table - it's a "this is what they put in the form" kind of save I think.

idavidmcdonald:
Note for observers, we decided to go for the quicker approach of merging what we have here as adding the ability to set the `updated_by` details would make this ticket much larger (an extra PR on the client, supplier FE and then on the API to remove backwards compatibility of not having an `updated_by`. We should still be able to work out who did the action if we really needed to by looking at audit events created at a similar time.

PR #760, opened 2018-03-12
tracing: bump digitalmarketplace-apiclient dependency to 15.0.0

risicle:
This should bring in support for zipkin span ids

https://trello.com/c/exuNFb7A/357-zipkin-headers-pt2-api-client
https://trello.com/b/ioXxSA1U/tech-debt


risicle:
I... think... so....

PR #764, opened 2018-03-19
Serialize audit events object type and id

Wynndow:
We weren't returning an audit events object type and id in the API
response. It's useful so we should!

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -61,6 +61,7 @@ def create_framework():
             lots=lots
         )
         db.session.add(framework)
+        db.session.flush()
         db.session.add(
             AuditEvent(
                 audit_type=AuditTypes.create_framework,
@@ -178,7 +179,11 @@ WHERE
                 audit_type=AuditTypes.framework_update,
                 db_object=framework,
                 user=updater_json['updated_by'],
-                data={'update': json_payload['frameworks']})
+                data={
+                    'update': json_payload['frameworks'],
+                    'frameworkSlug': framework.slug,
+                },
+            )
         )
         db.session.commit()
     except IntegrityError as e:
--- a/tests/main/views/test_audits.py
+++ b/tests/main/views/test_audits.py
@@ -629,7 +629,11 @@ class TestAuditEvents(BaseTestAuditEvents):
         assert AuditEvent.query.count() == count + 1
 
     def test_should_get_audit_event(self):
-        aid = self.add_audit_event(0)
+        supplier_id = self.setup_dummy_suppliers(1)[0]
+        supplier = Supplier.query.filter(Supplier.supplier_id == supplier_id).first()
+        object_id = supplier.id
+        aid = self.add_audit_event(0, db_object=supplier)
+
         response = self.client.get('/audit-events')
         data = json.loads(response.get_data())
         assert response.status_code == 200
@@ -639,6 +643,8 @@ class TestAuditEvents(BaseTestAuditEvents):
             'acknowledged': False,
             'user': '0',
             'data': {'request': 'data'},
+            'objectType': 'Supplier',
+            'objectId': object_id,
             'id': aid,
             'createdAt': mock.ANY
         }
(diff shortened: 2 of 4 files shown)


tests/main/views/test_audits.py
```diff
@@ -639,6 +643,8 @@ def test_should_get_audit_event(self):
             'acknowledged': False,
             'user': '0',
             'data': {'request': 'data'},
+            'objectType': 'Supplier',
+            'objectId': object_id,
```
galund: So suppliers is the weird one where `suppliers.id` != `supplier_id`.

Might this therefore be particularly confusing in that case? In most cases presumably the id is the id, but now we're exposing an identifier (for the first time) that otherwise, the API doesn't expose...

It's interesting that our tests seem to be covering the special-case model rather than any of the others... we probably should at least make sure the expected behaviour for other kinds of models is tested/documented somehow.

Maybe I am just confused :)

Wynndow: I see what you mean about the weirdness with `supplier.id` and `supplier.supplier_id`, but I think this is exhibiting the expected behaviour. As I understand it we expect the `object_id` of an audit event to be that objects id, which it is here? The only reason I used a supplier as an object is because the `add_audit_event` method is set up to create a `supplier_update` audit event. I've only used the `supplier_id` to make sure I'm getting the correct supplier back when querying the `Supplier` model.
I could change it to use a different object, like a `User` maybe which would be more straight forward, but then the audit event and the object don't really match and it's a bit weird?

galund: Sure, I've no idea what the expected behaviour is for audit events, but it makes sense to be as you describe... 

What's in that `data` dict? I'm struggling to see how this serialization can have been any use before, if it didn't include some kind of identifier for the thing that changed!!

Wynndow: It tends to be details of the audit event. For example, when updating the status of a framework it looks like this:
```
"data": {
    "update": {
        "status": "standstill"
    }
},
```
It was looking at exactly these events that prompted this PR. I was trying to work out when G-9 went from open to pending, and without the object id's of the frameworks, it was really difficult!

Other audit events have more details in them which make them more obvious, such as when updating a frameworks `frameworkAgreementDetails`. In that case the data blob includes stuff like `lotDescriptions` and `frameworkURL` so you can infer what the audit event was for more easily.

TheDoubleK:
Hmm, the service ID thing is a weird one, and especially if we're moving to use more obfuscated ID's, the internal database ID will be less useful.

There is a method around auditing that dies translation to public ID's. Maybe we can use that for the serialization? Out knowing the internal ID of, e.g. a service, is pretty much useless to us from the API, as the only way we can find the service from that is to log in to the database.

Wynndow:
For context Kev and I chatted about this this morning. We thought that it would be good to return an object's external ID (as apposed to it's table ID) or some identifiable piece of information in the audit event, but that it should go in the data blob. This is so the top level keys of the audit event are constant regardless of the object type the audit event is actually for.

I've added the `framework_slug` to the audit event for updating frameworks so they are more useful, but doing this for every audit event that may need it is probably outside the scope of this PR.

Returning the object_id and type is still potentially useful, so they'e stayed in.

app/main/views/frameworks.py
```diff
@@ -61,13 +61,6 @@ def create_framework():
             lots=lots
         )
         db.session.add(framework)
-        db.session.add(
-            AuditEvent(
-                audit_type=AuditTypes.create_framework,
-                db_object=framework,
-                user=updater_json['updated_by'],
-                data={'update': json_payload['frameworks']})
-        )
```
TheDoubleK: Rather than move it out of the transaction you can do (I think, but look it up) `db.session.flush()` which will get the ID you need but keep the things within a single transaction.

I'm pretty sure this is a thing we do elsewhere.

We definitely want the two things to be created in a single commit - so either both pass or both fail.  We don't want a situation where the data gets written but the audit event fails.

TheDoubleK: See eg here: https://github.com/alphagov/digitalmarketplace-api/blob/5cb28313a05b510bb9df04379b93ec7849ba3942/app/main/views/suppliers.py#L163

Wynndow: Ah nice. Did not know that was a thing. Will fix up.

PR #765, opened 2018-03-27
Allow setting framework close date using the API

TheDoubleK:
This is used by the "Reusing declarations" page both to display the date on the page and to select the most recent declaration to reuse.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -95,6 +95,7 @@ def update_framework(framework_slug):
         'status': 'status',
         'clarificationQuestionsOpen': 'clarification_questions_open',
         'frameworkAgreementDetails': 'framework_agreement_details',
+        'applicationCloseDate': 'application_close_date',
         'allowDeclarationReuse': 'allow_declaration_reuse',
     }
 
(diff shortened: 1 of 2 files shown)


app/main/views/frameworks.py
```diff
@@ -95,6 +95,7 @@ def update_framework(framework_slug):
         'status': 'status',
         'clarificationQuestionsOpen': 'clarification_questions_open',
         'frameworkAgreementDetails': 'framework_agreement_details',
+        'applicationCloseDate': 'application_close_date',
```
katstevens: Do we need to bother validating that the supplied date is in the right format anywhere before saving? I can't see a schema for a framework object, only the `frameworkAgreementDetails`.

TheDoubleK: It saves as a datetime field rather than a string.  You can post various formats that work (e.g. `2017-04-11 16:00`.  And if you post something that doesn't work the DB will choke on it, which I think we can live with as a validation.

katstevens: Fair enough 👍

PR #767, opened 2018-03-29
Remove serializeable isolation level from draft services

Wynndow:
We noticed that when under high load, we were seeing some transaction
errors from psycopg when updating draft briefs:
```
sqlalchemy.exc.OperationalError:
(psycopg2.extensions.TransactionRollbackError) could not serialize
access due to read/write dependencies among transactions
DETAIL:  Reason code: Canceled on identification as a pivot, during
commit attempt.
HINT:  The transaction might succeed if retried.
```

We think this was due to specifying that the transactions were fully
serializeable, which put a lock on more resources than necessary. By
updating the query to use `with_for_update` we can make sure only the
row we're interested in is locked.

We also noticed that the query for the draft service was joined, and so
each time we fetched the draft service from the db we were pulling back
_loads_ of stuff we didn't need. Frameworks, lots etc. Have chance to
lazy loading so we drastically reduce the cost of the query.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -1,10 +1,11 @@
 from dmapiclient.audit import AuditTypes
 from flask import jsonify, abort, request, current_app
 from sqlalchemy.exc import IntegrityError
+from sqlalchemy.orm import lazyload
 from sqlalchemy import asc
 
 from .. import main
-from ... import db, isolation_level
+from ... import db
 from ...validation import is_valid_service_id_or_400
 from ...models import Service, DraftService, Supplier, AuditEvent, Framework
 from ...utils import (
@@ -76,7 +77,6 @@ def copy_draft_service_from_existing_service(service_id):
 
 
 @main.route('/draft-services/<int:draft_id>', methods=['POST'])
-@isolation_level("SERIALIZABLE")
 def edit_draft_service(draft_id):
     """
     Edit a draft service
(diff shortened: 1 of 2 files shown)


app/__init__.py
```diff
@@ -45,21 +44,3 @@ def create_app(config_name):
...
-        def create_thing(id):
-            ...
-    """
-    def decorator(view):
-        @wraps(view)
-        def view_wrapper(*args, **kwargs):
-            db.session.connection(execution_options={'isolation_level': level})
-            return view(*args, **kwargs)
-        return view_wrapper
-    return decorator
```
risicle: I... think we should keep this. It's debugged, it works, it's not like we're never going to need to set an alternate serialization level.

Wynndow: If this PR goes in it's not used anywhere, and if we need it in the future it's in the commit history?

risicle: Nobody ever looks in the commit history - it'll be as good as gone and before you know it the next person who needs this will have done something godawful like injecting a manual `SET ...` statement in the beginning of a view or ... *something*

Wynndow: Fair enough. Have added un-removed it.

benvand:
Just to dive in on this if you google 'flask set transaction level per view' or anything similar you actually find @robyoung s gist for this code:
https://gist.github.com/robyoung/68e2121c045f117bea78

So not sure it's that important that we keep it just for the sake of history

risicle:
Shrug. It's one of those things that I'm amazed isn't provided out of the box.

Wynndow:
I'm gonna keep it. It's a thing that works and might be useful, and I guess there's no guarantee that Rob's gist will be there in the future.

PR #771, opened 2018-04-10
supplier audit events: add supplierId to most events

risicle:
this is the externally visible id, which is used in most parts of the api. far more handy than the internal id most the time.

A result of trying to track down a second line situation and finding this lacking: https://trello.com/c/11syiJPd/260-overzealous-notify-failure-causes-503-rather-than-graceful-handling


risicle:
Having it searchable would be a bonus, but not having the information at all makes certain things impossible without digging into the raw database.

`supplierId` is used in a number of the other events (in the same file).

PR #772, opened 2018-04-12
Copy services between frameworks

Wynndow:
For this trello ticket: https://trello.com/c/Nbv7N4lf

### Add 'copied_to_following_framework' column to services
We need to keep track of whether a service has been copied to a
following framework during applications. Adding this column will allow
this.

### Add copied_to_following_framework to service models
This new column will allow us to tell if a service has been copied to a
new framework. It needs to be dropped from the data when validated so it
doesn't prevent validation.

### Allow copying of draft services to new frameworks
This allows the existing `copy_draft_service_from_existing_service`
endpoint to copy existing services to different frameworks.

This endpoint doesn't appear to be currently used anywhere, however I've
maintained it's default behaviour so this isn't a breaking change to the
API.

### Add endpoint for copying multiple services
Some suppliers may have hundreds of services on a lot. To save on api
calls, this endpoint allows us to bulk copy services to drafts.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -2,12 +2,12 @@ from dmapiclient.audit import AuditTypes
 from flask import jsonify, abort, request, current_app
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import lazyload
-from sqlalchemy import asc
+from sqlalchemy import asc, desc
 
 from .. import main
 from ... import db
 from ...validation import is_valid_service_id_or_400
-from ...models import Service, DraftService, Supplier, AuditEvent, Framework
+from ...models import Service, DraftService, Supplier, AuditEvent, Framework, Lot
 from ...utils import (
     get_int_or_400,
     get_json_from_request,
@@ -100,6 +100,88 @@ def copy_draft_service_from_existing_service(service_id):
         abort(400, format(e))
 
     return single_result_response(RESOURCE_NAME, draft), 201
+
+
+@main.route('/draft-services/<framework_slug>/<lot_slug>/copy-published-from-framework', methods=['POST'])
+def copy_published_from_framework(framework_slug, lot_slug):
+    """
+    Copy all published services from a given framework/lot to a different framework's drafts.
+    :param framework_slug: The slug for the framework to create the new drafts in.
+    :param lot: The slug for the lot to copy services from/create drafts forself.
+    :return: The serialized drafts.
+    """
+    updater_json = validate_and_return_updater_request()
+    json_payload = get_json_from_request()
+
+    supplier_id = get_int_or_400(json_payload, 'supplierId')
+    source_framework_slug = json_payload.get('sourceFrameworkSlug')
+    questions_to_copy = json_payload.get('questionsToCopy')
+
+    if not source_framework_slug:
+        abort(400, "Required data missing: 'source_framework_slug'")
+    if not questions_to_copy:
+        abort(400, "Required data missing: 'questions_to_copy'")
+    if not isinstance(questions_to_copy, list):
+        abort(400, "Data error: 'questions_to_copy' must be a list")
+
+    target_framework = Framework.query.filter(
+        Framework.slug == framework_slug
+    ).first_or_404()
+
+    if target_framework.status != 'open':
+        abort(400, "Target framework is not open")
+
+    source_services = Service.query.filter(
+        Service.supplier_id == supplier_id,
+        Service.framework.has(
+            Framework.slug == source_framework_slug
+        ),
+        Service.lot.has(
+            Lot.slug == lot_slug
+        ),
+        Service.status == 'published',
+        Service.copied_to_following_framework == False,  # NOQA
+    ).order_by(
+        # Descending order so created drafts id's are sequential in reverse alphabetical order. Helps with ordering
+        # in the frontend (most recent id first).
+        desc(Service.data['serviceName'].astext)
+    ).all()
+
+    drafts_services = []
+    for service in source_services:
+        draft = DraftService.from_service(
+            service,
+            questions_to_copy,
+            target_framework_id=target_framework.id,
+        )
+        drafts_services.append((draft, service))
+
+        service.copied_to_following_framework = True
+        db.session.add(draft, service)
+
+    # Flush so the new drafts are assigned an ID, which is used below in the audit events.
+    db.session.flush()
+
+    for draft, service in drafts_services:
+        audit = AuditEvent(
+            audit_type=AuditTypes.create_draft_service,
+            user=updater_json['updated_by'],
+            data={
+                "draftId": draft.id,
+                "serviceId": service.id
+            },
+            db_object=draft
+        )
+
+        db.session.add(audit)
+
+    try:
+        db.session.commit()
+    except IntegrityError as e:
+        db.session.rollback()
+        abort(400, format(e))
+
+    return list_result_response(RESOURCE_NAME, [draft_service[0] for draft_service in drafts_services]), 201
 
 
 @main.route('/draft-services/<int:draft_id>', methods=['POST'])
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -10,6 +10,7 @@ from ...validation import is_valid_service_id_or_400
 from ...models import Service, DraftService, Supplier, AuditEvent, Framework
 from ...utils import (
     get_int_or_400,
+    get_json_from_request,
     get_request_page_questions,
     json_only_has_required_keys,
     list_result_response,
@@ -39,21 +40,46 @@ def copy_draft_service_from_existing_service(service_id):
     """
     is_valid_service_id_or_400(service_id)
     updater_json = validate_and_return_updater_request()
+    json_payload = get_json_from_request()
+
+    target_framework_slug = json_payload.get('targetFramework')
+    questions_to_copy = set(json_payload.get('questionsToCopy', []))
+
+    if target_framework_slug and not questions_to_copy:
+        abort(400, "Required data missing: 'questions_to_copy'")
 
     service = Service.query.filter(
         Service.service_id == service_id
     ).first_or_404()
 
+    if target_framework_slug:
+        target_framework = Framework.query.filter(
+            Framework.slug == target_framework_slug
+        ).first_or_404()
+        if not target_framework.status == 'open':
+            abort(400, "Target framework is not open")
+        target_framework_id = target_framework.id
+    else:
+        target_framework_id = service.framework.id
+
     draft_service = DraftService.query.filter(
         DraftService.service_id == service_id,
         DraftService.status.notin_(('not-submitted', 'submitted')),
     ).first()
-    if draft_service:
+
+    if draft_service and draft_service.framework.id == target_framework_id:
         abort(400, "Draft already exists for service {}".format(service_id))
 
-    draft = DraftService.from_service(service)
+    draft = DraftService.from_service(
+        service,
+        questions_to_copy=questions_to_copy,
+        target_framework_id=target_framework_id,
+    )
 
-    db.session.add(draft)
+    if target_framework_id != service.framework.id:
+        service.copied_to_following_framework = True
+
+    db.session.add(draft, service)
     db.session.flush()
 
     audit = AuditEvent(
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -1092,15 +1092,27 @@ class DraftService(db.Model, ServiceTableMixin):
                            default=None)
 
     @staticmethod
-    def from_service(service):
-        return DraftService(
-            framework_id=service.framework_id,
-            lot=service.lot,
-            service_id=service.service_id,
-            supplier=service.supplier,
-            data=service.data,
-            status=service.status
-        )
+    def from_service(service, questions_to_copy=None, target_framework_id=None):
+        draft_not_on_same_framework_as_service = target_framework_id and target_framework_id != service.framework.id
+        if draft_not_on_same_framework_as_service and not isinstance(questions_to_copy, (set, list, tuple)):
+            raise TypeError("'questions_to_copy' must be a list, set or tuple")
+
+        kwargs = {
+            'framework_id': service.framework_id if not target_framework_id else target_framework_id,
+            'lot': service.lot,
+            'service_id': service.service_id,
+            'supplier': service.supplier,
+            'data': {
+                key: value for key, value in service.data.items() if key in questions_to_copy
+            } if draft_not_on_same_framework_as_service else service.data,
+            'status': service.status if not target_framework_id or
+            target_framework_id == service.framework.id else 'not-submitted',
+        }
+
+        if draft_not_on_same_framework_as_service:
+            kwargs.pop('service_id')
+
+        return DraftService(**kwargs)
 
     def copy(self):
         if self.lot.one_service_limit:
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -890,6 +890,8 @@ class ServiceTableMixin(object):
     updated_at = db.Column(db.DateTime, index=False, nullable=False,
                            default=datetime.utcnow, onupdate=datetime.utcnow)
 
+    copied_to_following_framework = db.Column(db.Boolean, nullable=False, server_default=sql_false())
+
     @declared_attr
     def supplier_id(cls):
         return db.Column(db.BigInteger, db.ForeignKey('suppliers.supplier_id'),
@@ -944,7 +946,7 @@ class ServiceTableMixin(object):
             'supplierId', 'supplierName',
             'frameworkSlug', 'frameworkFramework', 'frameworkName', 'frameworkStatus',
             'lot', 'lotSlug', 'lotName',
-            'updatedAt', 'createdAt', 'links'
+            'updatedAt', 'createdAt', 'links', 'copiedToFollowingFramework'
         ])
 
         data = strip_whitespace_from_data(data)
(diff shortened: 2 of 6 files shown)


app/main/views/drafts.py
```diff
@@ -39,21 +40,42 @@ def copy_draft_service_from_existing_service(service_id):
...
         DraftService.service_id == service_id,
         DraftService.status.notin_(('not-submitted', 'submitted')),
     ).first()
-    if draft_service:
+
+    if draft_service and draft_service.framework.id == target_framework_id:
         abort(400, "Draft already exists for service {}".format(service_id))
 
-    draft = DraftService.from_service(service)
+    draft_kwargs = {
```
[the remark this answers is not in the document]
Wynndow: No, this was a hangover from when I was doing something silly, will remove.

app/main/views/drafts.py
```diff
@@ -76,6 +98,77 @@ def copy_draft_service_from_existing_service(service_id):
     return single_result_response(RESOURCE_NAME, draft), 201
 
 
+@main.route('/draft-services/<framework_slug>/<lot_slug>/copy-previous-framework-services-to-drafts', methods=['POST'])
```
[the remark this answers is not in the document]
Wynndow: Yep, something shorter will be better. Will address.

app/main/views/drafts.py
```diff
@@ -76,6 +98,77 @@ def copy_draft_service_from_existing_service(service_id):
...
+    target_framework = Framework.query.filter(
+        Framework.slug == framework_slug
+    ).first_or_404()
+
+    if not target_framework.status == 'open':
+        abort(400, "Target framework is not open")
+
+    source_services = Service.query.filter(
+        Service.supplier_id == supplier_id,
+        Service.framework.has(
```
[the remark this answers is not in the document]
Wynndow: Sadly not. You get this interesting error:
```
AttributeError: Neither 'InstrumentedAttribute' object nor 'Comparator' object associated with Service.framework has an attribute 'slug'
```

I've not dug hard into the why, but we use `has` in a bunch of other places.

app/main/views/drafts.py
```diff
@@ -76,6 +98,77 @@ def copy_draft_service_from_existing_service(service_id):
...
+        'target_framework_id': target_framework.id,
+    }
+
+    drafts = []
+    for service in source_services:
+        draft = DraftService.from_service(service, **draft_kwargs)
+        service.copied_to_following_framework = True
+
+        db.session.add(draft, service)
+        db.session.flush()
```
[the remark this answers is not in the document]
Wynndow: The flush is so the new draft gets an id, which is used in the audit event further in the loop.

app/models/main.py
```diff
@@ -894,6 +894,12 @@ class ServiceTableMixin(object):
     updated_at = db.Column(db.DateTime, index=False, nullable=False,
                            default=datetime.utcnow, onupdate=datetime.utcnow)
 
+    copied_to_following_framework = db.Column(
```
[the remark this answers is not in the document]
Wynndow: Yep

tests/main/views/test_drafts.py
```diff
@@ -50,12 +51,396 @@ def setup(self):
             **payload
         )
 
+    def teardown(self):
+        super().teardown()
```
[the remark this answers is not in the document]
Wynndow: yes indeed!

tests/main/views/test_drafts.py
```diff
@@ -50,12 +51,396 @@ def setup(self):
...
+        res = self.publish_draft_service(draft['id'])
+        assert res.status_code == 200
+
+        return res
+
+
+class TestCopyDraftServiceFromExistingService(DraftsHelpersMixin):
+    def test_reject_copy_with_no_updated_by(self):
+        res = self.client.put('/draft-services/copy-from/0000000000')
+        assert res.status_code == 400
```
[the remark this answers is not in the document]
Wynndow: I've just checked and you're right, it's actually failing because of an unexpected content type! These tests I just split out from the existing class (cos it was over 1000 lines) so I might need to look at if they're actually doing what they're supposed to.

app/main/views/drafts.py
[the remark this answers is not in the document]
Wynndow: `copy-published-from-framework` seems reasonable

app/main/views/drafts.py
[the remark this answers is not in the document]
samuelhwilliams: Interesting! Thanks for the explanation

app/main/views/drafts.py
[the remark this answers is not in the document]
samuelhwilliams: Ah gotcha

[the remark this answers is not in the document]
samuelhwilliams: Though we still may need to refactor this bit to not have multiple journeys back/forth from the DB. We could keep track of all the services and then submit the AuditEvents in a second batch afterwards.

Wynndow:
@samuelhwilliams Have just pushed changes to use the whitelist of questions instead of the blacklist. Also refactored the tests a bunch.

app/main/views/drafts.py
```diff
@@ -76,6 +100,78 @@ def copy_draft_service_from_existing_service(service_id):
     return single_result_response(RESOURCE_NAME, draft), 201
 
 
+@main.route('/draft-services/<framework_slug>/<lot_slug>/copy-published-from-framework', methods=['POST'])
+def copy_published_from_framework(framework_slug, lot_slug):
```
samuelhwilliams: A little docstring here might be nice to just expand on the name a little and give some more context so it's immediately obvious what this is doing.

tests/main/views/test_drafts.py
```diff
@@ -56,6 +62,457 @@ def service_count(self):
...
+            content_type='application/json')
+
+        # draft = json.loads(res.get_data(as_text=True))['services']
+        assert res.status_code == 201
+
+        draft = DraftService.query.filter(
+            DraftService.framework_id == 4,
+        ).first()
+
+        assert all(key in ['serviceBenefits', 'termsAndConditionsDocumentURL'] for key in draft.data.keys())
```
samuelhwilliams: Might be a bit cleaner doing a list/set comparison like `assert {'serviceBenefits', 'termsAndConditionalsDocumentURL'} == set(draft.data.keys())`

tests/helpers.py
```diff
@@ -217,15 +217,17 @@ def setup_dummy_service(self, service_id, supplier_id=1, data=None,
         db.session.commit()
         return service.id
 
-    def setup_dummy_services(self, n, supplier_id=None, framework_id=1,
-                             start_id=0, lot_id=1, model=Service):
+    def setup_dummy_services(self, n, supplier_id=None, framework_id=1, data=None,
```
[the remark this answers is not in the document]
Wynndow: No, no reason other than it was visually more pleasing. In hindsight, perhaps a bit reckless. But tests are passing!

tests/main/views/test_drafts.py
```diff
@@ -56,6 +62,457 @@ def service_count(self):
...
+        res = self.client.put(
+            '/draft-services/copy-from/{}'.format(self.service_id),
+            data=json.dumps({
+                **self.updater_json,
+                **self.basic_questions_to_copy,
+                'targetFramework': 'z-cloud'
+            }),
+            content_type='application/json')
+
+        assert res.status_code == 404
```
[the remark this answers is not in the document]
Wynndow: Yeah, I see what you're saying. Still feels like a resource you wanted to do something with doesn't exist though.

This isn't the original behaviour though, as we didn't have the concept of a target framework before.

app/models/main.py
```diff
@@ -1093,15 +1096,24 @@ class DraftService(db.Model, ServiceTableMixin):
...
+        draft_not_on_same_framework_as_service = target_framework_id and target_framework_id != service.framework.id
+
+        kwargs = {
+            'framework_id': service.framework_id if not target_framework_id else target_framework_id,
+            'lot': service.lot,
+            'service_id': service.service_id,
+            'supplier': service.supplier,
+            'data': {
+                key: value for key, value in service.data.items() if key in questions_to_copy
+            } if draft_not_on_same_framework_as_service else service.data,
```
samuelhwilliams: If we specify 'target_framework_id' but don't specify 'questions_to_copy' this will break in a weird way (doing an `in` check on `None`. Should probably firm this up a bit to have a check early on that `questions_to_copy` is a list of strings if `draft_not_on_same_framework_as_service` is True.

tests/main/views/test_drafts.py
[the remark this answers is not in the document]
samuelhwilliams: Yep, that's fair, and it clearly uses our existing convention of first_or_404, so good call.

tests/main/views/test_drafts.py
```diff
@@ -28,6 +31,9 @@ def setup(self):
             'lot': 'scs',
             'supplierId': 1
         }
+        self.basic_questions_to_copy = {
+            'questionsToCopy': ['serviceName']
```
samuelhwilliams: I think the name of this variable implies the value will just be a list rather than a dict to be unpacked into a method call. Probably this should just be `['serviceName']`

tests/main/views/test_drafts.py
```diff
@@ -56,6 +62,455 @@ def service_count(self):
...
+            }),
+            content_type='application/json')
+
+        assert res.status_code == 400
+        assert 'Target framework is not open' in res.get_data(as_text=True)
+
+
+class TestCopyPublishedFromFramework(DraftsHelpersMixin):
+    def post_to_copy_published_from_framework(
+        self, source_framework_slug="g-cloud-6", questions_to_copy='default', framework_slug='g-cloud-7'
```
samuelhwilliams: The default value here for `questions_to_copy` could be slightly confusing as we're providing a string, but if you want to override the value you'll need to provide a list. Without type hints here, it's not immediately obvious how to use this parameter correctly. A docstring would help, but I think maybe we could rename the variable `override_questions_to_copy` and have the default as None. Having the available types as `(None, list)` is a bit more standard than `(str, list)`, IMO. Especially because strings and lists are both iterables, but you generally don't want to iterate over a string - so it's better to have that be a None which will break when you try to iterate over it.

samuelhwilliams: This reminds me that I still do want to add type hints to our codebase one day...

samuelhwilliams: Also, `framework_slug` should probably be called `target_framework_slug` for the avoidance of any doubt (given we have a `source_framework_slug`).

tests/main/views/test_drafts.py
```diff
@@ -56,6 +62,455 @@ def service_count(self):
...
+            DraftService.supplier_id == 1,
+            DraftService.framework_id == 4,  # G-Cloud 7
+            DraftService.lot_id == 2,
+        ).all()
+
+        assert len(serialized_drafts) == 5
+        assert len(db_drafts) == 5
+        assert all(service['status'] == 'not-submitted' for service in serialized_drafts)
+        assert all(service['frameworkSlug'] == 'g-cloud-7' for service in serialized_drafts)
+        assert all(not service.get('serviceId') for service in serialized_drafts)
```
samuelhwilliams: This test might be more precise as `'serviceId' not in service` (or `service['serviceId'] is None` if we expect the key to be present with a None value).

tests/main/views/test_drafts.py
```diff
@@ -56,6 +62,455 @@ def service_count(self):
...
+
+        assert 'Target framework is not open' in res.get_data(as_text=True)
+
+    def test_should_only_copy_published_services_from_correct_framework_and_lot(self):
+        self.setup_dummy_services(3, supplier_id=1, lot_id=1, status='published')
+        self.setup_dummy_services(2, supplier_id=1, lot_id=1, status='disabled', start_id=3)
+        self.setup_dummy_services(4, supplier_id=1, lot_id=2, status='published', start_id=5)
+        self.setup_dummy_services(1, supplier_id=1, lot_id=2, status='disabled', start_id=9)
+        self.setup_dummy_services(2, supplier_id=1, lot_id=3, status='published', start_id=10)
+        self.setup_dummy_services(3, supplier_id=1, lot_id=3, status='enabled', start_id=12)
```
samuelhwilliams: Might be nice to test that only services from the same supplier are copied across. You could test that "for free" by creating a service here that's published on the right lot by a different supplier, but it might be clearer as a separate test. It would obviously be _completely broken_ if this condition wasn't met, but it's still good to explicitly test it.

tests/main/views/test_drafts.py
```diff
@@ -56,6 +62,455 @@ def service_count(self):
...
+
+        assert len(drafts) == 4
+        assert all(draft['frameworkSlug'] == 'g-cloud-7' for draft in drafts)
+        assert all(draft['lotSlug'] == 'paas' for draft in drafts)
+        assert all(draft['status'] == 'not-submitted' for draft in drafts)
+        for service, draft in zip(services, drafts):
+            assert service.data['serviceName'] == draft['serviceName']
+
+    def test_should_create_drafts_in_reverse_alphabetical_order_by_service_name(self):
+        for service_id, char in enumerate(('S', 'E', 'R', 'V', 'I', 'C', 'E')):
```
samuelhwilliams: You can just do `enumerate('SERVICE')`. The string itself is iterable and will get broken up in the exact same way.

tests/main/views/test_drafts.py
```diff
@@ -56,6 +62,455 @@ def service_count(self):
...
+                lot_id=2,
+                data={'serviceName': f'{char} service'},
+            )
+
+        res = self.post_to_copy_published_from_framework()
+        assert res.status_code == 201
+
+        drafts = json.loads(res.get_data(as_text=True))['services']
+
+        assert sorted(drafts, key=lambda draft: draft['serviceName'], reverse=True) == drafts
```
samuelhwilliams: Can we do a comparison against something static here? It's clearer what is happening here and what behaviour we actually want. Eg `assert [draft['serviceName'] for draft in drafts] == ['V service', 'S service', 'R service', 'I service', 'E service', 'E service', 'C service']`

app/main/views/drafts.py
```diff
@@ -39,21 +40,44 @@ def copy_draft_service_from_existing_service(service_id):
...
+    if draft_service and draft_service.framework.id == target_framework_id:
         abort(400, "Draft already exists for service {}".format(service_id))
 
-    draft = DraftService.from_service(service)
+    draft = DraftService.from_service(
+        service,
+        questions_to_copy=questions_to_copy,
+        target_framework_id=target_framework_id,
+    )
+    service.copied_to_following_framework = True
```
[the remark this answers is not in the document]
Wynndow: Yep. I'll add the check

app/models/main.py
Wynndow: Cool. I've added some logic, and also some tests to make sure this is doing what it should.

tests/main/views/test_drafts.py
Wynndow: I think it's nice having this defined like this, as it saves the boilerplate of writing out the containing dict for the list. I take your point about the name though. I'll change it to be more explicit. Something like `basic_questions_json` to be consistent with the other fixtures defined just above.

tests/main/views/test_drafts.py
Wynndow: Yeah, it felt a bit funky when I was writing it. I like your suggestions. Have updated, with a docstring to be really clear.

Wynndow: OK, I remember why I did it this way. Having the default for the kwarg as `None`, which then means that the `basic_questions_json` is used, makes it difficult to test the case when I actually want to test the case when `questions_to_copy` isn't supplied at all.

I've added a special case for when `override_questions_to_copy` is set to 'No questions' so that the `questionsToCopy` key can be skipped from the request data completely. Have explained in the docstring. This _might_  be a terrible idea. Probably is. Interested to know what you think.

tests/main/views/test_drafts.py
Wynndow: Have added the extra test

Wynndow:
@samuelhwilliams All done with this for now, let me know what you think.

tests/models/test_main.py
```diff
@@ -1614,6 +1614,102 @@ def test_update_from_json(self):
...
+        assert draft_service.supplier == service.supplier
+        assert draft_service.status == 'not-submitted'
+        assert draft_service.data == {
+            'serviceName': 'Service 1000000000',
+            'questionOne': 'Question one',
+            'questionThree': 'Question three',
+        }
+
+    @pytest.mark.parametrize(
+        ['questions_to_copy', 'should_raise'],
```
samuelhwilliams: Super pedantic, but this should be a tuple.

tests/main/views/test_drafts.py
```diff
@@ -56,6 +62,500 @@ def service_count(self):
...
+        :param target_framework_slug: The frameworks slug for where the drafts are being created.
+        :return: The json response from the api
+        """
+        data = {
+            **self.updater_json,
+            "sourceFrameworkSlug": source_framework_slug,
+            "supplierId": 1,
+        }
+
+        if override_questions_to_copy == 'No questions':
```
samuelhwilliams: It's still a bit dirty, but probably I'd be more comfortable having the special case of an empty list being passed in rather than a magic string. So we can do `if override_questions_to_copy == [] / elif override_questions_to_copy / else`

tests/models/test_main.py
Wynndow: You were supposed to stop working...

tests/main/views/test_drafts.py
Wynndow: Works for me.

tests/models/test_main.py
samuelhwilliams: So were you.

Wynndow: I got an email from github. Too tempting to ignore.

PR #774, opened 2018-04-21
Use `update_supplier_framework` for updating supplier frameworks

Wynndow:
We were using the `supplier_update` audit event previously which was
confusing. A new audit event has been created in the api client, which
is being pulled in in this commit.

We're also storing the `SupplierFramework` object in the audit event
instead of the supplier object.


Wynndow:
Yeah, I see your point regarding using the supplier object instead of the supplier_framework. Makes a lot of sense so I'll revert this back to using the supplier.

Wynndow:
Closing this PR until after G10 applications close.

PR #776, opened 2018-04-24
Store framework dates in DB

samuelhwilliams:
## Summary
At the moment we have strings describing dates for important events in
the framework lifecycle stored in the dm-frameworks repository in a YAML
file. We also duplicate one of these dates on the framework records in
the API (frameworkCloseDate). This leads to some duplication and the
possibility for these to get out of sync, or for one or the other to not
get populated (which indeed happened recently for G-Cloud 9).

We will store the dates in UTC, so changing it to a local timezone
format will be a job for the frontends. This feels like the safest
approach but does mean we need to be considerate around setting dates
during British Summer Time as we'll need to remove an hour from the
time that we want to save. I've named the new fields with a `UTC`
suffix to make this apparent, and temporarily duplicated the
`applicationCloseDate` field under `applicationCloseDateUTC` to be
consistent.

By storing these dates solely on the API, we can return to a single
source of truth and by storing them as datetimestamps rather than
strings, we can easily convert them into whatever format we need on the
frontend. It also leads us towards being able to submit these days e.g.
via the admin frontend, leading to management of more framework
configuration by non-devs.

We will remove `dates.yml` files eventually, as the fields added here
replace all of the dates we store. See:
https://github.com/alphagov/digitalmarketplace-frameworks/blob/e1f8e5c4a1a98d817e1bc9ff90ace85c1a0762c6/frameworks/g-cloud-10/messages/dates.yml

Follow-up PRs intend to make these non-nullable fields. We can enter
provisional dates before we pin down a timetable, as we currently do in
the frameworks repository, but as things in the frontend depend on dates
being available, it feels far safer to require that *some* date is
provided rather than not.


app/models/main.py
```diff
@@ -125,7 +125,12 @@ class Framework(db.Model):
         backref='frameworks'
     )
     allow_declaration_reuse = db.Column(db.Boolean, nullable=False, default=False)
+
     application_close_date = db.Column(db.DateTime, nullable=True)
+    intention_to_award_date = db.Column(db.DateTime, nullable=True)
+    clarifications_close_date = db.Column(db.DateTime, nullable=True)
+    clarifications_publish_date = db.Column(db.DateTime, nullable=True)
+    framework_live_date = db.Column(db.DateTime, nullable=True)
```
galund: If these really will be UTC then an `_utc` postfix on the column name is a good idea, as a reminder that something needs to happen before it's displayed to users.

galund: (And probably use `_time` rather than `_date`.)

samuelhwilliams: I've opted for `_at` rather than `_time`, as the latter feels like it could equally not carry date information - and `_datetime` feels a bit excessive.

app/models/main.py
```diff
@@ -125,7 +125,12 @@ class Framework(db.Model):
         backref='frameworks'
     )
     allow_declaration_reuse = db.Column(db.Boolean, nullable=False, default=False)
+
     application_close_date = db.Column(db.DateTime, nullable=True)
```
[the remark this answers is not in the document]
samuelhwilliams: It can, but it will need to be done in a few stages to do it cleanly so I'm going to think about that one a little later.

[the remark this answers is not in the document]
Wynndow: Could you create a duplicate column `applications_close_at_utc` which is a mirror of `application_close_date `. Then make the switch and drop the old column?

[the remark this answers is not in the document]
samuelhwilliams: Yep, that's basically the ticket.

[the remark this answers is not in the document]
Wynndow: Not worth adding that new column in this migration?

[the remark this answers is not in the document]
samuelhwilliams: It probably is, but this is still WIP. 😛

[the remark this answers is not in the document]
Wynndow: ah. I'll come back later!

samuelhwilliams:
Should be good to review now @Wynndow

app/models/main.py
```diff
@@ -145,6 +151,21 @@ def serialize(self):
             'applicationCloseDate': (
                 self.application_close_date and self.application_close_date.strftime(DATETIME_FORMAT)
             ),
+            'applicationsCloseAtUTC': (
```
galund: I'm not sure it makes sense to expose this if it won't actually have been filled out yet?

migrations/versions/1140_add_framework_lifecycle_dates.py
```diff
@@ -0,0 +1,36 @@
...
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('frameworks', 'intention_to_award_at_utc')
+    op.drop_column('frameworks', 'framework_live_at_utc')
+    op.drop_column('frameworks', 'clarifications_publish_at_utc')
+    op.drop_column('frameworks', 'clarifications_close_at_utc')
+    op.drop_column('frameworks', 'applications_close_at_utc')
+    # ### end Alembic commands ###
+
```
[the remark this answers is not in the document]
samuelhwilliams: I've got a script that will set all of the dates at once. It doesn't feel particularly important to mirror it at column created time, but if you think there's a good reason I'm all ears.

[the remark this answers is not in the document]
Wynndow: No good reason other that it's doable really.

app/models/main.py
samuelhwilliams: I've got a script to inject the dates when this PR gets merged. If I don't add these to the serialization now, I'll have to do a separate PR to do it later wouldn't I? Seems as easy to add it now as later.

samuelhwilliams: Are we talking theoretically or practically? The only thing that will be posting to the `AtUTC` variant before I tear out the old one will be my script to insert the dates. Given that I plan to rip the old one out quickly, I don't think it adds much ambiguity in reality.

Also the script just makes 20 or so posts to the API, so it will take like a few seconds. Don't think we need to worry about the impact there.

galund: I think I'm just saying that tearing out the old field could be a PR with a migration only, and all the code changes to ignore that field could potentially be here.

And I realised after I pressed send that this table is probably the least likely one to need to worry about performance impact :-)

Wynndow:
I just noticed there's nothing testing that the new dates are returned in the serialization with the expected format, `DATETIME_FORMAT `.

Other than that looks good to me.

samuelhwilliams:
Added a test for framework serialization.

tests/main/views/test_frameworks.py
```diff
@@ -244,15 +250,27 @@ def setup(self):
             'clarificationQuestionsOpen': False,
             'lots': ['saas', 'paas', 'iaas', 'scs'],
             'applicationCloseDate': '2023-04-11T16:00:00.000000Z',
+            'applicationsCloseAtUTC': '2023-04-11T16:00:00.000000Z',
+            'intentionToAwardAtUTC': '2023-04-25T00:00:00.000000Z',
+            'clarificationsCloseAtUTC': '2023-03-30T17:00:00.000000Z',
+            'clarificationsPublishAtUTC': '2023-04-04T17:00:00.000000Z',
+            'frameworkLiveAtUTC': '2023-05-01T00:00:00.000000Z',
+            'frameworkExpiresAtUTC': '2000-06-06T00:00:00.000000Z',
```
Wynndow: Slightly irrelevant, but did you mean to put `'2023-06-06T00:00:00.000000Z'`

samuelhwilliams: Ha. I knew you were going to spot this.

samuelhwilliams: Updated.

Wynndow:
I'm giving this PR my third thumbs up. 👍

PR #778, opened 2018-05-03
Use flush, not commit

samuelhwilliams:
## Summary
Just flushing the session rather than committing is enough to generate
an id to use for the Audit Event, and we don't create two transactions
that risks possibly locking the search without creating an associated
audit event.


risicle:
I'm getting these same failures on master.

samuelhwilliams:
Something super flaky has gone in, I think. They pass on my local machine, and I think if I re-run the Travis tests, they will pass. Will find out...

risicle:
i am bisecting

PR #779, opened 2018-05-09
Non-nullable framework dates

samuelhwilliams:
## Summary
We now store framework lifecycle datetimes on the API. These are values
that the frontends should be able to rely on existing, so we will set
some sane (but obviously-wrong) default values and make the columns
not-nullable. All existing frameworks have these dates populated.

We also remove the old `application_close_date` column that has been
replaced with a consistently-named equivalent.

We've had to do slightly interesting migrations here to satisfy our unit
tests, which currently rely on some data (mainly frameworks) provided by
historical migrations. Migration 1150 will affect only data in the test
database and adds default dates for frameworks where dates are not
already defined (g4-g7 and dos). Migration 1160 can then set the date
fields as not nullable as they now have values. On live systems, all of
the frameworks have dates assigned to them already, so migration 1150
will have no effect and migration 1160 can continue as normal to set the
columns as non-nullable.

 ## Ticket
https://trello.com/c/oiq4Z5ht/81-make-framework-model-lifecycle-datetimestamps-not-nullable


tests/conftest.py
```diff
@@ -217,31 +215,67 @@ def _framework_fixture_inner(request, app, slug, **kwargs):
     "slug": "g-cloud-8",
     "framework": "g-cloud",
     "framework_agreement_details": _generic_framework_agreement_details,
+    "applications_close_at_utc": "2000-01-01T00:00:00.000000Z",
+    "intention_to_award_at_utc": "2000-01-01T00:00:00.000000Z",
+    "clarifications_close_at_utc": "2000-01-01T00:00:00.000000Z",
+    "clarifications_publish_at_utc": "2000-01-01T00:00:00.000000Z",
+    "framework_live_at_utc": "2000-01-01T00:00:00.000000Z",
+    "framework_expires_at_utc": "2000-01-01T00:00:00.000000Z",
```
katstevens: Nice idea! Much better than a conditional migration 😸 I'll try running this locally and see how it behaves.

samuelhwilliams: Thanks! 👍 Yeah it seems to work for me. Let me know. :)

It doesn't, of course, get rid of our problem of relying on migration data in tests. But that feels like a separate (and bigger) issue/

Wynndow:
Nice. Could the migrations be combined into one?

tests/conftest.py
samuelhwilliams: Oops, I read this as the wrong thing.

We are actually still doing a conditional migration in 1150. Any framework that doesn't have dates set will get them assigned.

This stuff in conftest didn't work and I should probably remove it.

samuelhwilliams: Oops, I read this as the wrong thing.

We are actually still doing a conditional migration in 1150. Any framework that doesn't have dates set will get them assigned.

This stuff in conftest didn't work and I should probably remove it.

Wynndow:
No, it was just a tidyness thing. I was trying to think of reasons to have the separate but I couldn't really. Why will this lock the entire table, and how would splitting it more rectify it?

Wynndow:
1150 does that, but 1160 does a bunch of smaller transactions, i think. Is each `op.something()` a separate transaction? If so, you could tack 1160 onto the end of 1150 and it wouldn't make much difference?

risicle:
I *assume* (haven't checked) that migrations are always done in **single** transactions. If you consider the alternative it wouldn't make sense - what "migration version" would alembic be leaving the database in if `1160` failed? `1155`? Doesn't work like that. On next attempt to run migration `1160` it would want to try and run the whole thing again and the bits which had been committed would cause clashes (by name alone) and cause it to always fail.

samuelhwilliams:
Yeah I believe @risicle is correct - each migration is one transaction.

tests/conftest.py
samuelhwilliams: Except I won't remove it as it needs to be there if for frameworks getting created. So it's fine, but not the solution to our original problem :)

samuelhwilliams:
Happy with this as-is @Wynndow?

I don't think there's much point splitting the migrations out further and I'm pretty OK with them being two separate ones, for semantic as well as transactional reasons.

PR #780, opened 2018-05-09
Order services in tests

samuelhwilliams:
## Summary
We've had a number of intermittent unit test failures locally and on
travis due to results being returned from postgres in a non-reproducible
order. So now we'll order them consistently which should prevent these
occasional/flaky test failures.

Failure example: https://travis-ci.org/alphagov/digitalmarketplace-api/builds/374924411

### DIFF
--- a/tests/main/views/test_drafts.py
+++ b/tests/main/views/test_drafts.py
@@ -365,6 +365,8 @@ class TestCopyPublishedFromFramework(DraftsHelpersMixin):
             Service.supplier_id == 1,
             Service.framework_id == 1,  # G-Cloud 6
             Service.lot_id == 2,
+        ).order_by(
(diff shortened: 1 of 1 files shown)


tests/main/views/test_drafts.py
```diff
@@ -435,6 +437,8 @@ def test_should_only_copy_published_services_from_correct_framework_and_lot(self
             Service.framework_id == 1,  # G-Cloud 6
             Service.lot_id == 2,
             Service.status == 'published'
+        ).order_by(
+            desc(Service.data['serviceName'].astext)
```
risicle: You sure this is "unique" enough?

Wynndow: The service names are generated from the service's ID so should be.

PR #783, opened 2018-05-17
Add audit event for failed login

samuelhwilliams:
## Summary
We have an audit event type defined for `user_auth_failed` but this
doesn't seem to have been added to the appropriate API endpoint. We've
had a support query come in that we are unable to properly investigate
due to being unable to tie up login attempts with requests. This now
adds an audit event on failed login which will store the email address
of the account, the client IP making the request, the hashed password
and the number of failed login attempts currently on the account (after
adding one for the current requests).

 ## Ticket
https://trello.com/c/iSAzWgQk/466

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -40,6 +40,7 @@ def auth_user():
 
     if user is None:
         return jsonify(authorization=False), 404
+
     elif encryption.authenticate_user(json_payload['password'], user) and user.active:
         user.logged_in_at = datetime.utcnow()
         user.failed_login_count = 0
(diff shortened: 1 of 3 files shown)


risicle:
Another thing that would help would be binding the traceid to the `AuditEvent`

samuelhwilliams:
Have just realised that including a freshly-hashed password is pointless as it'll have a different salt and be completely incomparable. Do we think it's worth including a hashed version of the password using the same salt as set for the current user's password, or just drop this from the audit event altogether? I suspect it's probably not of much use, so maybe best to just not insert it.

benvand:
I think this is _badly_ non compliant with GDPR. We'd be collecting and storing personal data without _any_ defensible lawful basis.
https://ico.org.uk/for-organisations/guide-to-the-general-data-protection-regulation-gdpr/lawful-basis-for-processing/
Let me have a think and see what I can come up with.

samuelhwilliams:
Thanks for the chat offline.

This PR will only store audit events for users with accounts, which means they've agreed to our terms & conditions. As we've both concluded, this should cover us in terms of GDPR.

Let me know if you have any other issues with this, though :)

app/main/views/users.py
```diff
@@ -40,16 +40,35 @@ def auth_user():
...
         user.failed_login_count = 0
         db.session.add(user)
         db.session.commit()
 
         return single_result_response(RESOURCE_NAME, user), 200
+
     else:
         user.failed_login_count += 1
         db.session.add(user)
+        db.session.flush()
```
benvand: Could bulk add these I think but no biggie https://stackoverflow.com/a/31205155

samuelhwilliams: We don't seem to use bulk inserts very widely across the API at the moment, and I'm not overly concerned by the performance loss of two separate INSERTs, but it's a fair point to make.

PR #784, opened 2018-05-18
Add framework components to model

samuelhwilliams:
## Summary
We currently enable certain pages/journeys based on the name of the
framework (e.g. catalogue search for 'g-cloud' and competition for
'digital-outcomes-and-specialists'). This doesn't scale well and, based
on our current understanding, isn't really reflective of how frameworks
are composed.

This adds two booleans to the framework table: `has_direct_award` and
`has_further_competition`. We will set `has_direct_award` to True for
G-Cloud frameworks and `has_further_competition` to True for Digital
Outcomes and Specialists frameworks. We can then gate journeys based on
the values of these booleans. Based on our current understanding, a
framework could theoretically have both elements, allowing both
journeys to be used.

The migration for this initial PR will make the fields non-nullable and
give them a default of False. We will POST to the API once this is in to
assign the correct states. After that is done, a further migration will
add a constraint enforcing that at least one of the two components is
present on a framework record.

 ## Ticket
https://trello.com/c/WruHrcBf/456


tests/main/views/test_frameworks.py
```diff
@@ -161,6 +167,18 @@ def test_create_fails_if_slug_is_invalid(self):
...
+        framework = self.framework(slug="this is/invalid")
+        framework['frameworks']['hasDirectAward'] = False
+        framework['frameworks']['hasFurtherCompetition'] = False
+
+        response = self.client.post("/frameworks",
+                                    data=json.dumps(framework),
+                                    content_type="application/json")
+
+        assert response.status_code == 400
+        assert json.loads(response.get_data(as_text=True))["error"] == "Invalid slug value 'this is/invalid'"
```
katstevens: I think we're testing the wrong validation error here 😸

samuelhwilliams: 😂 

Concentrate Sam!

samuelhwilliams: Fixed 👍

tests/models/test_main.py
```diff
@@ -215,6 +219,7 @@ def test_framework_serialization_with_default_datetimes(self):
             slug='foo-109',
             framework='g-cloud',
             status='open',
+            has_further_competition=True,
```
katstevens: In the previous test we provide both args (with `True` and `False`) but here it's just one - not fussed which way but we should probably be consistent.

migrations/versions/1190_insert_framework_components_for_direct_.py
```diff
@@ -0,0 +1,32 @@
+"""Insert framework components for direct award and further competition
```
katstevens: Did this file name get truncated?

app/models/main.py
```diff
@@ -232,6 +237,16 @@ def validates_clarification_questions_open(self, key, value):
     def __repr__(self):
         return '<{}: {} slug={}>'.format(self.__class__.__name__, self.name, self.slug)
 
+    @validates('has_direct_award', 'has_further_competition')
+    def validates_framework_components(self, key, value):
+        has_direct_award = value if key == 'has_direct_award' else self.has_direct_award
+        has_further_competition = value if key == 'has_further_competition' else self.has_further_competition
+
+        if has_direct_award is False and has_further_competition is False:
+            raise ValidationError('At least one of `has_direct_award` and `has_further_competition` must be True.')
```
galund: These error messages talk about field names, but it seems from the tests that they make it through into potential API responses. In which case they are referring to the wrong field names - they would need to use the camelCased versions.

Not sure of the right solution here, maybe it has been done differently elsewhere tho?

migrations/versions/1190_insert_framework_components_for_direct_.py
samuelhwilliams: It did indeed.

tests/models/test_main.py
samuelhwilliams: I purposely left the latter out actually, to also test the default value as a little extra. Maybe I could do this more explicitly?

katstevens: Up to you really!

samuelhwilliams:
@galund @katstevens comments addressed 👍

app/models/main.py
```diff
@@ -232,6 +238,16 @@ def validates_clarification_questions_open(self, key, value):
     def __repr__(self):
         return '<{}: {} slug={}>'.format(self.__class__.__name__, self.name, self.slug)
 
+    @validates('has_direct_award', 'has_further_competition')
+    def validates_framework_components(self, key, value):
+        has_direct_award = value if key == 'has_direct_award' else self.has_direct_award
+        has_further_competition = value if key == 'has_further_competition' else self.has_further_competition
+
+        if has_direct_award is False and has_further_competition is False:
```
galund: The trouble with using `is False` here, is that `None` here is getting treated as `True`. Which - at database layer anyway - we aren't enforcing.

tests/main/views/test_frameworks.py
```diff
@@ -161,6 +167,31 @@ def test_create_fails_if_slug_is_invalid(self):
         assert response.status_code == 400
         assert json.loads(response.get_data(as_text=True))["error"] == "Invalid slug value 'this is/invalid'"
 
+    def test_create_fails_if_direct_award_and_further_competition_false(self):
+        framework = self.framework()
+        framework['frameworks']['hasDirectAward'] = False
+        framework['frameworks']['hasFurtherCompetition'] = False
```
katstevens: Related to George's comment above - could we test these with initial `None` values as well, to simulate the pre-default database state?

app/models/main.py
samuelhwilliams: This is required in order to set any values whatsoever, as otherwise the moment we update one of the values, validation will fail because the other is still None. We don't even technically need this validation as it's just an interim check that will last all of a couple of days.

tests/main/views/test_frameworks.py
samuelhwilliams: The `None` case will only exist for an interim period until the follow-up PR can make the column  non-nullable. Given it's a small table and nothing new will be going into it between now and (basically) 12 months from now, I don't think it's so important to test this case. Maybe?

katstevens: Hmm you're right, I guess it's probably not vital for this table.

galund: Yep if we're definitely going back and making this non-nullable then that's fine.

But in that case I'd expected to see a data migration that set values for any existing fields? (Wouldn't I? Maybe not - I understand we have a policy of not really doing data migrations, but I don't follow entirely why not having them would be better, especially in cases like this where if the UPDATEs aren't done, the migration to set the field to not-nullable will just fail.)

Anyhoo, happy to approve, my not understanding our migrations policy definitely isn't a blocker :-)

samuelhwilliams: Something like this? ;) https://github.com/alphagov/digitalmarketplace-api/pull/785/files#diff-3b681bde687c7d636bfdb6a9d03d6f14

PR #785, opened 2018-05-18
Enforce framework component constraint

samuelhwilliams:
## Summary
Enforces a constraint of the Framework model to ensure that each
framework record has at least one of the two components: direct award or
further competition. In theory a framework can have both, but the
Marketplace only supports these two frameworks so it must have at least
one of them or we will have no features to enable for it.

 ## Ticket
https://trello.com/c/WruHrcBf/456

 ## Note
Should be merged/deployed after https://github.com/alphagov/digitalmarketplace-api/pull/784 and after setting `has_direct_award`/`has_further_competition` on all existing frameworks.


samuelhwilliams:
@risicle so apparently Postgres isn't actually _applying_ the constraint during the migration and is just silently ignoring it... 😕

migrations/versions/2000_non_nullable_framework_components_and_.py
```diff
@@ -0,0 +1,49 @@
...
+    op.execute(frameworks.update().where(frameworks.c.framework == 'digital-outcomes-and-specialists').values(
+        has_direct_award=False,
+        has_further_competition=True,
+    ))
+
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('frameworks', 'has_direct_award', existing_type=sa.BOOLEAN(), nullable=False)
+    op.alter_column('frameworks', 'has_further_competition', existing_type=sa.BOOLEAN(), nullable=False)
+    sa.CheckConstraint('has_direct_award IS true OR has_further_competition IS true',
+                       name=op.f('ck_framework_has_direct_award_or_further_competition')),
```
risicle: ah but you see `sa.CheckConstraint(...)` isn't a *command* - it doesn't do anything by itself. Note the uses of it in my migration pass it as an argument to `op.create_table(...)`, and how `sa.CheckConstraint(...)` doesn't actually mention the target table anywhere, so couldn't work on its own.

samuelhwilliams: 🤦‍♂️ 

Yep that'll do it!

risicle:
(I thought process type was more of a property of the `Lot` than `Framework`?)

samuelhwilliams:
Interesting. I spoke to Louise briefly and that point wasn't raised.

I haven't implemented any checks on the frontends yet, so I guess I could change this all to add the property to Lots rather than Frameworks... But it also may be more effort than it's immediately worth, given we haven't got an immediate use for it on Lots or a known case for when it will be needed.

samuelhwilliams:
Actually I think I'm going to leave the PR as-is for now and make a note that it might need reviewing in the future. Our pages would need rearchitecting quite a bit, I feel, in order to have the 'process type' be related to the `Lot`, so for now I think `Framework` is the suitable place to hold it.

migrations/versions/1200_non_nullable_framework_components_and_.py
```diff
@@ -0,0 +1,50 @@
...
+    # Need to manually insert values for these columns so that the test database gets populated corrected.
+    # Live environments will be populated manually via the API before this migration goes in.
+    op.execute(frameworks.update().where(frameworks.c.framework == 'g-cloud').values(
+        has_direct_award=True,
+        has_further_competition=False,
+    ))
+    op.execute(frameworks.update().where(frameworks.c.framework == 'digital-outcomes-and-specialists').values(
+        has_direct_award=False,
+        has_further_competition=True,
+    ))
```
galund: As a general practice it's better to keep data migrations entirely separate to schema migrations, but it probably matters less in postgres than it did in MySQL, and I can't see any actual problems from these staying as they are.

samuelhwilliams:
Let's discuss.

PR #786, opened 2018-05-20
Disable user accounts on Notify callback

samuelhwilliams:
## Summary
When we send an email through Notify and Notify tells us that the email
address no longer exists, we should make sure we don't keep trying to
send that email. We will do this by deactivating the account, which
should mean the address is filtered out of email lists. If the mailbox
doesn't exist, we think it's reasonable to assert that the user is no
longer working on behalf of the organisation and therefore restricting
their account is a fair and appropriate action to take. Support will
remain able to re-activate the account if this turns out to be an
invalid assumption.

 ## Ticket
https://trello.com/c/vOAUDwnL/446

### DIFF
--- a/app/callbacks/__init__.py
+++ b/app/callbacks/__init__.py
@@ -10,4 +10,4 @@ callback_authentication_partial = partial(requires_authentication, module='callb
 callbacks.before_request(callback_authentication_partial)
 
 
-from .views import notify  # noqa
+from .views import notify
--- a/app/callbacks/__init__.py
+++ b/app/callbacks/__init__.py
@@ -1,0 +1,13 @@
+from functools import partial
+
+from flask import Blueprint
+
+from app.authentication import requires_authentication
+
+callbacks = Blueprint('callbacks', __name__, url_prefix='/callbacks')
+
+callback_authentication_partial = partial(requires_authentication, module='callbacks')
+callbacks.before_request(callback_authentication_partial)
+
+
+from .views import notify  # noqa
--- a/app/callbacks/views/notify.py
+++ b/app/callbacks/views/notify.py
@@ -19,11 +19,11 @@ def callbacks_root():
 def notify_callback():
     notify_data = get_json_from_request()
 
-    user = User.query.filter(
-        User.email_address == notify_data['to']
-    ).first()
-
     if notify_data['status'] == 'permanent-failure':
+        user = User.query.filter(
+            User.email_address == notify_data['to']
+        ).first()
+
         if user and user.active:
             user.active = False
             db.session.add(user)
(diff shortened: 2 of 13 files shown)


config.py
```diff
@@ -63,6 +64,7 @@ class Test(Config):
     ES_ENABLED = False
     SQLALCHEMY_DATABASE_URI = 'postgresql://localhost/digitalmarketplace_test'
     DM_API_AUTH_TOKENS = 'myToken'
+    DM_API_CALLBACK_AUTH_TOKENS = 'myCallbackToken'
```
lfdebrux: In Development the token is 'myToken' as for DM_API_AUTH_TOKENS, does it need to be different here?

samuelhwilliams: Nah. I think it'll be easier if the token is the same for developers so they don't have to remember two different tokens. But for tests, we need to differentiate in order to assert that each kind of token only works for its part of the API.

I'm happy to update the dev token to `myCallbackToken` if you think it's useful for devs to understand and always be aware that these are two different tokens.

app/callbacks/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+
+from app.authentication import requires_authentication
+
+callbacks = Blueprint('callbacks', __name__, url_prefix='/callbacks')
+
+callback_authentication_partial = partial(requires_authentication, module='callbacks')
+callbacks.before_request(callback_authentication_partial)
+
+
+from .views import notify  # noqa
```
benvand: Can we add `app/callbacks/__init__.py : E402` to `.flake8` instead of `#noqa` for the full line or is the `#noqa` because it triggers both an `E402` and an `E401`?

`E402 module level import not at top of file`
`F401 'notify' imported but unused`

benvand: Also we could glob match `__init__.py` files for F401 🤷‍♂️

app/callbacks/views/notify.py
```diff
@@ -0,0 +1,47 @@
...
+                data=notify_data,
+                db_object=user,
+            )
+            db.session.add(audit_event)
+
+            db.session.commit()
+
+    elif notify_data['status'] == 'technical-failure':
+        current_app.logger.error("Notify failed to deliver {reference} to {hashed_email}".format(
+            reference=notify_data['reference'],
```
benvand: Warn vs Error

app/main/__init__.py
```diff
@@ -14,5 +14,5 @@ def add_cache_control(response):
 
 
 from .views import suppliers, services, users, drafts, audits, frameworks, briefs, brief_responses, agreements,\
```
benvand: Specific exceptions rules

tests/callbacks/views/test_notify.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def setup(self):
+        super().setup()
+        self.app.wsgi_app = self.wsgi_app_callbacks
+
+    def test_200_on_notify_callback_root(self):
+        response = self.client.get('/callbacks')
+        assert response.status_code == 200
+
+    def test_user_account_deactivated_on_permanent_delivery_failure(self, user_role_supplier):
```
benvand: Chuck in user object

PR #789, opened 2018-05-25
Update 'export_users_for_framework' to include additional fields

CrystalPea:
Part of work for this trello card: https://trello.com/c/HR95DMjV/178-update-framework-manager-user-download-list-to-include-additional-fields

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -14,8 +14,6 @@ from ...models import (
     BuyerEmailDomain,
     ContactInformation,
     Framework,
-    FrameworkLot,
-    Lot,
     Service,
     Supplier,
     SupplierFramework,
@@ -294,8 +292,8 @@ def export_users_for_framework(framework_slug):
     users_for_each_supplier = {}
     for u in users:
         user_obj = {
-            'email address': u.email_address,
-            'user_name': u.name,
+            'email_address': u.email_address,
+            'name': u.name,
             'user_research_opted_in': u.user_research_opted_in,
         }
         # If supplier id key already exists, append the user to that supplier's list
@@ -367,18 +365,21 @@ def export_users_for_framework(framework_slug):
                 variations_agreed = ', '.join(sf.agreed_variations.keys()) if sf.agreed_variations else ''
 
             supplier_rows.append({
-                'users': users_for_each_supplier[supplier.supplier_id],
                 'supplier_id': supplier.supplier_id,
-                'declaration_status': declaration_status,
-                'application_status': application_status,
-                'framework_agreement': framework_agreement,
-                'application_result': application_result,
-                'variations_agreed': variations_agreed,
                 'supplier_name': supplier.name,
                 'supplier_organisation_size': supplier.organisation_size,
                 'duns_number': supplier.duns_number,
                 'registered_name': supplier.registered_name,
                 'companies_house_number': supplier.companies_house_number,
+                'application_result': application_result,
+                'application_status': application_status,
+                'declaration_status': declaration_status,
+                'framework_agreement': framework_agreement,
+                'variations_agreed': variations_agreed,
+                "published_services_count": {
+                    lot_slugs_by_id[lot_id]: service_counts_by_lot_by_supplier[supplier.supplier_id].get(lot_id, 0)
+                    for lot_id in lot_slugs_by_id.keys()
+                },
                 "contact_information": {
                     'contact_name': ci.contact_name,
                     'contact_email': ci.email,
@@ -388,10 +389,7 @@ def export_users_for_framework(framework_slug):
                     'address_postcode': ci.postcode,
                     'address_country': supplier.registration_country,
                 },
-                "published_services_count": {
-                    lot_slugs_by_id[lot_id]: service_counts_by_lot_by_supplier[supplier.supplier_id].get(lot_id, 0)
-                    for lot_id in lot_slugs_by_id.keys()
-                }
+                'users': users_for_each_supplier[supplier.supplier_id],
             })
 
     return jsonify(suppliers=[supplier for supplier in supplier_rows]), 200
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -331,7 +331,7 @@ def export_users_for_framework(framework_slug):
         User.id,
     ).all()
 
-    user_rows = []
+    supplier_rows = []
 
     for sf, u, ci in supplier_frameworks_and_users:
 
@@ -353,10 +353,14 @@ def export_users_for_framework(framework_slug):
             framework_agreement = bool(getattr(sf.current_framework_agreement, 'signed_agreement_returned_at', None))
             variations_agreed = ', '.join(sf.agreed_variations.keys()) if sf.agreed_variations else ''
 
-        user_rows.append({
-            'email address': u.email_address,
-            'user_name': u.name,
-            'user_research_opted_in': u.user_research_opted_in,
+        supplier_rows.append({
+            "users": [
+                {
+                    'email address': u.email_address,
+                    'user_name': u.name,
+                    'user_research_opted_in': u.user_research_opted_in,
+                }
+            ],
             'supplier_id': sf.supplier_id,
             'declaration_status': declaration_status,
             'application_status': application_status,
@@ -369,21 +373,24 @@ def export_users_for_framework(framework_slug):
             'duns_number': u.supplier.duns_number,
             'registered_name': u.supplier.registered_name,
             'companies_house_number': u.supplier.companies_house_number,
-            'contact_name': ci.contact_name,
-            'contact_email': ci.email,
-            'contact_phone_number': ci.phone_number,
-            'address_first_line': ci.address1,
-            'address_city': ci.city,
-            'address_postcode': ci.postcode,
-            'address_country': u.supplier.registration_country,
+            "contact_information": {
+                'contact_name': ci.contact_name,
+                'contact_email': ci.email,
+                'contact_phone_number': ci.phone_number,
+                'address_first_line': ci.address1,
+                'address_city': ci.city,
+                'address_postcode': ci.postcode,
+                'address_country': u.supplier.registration_country,
+            },
+            "published_services_count": {
+                "digital-outcomes": 0,
+                "digital-specialists": 0,
+                "user-research-studios": 3,
+                "user-research-participants": 0,
+            }
         })
-        for lot, supplier_services in published_service_count_by_supplier_and_lot.items():
-            if sf.supplier_id in supplier_services:
-                services_count = supplier_services[sf.supplier_id]
-            else:
-                services_count = 0
-            user_rows[-1][lot] = services_count
-    return jsonify(users=[user for user in user_rows]), 200
+
+    return jsonify(suppliers=[supplier for supplier in supplier_rows]), 200
 
 
 @main.route("/users/check-buyer-email", methods=["GET"])
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -286,6 +286,14 @@ def export_users_for_framework(framework_slug):
         User.supplier_id, User.name, User.email_address, User.user_research_opted_in
     ).filter(User.supplier_id.in_(suppliers_with_a_complete_service)).filter(User.active.is_(True))
 
+    users_by_supplier = {
+        u.supplier_id: {
+            'email address': u.email_address,
+            'user_name': u.name,
+            'user_research_opted_in': u.user_research_opted_in,
+        } for u in users
+    }
+
     lots = db.session.query(FrameworkLot.lot_id, Lot.slug).join(Lot, FrameworkLot.lot_id == Lot.id).filter(
         FrameworkLot.framework_id == framework.id
     ).distinct().all()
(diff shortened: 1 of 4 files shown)


app/main/views/users.py
```diff
@@ -271,24 +282,44 @@ def export_users_for_framework(framework_slug):
...
-    ).filter(
-        Service.status == 'published',
-        Service.framework_id == framework.id
-    ).group_by(
-        Service.supplier_id
-    ).all())
+    lots_ids = db.session.query(FrameworkLot.lot_id).filter(FrameworkLot.framework_id == framework.id).all()
+    lots = db.session.query(Lot).filter(Lot.id.in_(lots_ids)).all()
+    published_service_count_by_supplier_and_lot = {}
+
```
benvand: These 2 queries can be minimised to a single query with a join like so:
```
db.session.query(FrameworkLot.lot_id, Lot.slug).join(Lot, FrameworkLot.lot_id == Lot.id).filter(
        FrameworkLot.framework_id == 4
    ).distinct().all()
```

Joins are generally considered more performant than creating multiple queries (there are some exceptions).
https://dba.stackexchange.com/a/43018

benvand: Below you'd then have the for statement like so:

```
for lot_id, slug in lots:
    ...
```

app/main/views/users.py
```diff
@@ -271,24 +282,44 @@ def export_users_for_framework(framework_slug):
...
+        SupplierFramework, User, ContactInformation
     ).filter(
         SupplierFramework.supplier_id == User.supplier_id
     ).filter(
         SupplierFramework.framework_id == framework.id
     ).filter(
-        User.active.is_(True)
+        ContactInformation.supplier_id == User.supplier_id
+    ).filter(
+    User.active.is_(True)
```
benvand: Think you want an indent here?

app/main/views/users.py
```diff
@@ -271,24 +282,44 @@ def export_users_for_framework(framework_slug):
...
+        for supplier_id, service_count in v.items():
+            if supplier_id in supplier_id_published_service_count:
+                supplier_id_published_service_count[supplier_id] += service_count
+            else:
+                supplier_id_published_service_count[supplier_id] = service_count
+
 
     supplier_frameworks_and_users = db.session.query(
-        SupplierFramework, User
+        SupplierFramework, User, ContactInformation
```
benvand: It might be better to explicitly add `Supplier` here too. Accessing it though the attribute on user _could_ have some undesirable effects such as calling the DB once per supplier framework to find the corresponding supplier object.

This depends if the Supplier object is lazy loaded with the user. It might be fine.
Is there a specific reason you added `ContactInformation` and not `Supplier`? You should be able to do `u.supplier.contact_information.phone_number` for example (not that I'd recommend that, just curious)

app/main/views/users.py
CrystalPea: True, changing now :)

app/main/views/users.py
CrystalPea: As per our chat a few mins ago, supplier is lazy loaded on the user indeed. And just so we remember I am adding here that we can check if contact information is lazy loaded together with the supplier :).

app/main/views/users.py
```diff
@@ -271,24 +282,45 @@ def export_users_for_framework(framework_slug):
...
+        ).all())
+
+    supplier_id_published_service_count = {}
+    for k, v in published_service_count_by_supplier_and_lot.items():
+        for supplier_id, service_count in v.items():
+            if supplier_id in supplier_id_published_service_count:
+                supplier_id_published_service_count[supplier_id] += service_count
+            else:
+                supplier_id_published_service_count[supplier_id] = service_count
+
```
risicle: You're reinventing sql in slightly twisty-turny python. This is all doable in a single SQL statement.

tests/main/views/test_users.py
```diff
@@ -1411,6 +1411,22 @@ def _assert_things_about_export_response(self, row, parameters=None):
...
+            'contact_email': '1@contact.com',
+            'contact_phone_number': None,
+            'address_first_line': '7 Gem Lane',
+            'address_city': 'Cantelot',
+            'address_postcode': 'CN1A 1AA',
+            'address_country': 'country:GB',
+            'published_services_count_on_digital-outcomes_lot': 0,
+            'published_services_count_on_digital-specialists_lot': 0,
+            'published_services_count_on_user-research-studios_lot': 0,
+            'published_services_count_on_user-research-participants_lot': 0,
```
risicle: Why have we decided to flatten these counts into the top level object with the lot names string-formatted in? Why is it not

```
...
"published_services_count": {
    "digital-outcomes": 0,
    "digital-specialists": 0,
    "user-research-studios": 0,
    "user-research-participants": 0,
}
...
```

app/main/views/users.py
risicle: Whole thing more or less doable with two lines added:

```
db.session.query(
    Service.supplier_id,
    func.count(Service.id)
 ).filter(
    Service.status == 'published',
    Service.framework_id == framework.id
).group_by(
    Service.supplier_id
).all()
```

becomes

```
db.session.query(
    Service.supplier_id,
    Service.lot_id,
    func.count(Service.id)
 ).filter(
    Service.status == 'published',
    Service.framework_id == framework.id
).group_by(
    Service.supplier_id,
    Service.lot_id,
).all()
```

tests/main/views/test_users.py
CrystalPea: Yes, nested data like in your comment looks neater indeed. 

Things that would need to be explored are:
- how it would work when CSV is created in this view
- how framework managers at CCS are using this data

risicle: Mhm, it's the script/frontend's business whether it's turning it into a csv or not. If we're wed to it being output in csv-like format we should just output a csv directly.

app/main/views/users.py
risicle: Though there's actually an even neater way of doing it - let me think...

app/main/views/users.py
```diff
@@ -271,22 +282,42 @@ def export_users_for_framework(framework_slug):
...
+            Service.lot_id
+        ).all())
+
+    supplier_id_published_service_count = {}
+    for k, v in published_service_count_by_supplier_and_lot.items():
+        for supplier_id, service_count in v.items():
+            if supplier_id in supplier_id_published_service_count:
+                supplier_id_published_service_count[supplier_id] += service_count
+            else:
+                supplier_id_published_service_count[supplier_id] = service_count
```
risicle: I tell a lie - neater:
```
from itertools import groupby

lot_slugs_by_id = {lot.id: lot.slug for lot in framework.lots}

service_counts_by_lot_by_supplier = {
    supplier_id: {row[1]: row[2] for row in rows}
    for supplier_id, rows in groupby(
        db.session.query(
            Service.supplier_id,
            Service.lot_id,
            func.count(Service.id)
        ).filter(
            Service.status == 'published',
            Service.framework_id == framework.id
        ).group_by(
            Service.supplier_id,
            Service.lot_id,
        ).all(),
        key=lambda row: row[0],
    )
}
```

Then later:
```
user_rows.append({
    ...
    published_services_count = {
        lot_slugs_by_id[lot_id]: service_counts_by_lot_by_supplier[supplier_id].get(lot_id, 0)
        for lot_id in lot_slugs_by_id.keys()
    }
```

Bam, all done in ~10 additional, declarative lines.

risicle: Oh actually for backwards compatibility make the latter lines:

```
user_rows.append({
    ...
    "published_services_count": sum(service_counts_by_lot_by_supplier[supplier_id].values())
    "published_services_count_by_lot": {
        lot_slugs_by_id[lot_id]: service_counts_by_lot_by_supplier[supplier_id].get(lot_id, 0)
        for lot_id in lot_slugs_by_id.keys()
    }
})
```

tests/main/views/test_users.py
katstevens: I think the problem here is that we're trying to get a lot of (mostly duplicated!) data from one API call. The resulting CSV will display everything as per-user, but I think it'll probably be easier to split the per-supplier info (including service count per lot) out into a new API view, and then join that to each supplier user in the Admin FE (using @risicle's suggested code above). It does mean 2 API calls so it adds that overhead, but they should be smaller in size?

risicle: Um, yea? I don't really know the "points of use" of this endpoint in question so :man_shrugging: ... I guess that would also lead to a way of providing the ability to request the export in "cheap mode" without any aggregate data.

Offtopic... though it would sully the "JSON purity" of our wonderful API I think at some point we should really entertain producing csv directly from the API - csv can be produced directly from database output with near-zero memory overhead, the act of making this json uses potentially numerous MBs on the API instances simply in the dict-representation of the output, not to mention any intermediary dicts & list copies we use during preparation. Then we spend approximately that amount of memory *again* decoding the bytes back into dicts/lists on the frontend instances, manipulating them.... you get my drift. All seems like a bit of a silly dance when we start to run into problems with the size of our exports...

app/main/views/users.py
risicle: p.s. if you want me to explain how this all works and what the difference is, just ask me. Or, if it's obvious, ... don't.

app/main/views/users.py
```diff
@@ -271,71 +281,123 @@ def export_users_for_framework(framework_slug):
...
+        }
+        # If supplier id key already exists, append the user to that supplier's list
+        users_for_each_supplier.setdefault(u.supplier_id, []).append(user_obj)
+
+    lot_slugs_by_id = {lot.id: lot.slug for lot in framework.lots}
+
+    # Creates a dictionary of dictionaries, with published service counts per lot per supplier.
+    # To do this we use dictionary comprehension, SQL query and itertools' 'groupby' function.
+    # Grouping by supplier id is more efficient as this is what we will be iterating over later.
+    service_counts_by_lot_by_supplier = {
```
benvand: `groupby` is tricky. This will (I think) create an entry per supplier id - lot comination
Have a look at the output of:
```
[i for i in groupby([1,2,2,2,2,3,3,4,4,4,5,5,5,4,4,3,2,2,1,2,2,3,3,4,4,5,1,1,1,1,2,2,3,3,4,4,5,5,3,2,3,3,4,5])]
```
For an example. Is this what we want here?

risicle: Actually that made me notice that the query used here doesn't have an `order_by` set, and while I think the result of a `group_by` will almost always come out sorted by major group column first, I don't think it by law *has to*, so it would be better to add an `.order_by(Service.supplier_id)` clause here.

app/main/views/users.py
```diff
@@ -271,71 +281,123 @@ def export_users_for_framework(framework_slug):
...
+        supplier_id: {row[1]: row[2] for row in rows}
+        for supplier_id, rows in groupby(
```
benvand: Nested comprehensions like this are hard to read. This would look neated assigning the result of the query to a variable.
https://google.github.io/styleguide/pyguide.html#List_Comprehensions

benvand: At a glance it looks like you're starting a for loop here with:
```
for supplier_id, rows in groupby(
```
It's only when we get to line 325 that we realise it's a comprehension and have to navigate back to line 310 to start reading the whole thing over again.

app/main/views/users.py
```diff
@@ -271,71 +281,123 @@ def export_users_for_framework(framework_slug):
...
-            'variations_agreed': variations_agreed,
-            'published_service_count': supplier_id_published_service_count.get(sf.supplier_id, 0)
-        })
-
-    return jsonify(users=[user for user in user_rows]), 200
+    supplier_rows = []
+
+    for sf, supplier, ci in supplier_frameworks_and_users:
+        # Only export suppliers that have users
+        if users_for_each_supplier.get(supplier.supplier_id):
```
benvand: This would be a good case to use a `continue` statement. It will allow you to avoid indenting all the code after the if and it also works with the narrative. 
`if users_for_each_supplier.get(supplier.supplier_id)` is designed to stop us from exporting suppliers with no users so `if not:` `continue` works quite well.

`continue` statements are best used sparingly because when they are hidden in a big block of code they can obscure the flow but their use as a preconditional check at the top of a block is more obvious.

tests/main/views/test_users.py
```diff
@@ -1407,113 +1407,143 @@ def _setup(self, post_supplier=True, post_users=True, register_supplier_with_fra
...
         assert row['declaration_status'] == _parameters['declaration_status']
         assert row['framework_agreement'] == _parameters['framework_agreement']
-        assert row['published_service_count'] == _parameters['published_service_count']
 
-    ############################################################################################
+        assert row['published_services_count'] == _parameters['published_services_count']
+        assert row['contact_information'] == _parameters['contact_information']
+        assert row['users'] == _parameters['users']
 
+    ############################################################################################
```
benvand: I think this line can go unless it's a convention elsewhere in the codebase?

app/main/views/users.py
risicle: I don't think it at all looks like you're starting a for loop there - the simple fact that it's indented and inside braces means that it *can't* be a for loop.

As for nested comprehensions, breaking out to an additional variable requires the user to do more mental variable tracking, makes the reader wonder whether the intermediate has managed to be mutated (as it has leaked a reference to its inner working) and adds yet another thing to assign a name to (and, though in an ideal world we're all great at naming things, in reality "naming things" is one of the "hard problems")

risicle: but really, assigning the query to a variable isn't a huge change, not a big deal, won't fight tooth and nail, just think it's harder to read, making me go and link up the innards of the comprehension with where it's being used. Scrolly-uppy, scrolly-downy. Scrolly-uppy...

...and then wonder about the possibilities of whether I make a change to it whether something else is affected by it. Inside the comprehension it is *encapsulated*.

katstevens:
Thanks for the comments @risicle @benvand - we're going to rethink this story a little and close this PR. We didn't realise a script used this API endpoint as well as the admin CSV download, so to avoid unwanted side effects there, we're going to keep the existing view as is and create a new one specifically returning 'suppliers and their users' (there'll be some shared bits that we can refactor out into helpers).

risicle:
:dizzy:

katstevens:
Not going to delete the branch just yet - we'll open another PR.

PR #790, opened 2018-05-26
Preparing to remove `frameworkFramework`

samuelhwilliams:
## Summary
We want to deprecate the variable name `frameworkFramework` and replace
it with `frameworkFamily`, which is easier to understand. We need the
API to remain compatible with the frontends,  which will still expect to
find `frameworkFramework`, until we get around to updating them. So for
now, we will just add the new key `frameworkFamily` and once this goes
out we can start to transition the frontends over to using the name.

 ## Ticket
https://trello.com/c/iqFcVKpd/14


app/models/main.py
```diff
@@ -997,6 +998,7 @@ def serialize(self):
             'supplierName': self.supplier.name,
             'frameworkSlug': self.framework.slug,
             'frameworkFramework': self.framework.framework,
+            'frameworkFamily': self.framework.framework,
```
katstevens: I think we should add the `frameworkFamily` key to the `Framework.serialize()` method too. At the moment we just have `framework` in there - is the plan to eventually deprecate that field in favour of `framework_family` (as `self.framework.framework` is where all this confusion came from in the first place)?

samuelhwilliams: D'oh! Yep, that's a super important place to have it. 👍 

Will add.

PR #791, opened 2018-05-29
Add outcome model, creation and update views

risicle:
Trello https://trello.com/c/4JtM91iv/67-create-back-end-interface-for-award-flow

Here we add a `ProcessOutcome` model designed to generically represent the "outcome" (whether that be a successful outcome - an "award" - or an unsuccessful outcome) of a particular buying process. The buying process can be either a `DirectAwardProject` or a `Brief`, though in this PR only contains the creation process for a `ProcessOutcome` based on a `DirectAwardProject` and `Brief`s for now are left recording their awards in their own bespoke way. It is *intended* that eventually `Brief`s will be brought into line using `ProcessOutcome`s for this.

A `ProcessOutcome` can, as mentioned, be either an "awarded" or "unawarded" outcome. This is embodied in the `ProcessOutcome`'s `result` field. If the `result` is `awarded`, the `ProcessOutcome` is required to be further linked to the winning (`DirectAwardSearch` and) `ArchivedService` (in the "direct award" case) or winning `BriefResponse` (in the `Brief` case).

A number of database constraints are used to enforce consistency of the data model, chiefly ensuring that only a single `completed` `ProcessOutcome` can be referring to a particular process (be that `DirectAwardProject` or `Brief`) at once. Any number of non-`completed` `ProcessOutcome`s can exist for a process at once, though marking a `ProcessOutcome` as `completed` is intended to be permanent. It is also intended for a `ProcessOutcome`'s `result` or target to be immutable after creation time. The views included here enforce these restrictions.

Ignoring the one straightforward `get_process_outcome` view, there are two main views of interest included here. One (`create_award_process_outcome`) for creating an `awarded` ("successful") `ProcessOutcome` based on a `DirectAwardProject` (once given a "winning" `service_id`) and a general  `update_process_outcome` view. `create_award_process_outcome` creates a mostly empty and non-`completed` `ProcessOutcome` which the client/frontend is expected to then fill out further details of and finally "complete" using the `update_process_outcome` view.

The `@validates` validators added here are as problematic as any others we have - in that they have the undesired side-effect of requiring a particular order of setting interdependent fields because validators are triggered on setattr. i.e. it's not always possible to set all of a new instances field values in the constructor because it's uncertain what order sqlalchemy will set them in. Sometimes they have to be set after the fact and you can see examples of this happening in the tests. However my capacity for Introducing New Stuff was depleted by this point and so I continued following existing patterns.

Depends on https://github.com/alphagov/digitalmarketplace-apiclient/pull/139 though for now is aimed at that PR's specific branch name to make it an easy checkout...

*Not* included in this PR (but to follow soon) is:
 - inclusion of a `processOutcome` key on `DirectAwardProject`s and `Brief`s which have completed outcomes
 - a `DirectAwardProject`-based "unawarded" `ProcessOutcome` creation endpoint. This would presumably also "complete" the `ProcessOutcome` at creation time seeing as there is no additional information to be provided for this outcome type.

### DIFF
--- a/app/main/views/direct_award.py
+++ b/app/main/views/direct_award.py
@@ -25,11 +25,6 @@ from ...utils import (
 
 
 def get_project_by_id_or_404(project_id: int):
-    """During the transitional period, project_id may be either the internal or the public ID of the project."""
-    project = DirectAwardProject.query.filter(DirectAwardProject.id == project_id).first()
-    if project:
-        return project
-
     return DirectAwardProject.query.filter(DirectAwardProject.external_id == project_id).first_or_404()
 
 
@@ -227,7 +222,7 @@ def list_project_services(project_external_id):
     project = get_project_by_id_or_404(project_external_id)
 
     if not project.locked_at:
-        abort(400, 'Project has not been locked: {}'.format(project.id))
+        abort(400, 'Project has not been locked: {}'.format(project.external_id))
 
     # TODO: This should work for _all_ active searches, not just the first (although we currently enforce only one
     # TODO: active search) - SW 21/09/2017
@@ -236,7 +231,7 @@ def list_project_services(project_external_id):
         DirectAwardSearch.active == True  # noqa
     ).first()
     if not search:
-        abort(400, 'Project does not have a saved search: {}'.format(project.id))
+        abort(400, 'Project does not have a saved search: {}'.format(project.external_id))
 
     paginated_archived_services = search.archived_services.paginate(
         page=page,
--- a/app/main/views/direct_award.py
+++ b/app/main/views/direct_award.py
@@ -9,7 +9,7 @@ from dmutils.config import convert_to_boolean
 
 from .. import main
 from ... import db
-from ...models import User, AuditEvent, ArchivedService
+from ...models import User, AuditEvent, ArchivedService, Outcome
 from ...models.direct_award import DirectAwardProject, DirectAwardSearch
 from ...utils import (
     drop_all_other_fields,
@@ -343,3 +343,72 @@ def record_project_download(project_external_id):
     db.session.commit()
 
     return single_result_response("project", project), 200
+
+
+@main.route('/direct-award/projects/<int:project_external_id>/services/<string:service_id>/award', methods=['POST'])
+def create_award_outcome(project_external_id, service_id):
+    updater_json = validate_and_return_updater_request()
+
+    # Acquire share-lock on project and active search rows - what we're trying to assert here is that the "locked" field
+    # of the project and the "active" field of the search are set to appropriate values at the time we create this
+    # outcome
+    project = db.session.query(DirectAwardProject).options(
+        # this has to be performed as an inner join for the row-lock to work, just means we have to perform an extra
+        # database hit in the no-rows case to disambiguate the error message
+        db.joinedload(DirectAwardProject.active_search, innerjoin=True)
+    ).filter(
+        DirectAwardProject.external_id == project_external_id,
+    ).with_for_update(read=True).first()
+
+    if project is None:
+        if db.session.query(DirectAwardProject.id).filter_by(external_id=project_external_id).first():
+            # we reach the following conclusion by deduction...
+            abort(404, f"Project {project_external_id} doesn't have an active search")
+        else:
+            abort(404, f"Project {project_external_id} not found")
+
+    if not project.locked_at:
+        abort(400, f"Project {project_external_id} has not been locked")
+
+    if project.outcome is not None:
+        abort(410, "Project {} already has a completed outcome: {}".format(
+            project_external_id,
+            project.outcome.external_id,
+        ))
+
+    archived_service = project.active_search.archived_services.options(db.load_only("id")).filter_by(
+        service_id=service_id
+    ).first()
+
+    if not archived_service:
+        abort(404, f"Service {service_id} is not in project {project_external_id}'s active search search results")
+
+    # TODO? try-loop to assign ids, handling (unlikely) external_id collisions using sub-transaction so we don't lose
+    # our DirectAwardProject row-lock
+    outcome = Outcome(
+        result="awarded",
+        direct_award_project=project,
+        direct_award_search=project.active_search,
+        direct_award_archived_service=archived_service,
+    )
+
+    db.session.add(outcome)
+    # ensure database constraints are happy with the object (initially at least: some constraints are only verified
+    # on commit) and assign id
+    db.session.flush()
+
+    audit_event = AuditEvent(
+        audit_type=AuditTypes.create_outcome,
+        user=updater_json['updated_by'],
+        data={
+            "projectExternalId": project.external_id,
+            "searchId": project.active_search.id,
+            "archivedServiceId": archived_service.id,
+            "result": outcome.result,
+        },
+        db_object=outcome,
+    )
+    db.session.add(audit_event)
+    db.session.commit()
+
+    return single_result_response("outcome", outcome), 200
--- a/app/main/views/outcomes.py
+++ b/app/main/views/outcomes.py
@@ -1,0 +1,86 @@
+import datetime
+
+from flask import abort
+from dmapiclient.audit import AuditTypes
+
+from .. import main
+from ... import db
+from ...models import AuditEvent, Outcome
+
+from ...utils import (
+    get_json_from_request,
+    json_has_required_keys,
+    single_result_response,
+    validate_and_return_updater_request,
+)
+from ...validation import validate_outcome_json_or_400
+
+
+@main.route("/outcomes/<int:outcome_id>", methods=("GET",))
+def get_outcome(outcome_id):
+    outcome = Outcome.query.filter_by(external_id=outcome_id).first_or_404()
+
+    return single_result_response("outcome", outcome), 200
+
+
+@main.route("/outcomes/<int:outcome_id>", methods=("PUT",))
+def update_outcome(outcome_id):
+    uniform_now = datetime.datetime.utcnow()
+
+    updater_json = validate_and_return_updater_request()
+
+    json_payload = get_json_from_request()
+    json_has_required_keys(json_payload, ["outcome"])
+    outcome_json = json_payload["outcome"]
+
+    validate_outcome_json_or_400(outcome_json)
+
+    # fetch and lock Outcome row so we know writing this back won't overwrite any other updates to it that made
+    # it in in the meantime
+    outcome = db.session.query(Outcome).filter_by(
+        external_id=outcome_id,
+    ).with_for_update().first()
+
+    if not outcome:
+        abort(404, f"Outcome {outcome_id} not found")
+
+    outcome.update_from_json(outcome_json)
+
+    if outcome.completed_at is not None and outcome_json.get("completed") is False:
+        abort(400, f"Can't un-complete outcome")
+    if outcome.completed_at is None and outcome_json.get("completed") is True:
+        # make a cursory check for existing Outcome collisions. we're not actually totally relying on this
+        # to police the constraint - there is a database unique constraint that will do that for us in a transactionally
+        # bulletproof way. we do this check here manually too to be able to give a nicer error message in the 99% case.
+        if outcome.brief and outcome.brief.outcome:
+            abort(400, "Brief {} already has a complete outcome: {}".format(
+                outcome.brief_id,
+                outcome.brief.outcome.external_id,
+            ))
+        if outcome.direct_award_project and outcome.direct_award_project.outcome:
+            abort(400, "Direct award project {} already has a complete outcome: {}".format(
+                outcome.direct_award_project_id,
+                outcome.direct_award_project.outcome.external_id,
+            ))
+
+        outcome.completed_at = uniform_now
+        complete_audit_event = AuditEvent(
+            audit_type=AuditTypes.complete_outcome,
+            user=updater_json['updated_by'],
+            db_object=outcome,
+            data={},
+        )
+        complete_audit_event.created_at = uniform_now
+        db.session.add(complete_audit_event)
+
+    update_audit_event = AuditEvent(
+        audit_type=AuditTypes.update_outcome,
+        user=updater_json['updated_by'],
+        db_object=outcome,
+        data=outcome_json,
+    )
+    update_audit_event.created_at = uniform_now
+    db.session.add(update_audit_event)
+    db.session.commit()
+
+    return single_result_response("outcome", outcome), 200
--- a/app/models/direct_award.py
+++ b/app/models/direct_award.py
@@ -84,9 +84,13 @@ class DirectAwardSearch(db.Model):
     project = db.relationship(DirectAwardProject)
     user = db.relationship(User)
 
-    # Partial index on project_id,active==1. Enforces only one active search per project at a time.
-    __table_args__ = (db.Index('idx_project_id_active', project_id, active, unique=True,
-                               postgresql_where=db.Column('active')),)
+    __table_args__ = (
+        # this may appear tautological (id is a unique column *on its own*, so clearly the combination of id/project_id
+        # is), but is required by postgres to be able to make a compound foreign key to these together
+        db.UniqueConstraint(id, project_id, name="uq_direct_award_searches_id_project_id"),
+        # Partial index on project_id,active==1. Enforces only one active search per project at a time.
+        db.Index('idx_project_id_active', project_id, active, unique=True, postgresql_where=db.Column('active')),
+    )
 
     def serialize(self):
         resolved_search_url = urljoin(current_app.config['DM_SEARCH_API_URL'], self.search_url)
@@ -119,3 +123,11 @@ class DirectAwardSearchResultEntry(db.Model):
     id = db.Column(db.Integer, primary_key=True)
     search_id = db.Column(db.Integer, db.ForeignKey('direct_award_searches.id'), index=True, nullable=False)
     archived_service_id = db.Column(db.Integer, db.ForeignKey('archived_services.id'), index=True, nullable=False)
+
+    __table_args__ = (
+        db.UniqueConstraint(
+            archived_service_id,
+            search_id,
+            name="uq_direct_award_search_result_entries_archived_service_id_search_id",
+        ),
+    )
--- a/app/models/direct_award.py
+++ b/app/models/direct_award.py
@@ -1,15 +1,21 @@
 from datetime import datetime
 from urllib.parse import urljoin
 
-from sqlalchemy.orm import validates
+from sqlalchemy.orm import validates, foreign, remote
+from sqlalchemy.sql.expression import and_ as sql_and, true as sql_true
 from flask import current_app
 
 from app import db
 from app.utils import random_positive_external_id
 from app.url_utils import force_relative_url
-from app.models import User, ValidationError, ArchivedService
+from app.models import ValidationError
 
 from dmutils.formats import DATETIME_FORMAT
+
+# there is a danger of circular imports here. as such, it is not necessarily "safe" to expect all other models to be
+# present in `models` at import time, but it should be "safe" to reference any of them in this file from within a
+# function or an sqlalchemy expression declared as a lambda
+from app import models
 
 
 class DirectAwardProject(db.Model):
@@ -23,7 +29,29 @@ class DirectAwardProject(db.Model):
     downloaded_at = db.Column(db.DateTime, nullable=True)  # When the project's shortlist was last downloaded
     active = db.Column(db.Boolean, default=True, nullable=False)
 
-    users = db.relationship(User, secondary='direct_award_project_users', order_by=lambda: DirectAwardProjectUser.id)
+    users = db.relationship(
+        lambda: models.User,
+        secondary='direct_award_project_users',
+        order_by=lambda: DirectAwardProjectUser.id
+    )
+    active_search = db.relationship(
+        "DirectAwardSearch",
+        primaryjoin=lambda: (sql_and(
+            foreign(DirectAwardProject.id) == remote(DirectAwardSearch.project_id),
+            remote(DirectAwardSearch.active) == sql_true(),
+        )),
+        viewonly=True,
+        uselist=False,
+    )
+    outcome = db.relationship(
+        "Outcome",
+        primaryjoin=lambda: (sql_and(
+            foreign(DirectAwardProject.id) == remote(models.Outcome.direct_award_project_id),
+            remote(models.Outcome.completed_at).isnot(None),
+        )),
+        viewonly=True,
+        uselist=False,
+    )
 
     def serialize(self, with_users=False):
         data = {
@@ -78,11 +106,11 @@ class DirectAwardSearch(db.Model):
     # in a given state at a specific time. We can still access the live version of the service through the Archived
     # Service, but by storing this we have the flexibility to show either the live state or the state at the time
     # the user ran the search.
-    archived_services = db.relationship(ArchivedService, secondary='direct_award_search_result_entries',
+    archived_services = db.relationship(lambda: models.ArchivedService, secondary='direct_award_search_result_entries',
                                         order_by=lambda: DirectAwardSearchResultEntry.id, lazy='dynamic')
 
-    project = db.relationship(DirectAwardProject)
-    user = db.relationship(User)
+    project = db.relationship(DirectAwardProject, backref="searches")
+    user = db.relationship(lambda: models.User)
 
     __table_args__ = (
         # this may appear tautological (id is a unique column *on its own*, so clearly the combination of id/project_id
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -13,7 +13,7 @@ from sqlalchemy import func
 from sqlalchemy.dialects.postgresql import INTERVAL
 from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.ext.hybrid import hybrid_property
-from sqlalchemy.orm import validates, backref, mapper
+from sqlalchemy.orm import validates, backref, mapper, foreign, remote
 from sqlalchemy.orm.session import Session
 from sqlalchemy.sql.expression import (
     case as sql_case,
@@ -29,8 +29,8 @@ from sqlalchemy_utils import generic_relationship
 
 from dmutils.dates import get_publishing_dates
 from dmutils.formats import DATETIME_FORMAT
-from .. import db
-from ..models.buyer_domains import BuyerEmailDomain
+
+from app import db
 from app.utils import (
     drop_foreign_fields,
     link,
@@ -39,10 +39,15 @@ from app.utils import (
     strip_whitespace_from_data,
     url_for,
 )
-from ..validation import (
+from app.validation import (
     is_valid_service_id, get_validation_errors, buyer_email_address_has_approved_domain,
     admin_email_address_has_approved_domain
 )
+
+# there is a danger of circular imports here. as such, it is not necessarily "safe" to expect all other models to be
+# present in `models` at import time, but it should be "safe" to reference any of them in this file from within a
+# function or an sqlalchemy expression declared as a lambda
+from app import models
 
 
 class JSON(sqlalchemy.dialects.postgresql.JSON):
@@ -835,7 +840,7 @@ class User(db.Model):
 
     @validates('email_address')
     def validate_email_address(self, key, value):
-        existing_buyer_domains = BuyerEmailDomain.query.all()
+        existing_buyer_domains = models.BuyerEmailDomain.query.all()
         if value:
             if self.role == 'buyer' and not buyer_email_address_has_approved_domain(existing_buyer_domains, value):
                 raise ValidationError("invalid_buyer_domain")
@@ -845,7 +850,7 @@ class User(db.Model):
 
     @validates('role')
     def validate_role(self, key, value):
-        existing_buyer_domains = BuyerEmailDomain.query.all()
+        existing_buyer_domains = models.BuyerEmailDomain.query.all()
         if self.email_address:
             if value == 'buyer' and \
                     not buyer_email_address_has_approved_domain(existing_buyer_domains, self.email_address):
@@ -1317,6 +1322,16 @@ class Brief(db.Model):
         uselist=False,
         lazy='joined',
         viewonly=True
+    )
+
+    outcome = db.relationship(
+        "Outcome",
+        primaryjoin=lambda: (sql_and(
+            foreign(Brief.id) == remote(models.Outcome.brief_id),
+            remote(models.Outcome.completed_at).isnot(None),
+        )),
+        viewonly=True,
+        uselist=False,
     )
 
     @validates('users')
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -1658,6 +1658,12 @@ class BriefResponse(db.Model):
 
     award_details = db.Column(JSON, nullable=True, default={})
 
+    __table_args__ = (
+        # this may appear tautological (id is a unique column *on its own*, so clearly the combination of id/brief_id
+        # is), but is required by postgres to be able to make a compound foreign key to these together
+        db.UniqueConstraint(id, brief_id, name="uq_brief_responses_id_brief_id"),
+    )
+
     brief = db.relationship('Brief', lazy='joined')
     supplier = db.relationship('Supplier', lazy='joined')
 
--- a/app/models/outcomes.py
+++ b/app/models/outcomes.py
@@ -1,0 +1,299 @@
+from sqlalchemy.orm import validates, backref
+from sqlalchemy.sql.expression import (
+    and_ as sql_and,
+    case as sql_case,
+)
+
+from app import db
+from app.utils import random_positive_external_id
+from app.models import ValidationError
+
+from dmutils.formats import DATE_FORMAT, DATETIME_FORMAT
+
+
+class Outcome(db.Model):
+    __tablename__ = "outcomes"
+
+    RESULT_CHOICES = (
+        "awarded",
+        "cancelled",
+        "none-suitable",
+    )
+
+    id = db.Column(db.Integer, primary_key=True)
+    external_id = db.Column(db.BigInteger, nullable=False, default=random_positive_external_id, unique=True)
+    completed_at = db.Column(db.DateTime, nullable=True)
+
+    start_date = db.Column(db.Date, nullable=True)
+    end_date = db.Column(db.Date, nullable=True)
+    awarding_organisation_name = db.Column(db.String, nullable=True)
+    # should be enough scale to represent up to £1b - 1p
+    award_value = db.Column(db.Numeric(11, 2), nullable=True)
+
+    result = db.Column(db.String, nullable=False)
+
+    direct_award_project_id = db.Column(
+        db.Integer,
+        db.ForeignKey('direct_award_projects.id'),
+        nullable=True,
+    )
+    direct_award_search_id = db.Column(
+        db.Integer,
+        db.ForeignKey('direct_award_searches.id'),
+        nullable=True,
+    )
+    direct_award_archived_service_id = db.Column(
+        db.Integer,
+        db.ForeignKey('archived_services.id'),
+        nullable=True,
+    )
+
+    # NOTE though Outcome currently has the *ability* to be associated with a Brief/BriefResponse, at time of
+    # writing this is not yet *used* for reporting brief awards - those are still using their own mechanism, but should
+    # be ported over to this mechanism as soon as time allows.
+    brief_id = db.Column(
+        db.Integer,
+        db.ForeignKey('briefs.id'),
+        nullable=True,
+    )
+    brief_response_id = db.Column(
+        db.Integer,
+        db.ForeignKey('brief_responses.id'),
+        nullable=True,
+    )
+
+    # a concise wrapper to reduce the ugliness of the workaround for https://bitbucket.org/zzzeek/sqlalchemy/issues/3999
+    # (which should no longer be needed once we switch to sqlalchemy 1.2)
+    def c(expression):
+        return db.cast(expression, db.Boolean)
+
+    # The following constraints enforce this general scheme:
+    # All Outcomes must point at either a DirectAwardProject OR a Brief. If the Outcome's "result" is
+    # "awarded", then a DirectAwardProject-based Outcome must also point at the DirectAwardProject's active
+    # search and "winning" archived_service_id or a Brief-based Outcome must also point at the winning
+    # BriefResponse. Only one "completed" Outcome can point at its target (DirectAwardProject or Brief) at once.
+    #
+    # Overlapping compound foreign keys are used to ensure the correctness of the
+    # DirectAwardProject-DirectAwardSearch-ArchivedService and Brief-BriefResponse chains.
+    #
+    # These particular restrictions were chosen to be enforced at a database level because they're all part of an
+    # interrelated system of rules related to cross-table relationships. Contraints which only deal with intra-row
+    # rules can be handled safely (enough) with app-side validation.
+
+    __table_args__ = (
+        db.CheckConstraint(sql_case(
+            (
+                # if awarded, the "null-ness" of direct_award_project_id, direct_award_search_id &
+                # direct_award_archived_service_id must be the same
+                (result == "awarded", sql_and(
+                    (c(direct_award_project_id.is_(None)) == c(direct_award_search_id.is_(None))),
+                    (c(direct_award_project_id.is_(None)) == c(direct_award_archived_service_id.is_(None))),
+                )),
+            ),
+            # if not awarded, we shouldn't have direct_award_search_id or direct_award_archived_service_id set
+            else_=sql_and(c(direct_award_search_id.is_(None)), c(direct_award_archived_service_id.is_(None))),
+        ), name="ck_outcomes_direct_award_keys_nullable"),
+        db.CheckConstraint(sql_case(
+            (
+                # if awarded, the "null-ness" of brief_response_id and brief_id must be the same
+                (result == "awarded", c(brief_response_id.is_(None)) == c(brief_id.is_(None)),),
+            ),
+            # if not awarded, brief_response_id should not be set
+            else_=c(brief_response_id.is_(None)),
+        ), name="ck_outcomes_brief_keys_nullable"),
+        # the "null-ness" of direct_award_project_id and brief_id must *never* be the same
+        db.CheckConstraint(
+            (c(direct_award_project_id.is_(None)) != c(brief_id.is_(None))),
+            name="ck_outcomes_either_brief_xor_direct_award",
+        ),
+        # only one completed award can point at a DirectAwardProject at once
+        db.Index(
+            "idx_outcomes_completed_direct_award_project_unique",
+            direct_award_project_id,
+            postgresql_where=(completed_at.isnot(None)),
+            unique=True,
+        ),
+        # only one completed award can point at a Brief at once
+        db.Index(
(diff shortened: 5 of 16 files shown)


risicle:
I've updated this to remove specific mention of "unawarded" results and `unawarded-` prefixed `result` field values have simply lost their prefixes. We'll probably prefer the term "nonawarded" to describe the... non... awarded... results from here on.

app/models/main.py
```diff
@@ -1832,6 +1848,302 @@ def serialize(self):
...
+
+    id = db.Column(db.Integer, primary_key=True)
+    external_id = db.Column(db.BigInteger, nullable=False, default=random_positive_external_id, unique=True)
+    completed_at = db.Column(db.DateTime, nullable=True)
+
+    start_date = db.Column(db.Date, nullable=True)
+    end_date = db.Column(db.Date, nullable=True)
+    awarding_organisation_name = db.Column(db.String, nullable=True)
+    # should be enough scale to represent up to £1b - 1p
+    award_value = db.Column(db.Numeric(11, 2), nullable=True)
```
galund: This will have created a float won't it - don't you need `asdecimal=True`?

app/models/main.py
```diff
@@ -1832,6 +1848,302 @@ def serialize(self):
...
+        nullable=True,
+    )
+    brief_response_id = db.Column(
+        db.Integer,
+        db.ForeignKey('brief_responses.id'),
+        nullable=True,
+    )
+
+    # a concise wrapper to reduce the ugliness of the workaround for https://bitbucket.org/zzzeek/sqlalchemy/issues/3999
+    # (which should no longer be needed once we switch to sqlalchemy 1.2)
```
galund: We should get this onto the tech debt board

app/models/main.py
```diff
@@ -1832,6 +1848,302 @@ def serialize(self):
...
+                    "start_date",
+                    "end_date",
+                    "awarding_organisation_name",
+                    "award_value",
+                )
+                if getattr(self, fkey) is not None
+            )
+            if failures:
+                raise ValidationError(
+                    f"{result!r} ProcessOutcomes cannot have award-related data fields set, but {failures}"
```
galund: This feels like it might read weirdly in the `None` case, don't really know if we should care tho.

tests/main/views/test_direct_award.py
```diff
@@ -934,22 +935,285 @@ def test_record_project_download_creates_audit_event(self):
...
+                False,
+                True,
+                "awarded",
+                True,
+                410,
+                AnyStringMatching(r"Project \d+ already has a completed process outcome: \d+"),
+            ),
+        )
+    )
+    def test_most_paths(
```
galund: Once the test gets this name, maybe that's the point where we've taken parameterization too far ;-)

tests/main/views/test_process_outcomes.py
```diff
@@ -0,0 +1,918 @@
...
+                    "awardingOrganisationName": "Talbot de Malahide",
+                },
+            },
+            # expected_status_code
+            400,
+            # expected_response_data
+            {
+                "error": (
+                    "awarding_organisation_name cannot be set for 'none-suitable' "
+                    + "ProcessOutcomes. Attempted to set value 'Talbot de Malahide'"
```
galund: The plus isn't necessary

tests/main/views/test_process_outcomes.py
```diff
@@ -0,0 +1,918 @@
...
+            # put_values
+            {
+                "completed": True,
+            },
+            # expected_status_code
+            400,
+            # expected_response_data
+            {
+                "error": AnyStringMatching(
+                    r"(Direct award project|Brief) \d+ already has a complete process outcome: \d+"
```
galund: Feel like matching against a specific list of things isn't necessary here. We are more futureproof if we don't check the first part of the error message at all?

migrations/versions/1210_add_process_outcome_model.py
```diff
@@ -0,0 +1,126 @@
...
+        sa.Column('completed_at', sa.DateTime(), nullable=True),
+        sa.Column('start_date', sa.Date(), nullable=True),
+        sa.Column('end_date', sa.Date(), nullable=True),
+        sa.Column('awarding_organisation_name', sa.String(), nullable=True),
+        sa.Column('award_value', sa.Numeric(precision=11, scale=2), nullable=True),
+        sa.Column('result', sa.String(), nullable=False),
+        sa.Column('direct_award_project_id', sa.Integer(), nullable=True),
+        sa.Column('direct_award_search_id', sa.Integer(), nullable=True),
+        sa.Column('direct_award_archived_service_id', sa.Integer(), nullable=True),
+        sa.Column('brief_id', sa.Integer(), nullable=True),
```
galund: I don't think these should really be in there yet - they aren't used, and having them existing would be confusing, assuming we aren't going to start work on this straight away.

Which I think is a given - we would only refactor the existing briefs stuff when we next do major work on that flow, which isn't even slightly in a roadmap right now. They can stay, but commented out - but I realise that has implications for the tests you're writing. We shouldn't cover the briefs case at all for now, I think.

Happy to chat through on slack/hangouts/whatever if you like.

app/models/main.py
```diff
@@ -1319,6 +1320,16 @@ class Brief(db.Model):
         viewonly=True
     )
 
+    process_outcome = db.relationship(
```
galund: (So I'm proposing that this should not be in this PR either. - sorry this is my bad as obviously I wasn't clear at the design stage, that I was trying to get us to design something _suitable_ for Briefs, but I _wasn't_ trying to get us to implement anything at this stage.)

app/models/main.py
```diff
@@ -1658,6 +1669,11 @@ class BriefResponse(db.Model):
 
     award_details = db.Column(JSON, nullable=True, default={})
 
+    __table_args__ = (
+        # this may appear tautological, but is required to be able to make a compound foreign key to these together
+        db.UniqueConstraint(id, brief_id, name="uq_brief_responses_id_brief_id"),
```
galund: I'm a bit confused by what `brief_id` is here if it isn't the same as `id`?

app/models/main.py
risicle: Um.... let me check that - I didn't *think* so when I was playing with it...

app/models/main.py
risicle: :+1:

app/models/main.py
risicle: Yeah, `asdecimal=True` is the default: http://docs.sqlalchemy.org/en/latest/core/type_basics.html#sqlalchemy.types.Numeric

app/main/views/direct_award.py
```diff
@@ -348,3 +343,71 @@ def record_project_download(project_external_id):
...
+    # Acquire share-lock on project and active search rows - what we're trying to assert here is that the "locked" field
+    # of the project and the "active" field of the search are set to appropriate values at the time we create this
+    # process outcome
+    project = db.session.query(DirectAwardProject).options(
+        # this has to be performed as an inner join for the row-lock to work, just means we have to perform an extra
+        # database hit in the no-rows case to disambiguate the error message
+        db.joinedload(DirectAwardProject.active_search, innerjoin=True)
+    ).filter(
+        DirectAwardProject.external_id == project_external_id,
+    ).with_for_update(read=True).first()
```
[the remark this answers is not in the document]
risicle: It's more that this view can't possibly work properly if there's no active search, so I may as well signpost why it's not working instead of acting mysteriously (or possibly giving a misleading error message from further down in the view)

app/models/main.py
risicle: Actually, these will never get shown in the `None` case.

tests/main/views/test_process_outcomes.py
risicle: Just there as an obvious indicator that it's not a sequence of separate strings - commas are easy to miss.

galund: I thought you didn't need anything, just whitespace, but doesn't matter

migrations/versions/1210_add_process_outcome_model.py
risicle: We came up with a solution that works in a bulletproof way for both DAs and Briefs (through quite a bit of grief and debate), we should implement it. Not implement the braindead easy half and then leave it for the next person (perhaps not you or I) to come along and not quite get what the plan was and implement it in an asymmetric not-quite-joined-up way. I also think this is (certainly) one of those cases where making it "work" (to a releasable state) in the easy way and *then* the hard way is *significantly* more work than just making it all work straight off. Which it does. Look, it's here.

If we were just going to implement one half of this and then leave it with a question mark over it we could have completely skipped our wonderful debate.

app/models/main.py
risicle: I think leaving the design open "so the possibility exists" to also work with briefs is as good as not considering it in the first place. Whoever came to deal with that later would have their own ideas.

migrations/versions/1210_add_process_outcome_model.py
galund: The point of the debate was so that we agreed what we want this thing to look like - as a roadmap. So that there's a plan to follow, when we next touch the DOS code, to make it not be gloriously inconsistent. But yes I regret having that debate at all considering the confusion it has clearly caused - I honestly thought we were on the same page, sorry :-(

And it doesn't mean I want to have a load of code in the codebase that never gets run, and sits around as a weird stub - that would be _more_ confusing for people.

I take your point about people not-quite-getting the plan... but if the plan is good and clear, then really they should! Which is why the write-up phase of this is important (and I committed to getting that done). And I'd rather have it commented out than lose the work, there's definitely value there.

app/models/main.py
galund: This is an argument against doing any architecture at all, or even any technical roadmapping. I can't buy that argument, because it means everything has to be short-term. We know in this case that we _are_ going to need it, we just don't need it _yet_, and the question is just about what gets implemented as code _at this stage_, surely?

tests/main/views/test_process_outcomes.py
risicle: Oh no, you don't need anything. It's just that the only thing that tells you that it's *not* two separate strings is the *lack of* a little comma, so it's a stylistic quirk I sometimes do, but I'm not strongly attached to it.

migrations/versions/1210_add_process_outcome_model.py
risicle: In short, I went to the trouble of designing a solution for this and *making sure that it does actually work*, traversing any kinks involved, so I *must* make some kind of concrete manifestation of it, otherwise it was a waste of time.

risicle:
Oh, this is an unrelated point - do we think I should be giving these `ProcessOutcome` serializations a REST-y `"links"` key with e.g. the url for `"self"`?

Pros:
 - It would make the serialization of a `ProcessOutcome` make more sense if used in an abbreviated way e.g. if included on a `DirectAwardProject`. "Here's the bare bones information, for full info see this url..."

Cons:
 - It's a token gesture as next to nothing ever uses these links (and using these links would break through the apiclient's abstraction layer)

galund:
FWIW I don't think that worrying about the purity of the apiclient's abstraction is really a thing (I'd treat it more as a set of useful shortcuts than as an abstraction myself), but on the other hand yeah we really aren't gonna need it... self-documenting API's are nice tho...

migrations/versions/1210_add_process_outcome_model.py
galund: Understood, that's fine. Let's get this committed - most of my other comments still stand although are fairly minor, so once you've been through them, and someone else has approved the PR too, then I'm all like 👍

app/models/main.py
galund: OK.

I don't think the validation messages should include the model names either - they should be couched in terms of the API.

app/models/main.py
risicle: The reason it's tautological is that if `id` is unique (which we know it is, it's a primary key) then *of course* `(id, brief_id)` is. But postgres (at time of writing) don't care. It wants a unique index on the column combination it's been asked to make a foreign key to. Known postgres ~~~bug~~~ imperfection, can't find a reference link to right now.

risicle: (was that comment not clear enough?)

app/models/main.py
risicle: But of course, if you look at it, these error messages are also referring to the *model* field names, rather than the API field names, so the man behind the curtain is in fact in full view :tornado:

risicle: Agh trying to reformulate all these messages but finding something common that makes sense is hard - then again note how rarely these error messages make it out of the endpoint.

risicle:
Y'know what's missing here is actual *model* tests...

lfdebrux:
I don't know if this is relevant, but I find the order of commits in this PR is a bit weird.

The first three commits try to import the `ProcessOutcome` class from models, but this isn't added to models until the fourth commit?

app/main/views/direct_award.py
```diff
@@ -343,3 +343,71 @@ def record_project_download(project_external_id):
...
+
+    if project is None:
+        if db.session.query(DirectAwardProject.id).filter_by(external_id=project_external_id).first():
+            # by deduction...
+            abort(404, f"Project {project_external_id} doesn't have an active search")
```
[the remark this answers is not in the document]
risicle: Mmmmmmmyes, borderline. I think where I was going with this was that the sub-resource that "can't be found" here is the `.../<service_id>/...` and it can't be "found" because the service isn't in the active search's search results because it can't be because it doesn't *have* an active search.

app/main/views/direct_award.py
```diff
@@ -343,3 +343,71 @@ def record_project_download(project_external_id):
...
+            project_external_id,
+            project.process_outcome.external_id,
+        ))
+
+    archived_service = project.active_search.archived_services.options(db.load_only("id")).filter_by(
+        service_id=service_id
+    ).first()
+
+    if not archived_service:
+        abort(404, f"Service {service_id} is not a result of project {project_external_id}'s active search")
```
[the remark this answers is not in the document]
risicle: Er yes. I can see how it can be a bit ambiguous now.

risicle:
(oh shag, I might have screwed up my rebasing then...)

app/models/main.py
```diff
@@ -1832,6 +1837,276 @@ def serialize(self):
...
+    # a concise wrapper to reduce the ugliness of the workaround for https://bitbucket.org/zzzeek/sqlalchemy/issues/3999
+    # (which should no longer be needed once we switch to sqlalchemy 1.2)
+    def c(expression):
+        return db.cast(expression, db.Boolean)
+
+    __table_args__ = (
+        db.CheckConstraint(sql_case(
+            (
+                # if awarded, the "null-ness" of direct_award_project_id, direct_award_search_id &
+                # direct_award_archived_service_id must be the same
```
lfdebrux: In what situation should `result == awarded` and these all be null?

risicle: Hmmmmmm quite possibly.

app/models/main.py
```diff
@@ -1832,6 +1837,276 @@ def serialize(self):
...
+    @validates(
+        "start_date",
+        "end_date",
+        "awarding_organisation_name",
+        "award_value",
+    )
+    def _validates_data_complete_if_completed_at(self, key, value):
+        if key == "award_value":
+            if value is not None and value < 0:
+                raise ValidationError(f"{key} must be greater than or equal to zero, got {value!r}")
```
lfdebrux: If value is `None` then it will pass the validation, is that what you were going for?

lfdebrux:
I'm not sure about the endpoints. Current we have something like

   Verb   | Endpoint
--------|-------
CREATE | `/direct-award/projects/<int:project_external_id>/services/<string:service_id>/award`
READ | `/process-outcomes/<int:process_outcome_id>`
UPDATE | `/process-outcomes/<int:process_outcome_id>`

I'd expect a bit more consistency in the URLS?

tests/main/views/test_direct_award.py
lfdebrux: What is the advantage of parametrising the test like this?

app/main/views/direct_award.py
```diff
@@ -348,3 +343,71 @@ def record_project_download(project_external_id):
...
+        ))
+
+    archived_service = project.active_search.archived_services.options(db.load_only("id")).filter_by(
+        service_id=service_id
+    ).first()
+
+    if not archived_service:
+        abort(404, f"Service {service_id} is not in project {project_external_id}'s active search search results")
+
+    # TODO try-loop to assign ids, handling (unlikely) external_id collisions using sub-transaction so we don't lose
```
benvand: Is this a problem we have elsewhere in the codebase? Do we need a TODO if we can make a tech debt ticket? If it's urgent enough for a `TODO` shouldn't we be doing it now?

app/main/views/direct_award.py
```diff
@@ -348,3 +343,71 @@ def record_project_download(project_external_id):
...
+    # our DirectAwardProject row-lock
+    process_outcome = ProcessOutcome(
+        result="awarded",
+        direct_award_project=project,
+        direct_award_search=project.active_search,
+        direct_award_archived_service=archived_service,
+    )
+
+    db.session.add(process_outcome)
+    # ensure database constraints are happy with the object (initially at least) and assign id
```
benvand: The bracketed part of this comment adds confusion

app/main/views/direct_award.py
```diff
@@ -348,3 +343,71 @@ def record_project_download(project_external_id):
...
+        # this has to be performed as an inner join for the row-lock to work, just means we have to perform an extra
+        # database hit in the no-rows case to disambiguate the error message
+        db.joinedload(DirectAwardProject.active_search, innerjoin=True)
+    ).filter(
+        DirectAwardProject.external_id == project_external_id,
+    ).with_for_update(read=True).first()
+
+    if project is None:
+        if db.session.query(DirectAwardProject.id).filter_by(external_id=project_external_id).first():
+            # by deduction...
```
benvand: This comment is not helpful

benvand:
I am not keen on the name 'Process Outcome'. It is not descriptive.

I understand that our avoidance of the word 'procurement' on the Digital Marketplace has introduced challenges in naming our user flows but I think we can do better. 🏅 

There are some conventions for naming (both models and urls) in the award flow work for 'further competition' (DOS NOTDOS ⛔️) frameworks if we're talking about inspiration. 

Happy to put heads together and come up with something if you like.

app/main/views/process_outcomes.py
```diff
@@ -0,0 +1,86 @@
...
+
+    if not process_outcome:
+        abort(404, f"Process outcome {process_outcome_id} not found")
+
+    process_outcome.update_from_json(process_outcome_json)
+
+    if process_outcome.completed_at is not None and process_outcome_json.get("completed") is False:
+        abort(400, f"Can't un-complete process outcome")
+    if process_outcome.completed_at is None and process_outcome_json.get("completed") is True:
+        # make a cursory check for existing ProcessOutcome collisions. we're not actually totally relying on this
```
benvand: This comment is lengthy. You could just say we're checking to determine which error message to show 🤷‍♀️

risicle:
Hmmmmmmmmmmmmmmmmmmmmmmmmmmmm.

I think I also avoided "purchase"/"procurement" because this "outcome" isn't necessarily one of those (specifically, it could be the *absence of* one of those).

@galund?

risicle:
`BuyingProcessOutcome`?

lfdebrux:
Is there any reason not to go for simply `Outcome`?

app/models/direct_award.py
```diff
@@ -81,12 +100,15 @@ class DirectAwardSearch(db.Model):
...
 
-    project = db.relationship(DirectAwardProject)
+    project = db.relationship(DirectAwardProject, backref="searches")
     user = db.relationship(User)
 
-    # Partial index on project_id,active==1. Enforces only one active search per project at a time.
-    __table_args__ = (db.Index('idx_project_id_active', project_id, active, unique=True,
-                               postgresql_where=db.Column('active')),)
+    __table_args__ = (
+        # this may appear tautological, but is required to be able to make a compound foreign key to these together
```
benvand: - this may appear tautological, but is required to be able to make a compound foreign key to these together
+ this is required to be able to make a compound foreign key

Or something.

risicle: I might reword it but want to keep in the information that the casual reader might see it as pointless and want to delete it.

- [ ] TODO

risicle:
I think the only reason I didn't go with `Outcome` was the way we already use the term `Outcome` with DOS. e.g. the "Digital Outcomes" lot.

risicle:
re the endpoint URLs, the issue is the way they are created is not symmetrical. Depending on whether they are `awarded` or not, whether they point at a Brief or Project, they require different arguments. And really they are marking a different point in the lifecycle of their target object.

app/models/direct_award.py
```diff
@@ -81,12 +100,15 @@ class DirectAwardSearch(db.Model):
...
+        db.UniqueConstraint(id, project_id, name="uq_direct_award_searches_id_project_id"),
+        # Partial index on project_id,active==1. Enforces only one active search per project at a time.
+        db.Index('idx_project_id_active', project_id, active, unique=True, postgresql_where=db.Column('active')),
```
benvand: This is very cool but the comment could be clearer. Also we can use the code to be more descriptive with:

`db.Index('idx_project_id_active', project_id, active, unique=True, postgresql_where=(db.Column('active') == sql_true()))`

Something along the lines of:
`# Create a partial index on `project_id` and `active == True` to enforce a maximum of one active search per project`

risicle: oh ok i was going for brevity but yeah

katstevens:
Re: `Outcome` naming - we use this generally in the DOS award flow already, it hasn't seemed to cause problems so far? The lot name always has `digital-` as a prefix in our urls/content/tests etc.

app/models/main.py
```diff
@@ -1319,6 +1320,16 @@ class Brief(db.Model):
         viewonly=True
     )
 
+    process_outcome = db.relationship(
+        "ProcessOutcome",
+        primaryjoin=lambda: (sql_and(
+            foreign(Brief.id) == remote(ProcessOutcome.brief_id),
```
benvand: Seven lines above this we use the string notation to define a primary join. Here we pass it to a lambda.
We should pick one way.

tests/main/views/test_direct_award.py
risicle: It reduces repetition & copy-pastiness (which goes hand in hand with typos), and allows "expected conditions" -> "expected result" to be expressed as a simple tuple, making it extremely cheap to explore various combinations of options.

app/main/views/direct_award.py
risicle: Pass. It's very unlikely until we start to have *many* rows in the table. Don't know whether people want me to bother or not, but wanted to properly record how the design intends for it to be done when/if someone comes along to do it.

app/models/main.py
risicle: The lambda is only required when we're referencing models "lower down" in the file (i.e. haven't been declared in this scope yet), but using it where it's not required would just be line noise.

app/models/main.py
risicle: Because if you look at the number of times it is used in the following code, it bloats the expressions out to something horrible and scary. That's *why* it's only one character, because the lines overflow enough as it is. And syntactically it's supposed to be transparent (should be readable as though it's not there at all). I'd rather it *weren't* there at all - I *wanted* to use `_` but george didn't like that as it's normally used for translations.

app/models/main.py
```diff
@@ -1832,6 +1848,307 @@ def serialize(self):
...
+            (brief_response_id, brief_id,),
+            ("brief_responses.id", "brief_responses.brief_id",),
+            name="fk_process_outcomes_brief_response_id_brief_id",
+            deferrable=True,
+            initially="DEFERRED",
+        ),
+    )
+
+    # let's not leave our de-uglification wrapper hanging around as a method
+    del c
```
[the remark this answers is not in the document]
risicle: Because otherwise it would give all process outcome instances this weird unintentional `c` method which was never meant to be used outside this block. Trying not to leak a reference to something I don't need to. Encapsulation.

(actually come to think I haven't *checked* this actually does what I expect...)

risicle:
I'm not against `Outcome` personally

app/models/main.py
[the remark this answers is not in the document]
benvand: So what you're worried about is polluting the namespace. In this case we have the opportunity to just not.
I understand the point. I totally get it. But defining, then deleting, this weird single letter attribute _at class definition_ is not a good trade off for the sake of a couple of newlines in code. It's just going to confuse people. It's smart and neat but it's also confusing for people not familiar with how python classes are constructed or not familiar with the _incredibly rarely used_ `del` keyword.

Just the fact that we're having this conversation proves that. The newlines might not be as pretty but their ugliness wouldn't have started a conversation.

app/models/main.py
benvand: That's the same reason we're using a string 7 lines above. 
```
        primaryjoin="_and(Brief.id==BriefResponse.brief_id, BriefResponse.awarded_at.isnot(None))",
```

app/models/main.py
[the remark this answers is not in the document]
risicle: It's not "a few newlines", it's **significantly** bulking out most of the sql expressions used in this block with scary content that is actually _meaningless_, making them look much weirder than they actually are. If people don't know about the `del` keyword or how python classes are constructed, then hopefully they'll learn something - this isn't kindergarten.

migrations/versions/1210_add_process_outcome_model.py
benvand: I'm with @galund on this one. If we're not using it yet we shouldn't include it.

benvand: Re this comment:
```
    # NOTE though ProcessOutcome currently has the *ability* to be associated with a Brief/BriefResponse, at time of
    # writing this is not yet *used* for reporting brief awards - those are still using their own mechanism, but should
    # be ported over to this mechanism as soon as time allows.
```
I think this needs more discussion re a backlog/ tech debt ticket. We shouldn't dictate future decisions in the codebase like this. It may be de prioritised or even completely removed in the future and then the comment is untrue. 
The platform will still operate if this bit of work doesn't get done.

app/models/main.py
```diff
@@ -1832,6 +1848,307 @@ def serialize(self):
...
+    direct_award_archived_service_id = db.Column(
+        db.Integer,
+        db.ForeignKey('archived_services.id'),
+        nullable=True,
+    )
+
+    # NOTE though ProcessOutcome currently has the *ability* to be associated with a Brief/BriefResponse, at time of
+    # writing this is not yet *used* for reporting brief awards - those are still using their own mechanism, but should
+    # be ported over to this mechanism as soon as time allows.
+    brief_id = db.Column(
```
benvand: We are foreign keying to brief and brief response here when we can access brief through brief response. This kind of denormalisation is premature at this point and is likely to introduce more problems than it solves (misalignments). If we really must access brief on this model we can add a property. 

Also the point about it not being in use yet.

app/models/main.py
```diff
@@ -1832,6 +1848,307 @@ def serialize(self):
...
+
+    start_date = db.Column(db.Date, nullable=True)
+    end_date = db.Column(db.Date, nullable=True)
+    awarding_organisation_name = db.Column(db.String, nullable=True)
+    # should be enough scale to represent up to £1b - 1p
+    award_value = db.Column(db.Numeric(11, 2), nullable=True)
+
+    result = db.Column(db.String, nullable=False)
+
+    direct_award_project_id = db.Column(
```
benvand: Same as with `brief` and `brief_response` this can be accessed through the direct award search attribute?

tests/main/views/test_direct_award.py
benvand: This is very hard to read and understand. If we need this much data for a test then we need to find somewhere to explicitly store test data sets.

app/main/views/direct_award.py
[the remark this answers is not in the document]
galund: Yep, as long as the 404 relates to _something in the URI_ then I think that's ok.

app/main/views/direct_award.py
galund: This relates and can be seen as a continuation of the previous comments. It doesn't make sense on its own, but I think it's okay in context.

app/main/views/process_outcomes.py
galund: I think it's fair enough to note that this is technically duplicating something that exists elsewhere, and why.

app/models/direct_award.py
galund: I don't understand what we're compounding here, if there's actually only one column. Definitely agree this thing needs some more explanation, appreciate that underneath it all is some sqlalchemy limitation.

app/models/main.py
galund: Re the key - I think this is so that we can enforce here that there is only one non-draft outcome for each brief - if we only linked through brief_id then we wouldn't be able to do that.

So much comes down to the decision to insist on implementing all of these constraints in the database. Quite heated debate has happened about that already, there are compelling arguments on both sides - but given that decision, I think this is logical.

Re not being in use yet - I've cleared this with @louzoid-gds - I'm not massively happy either, but given there's some complexity to the design, and @risicle has now implemented it, after some debate/heartache I don't want to rip this out.

app/models/main.py
galund: No great need to argue about this seeing as it ought to go away, given some (fairly high priority I think) piece of tech debt getting done. I don't like the idea of us falling way being on ORM versions (as and when we are doing significant dev on the project anyway, erm.)

app/models/main.py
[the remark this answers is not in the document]
galund: See my previous comment - there's a TODO about why it can go away, so let's not fret about it - the intention here is clear enough I think. Even a newbie will appreciate that something that is `del`eted is going to go away, and if they didn't understand much about why it was there then maybe they don't need to - there are references to the issue above anyway.

migrations/versions/1210_add_process_outcome_model.py
galund: Well, I think it's okay to dictate design/architecture - that means we've done some thinking.

In this case, we needed to implement half of that piece of architecture, and some implementation decisions then have to be taken - which necessarily constrains that future work.

We could leave the thing out - but that doesn't actually take away from the fact that we've _already_ constrained the future implementation (because otherwise the existing implementation of the architecture doesn't make sense).

I don't like having code that doesn't do anything in there either. We've definitely taken on some tech debt here - although it's not super-bad - as you say, everything carries on working, it just doesn't have the greatest architecture. But that was true _before we did any work for G-Cloud_: the only alternative was a mechanism that collected outcomes for G-Cloud that didn't account for further competition at all _anyway_. There was never going to be a scenario where we could actually move over DOS and implement G-Cloud in one fell swoop - it's too big a piece of work.

Some I'm content that there's really no great choices here, and the working having been done means we can still choose to:

(a) do nothing for now - the situation is commented
(b) move DOS over as planned here and more-than-partially implemented
(c) decide some or other implementation decision here is wrong and come up with an entirely new idea for capturing this stuff.

Personally, long run, I don't think this is even the right platform for capturing outcomes - that's what Contracts Finder is for - but we are constrained by the organisational problems we all know about.

app/models/main.py
risicle: @benvand there is no misalignment possible here - that is the precise reason for it existing. The only reason for any denormalization existing in this model is for the ability it gives us to form constraints.

app/models/main.py
risicle: Access is not the role of this field. The purpose is to provide a column to add a unique constraint for. Same goes for `brief_id`.

app/models/main.py
risicle: The string method is slightly nastier & more vague (what are the scoping rules for those strings? what operators & functions are imported?). They also thwart the ability for "smart" IDEs that some people use to track variables or give comtextual information. If there are string-based clauses above written by *me*, it's because I didn't at that point know the lambda trick was supported.

app/main/views/process_outcomes.py
risicle: Lengthy comments are fine if they're relevant and in fact something that I think are sorely lacking in our code making it very hard to have confidence about the intention of the design.

app/main/views/direct_award.py
risicle: I will expand it though

tests/main/views/test_direct_award.py
risicle: So.. give the user the job of jumping back & forth between different files trying to piece together what argument corresponds to what, and then of course where would such a file go? We could discuss for hours. And then should it be json or python? How should it be imported or parsed?

app/models/direct_award.py
risicle: There's two columns - `id` and `project_id`.

app/main/views/process_outcomes.py
benvand: But this comment, specifically, is not concise. It uses too many words to say not enough important stuff.

From a 'why I wrote my code like this' point of view it's good. You do the check here instead of relying on the db constraint to return a more readable error message. Because otherwise someone might question why you're doing the check instead of relying on the constraint. Cool.

With your code comments it's worth putting yourself in the shoes of 'someone reading the code just to find out what it does' point of view. When I read this code I can see that you're doing a constraint check and raising an error. That's not a foreign pattern to me at all. I can even read the error message. It's obvious to me that you're checking whether the brief 'already has a complete process outcome' and erroring out if so. The variable names are descriptive, the check is pretty self explanatory and the error message is right there. That's good code.

app/models/main.py
benvand: When we pick a new way to do things it's usually worth updating existing instances at the same time. Otherwise we're increasing tech debt and inconsistency.

Alternatively, if that sounds too long, we can do the old way and put a tech debt ticket up to update to the new way. That way at least we're not increasing inconsistency.

Putting up these smaller tech debt tickets is really helpful for new starters and junior devs who need to get to know the codebase and need to be able to get stuck in quickly. Having a few small tickets like going through and updating ORM definitions provides exposure to the codebase and builds confidence.

The inconsistency might not throw you off because you know the codebase, in the same way that going through the model layer and updating all the string relationship definitions to lambdas sounds like a chore, but to a new starter or junior they are respectively a real problem and an opportunity.

app/main/views/process_outcomes.py
risicle: I had reason to go back and look at the source of my old (much bigger and more complicated) project a few days ago and it was down to comments like this that made me able to comprehend very intricate code that is now quite foreign to me, and the difference was quite stark compared to the opacity and often inexplicability of DM code.

More generally, I suspect it is this tendency to be *this fussy* about all comments that causes our whole source to be so uncommented. When someone hits the hash key they've got to weigh up whether they really want to have to put up with all the bikeshedding that will inevitably come their way.

app/models/main.py
risicle: And every extra line of code I unnecessarily touch is another thing that someone will potentially pull me up for, maybe the indentation, or potentially misunderstand what I'm doing and make a complaint about something that was in the original code that is being rewritten, or maybe stop everything and want to organize a meeting about "which way are we going to do this?". On the other hand *cautiously* doing things in "new ways" allows us to experiment and learn which ways are better or worse without selling the farm.

app/main/views/process_outcomes.py
benvand: But this comment, specifically, in this specific context, is not concise or helpful.

app/models/main.py
benvand: I don't agree with this at all when you're coding in a team. You're saying that not committing to a single good way of doing something is a good way to introduce changes?

tests/main/views/test_direct_award.py
```diff
@@ -934,22 +935,296 @@ def test_record_project_download_creates_audit_event(self):
...
+                "awarded",
+                True,
+                410,
+                {
+                    "error": AnyStringMatching(r"Project \d+ already has a completed process outcome: \d+"),
+                },
+            ),
+        )
+    )
+    def test_most_paths(
```
benvand: The unit tests doesn't test a unit. They test features. If a test fails on my test run called `test_most_paths` I have no idea what it tests or why it has failed. Part of having unit tests its to help us debug problems quickly.

Splitting the tests into units like audit event creation, paths for with and without active search, paths for successful and unsuccessful responses help with this and are descriptive. You can use the unit tests to determine what the code should be doing. You can often pinpoint exactly what has changed to trigger an error.

Updating the code then finding what part of the test to change and what parameters need to change is difficult when tests are too long.

app/models/main.py
benvand: It's not that I think that lambda > strings or vice versa it's that we should be pushing towards a more consistent codebase rather than less. 

If you're in a file and all the methods are in camelCase then you'd use camelCase, if everything is written in functions rather than classes then use functions, if strings are interpolated with `'%s' % var` instead of `'{}'.format(var)` use the former. 

This comment is not at all about which way is better. It's about picking a way and sticking to it because consistency is valuable in itself.

Inconsistent code is harder to read and harder to trust. If you're reading a file and 10% of the methods are in snake case and 90% in camel it feels... off? It throws you off, it doesn't quite look professional. Do these functions work together? Did the writers know where they were going with this? Did they know what eachother were doing? Why did they do it differently? Is there some sort of pitfall I'm missing here?

Put differently I wouldn't care if a python project used camel case for all of it's variables, as long as it did it consistently. It's not hurting anything. But if you want to start using snake case then convert the whole project. Otherwise you have a legacy code and new code split. New devs have to ask 'What are we up to here? Snake or camel? Why the split?'
What about if you treat every decision like that? Functions over classes? A Snake Function/ Camel Function/ Snake Class/ Camel Class split? Then you want to do string interpolation a new way so you have a Old String Snake function Old string Camel function New String Camel Snake exponential increase 🤷‍♂️ 

My only point is that doing this consistently would make this PR better. 🤷‍♂️

galund: C'mon.. the lettercase thing is a something of a straw man argument: it's not the scenario we're considering. I absolutely agree that we have to be consistent in a great many things including in how we name/case our variables/functions/API keys.

But here we're just saying that a library offers multiple ways to do a thing, and that's true all over with Python -- there's frequently multiple ways to do something, and sometimes one way may be more clear/neat/expressive and sometimes the other, and the tradeoffs may be different in different parts of the codebase, and frequently they are subjective anyway.

If we really want random unrelated counterexamples, then *for example* I wouldn't agree with a rule that said "all lambda's are banned because you could write them consistent with other functions and give them a name"! - what should be consistent is a matter of judgement and I think you and Rob might just need to agree to differ here ;-)

app/main/views/outcomes.py
```diff
@@ -0,0 +1,86 @@
...
+        abort(404, f"Outcome {outcome_id} not found")
+
+    outcome.update_from_json(outcome_json)
+
+    if outcome.completed_at is not None and outcome_json.get("completed") is False:
+        abort(400, f"Can't un-complete outcome")
+    if outcome.completed_at is None and outcome_json.get("completed") is True:
+        # make a cursory check for existing Outcome collisions. we're not actually totally relying on this
+        # to police the constraint - there is a database unique constraint that will do that for us in a transactionally
+        # bulletproof way. we do this check here manually too to be able to give a nicer error message in the 99% case.
```
katstevens: One thing we encountered during the DOS award flow was people putting in the wrong contract value (missing a decimal point and so on), and 2nd line being asked to change the value later. By changing it in the production DB 🙀 So we made the decision to allow an API user to change the contract details (they cannot change who won); the buyer can't change anything themselves from the frontend, they have to raise a support ticket. 

tl:dr - this might need to be less strict to allow for user mistakes.

app/main/views/outcomes.py
```diff
@@ -0,0 +1,86 @@
...
+
+    update_audit_event = AuditEvent(
+        audit_type=AuditTypes.update_outcome,
+        user=updater_json['updated_by'],
+        db_object=outcome,
+        data=outcome_json,
+    )
+    update_audit_event.created_at = uniform_now
+    db.session.add(update_audit_event)
+    db.session.commit()
```
katstevens: Curious why we need both an `update` and `complete` Audit Event? Would one or the other suffice?

app/models/outcomes.py
```diff
@@ -0,0 +1,328 @@
...
+    )
+    direct_award_archived_service_id = db.Column(
+        db.Integer,
+        db.ForeignKey('archived_services.id'),
+        nullable=True,
+    )
+
+    # NOTE though Outcome currently has the *ability* to be associated with a Brief/BriefResponse, at time of
+    # writing this is not yet *used* for reporting brief awards - those are still using their own mechanism, but should
+    # be ported over to this mechanism as soon as time allows.
```
katstevens: Is there a tech debt ticket for this, or is it going to be covered by mission work?

requirements-app.txt
```diff
@@ -11,7 +11,7 @@ SQLAlchemy==1.1.4
 SQLAlchemy-Utils==0.30.5
 
 git+https://github.com/alphagov/digitalmarketplace-utils.git@35.2.0#egg=digitalmarketplace-utils==35.2.0
-git+https://github.com/alphagov/digitalmarketplace-apiclient.git@15.0.0#egg=digitalmarketplace-apiclient==15.0.0
+git+https://github.com/alphagov/digitalmarketplace-apiclient.git@ris-process-outcomes-audit-types#egg=digitalmarketplace-apiclient==16.1.0
```
katstevens: Remember to update this once it's merged 👍

tests/main/views/test_direct_award.py
katstevens: I'd rename the test to `def test_direct_award_outcome_scenarios(...)`.

tests/main/views/test_outcomes.py
```diff
@@ -0,0 +1,918 @@
...
+                                "endDate": "2020-11-20",
+                            },
+                        }),
+                    },
+                ),
+            ),
+        )),
+        ids=(lambda val: "EMPTYDCT" if val == {} else None),
+    )
+    def test_most_paths(
```
katstevens: Again, I'd rename to `test_direct_award_and_brief_outcome_scenarios`, so it's clear what's different here to the test in the other module.

app/main/views/outcomes.py
risicle: Consider if we added more data to be collected on the outcome and/or decided we wanted to split the data collection process into multiple pages, we'd want to be able to hit the "update" view a number of times before the final "completion".

app/models/outcomes.py
risicle: Hopefully mission??

requirements-app.txt
risicle: Oh god yes.

tests/main/views/test_outcomes.py
risicle: :+1:

app/main/views/outcomes.py
katstevens: OK, but do we need another update *after* that final completion? I think that's what's happening here?

risicle: The request could have (and often will have) supplied updated data at the same time as marking it complete.

risicle: Just for clarity, I'm setting both audit events timestamps to the same `uniform_now` so it shouldn't matter which went into the db first.

katstevens: Makes sense 👍

app/models/outcomes.py
galund: (This should definitely be a tech debt ticket, but hopefully one that we should be able to address in a future mission - certainly next time we do anything outcomes/awards for briefs.)

PR #794, opened 2018-05-30
Revert callbacks API

samuelhwilliams:
## Summary
Notify only allows one callback per account, which means we can't
effectively segment our callbacks by environment.

Given this, it's probably better that we just use a daily job to
retrieve all notifications sent and run through them as a batch job to
process permanent failures.

This commit reverts PRs #786, #788, and #793.

 ## Ticket
https://trello.com/c/vOAUDwnL/446


samuelhwilliams:
I think we're going ahead with proper callbacks after all!

PR #796, opened 2018-06-06
Outcomes: add direct award non-awarded endpoint

risicle:
Trello https://trello.com/c/nhtrO8Qi/78-non-award-flow-back-end-api

This gives us a `POST`able view which will create a non-award `Outcome` and set it to be "completed" in a single call (because no extra information other than the non-awarding `reason` is needed).

URLs are `/direct-award/projects/<int:project_external_id>/none-suitable` and `/direct-award/projects/<int:project_external_id>/cancel`.

### DIFF
--- a/app/main/views/direct_award.py
+++ b/app/main/views/direct_award.py
@@ -412,3 +412,76 @@ def create_award_outcome(project_external_id, service_id):
     db.session.commit()
 
     return single_result_response("outcome", outcome), 200
+
+
+@main.route(
+    "/direct-award/projects/<int:project_external_id>/none-suitable",
+    methods=("POST",),
+    defaults={"nonawarded_reason": "none-suitable"},
+)
+@main.route(
+    "/direct-award/projects/<int:project_external_id>/cancel",
+    methods=("POST",),
+    defaults={"nonawarded_reason": "cancelled"},
+)
+def create_nonawarded_outcome(project_external_id, nonawarded_reason):
+    updater_json = validate_and_return_updater_request()
+    uniform_now = datetime.datetime.utcnow()
+
+    if nonawarded_reason not in ("cancelled", "none-suitable",):
+        abort(404, f"Unknown nonawarded_reason {nonawarded_reason!r}")
+
+    # Acquire share-lock on project row - what we're trying to assert here is that the "locked" field is true at the
+    # time we create this outcome
+    project = db.session.query(DirectAwardProject).filter_by(
+        external_id=project_external_id,
+    ).with_for_update(read=True).first()
+
+    if project is None:
+        abort(404, f"Project {project_external_id} not found")
+
+    if not project.locked_at:
+        abort(400, f"Project {project_external_id} has not been locked")
+
+    if project.outcome is not None:
+        abort(410, "Project {} already has a completed outcome: {}".format(
+            project_external_id,
+            project.outcome.external_id,
+        ))
+
+    # TODO? try-loop to assign ids, handling (unlikely) external_id collisions using sub-transaction so we don't lose
+    # our DirectAwardProject row-lock
+    outcome = Outcome(
+        result=nonawarded_reason,
+        direct_award_project=project,
+    )
+    outcome.completed_at = uniform_now
+
+    db.session.add(outcome)
+    # ensure database constraints are happy with the object (initially at least: some constraints are only verified
+    # on commit) and assign id
+    db.session.flush()
+
+    creation_audit_event = AuditEvent(
+        audit_type=AuditTypes.create_outcome,
+        user=updater_json['updated_by'],
+        data={
+            "projectExternalId": project.external_id,
+            "result": outcome.result,
+        },
+        db_object=outcome,
+    )
+    creation_audit_event.created_at = uniform_now
+    db.session.add(creation_audit_event)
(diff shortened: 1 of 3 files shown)


app/main/views/direct_award.py
```diff
@@ -412,3 +412,72 @@ def create_award_outcome(project_external_id, service_id):
...
+    methods=("POST",),
+    defaults={"nonawarded_reason": "cancelled"},
+)
+def create_nonawarded_outcome(project_external_id, nonawarded_reason):
+    updater_json = validate_and_return_updater_request()
+    uniform_now = datetime.datetime.utcnow()
+
+    # an extra guard against incorrect url configuration or unintended call circumstances
+    if nonawarded_reason not in ("cancelled", "none-suitable",):
+        abort(404, f"Unknown nonawarded-reason {nonawarded_reason!r}")
```
galund: This feels over-the-top, we only need one of the two ways of constraining this. But feel free to keep this one and just have one `@route` with a single argument, I don't mind which we keep. It's just that both seems unnecessary?

risicle: gone

tests/main/views/test_direct_award.py
```diff
@@ -1044,7 +1050,8 @@ def test_nonexistent_project_id(self):
                     "error": AnyStringMatching(r"Project \d+ already has a completed outcome: \d+"),
                 },
             ),
-        )
+        ),
+        ids=(lambda val: "ANYDICT" if val is _anydict else None),
```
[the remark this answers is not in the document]
risicle: Yeah ok it's to make pytest's printed representation of the parameter set more readable

tests/main/views/test_outcomes.py
```diff
@@ -670,6 +670,7 @@ class TestUpdateOutcome(BaseApplicationTest, FixtureMixin):
                 ),
             ),
         )),
+        # help pytest make its printed representation of the parameter set a little more readable
         ids=(lambda val: "EMPTYDCT" if val == {} else None),
```
galund: Pointing out the typo seems churlish but now that this is in three places, I wonder if it could done via ` __str__` on `AnySupersetOf`?

tests/main/views/test_direct_award.py
```diff
@@ -1228,3 +1240,305 @@ def test_direct_award_outcome_scenarios(
...
+                (endpoint_rel_path,) + base_scenario
+                for endpoint_rel_path, base_scenario in product(
+                    ("none-suitable", "cancel",),
+                    _base_direct_award_nonawarded_scenarios,
+                )
+            ),
+            (
+                # but we also include a couple of specific cases with invalid reasons
+                (
+                    "severe-chill",
```
galund: I don't quite get why we're testing specifically two cases here

tests/main/views/test_outcomes.py
risicle: If you're talking about `DCT` I'm ... just ... trying ... to make it ... ever ... so ... much ... shorter... (the full parameter string is quite long)

Using the `str` value would defeat that and would pytest would probably reject it as it has non-alphanumeric chars.

tests/main/views/test_direct_award.py
risicle: To assert that it's not just an implementation that's taking `<project_id>/<any_string>` and happily trying to use that as a `result`.

galund: But one test of some arbitrary string would confirm that wouldn't it?

tests/main/views/test_outcomes.py
galund: I'm still confused about what `ids` refers to or is used by. I'm suggesting you override `__str__` and hoping that pytest might use that, I don't see how it's in its gift to complain about what characters are in there.

tests/main/views/test_direct_award.py
risicle: I think I originally meant to test one of them as `awarded` but that idea got lost in the fog of my mind - will set it right...

tests/main/views/test_outcomes.py
risicle: e.g. when you run `py.test -vv` it will end up listing the test variant as something like `test_direct_award_nonawarded_scenarios[none-suitable-True-True-True-cancelled-True-410-expected_response_data7]` to disambiguate it. It tries to limit the character set used because the idea is that this is sometimes taken as a command line parameter ("run just this test") and it wants it to be as un-shell-hostile as possible. Hence no curly braces. This function allows it to switch the opaque `expected_response_data7` for `EMPTYDCT` in cases where it's... an empty dict.

tests/main/views/test_direct_award.py
```diff
@@ -1228,3 +1240,305 @@ def test_direct_award_outcome_scenarios(
...
+                    True,
+                    None,
+                    False,
+                    404,
+                    {
+                        "error": AnyStringMatching(r".*the requested url was not found.*", re.I),
+                    },
+                ),
+                (
+                    "awarded",
```
galund: Okay this kinda makes sense now but probably needs a comment - "Check that `awarded` is not a URI because awarding is done against a project URI (or, in future a brief)" or sommat like that maybe.

PR #797, opened 2018-06-07
Add a view that exports suppliers related to a framework

katstevens:
Trello: https://trello.com/c/gVkJo8Zj/70-update-framework-manager-user-download-list-to-include-additional-fields

- New API view to export supplier data for a particular framework
- Service counts for each lot
- TODO: Squash commits

See also: 
- API Client: https://github.com/alphagov/digitalmarketplace-apiclient/pull/140/files
- Admin FE: https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/392

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -127,9 +127,16 @@ def export_suppliers_for_framework(framework_slug):
                 supplier.supplier_id in suppliers_with_a_complete_service and
                 company_details_confirmed
         ) else 'no_application'
-        application_result = 'no result'
+        application_result = ''
         framework_agreement = False
         variations_agreed = ''
+
+        if framework.status != 'open':
+            if sf.on_framework is None:
+                application_result = 'no result'
+            else:
+                application_result = 'pass' if sf.on_framework else 'fail'
+            framework_agreement = bool(getattr(sf.current_framework_agreement, 'signed_agreement_returned_at', None))
 
         supplier_rows.append({
             "supplier_id": supplier.supplier_id,
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -88,6 +88,69 @@ def list_suppliers():
         abort(400, 'invalid framework')
 
 
+@main.route('/suppliers/export/<framework_slug>', methods=['GET'])
+def export_suppliers_for_framework(framework_slug):
+    # 400 if framework slug is invalid
+    framework = Framework.query.filter(Framework.slug == framework_slug).first()
+    if not framework:
+        abort(400, 'invalid framework')
+
+    if framework.status == 'coming':
+        abort(400, 'framework not yet open')
+
+    suppliers_and_framework = db.session.query(
+        SupplierFramework, Supplier, ContactInformation
+    ).filter(
+        SupplierFramework.supplier_id == Supplier.supplier_id
+    ).filter(
+        SupplierFramework.framework_id == framework.id
+    ).filter(
+        ContactInformation.supplier_id == Supplier.supplier_id
+    ).options(
+        lazyload(SupplierFramework.framework),
+        lazyload(SupplierFramework.prefill_declaration_from_framework),
+        lazyload(SupplierFramework.framework_agreements),
+    ).order_by(
+        Supplier.supplier_id
+    ).all()
+
+    supplier_rows = []
+
+    for sf, supplier, ci in suppliers_and_framework:
+        supplier_rows.append({
+            "supplier_id": supplier.supplier_id,
+            "supplier_name": supplier.name,
+            "supplier_organisation_size": supplier.organisation_size,
+            "duns_number": supplier.duns_number,
+            "registered_name": supplier.registered_name,
+            "companies_house_number": supplier.companies_house_number,
+            # TODO: framework application status
+            'application_result': 'no result',
+            'application_status': 'no_application',
+            'declaration_status': 'unstarted',
+            'framework_agreement': False,
+            'variations_agreed': '',
+            # TODO: service counts for each lot
+            "published_services_count": {
+                "digital-outcomes": 0,
+                "digital-specialists": 0,
+                "user-research-studios": 0,
+                "user-research-participants": 0,
+            },
+            "contact_information": {
+                'contact_name': ci.contact_name,
+                'contact_email': ci.email,
+                'contact_phone_number': ci.phone_number,
+                'address_first_line': ci.address1,
+                'address_city': ci.city,
+                'address_postcode': ci.postcode,
+                'address_country': supplier.registration_country,
+            }
+        })
+
+    return jsonify(suppliers=[supplier for supplier in supplier_rows]), 200
+
+
 @main.route('/suppliers/<int:supplier_id>', methods=['GET'])
 def get_supplier(supplier_id):
     supplier = Supplier.query.filter(
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -116,8 +116,21 @@ def export_suppliers_for_framework(framework_slug):
 
     supplier_rows = []
 
+    suppliers_with_a_complete_service = frozenset(framework.get_supplier_ids_for_completed_service())
+
     for sf, supplier, ci in suppliers_and_framework:
         declaration_status = sf.declaration.get('status') if sf.declaration else 'unstarted'
+        # For G10 we need to check the suppliers company detatils have been confirmed
+        company_details_confirmed = supplier.company_details_confirmed
+        application_status = 'application' if (
+                declaration_status == 'complete' and
+                supplier.supplier_id in suppliers_with_a_complete_service and
+                company_details_confirmed
+        ) else 'no_application'
+        application_result = 'no result'
+        framework_agreement = False
+        variations_agreed = ''
+
         supplier_rows.append({
             "supplier_id": supplier.supplier_id,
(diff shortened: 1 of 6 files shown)


app/main/views/suppliers.py
```diff
@@ -88,6 +90,114 @@ def list_suppliers():
...
+            "duns_number": supplier.duns_number,
+            "registered_name": supplier.registered_name,
+            "companies_house_number": supplier.companies_house_number,
+            # TODO: framework application result, framework agreement, variations agreed
+            'application_result': application_result,
+            'application_status': application_status,
+            'declaration_status': declaration_status,
+            'framework_agreement': framework_agreement,
+            'variations_agreed': variations_agreed,
+            # TODO: service counts for each lot
```
risicle: Can these `TODO`s be removed now?

app/main/views/suppliers.py
```diff
@@ -88,6 +90,114 @@ def list_suppliers():
...
+                'contact_email': ci.email,
+                'contact_phone_number': ci.phone_number,
+                'address_first_line': ci.address1,
+                'address_city': ci.city,
+                'address_postcode': ci.postcode,
+                'address_country': supplier.registration_country,
+            }
+        })
+
+    return jsonify(suppliers=[supplier for supplier in supplier_rows]), 200
```
risicle: What does this list comprehension do? Can't you just use `supplier_rows` directly?

katstevens: That is an extremely good question. We copied the pattern from the equivalent `users` export, which has been like that for... 3 years? https://github.com/alphagov/digitalmarketplace-api/commit/5dc89f003fe6a90346119487d6b5d02499c04933#diff-1b3a761e88acb979c136b48cf15b3976

katstevens: I've removed them in both export views.

app/main/views/suppliers.py
```diff
@@ -88,6 +90,112 @@ def list_suppliers():
...
+        abort(400, 'framework not yet open')
+
+    lot_slugs_by_id = {lot.id: lot.slug for lot in framework.lots}
+
+    # Creates a dictionary of dictionaries, with published service counts per lot per supplier.
+    # To do this we use dictionary comprehension, SQL query and itertools' 'groupby' function.
+    # Grouping by supplier id is more efficient as this is what we will be iterating over later.
+    service_counts_by_lot_by_supplier = {
+        supplier_id: {row[1]: row[2] for row in rows}
+        for supplier_id, rows in groupby(
```
benvand: The variable rows here isn't very descriptive. It feels like we're unessecarily complicating things by trying to do this in one line.

If we assign the query to a separate variable we can be more descriptive about what it is:

```
    supplier_id_lot_id_service_count_tuples = db.session.query(
        Service.supplier_id,
        Service.lot_id,
        func.count(Service.id)
    ).filter(
        Service.status == 'published',
        Service.framework_id == framework.id
    ).group_by(
        Service.supplier_id,
        Service.lot_id,
    ).all()
```

benvand: Then the comprehension/ groupby also seems a bit over involved and has a lot of `get`s used for setting default values. 
There is a method in collections called defaultdict that you can use to set defaults when unpacking stuff https://docs.python.org/2/library/collections.html#collections.defaultdict

The usage here would look something like this:
```
    # Convert the above to a dictionary like {supplier_id: {lot_1_slug: lot_1_count, lot_2_slug: lot_2_count}}
    service_counts_by_lot_by_supplier_id = defaultdict(dict)
    for supplier_id, lot_id, service_count in supplier_id_lot_id_service_count_tuples:
        service_counts_by_lot_by_supplier_id.setdefault(supplier_id, {}).update({lot_id: service_count})
```

benvand: You could alos unpack the lot slug here to save doing it further down:
```
    # Convert the above to a dictionary like {supplier_id: {lot_1_slug: lot_1_count, lot_2_slug: lot_2_count}}
    service_counts_by_lot_by_supplier_id = defaultdict(dict)
    for supplier_id, lot_id, service_count in supplier_id_lot_id_service_count_tuples:
        service_counts_by_lot_by_supplier_id.setdefault(supplier_id, {}).update({lot_slugs_by_id[lot_id]: service_count})

```

benvand: Then instead of the comprehension on line 181 you could set the defaults on that and away you go 🤷‍♀️

tests/bases.py
```diff
@@ -72,6 +72,13 @@ def teardown(self):
                 db.engine.execute(table.delete())
         FrameworkLot.query.filter(FrameworkLot.framework_id >= 100).delete()
         Framework.query.filter(Framework.id >= 100).delete()
+
```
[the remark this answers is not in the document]
benvand: Shame we're still doing this but it's a tricky one because the frameworks are added by the migrations :/

tests/main/views/test_suppliers.py
```diff
@@ -1932,3 +1932,298 @@ def test_agree_variation_various(
         assert audit_events[1].data["update"] == {
             "agreedUserId": 2,
         }
+
```
benvand: So I'm not going to suggest we rewrite these tests using model creation but I do want to take a bit of time to offer a critique of why I think they have some drawbacks.

First off, disclaimer, and I know they were inspired by the code that tests the user export, so it was a totally logical choice to use this pattern here and one which I think was, given the info at hand a good choice. 
I'm seriously all for following existing patterns and trusting previous bits of code. So in many ways using this pattern was the right thing to do. _That said_ I don't think it's a great pattern and here's why:

The specific problem I have with these tests is how they set up their test data.

In many cases they use views that already exist to set up test data. This might seem like a logical thing to do and really it sort of is. That's what the views are for after all. But it comes with quite a big drawback. They test _too much_.

These tests now will fail if someone changes something in the views on which they rely. Therefore they test not only the `supplier_export` but also all the associated views they hit. If they can fail when I change something then they are testing that thing.

So why does that matter? If I change something in the `draft-services` end point and tests in `test_suppliers` start failing then thats extra noise I really don't need and it detracts from the overall usefulness of the test suite. I have to find out why they're failing, and this happened with the debugging @katstevens had to do to find out what was happening to the framework object.

Basically if a unit test tests _everything_ then they don't test any one thing very effectively. 

Really useful unit tests obviously test only one unit of code. And one part of the codebase.
For these tests specifically that part of the codebase should be the relationship/ contract between what's in the database and what comes out of the api. It would only be testing the code between the database and the end of the view function (e the view function).

So if we only want to test the relationship between what's in the DB and what comes out of the api (so we know that the `export_suppliers` view has been written correctly) then how should we do it?

Have some input data (some test data that goes directly into the DB). Put it through our process (call the view) and compare the output to a set of expected values.

That way we know that whatever is causing the failures is in the code we wrote.

It's really tempting to test as much code as possible with unit tests but it makes them less useful overall if they fail every time a piece of code changes.
If you change a piece of code in the `supplier_export` view and 30 tests break then that is not useful. 
If you change a piece of code in the `supplier_export` view and 1 test called `test_correct_supplier_export_headings` breaks then that is super useful. 

Just my 2 cents. And as I say don't change them now!

app/main/views/suppliers.py
risicle: I strongly disagree with all these assertions about style. This is how I used to write things many years ago and there's a reason I migrated away from it - it is far harder to understand what high level operation the user is trying to achieve. Written in a functional style, the code structure tends to mirror the structure of the resulting output There is no such link having to follow lines and lines of imperative code which could be doing anything. *Describe the cake to me*, don't read me the long recipe.

However I'm not going to continue this discussion as it's not fruitful and is just bikeshedding in the extreme. But I think it should be kept in mind that these suggestions are just your *opinion*, which is not necessarily agreed with.

tests/main/views/test_suppliers.py
katstevens: I have to agree that I don't like using API methods in this way to set up test data. While it's a nice simulation of how the data gets into that state, we... don't care about that for these tests? We shouldn't make a habit of this pattern IMHO.

app/main/views/suppliers.py
risicle: (or, if you want another argument and you still have faith in them, look at the mccabe scores of the two approaches...)

tests/main/views/test_suppliers.py
risicle: > Basically if a unit test tests everything then they don't test any one thing very effectively.
> Really useful unit tests obviously test only one unit of code.

Though I agree that it's best not to use api endpoints to do your setup (mostly because of performance), you've made a broader philosophical statement so you're going to get a broader philosophical response...

I strongly disagree with dogmatic beliefs about the "technical" difference between unit tests and functional-integration-whatever-you-want-to-call-it-tests and don't think we should be propagating them. Nothing is purely a unit test. Nothing is purely a functional test. It's a spectrum. And also what is most effective in any particular situation varies massively. The important thing is whether it asserts things that you care about. The highly "technical" differences are largely made up by people who want to sell books. The reality is that if all you have is "pure" unit tests, you will *never* write enough of them to be able to properly assert the correct high level behaviour of your system and at that point they will take ages to run. They will also be close to useless come refactoring time. The same amount of code can usually be exercised in a much smaller & less fragile amount of test code with a *higher level* test. Testing is a place for pragma.

Insert "two passing unit tests, zero integration tests" image macro of kitchen drawers here.

PR #799, opened 2018-06-11
Update allowed failed login attempts number from 5 to 10

CrystalPea:
Trello ticket: https://trello.com/c/alSuFp7x/457-give-users-10-password-attempts


CrystalPea:
@galund Good idea, I did that now :).

PR #800, opened 2018-06-12
Outcome: change serialization of ids

risicle:
As discussed, embed ids in their own dictionaries as if we were just returning a very abbreviated version of these objects serializations.  This makes more sense and would mean that future changes to include more details from each object could be backwards compatible.

Also have slipped in a patch here to add `Outcome` and `BriefResponse` to `AUDIT_OBJECT_TYPES` so we're actually able to search by these in `list_audits`.

### DIFF
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -29,6 +29,8 @@ AUDIT_OBJECT_TYPES = {
     "frameworks": models.Framework,
     "users": models.User,
     "briefs": models.Brief,
+    "outcomes": models.Outcome,
+    "brief-responses": models.BriefResponse,
 }
 
 AUDIT_OBJECT_ID_FIELDS = {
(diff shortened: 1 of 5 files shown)


tests/main/views/test_direct_award.py
```diff
@@ -1248,10 +1248,18 @@ def test_direct_award_outcome_scenarios(
...
                     "resultOfDirectAward": {
-                        "projectId": project_external_id,
-                        "searchId": active_search_id,
-                        "serviceId": "2000000001",
-                        "archivedServiceId": chosen_archived_service_id,
+                        "project": {
+                            "id": project_external_id,
+                        },
+                        "search": {
+                            "id": active_search_id,
```
galund: It's interesting that we went to the effort of making proper public identifiers for projects, but never did it for searches?

app/main/views/audits.py
```diff
@@ -37,6 +39,7 @@
     "frameworks": models.Framework.slug,
     "users": models.User.id,
     "briefs": models.Brief.id,
+    "outcomes": models.Outcome.external_id,
```
samuelhwilliams: need to add `brief-responses` here

risicle: Do I? Didn't think `BriefResponse`s had a special external id.

risicle: Jeez, looks like I need to include it anyway...

PR #801, opened 2018-06-13
Remove frameworkStartDate and frameworkEndDate

samuelhwilliams:
## Summary
These two fields can be removed from the frameworkAgreementDetails blob
of data on a framework record as they duplicate information contained at
the top level under frameworkLiveAtUTC and frameworkExpiresAtUTC.

I have already removed these keys from the records across all of our
environments.

 ## Ticket
https://trello.com/c/82u630c9/80


samuelhwilliams:
Sorry yes, that's what I meant by "I have already removed these keys from the records across all of our environments."

I've removed them from the records in the DB that had them set (G-Cloud 9/10 and DOS1+2).

PR #802, opened 2018-06-18
Remove personal data methods

benvand:
** Requires https://github.com/alphagov/digitalmarketplace-apiclient/pull/142 **

https://trello.com/c/gsDnUsC3/208-delete-users-code-that-takes-the-user-and-strips-all-the-personal-data-out

We need a mechanism to remove personal data from the `users` and `contact_information` tables.

This PR:
* Creates a base class that adds a `personal_data_removed` flag
  * If this flag is set the object is made effectively read only by a listener that denies new changes
* Adds this base class to `User` and `ContactInformation`
* Adds methods to remove data to `User` and `ContactInformation`
* Adds views to trigger those methods and audit the action
* Adds tests for the new model methods and
* Adds tests for the new views

### DIFF
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -3,16 +3,15 @@
 import re
 from abc import ABCMeta, abstractmethod
 from datetime import datetime
-from six import string_types, iteritems
 from uuid import uuid4
 
 from flask import current_app
 from flask_sqlalchemy import BaseQuery
 
+import sqlalchemy.dialects.postgresql
 from sqlalchemy import Sequence
 from sqlalchemy import asc, desc, exists
 from sqlalchemy import func
-import sqlalchemy.dialects.postgresql
 from sqlalchemy.dialects.postgresql import INTERVAL
 from sqlalchemy.event import listen
 from sqlalchemy.ext.declarative import declared_attr
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -1,6 +1,8 @@
 # TODO split this file into per-functional-area modules
 
+from abc import ABCMeta, abstractmethod
 from datetime import datetime
+from uuid import uuid4
 
 import re
 import sqlalchemy.dialects.postgresql
@@ -10,10 +12,11 @@ from sqlalchemy import Sequence
 from sqlalchemy import asc, desc, exists
 from sqlalchemy import func
 from sqlalchemy.dialects.postgresql import INTERVAL
+from sqlalchemy.event import listen
 from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import validates, backref, mapper, foreign, remote
-from sqlalchemy.orm.session import Session
+from sqlalchemy.orm.session import Session, object_session
 from sqlalchemy.sql.expression import (
     case as sql_case,
     cast as sql_cast,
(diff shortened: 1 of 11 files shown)


app/models/main.py
```diff
@@ -60,6 +60,35 @@ def __init__(self, astext_type=None):
...
+    @staticmethod
+    def validate_personal_data_removed(mapper, connection, instance):
+        # """The only time we should be able to update an object with """
+        session = object_session(instance)
+        model_class = instance.__class__
+        if session.query(model_class.personal_data_removed).filter(model_class.id == instance.id).scalar():
+            raise ValidationError("Cannot update an object once personal data has been removed")
+
+
+listen(
```
benvand: Had to manually write an event listener instead of using the `@validates` decorator because we need this to be triggered on every save of an object.

tests/helpers.py
```diff
@@ -50,9 +50,11 @@ def fixture_params(fixture_name, params):
```
benvand: Changed this to be setable by individual tests so we can change it depending on what test data we're using

tests/models/test_main.py
```diff
@@ -145,6 +145,174 @@ def test_user_with_invalid_email_domain_cannot_be_changed_to_admin_role(self, ro
             valid_buyer_user.role = role
```
benvand: Test views

tests/main/views/test_suppliers.py
```diff
@@ -790,6 +790,45 @@ def test_update_without_updated_by(self):
         assert response.status_code == 400
```
benvand: Test model methods

app/models/main.py
benvand: `@validates` decorators only trigger when their field is changed

app/models/main.py
```diff
@@ -60,6 +60,35 @@ def __init__(self, astext_type=None):
         super(JSON, self).__init__(none_as_null=True, astext_type=astext_type)
 
 
+class RemovePersonalDataModel:
```
katstevens: As discussed, let's rename this to make it clearer that it's a mixin and not its own table (and update the docstring for the method).

tests/main/views/test_users.py
```diff
@@ -1253,6 +1253,117 @@ def test_400_returned_if_role_param_is_not_valid(self):
...
+            content_type='application/json'
+        )
+        assert response.status_code == 200
+
+        data = json.loads(response.get_data())
+        assert data['users']['active'] is False
+        assert data['users']['name'] == '<removed>'
+        assert data['users']['phoneNumber'] == '<removed>'
+        assert data['users']['emailAddress'] == '<removed>@digital.cabinet-office.gov.uk'
+        assert data['users']['userResearchOptedIn'] is False
```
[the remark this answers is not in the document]
benvand: Done 👍

app/models/main.py
benvand: Done 👍

tests/main/views/test_suppliers.py
```diff
@@ -790,6 +790,45 @@ def test_update_without_updated_by(self):
...
+        assert response.status_code == 200
+
+        data = json.loads(response.get_data())
+
+        assert data['contactInformation']['address1'] == '<removed>'
+        assert data['contactInformation']['city'] == '<removed>'
+        assert data['contactInformation']['contactName'] == '<removed>'
+        assert data['contactInformation']['email'] == '<removed>'
+        assert data['contactInformation']['phoneNumber'] == '<removed>'
+        assert data['contactInformation']['postcode'] == '<removed>'
```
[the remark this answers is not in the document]
benvand: Done 👍

PR #804, opened 2018-06-20
Update supplier framework details on confirmation 

samuelhwilliams:
Commits 2+3 are refactor only. Commit 1 sets ContactInformation to be returned on a Supplier record in a consistent order.

 ## Summary
A supplier has to confirm that their company details are correct as part
of an application to a framework. They need to do this every time they
apply to a framework to ensure that they are still the same company. To
track this, we need to have two concepts of company details being
confirmed:

1) Company details get confirmed at the account level. This confirms
that their account is configured correctly to represent the entity they
will be applying with (specifically this locks down four attributes for
now: company name and number, vat number, and duns number).
2) Company details are confirmed as being accurate for a specific
framework application. They must confirm the details are still correct
for the application they are making, as some of these things change over
time. If a significant change happens (e.g. new duns number), they will
need to apply with a new supplier account.

An application is made when the details are confirmed for that framework
application, so we will update the logic for these checks as well.

For G-Cloud 10, we injected all supplier company details into the
SupplierFramework when the Framework was set to pending. This is no
longer required now that we are injecting the details on every relevant
update.

 ## Ticket
https://trello.com/c/wrz3NRG3/171


app/supplier_utils.py
```diff
@@ -37,3 +44,80 @@ def check_supplier_role(role, supplier_id):
...
+    return (
+        supplier_framework.supplier.company_details_confirmed
+        if framework_slug == 'g-cloud-10' else
+        supplier_framework.application_company_details_confirmed
+    )
+
+
+def update_open_declarations_with_company_details(db, supplier_id, updater_json, specific_framework_slug=None):
+    """Expected to be called within a view"""
+    open_supplier_frameworks_query = SupplierFramework.query.filter(
```
Wynndow: Could rename this var `supplier_frameworks_query` to be clearer. Or could add the check for frameworks being open here and then not worry about the `else` below.

app/supplier_utils.py
```diff
@@ -37,3 +44,80 @@ def check_supplier_role(role, supplier_id):
...
+        SupplierFramework.supplier_id == supplier_id,
+    )
+
+    if specific_framework_slug:
+        open_supplier_frameworks_query = open_supplier_frameworks_query.filter(
+            SupplierFramework.framework.has(Framework.slug == specific_framework_slug),
```
Wynndow: This doesn't actually check the framework is open? As above, could make the initial query look for open frameworks, and then just have the check for `specific_framework_slug` here and drop the `else` below.

app/supplier_utils.py
```diff
@@ -37,3 +44,80 @@ def check_supplier_role(role, supplier_id):
...
+        )
+    else:
+        open_supplier_frameworks_query = open_supplier_frameworks_query.filter(
+            SupplierFramework.framework.has(Framework.status == 'open'),
+        )
+
+    # We need to ensure that SupplierFramework and Framework haven't changed during this transaction, so grab
+    # the affected rows with a for update lock.
+    open_supplier_frameworks = open_supplier_frameworks_query.options(lazyload('*')).with_for_update().all()
```
Wynndow: Is the Framework going to be loaded here with `with_for_update`? I'm guessing it is because the query's above include a check on the Framework (namely its status) and so they will be loaded with the `lazyload` option? Otherwise would you need to explicitly exclude Framework from the lazyload option?

app/supplier_utils.py
```diff
@@ -37,3 +44,80 @@ def check_supplier_role(role, supplier_id):
...
+            supplier_constants.KEY_VAT_NUMBER: supplier.vat_number,
+            supplier_constants.KEY_TRADING_STATUS: supplier.trading_status,
+            supplier_constants.KEY_ORGANISATION_SIZE: supplier.organisation_size,
+            supplier_constants.KEY_REGISTRATION_COUNTRY: supplier.registration_country,
+            supplier_constants.KEY_REGISTRATION_BUILDING: supplier.contact_information[0].address1,
+            supplier_constants.KEY_REGISTRATION_TOWN: supplier.contact_information[0].city,
+            supplier_constants.KEY_REGISTRATION_POSTCODE: supplier.contact_information[0].postcode,
+        }
+
+        company_details_without_nones = {k: v for k, v in company_details.items() if v is not None}
```
Wynndow: Could this be just `{k: v for k, v in company_details.items() if v}`

app/supplier_utils.py
```diff
@@ -37,3 +44,80 @@ def check_supplier_role(role, supplier_id):
...
+
+        open_supplier_framework.declaration = {
```
Wynndow: Why not `open_supplier_framework.declaration.update(company_details_without_nones)`?

Wynndow: Also, hypothetical situation, but if a supplier initially put a `companies_house_number`, and that got added to the declaration, then they realised that it should have been a `other_company_registration_number`, or visa versa, it there a risk of the old key being left in the declaration?

app/supplier_utils.py
```diff
@@ -37,3 +44,80 @@ def check_supplier_role(role, supplier_id):
...
+
+        open_supplier_framework.company_details_confirmed = False
+
+        db.session.add(open_supplier_framework)
+
+        calling_function = inspect.stack()[1].function
+        db.session.add(
+            AuditEvent(
+                audit_type=AuditTypes.answer_selection_questions,
+                db_object=open_supplier_framework,
```
Wynndow: Potentially the object should be the supplier for ease of search, as we can't currently search the data?

tests/main/views/test_suppliers.py
```diff
@@ -1542,6 +1605,78 @@ def test_get_supplier_framework_info_with_non_existent_supplier(self, supplier_f
 
         assert response.status_code == 404
 
+    def test_set_application_company_details_confirmed_updates_declaration(self, supplier_framework):
```
Wynndow: Maybe this should go in the class below.

app/models/main.py
```diff
@@ -529,6 +534,12 @@ class SupplierFramework(db.Model):
         nullable=True,
     )
 
+    # Suppliers must confirm that their details are accurate for every framework application they make.
+    # The flag `company_details_confirmed` is set on the Supplier object the first time a supplier confirms their
+    # details (company name+number, vat number, duns number). The below SupplierFramework flag is set when a
+    # supplier confirms their details with an open application to an iteration of a framework.
+    application_company_details_confirmed = db.Column(db.Boolean, index=False, default=False, nullable=True)
```
Wynndow: Do you need to specify `False` for index? Isn't it the default?
Is `default=False` the kind of default that isn't going to lock everything when the migration is run?

app/supplier_utils.py
```diff
@@ -1,6 +1,13 @@
 from flask import abort
+import inspect
```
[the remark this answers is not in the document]
samuelhwilliams: Yep

migrations/versions/1230_add_company_details_confirmed_to_.py
```diff
@@ -0,0 +1,28 @@
+"""add company_details_confirmed to supplier framework
```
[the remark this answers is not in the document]
samuelhwilliams: Yep, I'm fine leaving it though.

app/supplier_utils.py
samuelhwilliams: Kept it as-is after moving the `open` filter up to here.

app/supplier_utils.py
samuelhwilliams: 👍

app/supplier_utils.py
samuelhwilliams: Could be, but there's not a lot of difference 🤷‍♂️

Wynndow: It made me start to wonder if some of they keys could be `False` and this was a way to differentiate?

app/supplier_utils.py
samuelhwilliams: The other place that does an `answer_selection_questions` audit event has the object as the supplier framework, so this is done for consistency's sake.

tests/main/views/test_suppliers.py
samuelhwilliams: Yep 👍

app/models/main.py
samuelhwilliams: Don't need to specify it, but explicit is better than implicit 🤷‍♂️ 

`default=False` won't lock the table here as we're setting it to nullable. This just means in the future it will default to False, but it won't touch existing rows.

app/supplier_utils.py
samuelhwilliams: `open_supplier_framework.declaration.update(company_details_without_nones)` - this encounters the issue with Postgres/SQLAlchemy not recognising changes within JSON blobs. Creating a new dict altogether means that the changes are recognised and persisted to the db correctly.

app/supplier_utils.py
samuelhwilliams: Theoretically, if some of these keys were set to False, that would imply it's data provided by the supplier so still want to keep them - so the explicit check vs None is better here.

app/supplier_utils.py
samuelhwilliams: This is the emitted statement, which selects from both tables, so they should both get locked.

```SQL
SELECT supplier_frameworks.supplier_id AS supplier_frameworks_supplier_id, supplier_frameworks.framework_id AS supplier_frameworks_framework_id, supplier_frameworks.declaration AS supplier_frameworks_declaration, supplier_frameworks.on_framework AS supplier_frameworks_on_framework, supplier_frameworks.agreed_variations AS supplier_frameworks_agreed_variations, supplier_frameworks.prefill_declaration_from_framework_id AS supplier_frameworks_prefill_declaration_from_framework_id, supplier_frameworks.application_company_details_confirmed AS supplier_frameworks_application_company_details_confirmed
FROM supplier_frameworks
WHERE supplier_frameworks.supplier_id = %(supplier_id_1)s AND (EXISTS (SELECT 1
FROM frameworks
WHERE frameworks.id = supplier_frameworks.framework_id AND frameworks.status = %(status_1)s)) AND (EXISTS (SELECT 1
FROM frameworks
WHERE frameworks.id = supplier_frameworks.framework_id AND frameworks.slug = %(slug_1)s)) FOR UPDATE
```

app/supplier_utils.py
Wynndow: Cool. But regarding my point about companies house numbers, if the declaration had already been updated with a companies house number, is the unpacking method going to remove the old key? I don't think either way would...

migrations/versions/1230_add_company_details_confirmed_to_.py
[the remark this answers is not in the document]
Wynndow: really? All the other migrations you can get a reasonable idea of what they do just be reading the file name. Feels bad to leave this like it is.

app/supplier_utils.py
samuelhwilliams: Sorry, forgot to address that bit.

When a supplier updates either of these numbers, the other one is wiped out from the Supplier record. This method then denormalizes the two fields (specifically, taking the value of whichever is set) under a single name 'supplierCompanyRegistrationNumber', which can be _either_ a companies house number or 'another' company registration number.

So no, there's no chance of both of them ending up here in the declaration at the same time.

migrations/versions/1230_add_company_details_confirmed_to_.py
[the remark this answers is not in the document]
samuelhwilliams: There are plenty of other truncated migration filenames... But sure, I'll fix this up.

PR #806, opened 2018-06-21
[WIP] Strawman - centralise application status

samuelhwilliams:
## Summary
Total strawman for centralising the logic around whether an application
has been made to a framework. This logic is currently fractured within
the API and also duplicated on the supplier-frontend. It's a prime
candidate for bringing together to one place. This feels really
important given how core the idea of making, or not making, an
application is to the platform.

The two endpoints in the API that use this piece of information are
`export_suppliers_for_framework` and `export_users_for_framework`. Part
of the reason for this logic not being on the SupplierFramework before
is probably to keep these endpoints performant by doing one of the
checks (whether a supplier has submitted a service for the framework)
performant by running it only once, rather than checking for each
individual application. This is a valid concern and I'm open to ideas
as to how we can keep this endpoint returning in an acceptable timespan
while also keeping the application_status logic in a single place.

One of the tests here breaks because it submits a DOS application and
asserts that no application has been made, expecting it to fail on the
fact that company details must be confirmed. This is a slight misnomer,
as company details didn't need confirming for DOS1. The code is behaving
correctly for the framework under test; the test is currently invalid.
Fixing it may also require a larger refactor.

Alternatively should we store this as a legitimate top-level attribute on the SupplierFramework model so it can get locked in, rather than being dynamically generated?


app/models/main.py
```diff
@@ -577,6 +564,32 @@ def serialize_agreed_variation(agreed_variation, with_users=False):
...
+        # These 'old' frameworks had no concept of company details needing to be confirmed.
+        old_framework_slugs = {
+            'g-cloud-4', 'g-cloud-5', 'g-cloud-6', 'g-cloud-7', 'g-cloud-8', 'g-cloud-9',
+            'digital-outcomes-and-specialists', 'digital-outcomes-and-specialists-2',
+        }
+
+        company_details_confirmed_if_required_for_framework = (
+            True
+            if self.framework.slug in old_framework_slugs else
+            self.supplier.company_details_confirmed
```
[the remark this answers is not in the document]
samuelhwilliams: Ah yep that's a fair point 👍

app/models/main.py
```diff
@@ -577,6 +564,32 @@ def serialize_agreed_variation(agreed_variation, with_users=False):
...
+        )
+
+        return 'complete' if (
+            self.declaration.get('status', 'unstarted') == 'complete' and
+            submitted_services > 0 and
+            company_details_confirmed_if_required_for_framework
+        ) else 'incomplete'
+
```
risicle: _Suspect_ this whole thing is probably implementable as a `hybrid_property`

katstevens:
Yep, this idea looks good to me. Agree with Rob that a `hybrid_property` would be useful so we can make use of the SQL stuff in the same way we do for the other `status` properties.

Good work on removing the duplicate code from the export views, that was a forgotten item on my to-do list. The user export is fairly slow at the moment anyway, we deprioritised optimising it as it's not going to be frequently used.

app/models/main.py
```diff
@@ -635,6 +622,32 @@ def serialize_agreed_variation(agreed_variation, with_users=False):
             "agreedUserEmail": user.email_address,
         })
 
+    @property
+    def application_status(self):
```
galund: So perhaps the issue here is that making something a `property` can quite strongly suggest that it's quite "quick" to run (for some value of quick). But here we're running a DB query, so that isn't necessarily the case.

So maybe this is a nitpick but I think probably it shouldn't be a property. Not unless we can reasonably cache the value, anyway, which might add some unnecessary complexity.

app/models/main.py
```diff
@@ -635,6 +622,32 @@ def serialize_agreed_variation(agreed_variation, with_users=False):
...
+        submitted_services = DraftService.query.filter(
+            DraftService.status.in_(('submitted', 'failed',)),
+            DraftService.framework_id == self.framework_id,
+            DraftService.supplier_id == self.supplier_id
+        ).count()
```
galund: We might not actually need to run this query at all depending on whether company details are required, so some re-ordering of the logic here might make things much more efficient in certain cases.

app/models/main.py
samuelhwilliams: I suppose - although it basically acts the same as a lazy loaded relationship on the model, which we have in plenty of other places? All those would need to hit the DB to load the information they need.

samuelhwilliams:
Yep, by all means this is just to get a conversation going. I think after the tech debt meeting on Monday it might be more sensible for this to be a locked-in value rather than a dynamically-derived one, anyway, so I guess that's what I'm more interested in hashing out here as opposed to the quick code I slapped together.

app/models/main.py
risicle: ^this. Simply accessing a lazy loaded relationship (of which we have plenty) can cause signifacant database hits. Done as a hybrid property has the advantage that we can also filter by the property without having to pull extra things out of the db.

katstevens:
Going to close this PR for now - I think between the comments here and the Trello card https://trello.com/c/Anqo0IUv/492-no-centralised-concept-of-an-application-to-a-framework-having-been-made-or-not we have enough detail to decide if/when to prioritise it.

PR #807, opened 2018-06-22
Create an 'update project' endpoint

lfdebrux:
This PR adds the endpoint `PATCH /direct-award/projects/<id>`.

Currently the only column that can be updated is `stillAssessingAt`;
this was added for the buyer-frontend ticket [#110][ticket].

[ticket]: https://trello.com/c/5Lrp1lOK

### DIFF
--- a/app/main/views/direct_award.py
+++ b/app/main/views/direct_award.py
@@ -23,6 +23,8 @@ from ...utils import (
     validate_and_return_updater_request,
 )
 
+from ...validation import validate_direct_award_project_json_or_400
+
 
 def get_project_by_id_or_404(project_id: int):
     return DirectAwardProject.query.filter(DirectAwardProject.external_id == project_id).first_or_404()
@@ -139,6 +141,47 @@ def get_project(project_external_id):
     project = get_project_by_id_or_404(project_external_id)
 
     return single_result_response("project", project, serialize_kwargs={"with_users": True}), 200
+
+
+@main.route('/direct-award/projects/<int:project_external_id>', methods=['PATCH'])
+def update_project(project_external_id):
+    uniform_now = datetime.datetime.utcnow()
+
+    updater_json = validate_and_return_updater_request()
+
+    json_payload = get_json_from_request()
+    json_has_required_keys(json_payload, ["project"])
+    project_json = json_payload["project"]
+
+    validate_direct_award_project_json_or_400(project_json)
+
+    # fetch and lock row
+    project = db.session.query(DirectAwardProject).filter_by(
+        external_id=project_external_id,
+    ).with_for_update().first()
+
+    if not project:
+        abort(404, f"Project {project_external_id} not found")
+
+    # TODO: project.update_from_json(project_json)
+
+    if project_json.get('stillAssessing'):
+        if project.outcome:
+            abort(400, "Cannot update stillAssessing when an outcome is present")
+        project.still_assessing_at = uniform_now
+
+    update_audit_event = AuditEvent(
+        audit_type=AuditTypes.update_project,
+        user=updater_json['updated_by'],
+        db_object=project,
+        data=project_json,
+    )
+    update_audit_event.created_at = uniform_now
+    db.session.add(update_audit_event)
+
+    db.session.commit()
+
+    return single_result_response('project', project), 200
 
 
 @main.route('/direct-award/projects/<int:project_external_id>/searches', methods=['GET'])
--- a/tests/main/views/test_direct_award.py
+++ b/tests/main/views/test_direct_award.py
@@ -26,7 +26,12 @@ from app.models.direct_award import (
     DirectAwardSearchResultEntry
 )
 from ...helpers import (
-    DIRECT_AWARD_SEARCH_URL, DIRECT_AWARD_PROJECT_NAME, DIRECT_AWARD_FROZEN_TIME, load_example_listing, FixtureMixin
+    DIRECT_AWARD_SEARCH_URL,
+    DIRECT_AWARD_PROJECT_NAME,
+    DIRECT_AWARD_FROZEN_TIME,
+    DIRECT_AWARD_FROZEN_TIME_DATETIME,
+    load_example_listing,
+    FixtureMixin
 )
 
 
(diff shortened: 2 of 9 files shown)


app/main/views/direct_award.py
```diff
@@ -125,6 +127,45 @@ def get_project(project_external_id):
     return single_result_response("project", project, serialize_kwargs={"with_users": True}), 200
 
 
+@main.route('/direct-award/projects/<int:project_external_id>', methods=['PUT'])
```
galund: This should be POST as you aren't completely replacing the resource, merely updating bits of it.

lfdebrux: Hmmm, so the other update views I looked at use PUT

risicle: Yeees, we seem to have mixed opinions about this in DM, I've always thought it was the other way around but I'm not hugely fussed either way...

galund: I didn't realise there was any controversy. There isn't in terms of the specs. https://stackoverflow.com/questions/19732423/why-isnt-http-put-allowed-to-do-partial-updates-in-a-rest-api

I generally prefer consistency over most other considerations, but in the case of basic HTTP stuff I tend to get a bit militant/purist :-)

samuelhwilliams: My understanding is that POSTs shouldn't occur at the individual resource level either, so doing partial updates has always been really confusing for me. Open to be taught that this is fine, but everything I can recall reading says that POST is for /direct-award/projects, PUT is for overwriting /direct-award/projects/<project_id>, PATCH is for instructions to modify /direct-award/projects/<project_id>.

lfdebrux: I'm going to change this to `PATCH`, based on the advice of http://jsonapi.org/format/#crud.

I agree with @samuelhwilliams that `POST` shouldn't be done at the individual resource level, but I also agree with @galund that `PUT` should not be used for partial updates. To me using `PATCH` seems like a good compromise, as by being neither PUT nor POST it indicates that it is new semantics.

This will of course be inconsistent with the rest of the codebase, which currently doesn't use PATCH anywhere. However, we are already internally inconsistent, so I think that means we can do as @galund said and put consistency with the HTTP spec first.

Of course, if anyone disagrees with any of this, please do make your thoughts known.

benvand: So I'd have to weigh in against introducing a new way of doing things without blanket updating the code. I think adding a single `PATCH` route to our API increases complexity, inconsistency and tech-debt unnecessarily. If we really want to change the patterns we use for making api requests then we should do it globally, and we should make the decision that it is worth doing globally.
We have a consistent way of updating objects at the moment (`POST` requests). Just because there is a better way doesn't mean we should immediately leap to using it without considering the longer term implications.

If we start using PATCH then we are committing our future selves to do work that we don't think is worth doing _right now_ (ie updating all our routes to have consistent request methods).

However, if we all agree that we should have end points consistent with the jsonapi spec, and we agree that that work is worth doing, then we should update them all now.

lfdebrux:
So I'm going to put myself out on a limb here and plump for `PATCH`.

If I can summarise the conversations we've had online and offline:
- we use both POST and PUT (inconsistently) for partial update endpoints
- using PUT for a partial update endpoint is definitely against the HTTP spec
- using POST for a single item in a collection is _arguably_ against the HTTP spec
- POST is often used in instances where no other verb is appropriate anyway
- PATCH was put in the spec for partial update of collection items
- PATCH data is supposed to strictly follow certain semantics depending on the content-type, which we don't want to use

My take away from this is that there are some good arguments for PATCH and some good arguments 
against, and the only argument _for_ using POST is that it usually used as a catch-all anyway. For me, the benefits of having a future consistent semantics outweighs any downsides of using PATCH in a laissez-faire way.

Obviously, in the short term there will be more inconsistency in the api, but I have created a tech debt ticket (https://trello.com/c/ud2H7BKJ/496-make-api-verbs-consistent) to go through and bring consistency that I will take up once this PR has been merged.

tl;dr let's do PATCH

risicle:
Damn you I'm going to have to add another symbol to my log diagrammer.

tests/main/views/test_direct_award.py
```diff
@@ -1691,3 +1696,66 @@ def test_direct_award_nonawarded_scenarios(
...
+        db.session.commit()
+
+        res = self.client.patch(f'/direct-award/projects/{project.external_id}',
+                                content_type='application/json',
+                                data=json.dumps({
+                                    'project': {},
+                                    'updated_by': 'example@example.com',
+                                }))
+
+        assert res.status_code == 200
```
risicle: You could also assert here very cheaply that `stillAssessingAt` *wasn't* updated by this.

app/main/views/direct_award.py
```diff
@@ -141,6 +143,45 @@ def get_project(project_external_id):
...
+    project = db.session.query(DirectAwardProject).filter_by(
+        external_id=project_external_id,
+    ).with_for_update().first()
+
+    if not project:
+        abort(404, f"Project {project_external_id} not found")
+
+    # TODO: project.update_from_json(project_json)
+
+    if project_json.get('stillAssessing'):
```
risicle: Here's a thought - should we be allowing `stillAssessing` to be set if we've already got an `outcome` for this project?

tests/main/views/test_direct_award.py
lfdebrux: I decided not to do that because it kind of goes against the 'test one thing' philosophy.

I see the value of testing that `stillAssessingAt` doesn't change though, so I will create a separate one that does this.

app/main/views/direct_award.py
lfdebrux: Will do

lfdebrux: Should we then make it so that `stillAssessingAt` is set to `None` when an outcome is created?

tests/main/views/test_direct_award.py
risicle: Ok. Twice the running time and twice the volume of code to maintain though.

app/main/views/direct_award.py
risicle: I would say no, we still want to retain that data for our records. If you think it feels weird it perhaps would make sense to think of the `stillAssessingAt` field as "was still assessing at <this time>" rather than a field flagging that the project _is_ currently in a particular state.

tests/main/views/test_direct_award.py
galund: (personally I haven't ever been convinced by the "test one thing" argument, and you're unlikely to see me insist on it)

lfdebrux: Done

app/main/views/direct_award.py
lfdebrux: Done

katstevens:
Agree with PATCH for partial updates, as long as it's consistently rolled out across both the API and the Search API. 

The [JSON API spec](http://jsonapi.org/format/#crud-updating) has sensible recommendations about nulls/missing fields.

We can add a new `foo_api_client._patch` method on the base API client to go alongside the `._put` methods to ease the transition.

tests/main/views/test_direct_award.py
samuelhwilliams: "Twice the running time" coming as a comment from the person who adds 100s to 1000s of tests at a time with parametrization? 😛

app/main/views/direct_award.py
```diff
@@ -141,6 +143,45 @@ def get_project(project_external_id):
     return single_result_response("project", project, serialize_kwargs={"with_users": True}), 200
 
 
+@main.route('/direct-award/projects/<int:project_external_id>', methods=['PATCH'])
+def update_project(project_external_id):
+    uniform_now = datetime.datetime.utcnow()
```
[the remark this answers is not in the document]
lfdebrux: I'm not sure I understand what you mean exactly, so if you could create the tech debt ticket that would be appreciated.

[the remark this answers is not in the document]
risicle: You'll see it in most things I write that use a notion of "current time". Haven't gone as far as pulling it from the db because most the time I do this kind of thing outside the api. But then again, i'm not sure we really currently use db time at all - all our "default now"s use `datetime` don't they?

tests/main/views/test_direct_award.py
risicle: And each of those would be 5 or 10 times longer if they didn't make multiple assertions per pass.

PR #808, opened 2018-06-25
list_projects: add ability to filter by `having_outcome`

risicle:
This is to enable https://trello.com/c/4Ub332Ty/70-fe-7-you-need-to-tell-us-the-outcome-for-1-saved-search-to-be-added-dynamic to be performed efficiently.

I can't think of anything else to say about it.

### DIFF
--- a/app/main/views/direct_award.py
+++ b/app/main/views/direct_award.py
@@ -52,6 +52,14 @@ def list_projects():
     if user_id:
         projects = projects.filter(DirectAwardProject.users.any(id=user_id))
 
+    if "having-outcome" in request.args:
+        projects = projects.filter(
+            db.cast(
+                DirectAwardProject.outcome.has(),
+                db.Boolean,
+            ) == convert_to_boolean(request.args.get("having-outcome"))
+        )
+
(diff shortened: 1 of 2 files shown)


risicle:
Actually come to think, I'm going to need to be able to filter by the project being locked too. So I may as well add that.

app/main/views/direct_award.py
```diff
@@ -52,6 +52,22 @@ def list_projects():
     if user_id:
         projects = projects.filter(DirectAwardProject.users.any(id=user_id))
 
+    if "having-outcome" in request.args:
+        projects = projects.filter(
+            db.cast(
+                DirectAwardProject.outcome.has(),
+                db.Boolean,
+            ) == convert_to_boolean(request.args.get("having-outcome"))
```
galund: I suppose you could use square brackets here seeing as you already tested it's there, whatevs :)

risicle: Um yes I guess I could have

risicle:
It's the sort of thing where I *bet* it would be the only ever use in the code

PR #809, opened 2018-06-26
Registered postcode can't be longer than 10 characters now

CrystalPea:
This has been done for consistency: labAddressPostcode field
is already validated in such a way.

Connected supplier-frontend pull request: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/904

Trello ticket: https://trello.com/c/kOGpfxcw/475-supplier-fe-postcode-field-validation-is-more-lax-than-api-and-varies-between-two-forms


samuelhwilliams:
We'll need to do a validation piece before this gets merged to see how many records (if any) break this limit and why. Unless you've already done that?

CrystalPea:
@samuelhwilliams I have not done this yet, no, but I thought some kind of check might be needed. How do we do it?

katstevens:
If we do have existing null values in the postcode field then we should probably remove the `minLength` validation (at least for now). In general, requiring a postcode feels like a framework/frontend specific thing rather than a database thing - not every address has a postcode.

PR #810, opened 2018-06-27
Update `framework_expires_at_utc` when framework is expired.

Wynndow:
This may or may not be a good idea. Do we want these dates to reflect
what we _thought_ they should be initially, or what they actually turned 
out to actually be? Once they're set, should they remain consistent? If
this is a good idea, should they all be updated on framework lifecycle
changes?

This date is used in content in the buyer frontend after a framework is
expired. If this date is wrong for any reason, the content will be wrong
and confusing. This ensures that the date is accurate.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -135,6 +135,22 @@ def update_framework(framework_slug):
             enforce_required=True
         )
 
+    # We can ensure that some of the timestamps we have for a framework are accurate, as they are tied to state changes.
+    status_timestamp_audit_data = {}
+    if json_framework.get('clarificationQuestionsOpen') is False and framework.clarification_questions_open is True:
+        framework.clarifications_close_at_utc = datetime.datetime.utcnow()
+        status_timestamp_audit_data = {"clarifications_close_at_utc set": "clarification questions closed"}
+
+    if json_framework.get('status') == 'pending' and framework.status != 'pending':
+        framework.applications_close_at_utc = datetime.datetime.utcnow()
+        status_timestamp_audit_data = {"applications_close_at_utc set": "framework status set to 'pending'"}
+    elif json_framework.get('status') == 'live' and framework.status != 'live':
+        framework.framework_live_at_utc = datetime.datetime.utcnow()
+        status_timestamp_audit_data = {"framework_live_at_utc set": "framework status set to 'live'"}
+    elif json_framework.get('status') == 'expired' and framework.status != 'expired':
+        framework.framework_expires_at_utc = datetime.datetime.utcnow()
+        status_timestamp_audit_data = {"framework_expires_at_utc set": "framework status set to 'expired'"}
+
     for whitelisted_key, value in FRAMEWORK_UPDATE_WHITELISTED_ATTRIBUTES_MAP.items():
         if whitelisted_key in json_framework:
             setattr(framework, value, json_framework[whitelisted_key])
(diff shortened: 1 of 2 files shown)


app/main/views/frameworks.py
```diff
@@ -136,6 +136,13 @@ def update_framework(framework_slug):
             enforce_required=True
         )
 
+    if json_framework.get('status') == 'expired' and framework.status != 'expired':
+        # framework_expires_at_utc is used in the buyer frontend in content that is displayed once the framework is
+        # expired. It's important that it's accurate. This ensures that the date is accurate with the framework is
+        # expired.
+        framework.framework_expires_at_utc = datetime.datetime.utcnow()
+
+    import pdb; pdb.set_trace()
```
samuelhwilliams: whoops

app/main/views/frameworks.py
```diff
@@ -136,6 +136,13 @@ def update_framework(framework_slug):
             enforce_required=True
         )
 
+    if json_framework.get('status') == 'expired' and framework.status != 'expired':
+        # framework_expires_at_utc is used in the buyer frontend in content that is displayed once the framework is
+        # expired. It's important that it's accurate. This ensures that the date is accurate with the framework is
+        # expired.
+        framework.framework_expires_at_utc = datetime.datetime.utcnow()
```
samuelhwilliams: We probably would want this recorded in the/an audit event as well

Wynndow: Cool, have updated.

samuelhwilliams:
I think it's reasonable to extend this PR to all of the other currently recorded datetimestamps that relate to state changes (including clarification questions, maybe). If you're happy to do that?

Wynndow:
Yeah, totally happy. I'm just not sure if there are knock on consequences. I'm assuming that for the expiration date it won't, as it's currently in the past for G9 on prod. So we'd have seen any issues. My hunch is not but I might have a look around and see where the dates are being used.

app/main/views/frameworks.py
```diff
@@ -136,7 +136,23 @@ def update_framework(framework_slug):
...
+
+    if json_framework.get('status') == 'live' and framework.status != 'live':
+        framework.framework_live_at_utc = datetime.datetime.utcnow()
+        status_timestamp_audit_data = {"framework_live_at_utc set": "framework status set to 'live'"}
+
+    if json_framework.get('status') == 'expired' and framework.status != 'expired':
+        framework.framework_expires_at_utc = datetime.datetime.utcnow()
+        status_timestamp_audit_data = {"framework_expires_at_utc set": "framework status set to 'expired'"}
+
     if json_framework.get('status', None) == 'pending' and framework.status != 'pending':
```
samuelhwilliams: These could probably be an `if/elif/elif` change rather than distinct `if`s

samuelhwilliams: Also one of these sets a distinct `None` default value while the rest don't

Wynndow: Yeah, that was already there for the supplier details injection stuff. But I don't think it's necessary. `None` is the default return value for `.get()`.

Wynndow: And yep, the statuses can if `if/elif`, although as the status can only ever be one thing it seems unnecessary?

samuelhwilliams: That's kinda why it's more appropriate to do `if/elif` - there's no sense having them as distinct statements because it lightly implies that they could all match in some circumstance.

PR #812, opened 2018-06-29
Registered postcode can't be longer than 15 characters now

CrystalPea:
This has been done for consistency: labAddressPostcode field
is already validated in such a way. Although there maximum
length is 10 characters. We decided for 15 characters for
supplier registered postcode in the end because some US
postcodes are 13 characters long.

Connected supplier-frontend pull request: alphagov/digitalmarketplace-supplier-frontend#904

Trello ticket: https://trello.com/c/kOGpfxcw/475-supplier-fe-postcode-field-validation-is-more-lax-than-api-and-varies-between-two-forms

### DIFF
--- a/json_schemas/contact-information.json
+++ b/json_schemas/contact-information.json
@@ -30,6 +30,8 @@
       "type": "boolean"
     },
     "postcode": {
+      "maxLength": 15,
+      "minLength": 1,
       "type": "string"
     },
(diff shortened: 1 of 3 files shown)


katstevens:
Cool - shall we change one of the tests to be `15Ch4r4ctrsL0ng` as well?

PR #813, opened 2018-06-29
digitalmarketplace-apiclient dependency: bump to 17.4.0

risicle:
This should bring `childSpanId` logging.


risicle:
Already superseded

PR #814, opened 2018-07-02
New 'readyToAssess' flag and associated datetime column.

galund:
This records the user claiming to have read the guidance and thus being ready to start assessing services in their Direct Award procurement.

https://trello.com/c/7wlBbGU0/138-add-step-3-assessing-your-service-and-badges

**Don't merge until we know if the frontend part of this will land.**


requirements-app.txt
```diff
@@ -10,7 +10,9 @@ psycopg2==2.7.3
 SQLAlchemy==1.1.4
 SQLAlchemy-Utils==0.30.5
 
-git+https://github.com/alphagov/digitalmarketplace-utils.git@39.3.0#egg=digitalmarketplace-utils==39.3.0
+git+https://github.com/alphagov/Flask-FeatureFlags.git@1.0#egg=Flask-FeatureFlags==1.0
```
katstevens: Do we have to bring this in because of the utils bump?

app/main/views/direct_award.py
```diff
@@ -170,6 +170,11 @@ def update_project(project_external_id):
             abort(400, "Cannot update stillAssessing when an outcome is present")
         project.still_assessing_at = uniform_now
 
+    if project_json.get('readyToAssess') and project.ready_to_assess_at is None:
```
risicle: Neat - so we only set the timestamp the *first* time.

lfdebrux:
@katstevens that's a good point.

Currently (digitalmarketplace-buyer-frontend#809) the tasklist will show the 'ready to assess' task as not completed if readyToAssess is None, which will be an issue for projects where we have already been told the outcome; however I think that is probably a small number of projects.

Rather than having a default, I think I will change the tasklist so the step will be shown as completed if the next step has been completed; I think this is cleaner than a big migration step.

PR #818, opened 2018-07-18
Remove restriction on changing contact information after personal data stripped

benvand:
> We need to allow this because it is conceivable to want to resurrect a supplier by inviting a new contributor to their account

Remove restriction on changing contact information after personal data is stripped
* This has been found to be unnecessary and stops suppliers from reactivating/ adding a new user once all users are removed.

Update test

### DIFF
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -77,23 +77,6 @@ class RemovePersonalDataModelMixin:
         """
         pass
 
-    @staticmethod
-    def validate_personal_data_removed(mapper, connection, instance):
-        # """The only time we should be able to update an object with """
-        session = object_session(instance)
-        model_class = instance.__class__
-        if session.query(model_class.personal_data_removed).filter(model_class.id == instance.id).scalar():
-            raise ValidationError("Cannot update an object once personal data has been removed")
-
-
-listen(
-    RemovePersonalDataModelMixin,
-    'before_update',
-    RemovePersonalDataModelMixin.validate_personal_data_removed,
-    propagate=True,
-    active_history=True
-)
-
 
 class FrameworkLot(db.Model):
     __tablename__ = 'framework_lots'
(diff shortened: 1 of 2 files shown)


app/models/main.py
```diff
@@ -982,9 +965,25 @@ def remove_personal_data(self):
         self.name = '<removed>'
         self.phone_number = '<removed>'
 
+        self.failed_login_count = 0
         self.password = encryption.hashpw(str(uuid4()))
         self.user_research_opted_in = False
 
+    @staticmethod
+    def validate_personal_data_removed(mapper, connection, instance):
+        # """The only time we should be able to update an object with """
```
lfdebrux: Is this meant to be here?

app/models/main.py
```diff
@@ -77,23 +77,6 @@ def remove_personal_data(self):
         """
         pass
 
-    @staticmethod
-    def validate_personal_data_removed(mapper, connection, instance):
```
lfdebrux: Rather than moving the validator definition out of the mixin, can we leave the function here but only add the listener on the relevant class?

lfdebrux:
If the contact information is re-added after the user data is removed, the `personal_data_removed` flag will remain true, right? Is this what we want?

app/models/main.py
benvand: So I did have a think about this. But I think it's kind of misleading to have it on the mixin if it's not applied to everything.
The method now technically only applies to User objects. It's whole purpose has changed, from validating all personal data models to validating only user objects. That's because we've decided that objects that have had their personal data removed _can_ be changed but Users cannot be changed once their personal data is removed. 
In the future it's possible that this will be re implemented but I think, although it's a nifty bit of code, it'll just be a confusing relic for people to read. Plus I think the code represents the decision as is.

app/models/main.py
benvand: The comment? No, no it is not. :p

app/models/main.py
lfdebrux: Fair point

benvand:
@lfdebrux Great catch, I've updated the model to reset the flag to false whenever personal data is changed.

PR #826, opened 2018-08-13
Make audit events searchable by supplier

Wynndow:
For [this trello ticket.](https://trello.com/c/AdtcNmBR)
See [this associated PR on the apiclient.
](https://github.com/alphagov/digitalmarketplace-apiclient/pull/154)
### Add index to audit event data  …
To improve the searchability of audit events, `supplierId` has been added
to draft brief audit events as a top level key in the audit event data.
To make this searchable in an effective way, this adds an index on it.

To do this, audit event data needed to be converted to jsonb. A partial
expression index is then added. Partial because it only indexes rows
that have the `supplierId` field in the data, and an expression index as
it only indexes the `supplierId` field instead of the whole data blob.

### Allow searching of audit events by supplier id  …
This adds a parameter to the `list_audits` view that allows passing in a
supplier id to search for in the data of audit events.

### DIFF
--- a/tests/main/views/test_audits.py
+++ b/tests/main/views/test_audits.py
@@ -18,16 +18,16 @@ from tests.helpers import FixtureMixin
 
 class BaseTestAuditEvents(BaseApplicationTest, FixtureMixin):
     @staticmethod
-    def audit_event(user=0, type=AuditTypes.supplier_update, db_object=None):
+    def audit_event(user=0, type=AuditTypes.supplier_update, db_object=None, data={"request": "data"}):
         return AuditEvent(
             audit_type=type,
             db_object=db_object,
             user=user,
-            data={'request': "data"}
+            data=data,
         )
 
-    def add_audit_event(self, user=0, type=AuditTypes.supplier_update, db_object=None):
-        ae = self.audit_event(user, type, db_object)
+    def add_audit_event(self, user=0, type=AuditTypes.supplier_update, db_object=None, data={"request": "data"}):
+        ae = self.audit_event(user, type, db_object, data)
         db.session.add(
             ae
         )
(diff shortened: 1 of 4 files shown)


risicle:
Neat! I think this is the right approach to follow for the index, certainly for now when what we initially want to be able to do is so straightforward. I'll put it through its paces later...

risicle:
Ah so we're handling this data that comes out of `data["supplierId"]` as a string are we? I wonder if we're going to have to be more explicit about casting it to a type because it's possible many of the audit events add `supplierId` as an integer (well, json "number") and some may add it as a string...

risicle:
I think the trick might be to use the postgres `->>` operator instead of `->` because the former always outputs its result as text. I'm not exactly sure how to access the operator from within sqlalchemy but it shouldn't be too hard.

tests/main/views/test_audits.py
```diff
@@ -887,6 +887,18 @@ def test_paginated_audit_with_invalid_custom_page_size(self):
...
+        for id, info in [(0, 'miss'), (3, 'hit'), (2, 'miss'), (7, 'miss'), (3, 'hit'), (1, 'miss')]:
+            self.add_audit_event(data={'supplierId': id, 'info': info})
+
+        response = self.client.get('/audit-events?data-supplier-id=3')
+        audit_events = json.loads(response.get_data())['auditEvents']
+
+        assert response.status_code == 200
+        assert len(audit_events) == 2
+        assert all(audit_event['data']['supplierId'] == 3 for audit_event in audit_events)
+        assert all(audit_event['data']['info'] == 'hit' for audit_event in audit_events)
```
katstevens: What happens when an AuditEvent doesn't have `supplierId` in `data`?

risicle:
(yeah, according to http://docs.sqlalchemy.org/en/latest/dialects/postgresql.html#sqlalchemy.dialects.postgresql.JSON it's just `data_table.c.data['some key'].astext == 'some value'`)

tests/main/views/test_audits.py
Wynndow: Nothing (hopefully)! I've added a test to show this, but the index is set up to only index rows where supplier id is set.

katstevens: Nice 👍

PR #832, opened 2018-08-24
digitalmarketplace-utils dependency: bump to 43.0.0, remove featureflags references, fix error handlers for flask 0.12

risicle:
Trello https://trello.com/c/ZH9trgjH/419-snyk-alert-for-flask-010x-vulnerability

https://github.com/alphagov/digitalmarketplace-utils/pull/447

Only thing that stands out about this app is the changes I had to make to `generic_error_handler` to accommodate flask 0.12.

### DIFF
--- a/app/errors.py
+++ b/app/errors.py
@@ -1,4 +1,5 @@
 from flask import jsonify
+from werkzeug.exceptions import default_exceptions
 
 from .main import main
 from .callbacks import callbacks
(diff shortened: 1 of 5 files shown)


app/errors.py
```diff
@@ -25,5 +26,6 @@ def generic_error_handler(e):
 
 
 for code in range(400, 599):
-    main.app_errorhandler(code)(generic_error_handler)
-    callbacks.app_errorhandler(code)(generic_error_handler)
+    if code in default_exceptions:  # flask complains if we attempt to register a handler for status code its unaware of
+        main.app_errorhandler(code)(generic_error_handler)
+        callbacks.app_errorhandler(code)(generic_error_handler)
```
katstevens: Can we just do `for code in default_exceptions:` and ditch the `if`?

risicle: Mmmmmmmmmaybe... as long as we're 100% certain that werkzeug will only ever put 400 -> 599 status codes in there?

katstevens: If they do put extra ones in, we should probably handle them 😹

PR #836, opened 2018-09-03
Create endpoint for transitioning DOS frameworks

Wynndow:
For [this trello ticket.](https://trello.com/c/qGsP2FS0)

When we transition from one DOS framework to another, there is no
overlap period. One framework is expired at the same time the other 
is made live. A buyer publishes a requirement, and it will either be
on one framework or the other. Suppliers are able to apply to an
opportunity even if the framework it was published on is expired, as
long as the brief is still open (within 1 or 2 weeks of its publication
date).

The issue is that a brief is tied to a framework at the point it is
created (as a draft). If a framework is expired, any drafts that belong
to it are no longer editable/publishable. We don't want a buyer to lose
a draft.

This endpoint will update the framework statuses, move the draft briefs
belonging to that framework to the incoming framework, and create audit
events for it all, in a transactionally safe way.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -3,6 +3,7 @@ import datetime
 from flask import jsonify, abort, request
 from sqlalchemy import func, orm, case
 from sqlalchemy.exc import IntegrityError, DataError
+from sqlalchemy.orm import lazyload
 from dmapiclient.audit import AuditTypes
 from dmutils.config import convert_to_boolean
 
@@ -16,6 +17,7 @@ from ...models import (
     Supplier,
     SupplierFramework,
     User,
+    Brief,
 )
 from ...utils import (
     get_json_from_request,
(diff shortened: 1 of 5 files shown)


tests/main/views/test_frameworks.py
```diff
@@ -1110,3 +1110,221 @@ def test_a_404_is_raised_if_it_does_not_exist(self):
...
+            slug="g-cloud-10", framework_family="g-cloud", lots=[], id=101,
+        )
+        self.setup_dummy_framework(
+            slug="digital-outcomes-and-specialists-2", framework_family="digital-outcomes-and-specialists", id=102,
+        )
+
+        response = self.client.post(
+            f"/frameworks/transition-dos/{going_live_slug}",
+            data=json.dumps({
+                "updated_by": "🤖",
```
benvand: LOL

app/main/views/frameworks.py
```diff
@@ -330,3 +331,97 @@ def get_framework_interest(framework_slug):
...
+    updater_json = validate_and_return_updater_request()
+    json_payload = get_json_from_request()
+    json_has_required_keys(json_payload, ['expiringFramework'])
+
+    going_live_framework = Framework.query.filter(
+        Framework.slug == framework_slug
+    ).first_or_404()
+    expiring_framework = Framework.query.filter(
+        Framework.slug == json_payload['expiringFramework']
+    ).first_or_404()
```
risicle: Should probably use `.with_for_update()` for these selects to lock the rows.

app/main/views/frameworks.py
```diff
@@ -330,3 +331,97 @@ def get_framework_interest(framework_slug):
...
+              f" '{expiring_framework.framework}' must both be 'digital-outcomes-and-specialists'")
+
+    if going_live_framework.status != 'standstill' or expiring_framework.status != 'live':
+        abort(400, f"'going_live_framework' status ({going_live_framework.status}) must be 'standstill', and "
+              f"'expiring_framework' status ({expiring_framework.status}) must be 'live'")
+
+    expiring_framework_draft_briefs = Brief.query.filter(
+        Brief.status == 'draft',
+        Brief.framework_id == expiring_framework.id,
+    ).all()
```
risicle: Again, should probably lock all of these rows `.with_for_update()` else there's a possibility we could clobber an update made to the brief after we've fetched it but before we issued the `UPDATE` to it in the loop below.

app/main/views/frameworks.py
```diff
@@ -330,3 +331,97 @@ def get_framework_interest(framework_slug):
...
+        audit = AuditEvent(
+            audit_type=AuditTypes.update_brief_framework_id,
+            user=updater_json['updated_by'],
+            data={
+                'briefId': brief.id,
+                'previousFrameworkId': expiring_framework.id,
+                'newFrameworkId': going_live_framework.id,
+            },
+            db_object=brief,
+        )
```
risicle: I would consider giving all these audit events (including those below) a common uniform timestamp.

Wynndow: These should all occur within a very short time period. What would the benefit of a common timestamp be rather than when it actually happens?

risicle: I suppose to kinda record that the logical order these happened in was basically simultaneous. Because it basically is, seeing that all the changes only become visible to the rest of the world at the same time.

Wynndow: hmmm....interesting. That does make sense.

Wynndow: After some jiggery-pokery I've realised that AuditEvents don't take a timestamp on it, they just use a default.
Rather than alter the model, I'm just going to leave as is. I don't think it's going to be an issue.

risicle: Yeah I've run into that before the issue is just that the *constructor* doesn't expose the timestamp as an argument. You can still set the timestamp on the instance just after you've constructed it, which is what I've done elsewhere when I've done this.

Wynndow: grumble grumble messy grumble

app/main/views/frameworks.py
```diff
@@ -330,3 +332,110 @@ def get_framework_interest(framework_slug):
     supplier_ids = [supplier_framework.supplier_id for supplier_framework in supplier_frameworks]
 
     return jsonify(interestedSuppliers=supplier_ids), 200
+
+
+@main.route('/frameworks/transition-dos/<string:framework_slug>', methods=['POST'])
+def transition_dos_framework(framework_slug):
```
katstevens: Can we add a brief docstring to explain the view (the description from this PR would be ideal)?

katstevens:
I've just had a quick nose around what we did for draft briefs in DOS1 -> DOS2, which was done via several data migrations, one of which I don't understand: https://github.com/alphagov/digitalmarketplace-api/blob/master/migrations/versions/870_remove_invalid_draft_dos2_brief_dates.py Do we need to worry about this for DOS2 -> DOS3?

app/main/views/frameworks.py
Wynndow: Yep, will do.

Wynndow:
It looks like the 'latest start date' question changed from being text format to a date format at some point. By the look of it the migration is removing that answer to that question for anything that wasn't published yet, so that it can be re-answered with a date format before it was published. We've not made any changes like this for DOS3 so should be good.

Here's the PR that change the format: https://github.com/alphagov/digitalmarketplace-frameworks/pull/413

And here's the commit message which explains a bit more: https://github.com/alphagov/digitalmarketplace-api/commit/e387ccfd85484fe56e10d1ba7a6bb6a4cdc54190

katstevens:
@Wynndow what a lovely descriptive commit message 😸 Cool, that all seems fine then.

PR #838, opened 2018-09-13
Reduce locking potential on DOS transition endpoint

Wynndow:
Frameworks are now committed in a separate transaction, before the
briefs. This is to prevent the framework rows from being locked for any
longer than necessary. If there are lots of briefs to update, the
frameworks could be locked unnecessarily. The Framework rows are used
heavily by lots of api calls so that would be bad.

This also changes the way the briefs are loaded to only fetch the id and
framework_id columns. There's no need to pull all the data. This should
make the process faster.

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -3,7 +3,7 @@ import datetime
 from flask import jsonify, abort, request
 from sqlalchemy import func, orm, case
 from sqlalchemy.exc import IntegrityError, DataError
-from sqlalchemy.orm import lazyload
+from sqlalchemy.orm import lazyload, load_only
 from dmapiclient.audit import AuditTypes
 from dmutils.config import convert_to_boolean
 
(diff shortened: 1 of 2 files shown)


lfdebrux:
Is the thing with the mock you were asking about the other day? How did you get it working?

Wynndow:
Yeah it is, you can see how it works now here: https://github.com/alphagov/digitalmarketplace-api/pull/838/files?utf8=%E2%9C%93&diff=unified#diff-baaa491be1f5bd836bc99dac51719bd4R1320

Turns out that when you give `side_effect` an iterable, if one of the items is a called function, then that function is called immediately and not when the mock is called. so I was giving the mock `mock.side_effect = [db.app.commit(), IntegrityError]` which meant that commit was being called as soon as the tests were loaded, and not at the point the mock was called in the view.

By giving `side_effect` a generator, we can control when the function is called. Props to @risicle for that one.

lfdebrux:
> By giving `side_effect` a generator, we can control when the function is called. Props to @risicle for that one.

Nice 💯

PR #840, opened 2018-10-10
Test that json fields are mutable and updated

benvand:
https://trello.com/c/9IfzOZwE/187-spike-updating-json-blob-keys-have-the-required-libraries-been-updated-yet

This addressed the concerns on :
https://trello.com/c/1rbvLCNL/31-updating-json-blob-keys-will-not-save-data-to-sqlalchemy

And fixes the concerns that this issue will trip developers up 🤷‍♀️


tests/models/test_main.py
```diff
@@ -20,6 +20,22 @@
...
+        db.session.add(test_object)
+        db.session.commit()
+
+        assert Lot.query.get(test_object.id).data == {'test': 'data'}
+
+        test_object.data['test'] = 'update'
+        db.session.add(test_object)
+        db.session.commit()
+
+        assert Lot.query.get(test_object.id).data == {'test': 'update'}
```
katstevens: For my peace of mind, can we try it with nested data as well?

benvand: Works now!

benvand: Hmmmm 🤔

benvand: Works....... now!

requirements.txt
```diff
@@ -36,9 +37,9 @@ docutils==0.14
 Flask-Login==0.4.1
 Flask-Script==2.0.6
 Flask-WTF==0.14.2
-future==0.16.0
+future==0.17.0
 idna==2.7
-itsdangerous==0.24
+itsdangerous==1.1.0
```
katstevens: Noooo 😸

tests/models/test_main.py
benvand: OK this doesn't work at all.

app/models/main.py
```diff
@@ -63,6 +64,9 @@ def __init__(self, astext_type=None):
         super(JSON, self).__init__(none_as_null=True, astext_type=astext_type)
 
 
+NestedMutable.associate_with(JSON)
```
katstevens: I think a comment here about why it's needed would be useful - something like `Enable tracking of nested updates/changes for all subsequent instances of the JSON class`.

PR #841, opened 2018-10-22
Delete declaration

jonodrew:
This PR enables us to remove supplier declarations from particular frameworks

# Why
It is important that we do this for GDPR, allowing us to remove user's declaration when they ask or after a set period

# What
A new route, `'suppliers/<supplier_id>/frameworks/<framework_slug>'` has been created. POST requests to the endpoint with valid `supplier_id` and `framework_slug` will replace the supplier's declaration with an empty dictionary.

[Ticket](https://trello.com/c/h3VPjiql/209-new-route-for-removing-supplier-declarations)

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -479,6 +479,41 @@ def set_a_declaration(supplier_id, framework_slug):
     return jsonify(declaration=supplier_framework.declaration), status_code
 
 
+@main.route('/suppliers/<int:supplier_id>/frameworks/<framework_slug>/declaration', methods=['POST'])
+def remove_a_declaration(framework_slug, supplier_id):
+    """
+    This route will replace unsuccessful applicants's declarations with an empty dict and return the altered object.
+    :param framework_slug: a string describing the framework
+    :type framework_slug: string
+    :param supplier_id: a way of identifying the supplier whose declaration will be removed
+    :type supplier_id: int
+    :return: The altered SupplierFramework object
+    :rtype: SupplierFramework
+    """
+    updater_json = validate_and_return_updater_request()
+    supplier_framework = SupplierFramework.find_by_supplier_and_framework(
+        supplier_id, framework_slug
+    ).first()
+
+    supplier_framework.declaration = {}
+
+    audit_event = AuditEvent(
+        audit_type=AuditTypes.update_supplier_framework,
+        db_object=supplier_framework,
+        user=updater_json['updated_by'],
+        data={}
+    )
+
+    db.session.add(supplier_framework, audit_event)
+    try:
+        db.session.commit()
+    except IntegrityError as e:
+        db.session.rollback()
+        abort(400, format(e))
+
+    return jsonify(message="Successful!"), 200
+
+
 @main.route('/suppliers/<int:supplier_id>/frameworks/interest', methods=['GET'])
 def get_registered_frameworks(supplier_id):
     supplier_frameworks = SupplierFramework.query.filter(
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -482,13 +482,14 @@ def set_a_declaration(supplier_id, framework_slug):
 @main.route('/suppliers/<int:supplier_id>/frameworks/<framework_slug>/declaration', methods=['POST'])
 def remove_a_declaration(framework_slug, supplier_id):
     """
-    This route will replace unsuccessful applicants' declarations with an empty dict and returns an empty body
+    This route will replace unsuccessful applicants' declarations with an empty dict and returns the updated object,
+    serialized in JSON format
     :param framework_slug: a string describing the framework
     :type framework_slug: string
     :param supplier_id: a way of identifying the supplier whose declaration will be removed
     :type supplier_id: int
-    :return: An empty body and a status code
-    :rtype: Tuple(Dict, int)
+    :return: The serialized SupplierFramework and a status code
+    :rtype: Response
     """
     updater_json = validate_and_return_updater_request()
     supplier_framework = SupplierFramework.find_by_supplier_and_framework(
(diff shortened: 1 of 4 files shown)


app/main/views/suppliers.py
```diff
@@ -479,6 +479,42 @@ def set_a_declaration(supplier_id, framework_slug):
...
+        supplier_id, framework_slug
+    ).first()
+
+    supplier_framework.declaration = {}
+
+    audit_event = AuditEvent(
+        audit_type=AuditTypes.update_supplier_framework,
+        db_object=supplier_framework,
+        user=updater_json['updated_by'],
+        data={}
```
risicle: I wonder if it will be obvious enough to us that a `data` of `{}` here means "the declaration was deleted". There's always the possibility of creating a specific `AuditType` for this operation. Minor pain in the ass though.

app/main/views/suppliers.py
```diff
@@ -479,6 +479,42 @@ def set_a_declaration(supplier_id, framework_slug):
...
+
+    db.session.add(supplier_framework, audit_event)
+    try:
+        db.session.commit()
+    except (IntegrityError, DataError):
+        db.session.rollback()
+        error_msg = "Could not remove declaration data from supplier framework: supplier_id {}, framework {}"
+        abort(400, error_msg.format(supplier_id, framework_slug))
+
+    return jsonify(message="Successful!"), 200
```
risicle: Thought it was going to "return the altered object"? Am I missing something?

jonodrew: Yes! Excellent catch, thanks Rob

app/main/views/suppliers.py
```diff
@@ -479,6 +479,42 @@ def set_a_declaration(supplier_id, framework_slug):
...
+    :type framework_slug: string
+    :param supplier_id: a way of identifying the supplier whose declaration will be removed
+    :type supplier_id: int
+    :return: The altered SupplierFramework object
+    :rtype: SupplierFramework
+    """
+    updater_json = validate_and_return_updater_request()
+    supplier_framework = SupplierFramework.find_by_supplier_and_framework(
+        supplier_id, framework_slug
+    ).first()
```
risicle: (I think) you can use `.first_or_404()` here to give a nice behaviour when the supplier/framework combination doesn't exist (rather than a 500 which is what would happen here).

Personally, when doing this sort of operation, I would also take out a write lock on the `SupplierFramework` row by adding a use of `.with_for_update()`, specifically in this case `.with_for_update(of=SupplierFramework)`. This prevents any race conditions over the modification of the row.

risicle:
Oh but let me run it on my local machine to make sure.

risicle:
Yup WFM

app/main/views/suppliers.py
```diff
@@ -479,6 +479,42 @@ def set_a_declaration(supplier_id, framework_slug):
     return jsonify(declaration=supplier_framework.declaration), status_code
 
 
+@main.route('/suppliers/<int:supplier_id>/frameworks/<framework_slug>/declaration', methods=['POST'])
+def remove_a_declaration(framework_slug, supplier_id):
+    """
+    This route will replace unsuccessful applicants's declarations with an empty dict and return the altered object.
```
katstevens: Typo in `applicants's`

app/main/views/suppliers.py
katstevens: Agree that we should probably have a new AuditType. I am happy to knock one up right now 😸

tests/main/views/test_suppliers.py
```diff
@@ -1914,6 +1914,57 @@ def test_set_application_company_details_confirmed_updates_declaration(
         }
 
 
+class TestDeleteUnsuccessfulApplicantDeclarations(BaseApplicationTest, FixtureMixin,
+                                                  PutDeclarationAndDetailsAndServicesMixin):
+    def setup(self):
+        """
+        This sets up the class. It creates a dummy framework and a supplier that applied to that framework, with a
```
katstevens: We don't actually create a dummy framework here - hilariously we just never delete them when we teardown API tests, so they're just 'there'. Obviously not ideal, but let's not make a confusing situation more confusing in this docstring: how about `Set up test data for a supplier and their successful application to a framework.` (We also don't set up the declaration here, but in the test itself)

tests/main/views/test_suppliers.py
```diff
@@ -1914,6 +1914,57 @@ def test_set_application_company_details_confirmed_updates_declaration(
...
+        super(TestDeleteUnsuccessfulApplicantDeclarations, self).setup()
+
+        self.supplier_id = self.setup_dummy_suppliers(1)[0]
+        self.framework_slug = 'digital-outcomes-and-specialists'
+        self.set_framework_status('digital-outcomes-and-specialists', 'open')
+        self.updater_json = {'updated_by': 'Joe Bloggs'}
+        self._register_supplier_with_framework()
+
+    def test_a_supplier_framework_object_is_returned_with_an_empty_declaration(self):
+        self._put_declaration("Success!")
```
katstevens: I know I suggested this delightful string but we should probably use a valid declaration status (`started` or `complete`).

katstevens: I also wonder if we should test what happens if the declaration is already empty, e.g. if a careless dev hits the endpoint twice (hopefully the endpoint won't care?)

jonodrew:
This has been updated - @risicle and @katstevens can I ask for a quick once-over to ensure I've not missed anything? 

I've not included the race condition code as it throws an error - it doesn't like the `DISTINCT` generated in the SQL. Is it worth continuing to plug away and make it work?

app/main/views/suppliers.py
```diff
@@ -479,6 +479,42 @@ def set_a_declaration(supplier_id, framework_slug):
...
+
+    db.session.add(supplier_framework, audit_event)
+    try:
+        db.session.commit()
+    except (IntegrityError, DataError):
+        db.session.rollback()
+        error_msg = "Could not remove declaration data from supplier framework: supplier_id {}, framework {}"
+        abort(400, error_msg.format(supplier_id, framework_slug))
+
+    return jsonify(), 200
```
katstevens: As discussed offline - let's return `single_result_response` for the updated object.

PR #844, opened 2018-10-31
Upgrade sqlalchemy to 1.2.12

benvand:
See commit comments for changes, part of this ticket:
https://trello.com/c/juqZDbC1/217-updating-json-attributes-on-our-models-does-not-persist-to-the-database

### DIFF
--- a/app/main/views/frameworks.py
+++ b/app/main/views/frameworks.py
@@ -2,7 +2,7 @@ import datetime
 
 from flask import jsonify, abort, request
 from sqlalchemy import func, orm, case
-from sqlalchemy.exc import IntegrityError, DataError
+from sqlalchemy.exc import IntegrityError, DataError, StatementError
 from sqlalchemy.orm import lazyload, load_only
 from dmapiclient.audit import AuditTypes
 from dmutils.config import convert_to_boolean
@@ -96,7 +96,7 @@ def create_framework():
         db.session.rollback()
         abort(400, "Invalid framework")
 
-    except IntegrityError as error:
+    except StatementError as error:
         db.session.rollback()
         abort(400, format_framework_integrity_error_message(error, json_framework))
 
--- a/app/framework_utils.py
+++ b/app/framework_utils.py
@@ -1,3 +1,5 @@
+import re
+
 from flask import abort
 from .validation import get_validation_errors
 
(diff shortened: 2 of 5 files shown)


app/main/views/frameworks.py
```diff
@@ -96,7 +96,7 @@ def create_framework():
         db.session.rollback()
         abort(400, "Invalid framework")
 
-    except IntegrityError as error:
+    except (IntegrityError, StatementError) as error:
```
katstevens: We catch `IntegrityError` in a whole bunch of create/update views - do we need to add this in anywhere else?

risicle: Oh, is this just raised when we use an invalid `enum`? If so, I guess it's possible for us to make this error every time we insert or update an audit log...

benvand: Sort of...

So as part of SQLAlchemy fixing a bug in how python booleans are converted to their database value counterpart here:
https://github.com/zzzeek/sqlalchemy/commit/c63658973c95996b73251e0953cf4c598cdb262b

... they made it possible to raise a `TypeError` or `ValueError` inside code that predominantly raises `DBAPI` exceptions, exceptions caught directly from the DB engine sqlalchemy is communicating with.

These DBAPI exceptions are quite helpful and contain various bits of info about the sql that SQLAlchemy was trying to run, and a descriptive error code. However... if the error is not of type DBAPI error then it is reraised here as a `StatementError`:
https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/exc.py#L380

The main point of the above then is that SQLAlchemy uses `StatementError` as its standard 'the database has a problem with what you just tried to do' error. Whereas we only catch a subset of those errors in our views `IntegrityError` (which, as defined here, subclasses `StatementError` https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/exc.py#L446)

We should be catching `StatementError` not `IntegrityError` (which will include all `IntegrityError`s). But we should always have been doing that.

benvand: The question is should I do that as part of this ticket? It seems like it could be a big change.

benvand: Hmmm, I'm going to have to go with 'no' on that. There are a load of places we're also catching both `(IntegrityError, DataError)`. So the error handling needs looking at separately.

lfdebrux: Aw boo

PR #845, opened 2018-11-12
Update requirements-dev.txt

lfdebrux:


requirements-dev.txt
```diff
@@ -1,19 +1,19 @@
...
-testfixtures==6.0.2
+flake8-per-file-ignores
+freezegun
+hypothesis<4.0.0,>=3.0.0
+mock
+pytest
+pytest-cov
+requests-mock
+testfixtures<7.0.0,>=6.0.2
+
```
risicle: I'm getting less and less keen on the removal of version pinning. And I was never _keen_.

lfdebrux: Discuss at stand-up?

lfdebrux: Okay, agreement is that we will pin dev requirements, but @jonodrew will look at setting up something like depend-a-bot to keep our dependencies up-to-date.

lfdebrux:
Tests are failing because of pip behaviour;

- `pytest==3.10.1` has a requirement `attrs>=17.4.0`
- however, `hypothesis==3.82.1` comes first in the `requirements-dev.txt` file and has as a requirement `attrs>=16.0.0`
- therefore pip only checks whether `attrs>=16.0.0` is installed
- Travis CI virtualenv already has a version of `attrs` installed that is between `16.0.0` and `17.4.0`, so pip treats the requirements as fulfilled and does not install a newer version of `attrs`

There are a couple of solutions to this issue:

1. Freeze the requirements in the same way as we do for `requirements-app.txt`; this is a heavyweight solution, but with @benvand's new script it should be easy to do
2. Pin pytest to an older version that does not require `attrs>=17.4.0`; this is not a long term fix
3. Arrange for `attrs>=17.4.0` to be installed, either by explicitly adding it as a requirement to `requirements-dev.txt` or by moving the `hypothesis` line to below `pytest`; this is a lightweight change but possibly brittle
4. Install dev requirements with the `-U` so that the latest versions of sub-requirements are always installed; similar to the above but might affect other packages
5. Don't pin dev requirements

risicle:
> Don't pin dev requirements

I don't see this fixing the general case - there's nothing to say that the "latest version" of all dependencies won't have a similar version situation going on, and it interferes with our ability to go through the "change *one* thing" process when debugging.

lfdebrux:
> > Pin pytest to an older version that does not require attrs>=17.4.0; this is not a long term fix
> 
> Nothing when it comes to pinning is a long term fix. But that mirrors reality. Package versions can have conflicting dependencies - it's part of life. Those need to be negotiated and the solution committed until next time.

Is this your preferred solution then @risicle?

risicle:
Well I mean _clearly_ my preferred option would involve :nix: but of the options presented, yes.

requirements-dev.txt
```diff
@@ -1,19 +1,19 @@
...
-coverage==3.7.1
-flake8-per-file-ignores==0.4.0
+coverage==4.5.2
+coveralls==1.5.1
 flake8==3.5.0
-freezegun==0.3.8
-hypothesis==2.0.0
+flake8-per-file-ignores==0.6.0
+freezegun==0.3.11
+hypothesis==3.82.1
```
katstevens: As discussed offline - let's try bumping this back down to v3.44.21 which will hopefully avoid the `attrs` problem.

lfdebrux:
@katstevens hypothesis is still interfering :( Going to back out the commit, nice idea though.

PR #847, opened 2018-11-22
New admin user role for CCS Data Controller

katstevens:
Trello: https://trello.com/c/tZJCxC87/249-new-admin-search-box-for-suppliers

The DOS3 Admin mission identified a need for a new CCS admin role, 'data controller'. Users with this role will have new permissions in the Admin FE to edit a limited subset of supplier details: name, companies house registration number, registered name, registered address, and DUNS number (this must still be a unique value). 

The role will also have the following permissions, currently available to the CCS Category role:
- search for suppliers
- view framework agreements for a supplier
- view a list of users for a supplier 
- download 'Official supplier details' CSV

Note that the `ENUM` migration is copied from 1050 and 1060 (adding framework manager and admin manager roles).

### DIFF
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -839,12 +839,12 @@ class User(db.Model, RemovePersonalDataModelMixin):
     __tablename__ = 'users'
 
     ADMIN_ROLES = [
-        'admin',  # a general admin user, with permission to do most (but not all) admin actions.
-        'admin-ccs-category',  # generally restricted to read-only access to admin views.
-        'admin-ccs-sourcing',  # can perform admin actions involving supplier acceptance.
-        'admin-manager',  # can add, edit and disable other types of admin user
-        'admin-framework-manager',  # can perform admin actions involving framework applications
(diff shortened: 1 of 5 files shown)


risicle:
Are we going to get to the point where we have to reconsider our permissions model to make roles non-exclusive? I imagine we have a number of users who need to have an array of different accounts with the different roles...

katstevens:
It's a good question. This particular role's main function is editing supplier info, which doesn't yet exist in the admin (our mission will be building it in the next 3-4 weeks). The Category role is what they're using at the moment, so the additional permissions are there for some continuity - hopefully they wont need it once they've got their new 'Edit Company Details' screen and all the data is there in one place.

We did some work in Q3 2017 on a possible multi-role permission structure and decided not to do it (I can attach the closedown doc to the Trello card), and the situation hasn't changed much since then apart from the emergence of this Data Controller role (which we somehow missed when we revamped the admin roles the last time round).

PR #849, opened 2018-11-27
Search supplier by companies house number

katstevens:
Trello: https://trello.com/c/tZJCxC87/249-new-admin-search-box-for-suppliers

Admin users want to search for suppliers by Companies House number.  Happily this is very easy for us to do.

Currently users do not need to search by an overseas registration number, which simplifies things. 

Corresponding API client changes: https://github.com/alphagov/digitalmarketplace-apiclient/pull/167

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -47,6 +47,8 @@ def list_suppliers():
 
     duns_number = request.args.get('duns_number')
 
+    company_registration_number = request.args.get('company_registration_number')
+
     if framework:
         is_valid_string_or_400(framework)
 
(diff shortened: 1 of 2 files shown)


app/main/views/suppliers.py
```diff
@@ -64,11 +66,18 @@ def list_suppliers():
     else:
         suppliers = Supplier.query.order_by(Supplier.name, Supplier.supplier_id)
 
+    # Can search by either DUNS or Company Registration number but not both
```
lfdebrux: Should we raise an error if the user tries to search using both?

app/main/views/suppliers.py
```diff
@@ -64,11 +66,18 @@ def list_suppliers():
...
     if duns_number:
         is_valid_string_or_400(duns_number)
         suppliers = suppliers.filter(
             Supplier.duns_number == duns_number
         )
+    elif company_registration_number:
+        is_valid_string_or_400(company_registration_number)
+        # TODO: enable search by other registration number
```
lfdebrux: This is a really minor point, but I feel this TODO is in the wrong place? If we were to add searching by other companies numbers, wouldn't it be as part of the if..elif block?

katstevens: I removed that in the end - we're not going to need to search by the other number.

app/main/views/suppliers.py
katstevens: I've just ignored it for now, the apps aren't ever likely to search by both at once. I think in general we ignore 'bad' query strings rather than raising errors. Happy to change if people feel differently?

app/main/views/suppliers.py
lfdebrux: Okay, no worries

PR #854, opened 2019-01-07
Test API stubs

katstevens:
Trello: https://trello.com/c/FYPvix7N/270-update-api-stubs-in-dm-utils

Basic tests for the API stubs added to `dmtestutils` (see https://github.com/alphagov/digitalmarketplace-test-utils/pull/16)

On the `Supplier` serialization tests I've removed `otherCompanyRegistrationNumber` from the serialized payload, as this won't realistically be included alongside a `companiesHouseNumber`.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -13,7 +13,7 @@ pytest-cov==2.6.0
 requests-mock==1.5.2
 testfixtures==6.3.0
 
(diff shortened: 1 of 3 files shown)


katstevens:
The reasoning behind that is because it's this repo where the `serialize()` method for each model is defined. Therefore if a model changes (eg we add a new field and include it in that model's API response), it's immediately obvious that the stub needs updating too.

PR #856, opened 2019-01-15
Test BriefResponse API stub

katstevens:
See https://github.com/alphagov/digitalmarketplace-test-utils/pull/19

Also updates the test for BriefStub to handle `with_users` and `with_clarification_questions`.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -13,7 +13,7 @@ pytest-cov==2.6.0
 requests-mock==1.5.2
 testfixtures==6.3.0
 
-git+https://github.com/alphagov/digitalmarketplace-test-utils.git@2.1.1#egg=digitalmarketplace-test-utils==2.1.1
+git+https://github.com/alphagov/digitalmarketplace-test-utils.git@2.2.0#egg=digitalmarketplace-test-utils==2.2.0
 
 # For schema generation
 alchemyjsonschema==0.5.0
--- a/tests/models/test_main.py
+++ b/tests/models/test_main.py
@@ -20,7 +20,7 @@ from tests.bases import BaseApplicationTest
 from tests.helpers import FixtureMixin
 
 from dmtestutils.api_model_stubs import (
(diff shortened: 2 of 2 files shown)


requirements-dev.txt
```diff
@@ -13,7 +13,7 @@ pytest-cov==2.6.0
 requests-mock==1.5.2
 testfixtures==6.3.0
 
-git+https://github.com/alphagov/digitalmarketplace-test-utils.git@2.1.1#egg=digitalmarketplace-test-utils==2.1.1
+git+https://github.com/alphagov/digitalmarketplace-test-utils.git@ks-api-stubs-brief-response#egg=digitalmarketplace-test-utils==2.2.0
```
benvand: dev commit :)

tests/models/test_main.py
```diff
@@ -672,7 +672,9 @@ def test_brief_serialize_keys_match_api_stub_keys(self):
         brief_stub = BriefStub(
             framework_slug=self.framework.slug, status='open', lot=self.lot, clarification_questions_closed=True,
         )
-        assert sorted(brief.serialize(with_users=True).keys()) == sorted(brief_stub.response().keys())
+        assert sorted(brief.serialize().keys()) == sorted(brief_stub.response().keys())
+        assert sorted(brief.serialize(with_users=True, with_clarification_questions=True).keys()) == \
+            sorted(brief_stub.single_result_response()['briefs'].keys())
```
benvand: Can use parenthesis implied line continuation rather than backslash
https://www.python.org/dev/peps/pep-0008/#maximum-line-length

```
        assert(
            sorted(brief.serialize(with_users=True, with_clarification_questions=True).keys()) ==
            sorted(brief_stub.single_result_response()['briefs'].keys())
        )
```

requirements-dev.txt
katstevens: It's so the CI passes while I'm waiting for the test utils PR to be merged 😸

tests/models/test_main.py
katstevens: No problem - will fixup

PR #857, opened 2019-01-15
Include registered supplier name in prefix search

katstevens:
Trello: https://trello.com/c/aSOduJAQ/319-admin-search-for-registered-name-as-well-as-supplier-name

Filters by either `Supplier.name` or `Supplier.registered_name`, given a query string prefix.


risicle:
D'ya think we should add expression indexes on these two columns for `lower(name)` so we can get fast lookups? Or do we have too few to worry about?

katstevens:
Hmm, good question. Shall we keep an eye on this endpoint for slow requests?

risicle:
Oh hold on I haven't run the tests myself yet.

risicle:
They do :+1:

PR #860, opened 2019-01-22
set_a_declaration view: add handling of PATCH method 

risicle:
https://trello.com/c/rMwWZyBv

This PR also safens-up a couple of existing views by taking row locks where appropriate and reduces the amount of unnecessary eager loading that's taking place. Also refactored tests for this view which were previously covering only a subset of its functionality.

This shouldn't go anywhere near merging until https://trello.com/c/X4vuC79J has been dealt with satisfactorily.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -450,19 +450,25 @@ def remove_contact_information_personal_data(supplier_id, contact_id):
 
 @main.route('/suppliers/<int:supplier_id>/frameworks/<framework_slug>/declaration', methods=['PUT'])
 def set_a_declaration(supplier_id, framework_slug):
-    framework = Framework.query.filter(
-        Framework.slug == framework_slug
-    ).first_or_404()
-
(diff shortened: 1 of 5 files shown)


risicle:
> There are a couple of test class bases that aren't used (leftover from a refactor?)

Um, no? If there are it's probably a mistake.

katstevens:
Oh. I meant `TestSetSupplierDeclarationsBasicPatch` and `TestSetSupplierDeclarationsBasicPut` -  I can't see where they're used?

risicle:
Ah - they're actual tests - they use `JSONUpdateTestMixin` which is something I didn't know about before. Was looking at the tests in them and thought it would be good if I found a way to ensure none of their calls wrote to the db. Mocked out `.commit()` or something? Dunno.

risicle:
I've pushed some commits which add a hacky explicit self re-assignment of the declaration simply to ensure the validators are run until we can get https://trello.com/c/X4vuC79J sorted

PR #862, opened 2019-02-05
Test service stubs against the API

lfdebrux:
https://github.com/alphagov/digitalmarketplace-test-utils/pull/20 adds stubs for ArchivedService, DraftService, and Service, this PR tests them against the API.

This is part of the work I've been doing to [clean up the signature page generation script](https://trello.com/c/jEZzVo2u/326-tidy-signature-page-generation-script)


Makefile
```diff
@@ -33,7 +30,12 @@ freeze-requirements: virtualenv requirements-dev requirements-app.txt
 	${VIRTUALENV_ROOT}/bin/python -m dmutils.repoutils.freeze_requirements requirements-app.txt
 
 .PHONY: test
-test: test-requirements test-flake8 test-migrations test-unit
+test: test-bootstrap test-requirements test-flake8 test-migrations test-unit
+
+.PHONY: test-bootstrap
+test-bootstrap: virtualenv
+	dropdb -h ${DATABASE_HOST} --if-exists digitalmarketplace_test
+	createdb -h ${DATABASE_HOST} digitalmarketplace_test
```
risicle: Putting this in as a prerequisite for `test` seems like it's just going to cause trouble with e.g. someone who doesn't use the default user for their db or needs to use a postgres that's on a different port or something. Previously `bootstrap.sh` was never called automatically.

tests/models/test_main.py
```diff
@@ -2520,3 +2595,49 @@ def test_current_framework_agreement_with_countersigned_and_new_signed(self):
...
+        self.framework_id = \
+            self.setup_dummy_framework("g-cloud-11", "g-cloud", "G-Cloud 11", id=111)
+        self.supplier_id = self.setup_dummy_suppliers(1)[0]
+        self.service_id = "1000000000"
+        self.setup_dummy_service(
+            data={"serviceName": "Cloud Pies"},
+            service_id=self.service_id,
+            supplier_id=self.supplier_id,
+            status="published",
+            lot_id=9,  # cloud-hosting
```
risicle: This number isn't _actually_ fixed (or at least, it shouldn't be).

risicle: Oh ignore me - this is the test dataset of course.

risicle:
I've pushed a tiny changeset on top of this fixing the flake8 complaint - squash if you like.

risicle:
(but of course travis still isn't passing because of the missing `-test-utils` 2.3.0, but tests pass locally for me)

PR #895, opened 2019-02-18
Remove VAT number from data to copy into supplier declaration

lfdebrux:
We no longer include company VAT number in the information we ask suppliers for, so this PR removes an unneccesary copy.

As well as running unit tests locally I ran the functional tests against a dmrunner deployment, everything passed.

Closes ticket https://trello.com/c/cscYDkOj


app/supplier_utils.py
```diff
@@ -94,7 +94,6 @@ def update_open_declarations_with_company_details(db, supplier_id, updater_json,
                 supplier.companies_house_number or supplier.other_company_registration_number
             ),
             supplier_constants.KEY_DUNS_NUMBER: supplier.duns_number,
-            supplier_constants.KEY_VAT_NUMBER: supplier.vat_number,
```
katstevens: We can also remove the KEY_VAT_NUMBER line from `supplier_constants.py`.

lfdebrux: 👍

PR #897, opened 2019-02-18
Update requirements-dev

lfdebrux:
Bumps updated dependencies in `requirements-dev.txt`.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -4,8 +4,9 @@
 # F401: 'identifier' imported but unused
 # E402: module level import not at top of file
 # W503: line break before binary operator
+# W504: line break after binary operator
 exclude = venv*,__pycache__,node_modules,migrations,.git
-ignore = D203,W503
+ignore = D203,W503,W504
 max-complexity = 24
 max-line-length = 120
 per-file-ignores =
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ requirements: virtualenv test-requirements requirements.txt
 
 .PHONY: requirements-dev
 requirements-dev: virtualenv requirements-dev.txt
-	${VIRTUALENV_ROOT}/bin/pip install -r requirements-dev.txt
+	${VIRTUALENV_ROOT}/bin/pip install -Ur requirements-dev.txt
 
 .PHONY: freeze-requirements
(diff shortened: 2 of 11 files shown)


jonodrew:
The error is the same on you helped me fix last week - we had to reorder the file: https://github.com/alphagov/digitalmarketplace-utils/pull/490/commits/368e50fb5396749076eba28505cb45c5d6642efb

lfdebrux:
> The error is the same on you helped me fix last week - we had to reorder the file: [alphagov/digitalmarketplace-utils@368e50f](https://github.com/alphagov/digitalmarketplace-utils/commit/368e50fb5396749076eba28505cb45c5d6642efb)

Ah yes that makes sense, thanks

lfdebrux:
> > The error is the same on you helped me fix last week - we had to reorder the file: [alphagov/digitalmarketplace-utils@368e50f](https://github.com/alphagov/digitalmarketplace-utils/commit/368e50fb5396749076eba28505cb45c5d6642efb)
> 
> Ah yes that makes sense, thanks

I actually decided to fix this a different way, which I think will be more sustainable; adding the `-U` flag to `pip` for `make requirements-dev` makes sure we get the latest valid version of sub-dependencies.

Any objections @jonodrew, @risicle?

jonodrew:
> > > The error is the same on you helped me fix last week - we had to reorder the file: [alphagov/digitalmarketplace-utils@368e50f](https://github.com/alphagov/digitalmarketplace-utils/commit/368e50fb5396749076eba28505cb45c5d6642efb)
> > 
> > 
> > Ah yes that makes sense, thanks
> 
> I actually decided to fix this a different way, which I think will be more sustainable; adding the `-U` flag to `pip` for `make requirements-dev` makes sure we get the latest valid version of sub-dependencies.
> 
> Any objections @jonodrew, @risicle?

None from me

risicle:
Yeah I think that seems like the least wrong way of doing it. Or rather - it's one of those things that seems like pip should be doing by default, so let's force it.

risicle:
FWIW we do have across the project a number of places with regexes not using raw strings - I fix them where I find them, but I'm sure there are more lurking.

PR #898, opened 2019-02-19
auth_user endpoint: lock user row FOR UPDATE to prevent race condition when incrementing user.failed_login_count

risicle:
Fixes https://trello.com/c/ShKl24A0

Also remove `User.get_by_email_address()` as it is restrictive and ... not very useful.

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -41,7 +41,14 @@ def auth_user():
     json_payload = json_payload["authUsers"]
     validate_user_auth_json_or_400(json_payload)
 
-    user = User.get_by_email_address(json_payload['emailAddress'].lower())
+    user = User.query.filter(
+        User.email_address == json_payload['emailAddress'].lower()
+    ).options(
+        # specifying a lazyload here to satisfy the FOR UPDATE OF's limitations, but inevitably the supplier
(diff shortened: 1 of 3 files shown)


app/main/views/users.py
```diff
@@ -41,7 +41,14 @@ def auth_user():
...
 
-    user = User.get_by_email_address(json_payload['emailAddress'].lower())
+    user = User.query.filter(
+        User.email_address == json_payload['emailAddress'].lower()
+    ).options(
+        # specifying a lazyload here to satisfy the FOR UPDATE OF's limitations, but inevitably the supplier
+        # will be fetched on-demand when the result comes to be serialized.
+        # the FOR UPDATE lock is required to prevent a race condition when incrementing user.failed_login_count
+        lazyload(User.supplier),
+    ).with_for_update(of=User).first()
```
katstevens: We use `with_for_update(read=True)` in one of our other queries, would it be useful here? 🤷‍♀️

risicle: I think we actually end up writing in every case other than 404 - in the success case we reset their `failed_login_count` and update `logged_in_at`.

katstevens: Fair enough.

PR #899, opened 2019-02-22
Bump digitalmarketplace-utils dependency to v46.1.2, update misc other dependencies

risicle:
This is non-essential but should bring in the improved `/_status` logging noted as a problem in https://trello.com/c/CqTPp8E0

### DIFF
--- a/requirements-app.txt
+++ b/requirements-app.txt
@@ -7,12 +7,12 @@ Flask-Migrate==2.0.3
 Flask-SQLAlchemy==2.1
 itsdangerous==0.24 # pyup: ignore
 --no-binary=psycopg2
-psycopg2==2.7.3
-SQLAlchemy==1.2.12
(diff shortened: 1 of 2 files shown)


risicle:
Interesting that it's failing the snyk test - the issue that snyk is citing *actually* appears to apply to *all* 1.2 versions of sqlalchemy, but they've got it in their db as 1.2.17+

risicle:
(I've dropped them a note about it)

risicle:
Think I should "ignore" that vuln for now?

katstevens:
If we're already vulnerable then ignoring would leave us no worse off?

risicle:
The vulnerability is a bit of a "sqlalchemy lets you shoot yourself in the foot" vulnerability, and I don't think we use the functionality that is vulnerable. There is a fix merged to master but it hasn't made it to a release yet.

PR #911, opened 2019-03-12
auth failures: don't bother keeping IP information in audit log, instead rely on trace/span ids

risicle:
https://trello.com/c/9KsjF8Ns

The way the APIs are served these days, it'll be hard to get a single useful IP for a client from the API process. it's probably much better to rely on our regular logs and instead make the association as strong as possible by adding the request's span id as well as trace id.


risicle:
Should have checked the tests shouldn't I?

PR #919, opened 2019-04-15
Update copied_to_following_framework flag if supplied

katstevens:
Trello: https://trello.com/c/u0QnZn8y/459-deleting-a-copied-draft-service-should-reset-the-source-service-flag

Currently a call to `update_service` can only update the service `data` itself, not the `copied_to_following_framework` flag. This PR checks for the flag in the request payload and sets the flag if necessary.

To keep changes minimal, the service is still archived and re-indexed via the helper function, as per any other service update.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -144,8 +144,11 @@ def update_service(service_id):
     update_details = validate_and_return_updater_request()
     update = validate_and_return_service_request(service_id)
 
-    updated_service = update_and_validate_service(service, update)
+    # Check for an update to the copied_to_following_framework flag on the service object
+    if 'copiedToFollowingFramework' in update and isinstance(update['copiedToFollowingFramework'], bool):
+        service.copied_to_following_framework = update['copiedToFollowingFramework']
 
+    updated_service = update_and_validate_service(service, update)
     audit_type = (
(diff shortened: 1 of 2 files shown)


app/main/views/services.py
```diff
@@ -144,8 +144,11 @@ def update_service(service_id):
     update_details = validate_and_return_updater_request()
     update = validate_and_return_service_request(service_id)
 
-    updated_service = update_and_validate_service(service, update)
+    # Check for an update to the copied_to_following_framework flag on the service object
+    if 'copiedToFollowingFramework' in update and isinstance(update['copiedToFollowingFramework'], bool):
```
risicle: It would probably be nicer to throw a 400 if it's not a bool rather than silently ignore it.

katstevens: It would. I was looking at adding it to one of the schemas but the service data validation happens elsewhere, and the `validate_and_return_updater_request` and `validate_and_return_service_request` are just checking for `updated_by` and the top level `services` key. 

tl:dr I'll add a quick commit to raise on non-bools.

matthewford:
This prevented me from editing/updating an existing service copied from G10 for a G11 submission, in the end, I needed to copy and paste to replicate the service.

As an aside, I was able to copy the broken service, it throws the 400 error but still duplicated the service.

PR #928, opened 2019-05-13
Exclude DOS services from deindexing

katstevens:
Trello: https://trello.com/c/8U5bZl2b/778-removing-a-dos-service-shouldnt-try-to-de-index-it-from-elasticsearch

This has been causing some 500s when admins remove services for a DOS supplier, which aren't in Elasticsearch to begin with.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -334,7 +334,6 @@ def update_service_status(service_id, status):
                                            'new_status': status})
 
     if prior_status != status:
-
         # If it's being unpublished, delete it from the search api.
         if prior_status == 'published':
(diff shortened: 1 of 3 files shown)


lfdebrux:
So previously `index_service` was being called with DOS services, but it ignored them because it tests `framework.framework` https://github.com/alphagov/digitalmarketplace-api/blob/master/app/service_utils.py#L134-L138; for symmetries sake would it be better to put a similar check in `delete_service_from_index`, rather than in the view?

katstevens:
@lfdebrux Yep that's a good shout. Maybe that'll be easier to test as well 😸

PR #934, opened 2019-05-29
Reset copiedtoFollowingFramework flag on service when draft is deleted

katstevens:
Trello: https://trello.com/c/n7cCL1ZM/411-if-a-user-copies-a-service-from-the-previous-framework-then-deletes-that-service-they-will-not-be-able-to-copy-it-again

- When a draft service (the 'copy') is created, add in the ID of the source service to the copy's `data` blob
- If the copy is deleted, look up the source service and reset the `copiedToFollowingFramework` flag.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -77,6 +77,8 @@ def copy_draft_service_from_existing_service(service_id):
     )
 
     if target_framework_id != service.framework.id:
+        # TODO: convert data['copiedFromServiceId'] to a foreign key field on the model, linking the new draft with the
+        # source service. That way if the draft is deleted, the relationship to the source service will also be removed.
         service.copied_to_following_framework = True
         draft.data['copiedFromServiceId'] = service_id
 
@@ -316,6 +318,7 @@ def delete_draft_service(draft_id):
     # a service is copied, it's removed from the list of available services to copy from.
     # Resetting the .copied_to_following_framework flag on the source when the copy is deleted allows
     # the user to try again with a fresh copy.
+    # TODO: remove this when draft.data['copiedFromServiceId'] is converted to a foreign key field
     source_service, source_service_audit = None, None
     if 'copiedFromServiceId' in draft.data:
         source_service_id = draft.data['copiedFromServiceId']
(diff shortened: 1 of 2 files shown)


risicle:
Now, y'see, if this were an actual database field with a foreign key constraint, it would make it impossible for it to not have this property reset when the draft is deleted...

app/main/views/drafts.py
```diff
@@ -311,6 +314,31 @@ def delete_draft_service(draft_id):
...
+            source_service.copied_to_following_framework = False
+            source_service_audit = AuditEvent(
+                audit_type=AuditTypes.update_service,
+                user=updater_json['updated_by'],
+                data={
+                    "serviceId": draft.service_id,
+                    "supplierId": draft.supplier_id,
+                    "copiedToFollowingFramework": False
+                },
+                db_object=None
```
risicle: I think you can still add it as the `db_object` - it's not actually enforced as a foreign key (can't be). Might need to add the "type" and "id" manually though so sqlalchemy doesn't try and be too clever and check its existence...

risicle: Oh hold on I'm looking at the wrong audit event here - if it's the audit event for the service, it should _definitely_ have the object set shouldn't it?

tests/main/views/test_drafts.py
```diff
@@ -1115,9 +1128,61 @@ def test_should_delete_a_draft(self):
         assert data['auditEvents'][1]['type'] == 'delete_draft_service'
         assert data['auditEvents'][1]['data']['serviceId'] == self.service_id
 
+        # Check the draft has gone
+        fetch_again = self.client.get('/draft-services/{}'.format(draft_id))
+        assert fetch_again.status_code == 404
```
risicle: :+1:

app/main/views/drafts.py
katstevens: Good spot. Confession: I copied and pasted this from somewhere else. Will take a closer look.

katstevens: Yes it should. And it should get the data from the `source_service`, not the draft 🤦‍♀ Commit pushed to fix.

PR #940, opened 2019-06-12
SupplierFramework: add allow_declaration_reuse

risicle:
This is to allow https://trello.com/c/V7cUDIRW/, and adds an `allow_declaration_reuse` flag to `SupplierFramework` that should work in the same way as the existing `Framework.allow_declaration_reuse`.

I've also made the endpoint perform some checks that the setting of `prefillDeclarationFromFrameworkSlug` in `update_supplier_framework` doesn't violate these rules. This isn't done in a 100% foolproof way - a request that comes in milliseconds later to change a relevant `allow_declaration_reuse` flag's value will succeed - the "reverse-check" is not implemented. If we cared about that, I really think that's the point at which we'd be better off looking at real db constraints.

The migration adding the column and then back-populating default values is performed in several individual steps to reduce locking time. More information available in commit messages.

Also depends on https://github.com/alphagov/digitalmarketplace-test-utils/pull/24 which I don't want to merge until I'm more sure on the shape of this.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -12,7 +12,7 @@ pytest-cov==2.7.1
 requests-mock==1.6.0
 testfixtures==6.8.2
 
-git+https://github.com/alphagov/digitalmarketplace-test-utils.git@2.3.0#egg=digitalmarketplace-test-utils==2.3.0
+git+https://github.com/alphagov/digitalmarketplace-test-utils.git@2.6.0#egg=digitalmarketplace-test-utils==2.6.0
 
 # For schema generation
 alchemyjsonschema==0.5.0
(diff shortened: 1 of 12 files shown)


katstevens:
This all looks reasonable to me.

requirements-dev.txt
```diff
@@ -12,7 +12,7 @@ pytest-cov==2.7.1
 requests-mock==1.6.0
 testfixtures==6.8.2
 
-git+https://github.com/alphagov/digitalmarketplace-test-utils.git@2.3.0#egg=digitalmarketplace-test-utils==2.3.0
+git+https://github.com/alphagov/digitalmarketplace-test-utils.git@ris-sf-allow-declaration-reuse#egg=digitalmarketplace-test-utils==2.6.0
```
benvand: `ris-sf-allow-declaration-reuse` ;)

risicle: Yeah it's to allow people to test it - haven't merged it to test-utils yet.

PR #946, opened 2019-06-26
Add suppliers 'active' flag

lfdebrux:
https://trello.com/c/t85UtxUa/560-inactive-suppliers

We want to be able to indicate in the database whether a supplier is inactive (i.e. they have stopped existing, have merged with another supplier). This commit adds a simple boolean flag to the supplier table.  (@agalamatis actually wrote it, I'm just commiting it :P).

The migration is a bit involved, thanks @risicle for taking me through it. You can see the SQL that will be produced below:

```
$ ./application.py db upgrade 1320:1330 --sql
-- Running upgrade 1320 -> 1330

ALTER TABLE suppliers ADD COLUMN active BOOLEAN;

ALTER TABLE suppliers ALTER COLUMN active SET DEFAULT true;

UPDATE suppliers SET active = true WHERE active = NULL;

ALTER TABLE suppliers ALTER COLUMN active SET NOT NULL;

UPDATE alembic_version SET version_num='1330' WHERE alembic_version.version_num = '1320';

COMMIT;
```

### DIFF
--- a/migrations/versions/1330_add_suppliers_active_flag.py
+++ b/migrations/versions/1330_add_suppliers_active_flag.py
@@ -1,0 +1,29 @@
+"""add suppliers active flag
+
+Adds a new column to the suppliers table, 'active', that indicates whether a
+supplier is currently valid.
+
+Revision ID: 1330
+Revises: 1320
+Create Date: 2019-06-26 11:14:56.085586
+
+"""
+from alembic import context, op
+import sqlalchemy as sa
+
+
+
+# revision identifiers, used by Alembic.
+revision = '1330'
+down_revision = '1320'
+
+
+def upgrade():
+    # We want this column to be DEFAULT true, but creating a column
+    # with defaults is a slow operation, so we split it up a bit.
+    op.add_column('suppliers', sa.Column('active', sa.Boolean(), nullable=True))
+    op.alter_column('suppliers', 'active', server_default=sa.text('true'))
+
+
+def downgrade():
+    op.drop_column('suppliers', 'active')
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -409,6 +409,9 @@ class Supplier(db.Model):
     # SupplierFramework.application_company_details_confirmed
     company_details_confirmed = db.Column(db.Boolean, index=False, default=False, nullable=False)
 
+    # This flag indicates if a supplier is no longer providing any services.
+    active = db.Column(db.Boolean, default=True, server_default=sql_true(), nullable=False)
+
     @validates('trading_status')
     def validates_trading_status(self, key, value):
         if value not in self.TRADING_STATUSES:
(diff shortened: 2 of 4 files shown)


lfdebrux:
> If we have auto-commit turned off, it looks great. Thanks

What is auto-commit?

migrations/versions/1330_add_suppliers_active_flag.py
```diff
@@ -0,0 +1,34 @@
...
+
+def upgrade():
+    # We want this column to be DEFAULT true and NOT NULLABLE, but creating a column
+    # with defaults is a slow operation, so we split it up a bit.
+    op.add_column('suppliers', sa.Column('active', sa.Boolean(), nullable=True))
+    op.alter_column('suppliers', 'active', server_default=sa.text('true'))
+
+    # Set any NULL values
+    # NULLs are active suppliers (i.e. they have not been made inactive)
+    op.execute("UPDATE suppliers SET active = true WHERE active = NULL")
```
risicle: To be useful this has to be done in a separate transaction (i.e. migration)

lfdebrux: I experimented with this but it seems that alembic puts all migrations applied into one transaction anyway?

risicle:
It's not useful performing the column addition in several steps if you're not going to do them as separate transactions, because transactions never really "downgrade" their locks. If a transaction takes a strong lock, (as it will when adding a column), it will only drop that lock when the transaction ends. In this migration, the "table exclusive" lock will be held right from beginning to end. This is why I do it as 3 transactions.

lfdebrux:
> It's not useful performing the column addition in several steps if you're not going to do them as separate transactions, because transactions never really "downgrade" their locks. If a transaction takes a strong lock (as it will when adding a column), it will only drop that lock when the transaction ends. In this migration, the "table exclusive" lock will be held right from beginning to end. This is why I do it as 3 migrations.

Hmm, I guess we'll have to do this as two or three PRs then

risicle:
Why? Single PR, 3 transactions - easy.

risicle:
https://stackoverflow.com/a/54541984 if we don't have it set to 1 transaction per migration we definitely need to fix that.

risicle:
ok the change is simple enough and it seems to work:

```
--- a/migrations/env.py
+++ b/migrations/env.py
@@ -37,7 +37,10 @@ def run_migrations_offline():
 
     """
     url = config.get_main_option("sqlalchemy.url")
-    context.configure(url=url)
+    context.configure(
+        transaction_per_migration=True,
+        url=url,
+    )
 
     with context.begin_transaction():
         context.run_migrations()
@@ -56,6 +59,7 @@ def run_migrations_online():
 
     connection = engine.connect()
     context.configure(
+                transaction_per_migration=True,
                 connection=connection,
                 target_metadata=target_metadata
                 )
```

migrations/versions/1330_add_suppliers_active_flag.py
lfdebrux: Okay I've added your patch and split the migration in two.

PR #947, opened 2019-06-26
Fix migration 1340

lfdebrux:
The last PR had bug where I was using the wrong type of comparison to NULL.

Our preview DB is on version 1330 so I don't think we need to do anything more complicated than fix the bad migration script.


risicle:
Really, add a new migration - it's not worth potential confusion - "well _I've_ run migration 1340" - "er, so have i - but it's not working for me"...

risicle:
The existing migration will have just executed on zero rows, so it's not really making migrations do extra work.

lfdebrux:
The existing migration crashes

risicle:
Um... it didn't for me...?

lfdebrux:
> Um... it didn't for me...?

https://ci.marketplace.team/job/database-migration-paas/665/consoleFull

risicle:
Huh, actually it does... why didn't my tests fail then?

PR #950, opened 2019-07-09
Acknowledge audit events for deleting a copied service

lfdebrux:
https://trello.com/c/JSPtBsU5/933-bug-in-admin-frontend

Admins should not need to approve a reset of the copiedToFollowingFramework flag.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -4,6 +4,8 @@ from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import lazyload
 from sqlalchemy import asc, desc
 
+from datetime import datetime
+
 from .. import main
 from ... import db
 from ...validation import is_valid_service_id_or_400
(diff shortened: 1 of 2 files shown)


app/main/views/drafts.py
```diff
@@ -350,6 +353,10 @@ def delete_draft_service(draft_id):
         db_object=None
     )
 
+    audit.acknowledged = True
+    audit.acknowledged_by = "acknowledged_at_creation"
+    audit.acknowledged_at = datetime.utcnow()
```
katstevens: It's not this audit we want to change - it's the `source_service_audit` above

app/main/views/drafts.py
```diff
@@ -350,6 +353,10 @@ def delete_draft_service(draft_id):
         db_object=None
     )
 
+    audit.acknowledged = True
+    audit.acknowledged_by = "acknowledged_at_creation"
```
katstevens: I'd put something like `acknowledging service copy reset` (which is what is happening).

lfdebrux: Good idea

PR #951, opened 2019-07-10
Strengthen DUNS validation 💪

jonodrew:
The API now validates that DUNS numbers are the right length, as well as the right kind of characters. I've written a test for both of the validators that validate supplier JSON. [Ticket](https://trello.com/c/AtOz5X7I/414-duns-number-validation-in-the-api-is-weak-sept-17)

### DIFF
--- a/json_schemas/suppliers.json
+++ b/json_schemas/suppliers.json
@@ -26,7 +26,9 @@
     },
     "dunsNumber": {
       "pattern": "^[0-9]+$",
-      "type": "string"
+      "type": "string",
+      "minLength": 9,
+      "maxLength": 9
     },
     "id": {
       "type": "integer"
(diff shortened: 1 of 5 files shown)


json_schemas/suppliers.json
```diff
@@ -26,7 +26,9 @@
     },
     "dunsNumber": {
       "pattern": "^[0-9]+$",
-      "type": "string"
+      "type": "string",
+      "minLength": 9,
+      "maxLength": 9
```
katstevens: Are there any existing suppliers with a DUNS number of the wrong length? Suppliers can't change this value once they've created an account, so they won't be able to edit any other days if their DUNS doesn't validate.

jonodrew:
I've closed this in favour of [this new PR](https://github.com/alphagov/digitalmarketplace-api/pull/952)

PR #953, opened 2019-07-10
Change indices to unique constraints

jonodrew:
It seems that unique constraints are faster than indices, so here's a migration and a small code change that implements unique where possible. The only index not converted here is `suppliers.supplier_id`. Changing this would require dropping a lot of foreign keys, so we're not going to do that today. [Ticket](https://trello.com/c/xzi2dA03/600-make-sure-were-using-unique-constraints-over-unique-indexes)


risicle:
(For more on the _subtle_ difference between unique indexes and unique constraints see https://medium.com/flatiron-engineering/uniqueness-in-postgresql-constraints-versus-indexes-4cf957a472fd)

risicle:
Cool WFM.

I imagine it will be fine but a bit of me wonders whether we should split each of these conversions up into separate migrations, considering that writes to the affected tables will be blocked while the new indexes are being built. By the end of this migration's transaction we'll have 4 tables blocked for writes.

jonodrew:
@risicle - different commits or different PRs entirely?

risicle:
Neither - just separate migration files. But what do others think...?

katstevens:
My gut says split 'em up. Being able to upgrade/downgrade in a granular way is a good rule of thumb.

PR #954, opened 2019-07-17
Add configuration to allow searching audit events by draft service ID

lfdebrux:
Trello: https://trello.com/c/5MAl7BHa/607-auditevents-allow-draftservices-as-an-object-id

### DIFF
--- a/tests/main/views/test_drafts.py
+++ b/tests/main/views/test_drafts.py
@@ -162,7 +162,7 @@ class TestCopyDraftServiceFromExistingService(DraftsHelpersMixin):
             content_type='application/json')
         assert res.status_code == 201
 
-        audit_response = self.client.get('/audit-events')
+        audit_response = self.client.get('/audit-events?object-type=draft-services')
         assert audit_response.status_code == 200
 
         data = json.loads(audit_response.get_data())
@@ -767,7 +767,7 @@ class TestDraftServices(DraftsHelpersMixin):
         data = json.loads(res.get_data())
         draft_id = data['services']['id']
 
-        audit_response = self.client.get('/audit-events')
+        audit_response = self.client.get('/audit-events?object-type=draft-services')
         assert audit_response.status_code == 200
         data = json.loads(audit_response.get_data())
         assert len(data['auditEvents']) == 1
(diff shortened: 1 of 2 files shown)


tests/main/views/test_drafts.py
```diff
@@ -162,7 +162,7 @@ def test_create_draft_from_existing_should_create_audit_event(self):
             content_type='application/json')
         assert res.status_code == 201
 
-        audit_response = self.client.get('/audit-events')
+        audit_response = self.client.get('/audit-events?object-type=draft-services')
```
katstevens: Do we need to add the query string to all these tests? I think for at least some of them we should be making sure that the test is only creating 1 audit event (of any type).

lfdebrux: Hmm fair point...

lfdebrux: My thinking was that we would want to add the query string to check that all the endpoints were adding draft service object to the audit event.

katstevens: Sure - could we cover that scenario with one new test instead?

lfdebrux: @katstevens I think I may have found an issue, which is that currently the `delete_draft_service` and `publish_draft_service` audit events don't have draft services attached, so they wouldn't appear in a query by draft service object type.

delete_draft_service sets the db_object to `None`. publish_draft_service sets the db_object to the newly created service, rather than the draft service.

lfdebrux: I think I'm going to close this PR and instead create a new one that adds a parameter to the audits endpoint called data-draft-id, similar to data-supplier-id, that allows filtering by draftId in the audit event data. It looks like the audit events already include this data so this wouldn't be a breaking change.

katstevens: OK. Let's have a chat about this

lfdebrux:
Closing this in favour of https://github.com/alphagov/digitalmarketplace-api/pull/955

lfdebrux:
Closing this in favour of https://github.com/alphagov/digitalmarketplace-api/pull/955

PR #955, opened 2019-07-18
Add data-draft-service-id parameter to audit-events endpoint

lfdebrux:
Trello: https://trello.com/c/5MAl7BHa/607-auditevents-allow-draftservices-as-an-object-id

We want to allow searching for audit events relating to a specific draft
service. The easiest way to do this is to add a new parameter to the
audit-events endpoint, `data-draft-service-id`, that when provided will
filter the audit events to those which contain `draftId` of that value
in their data.

### DIFF
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -97,6 +97,12 @@ def list_audits():
             ) == data_supplier_id
         )
 
+    data_draft_service_id = request.args.get('data-draft-service-id')
+    if data_draft_service_id:
+        audits = audits.filter(
+            AuditEvent.data['draftId'].astext == data_draft_service_id
+        )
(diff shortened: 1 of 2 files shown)


katstevens:
I love it. Let's get a unit test in there 👍

lfdebrux:
> I love it. Let's get a unit test in there 👍

Tests added.

risicle:
Oh I guess I missed the boat on this one - but if we keep adding funny special cases like this one, where will it end? This is how we end up with dysfunctional APIs.

risicle:
Also this filter has **no index**. Have you tried it on production with the n million audit events we have? I suspect it will bring it to its knees.

PR #960, opened 2019-08-06
AuditEvent: add partial index for data['draftId']

risicle:
Similar to that for `supplierId`/`supplier_id`, but simpler because we only use one key for `draftId`.

Also some tidying up of the queries used in the `list_audits` view.

### DIFF
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -87,20 +87,25 @@ def list_audits():
             AuditEvent.user == user
         )
 
-    data_supplier_id = request.args.get('data-supplier-id')
+    # note in the following that even though the supplier and draft ids *are* integers, we're doing the searches as
+    # strings because of the postgres static type system's awkwardness with json types. we first let args.get normalize
+    # them into actual integers though
+
+    data_supplier_id = request.args.get('data-supplier-id', type=int)
     if data_supplier_id:
-        #  This filter relies on index `idx_audit_events_data_supplier_id`. See `app..models.main` for its definition.
+        # This filter relies on index `idx_audit_events_data_supplier_id`. See `app..models.main` for its definition.
         audits = audits.filter(
             func.coalesce(
-                AuditEvent.__table__.c.data['supplierId'].astext,
-                AuditEvent.__table__.c.data['supplier_id'].astext,
-            ) == data_supplier_id
+                AuditEvent.data['supplierId'].astext,
+                AuditEvent.data['supplier_id'].astext,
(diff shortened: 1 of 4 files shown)


app/main/views/audits.py
```diff
@@ -87,20 +87,25 @@ def list_audits():
             AuditEvent.user == user
         )
 
-    data_supplier_id = request.args.get('data-supplier-id')
+    # note in the following that even though the supplier and draft ids *are* integers, we're doing the searches as
+    # strings because of the postgres static type system's awkwardness with json types. we first let args.get normalize
+    # them into actual integers though
+
+    data_supplier_id = request.args.get('data-supplier-id', type=int)
```
katstevens: We have an existing test for the `data-supplier-id` param (https://github.com/alphagov/digitalmarketplace-api/blob/master/tests/main/views/test_audits.py#L904) - does it need updating or does it already cover the expected change?

app/main/views/audits.py
```diff
@@ -87,20 +87,25 @@ def list_audits():
...
-                AuditEvent.__table__.c.data['supplierId'].astext,
-                AuditEvent.__table__.c.data['supplier_id'].astext,
-            ) == data_supplier_id
+                AuditEvent.data['supplierId'].astext,
+                AuditEvent.data['supplier_id'].astext,
+            ) == str(data_supplier_id)
         )
 
-    data_draft_service_id = request.args.get('data-draft-service-id')
+    data_draft_service_id = request.args.get('data-draft-service-id', type=int)
```
katstevens: Our test coverage for `data-draft-service-id` is limited to the happy path (used as an assert in `test_drafts`). Might be worth adding something into `test_audits` - up to you.

app/main/views/audits.py
risicle: It would only make a difference for cases where it e.g. zero-prefixed the number for some reason.

app/main/views/audits.py
risicle: Mayyyyyyyybe

risicle: Yeah after being initially surprised by the testing of this facet of this view being in `test_drafts.py` I later figured it might be because setting up the state with draft service audit events might be a pain in the ass, and so I wasn't about to go making this story bigger than it needs to be by uprooting it all. However it doesn't actually look like it should be difficult to add a couple of extra artificial tests to `test_audits.py`.

katstevens: Sounds good to me 👍

PR #966, opened 2019-09-02
update_service view: handle `wait-for-index` query argument

risicle:
https://trello.com/c/LABe8BKA

Depends on https://github.com/alphagov/digitalmarketplace-apiclient/pull/222

With this argument set as `false`, the request should trigger an reindexing request corresponding to the service update but not wait for its result. though of course it has the same danger any async request has in that overuse can cause a large number of requests to be sent to the target (in this case the search api) in a very short amount of time. hopefully the `update_service` view itself is slow enough to cause this to be throttled to some extent.

This follows `user-role` (in this view) in being a Bit Weird because it's accepting query string arguments in a POST request. but given the way we use the body of a request implicitly as the contents of the new object, supplying it there would be an even weirder option.

### DIFF
--- a/requirements-app.txt
+++ b/requirements-app.txt
@@ -2,7 +2,7 @@
 # with package version changes made in requirements-app.txt
 
 git+https://github.com/alphagov/digitalmarketplace-utils.git@48.4.0#egg=digitalmarketplace-utils==48.4.0
-git+https://github.com/alphagov/digitalmarketplace-apiclient.git@21.0.0#egg=digitalmarketplace-apiclient==21.0.0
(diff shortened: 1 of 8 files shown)


requirements-app.txt
```diff
@@ -2,7 +2,7 @@
 # with package version changes made in requirements-app.txt
 
 git+https://github.com/alphagov/digitalmarketplace-utils.git@48.4.0#egg=digitalmarketplace-utils==48.4.0
-git+https://github.com/alphagov/digitalmarketplace-apiclient.git@21.0.0#egg=digitalmarketplace-apiclient==21.0.0
+git+https://github.com/alphagov/digitalmarketplace-apiclient.git@ris-ignore-response#egg=digitalmarketplace-apiclient==999
```
risicle: I realize this needs to be finalized before merge...

PR #967, opened 2019-09-03
219-alert-when-the-admin-manager-password-is-changed

benvand:
Add a log statement when the 'admin-manager' password is changed so we can alert on this.

https://trello.com/c/xeZofjDd/219-alert-when-the-admin-manager-password-is-changed

### DIFF
--- a/tests/main/views/test_users.py
+++ b/tests/main/views/test_users.py
@@ -1,9 +1,10 @@
 from datetime import datetime
-
-from flask import json
-from freezegun import freeze_time
+from logging import Logger
 import mock
 import pytest
+
+from freezegun import freeze_time
+from flask import json
 from sqlalchemy.exc import DataError, IntegrityError
 
 from app import db, encryption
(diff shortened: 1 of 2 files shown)


tests/main/views/test_users.py
```diff
@@ -1166,6 +1167,83 @@ def test_can_update_phone_number(self):
...
+        """Ensude that the correct message is output when an admin manager changes their password."""
+        admin_user = self._create_admin_user('admin-manager')
+        update_data = {"updated_by": "a.user", 'users': {'password': 'new_password'}}
+
+        with self.app.test_request_context('/'):
+            self.client.post(f'/users/{admin_user.id}', data=json.dumps(update_data), content_type='application/json')
+
+        assert self.admin_manager_warning_log_message in self.app.logger.warning.mock_calls
+
+    @pytest.mark.parametrize('role', [i for i in User.ROLES if i != 'admin-manager'])
```
risicle: neat

tests/main/views/test_users.py
```diff
@@ -1166,6 +1167,83 @@ def test_can_update_phone_number(self):
...
+        "{code}: Password reset requested for {user_role} user '{email_hash}'",
+        extra={
+            'code': 'update_user.password.role_warning',
+            'email_hash': 'W-jA20fzplHyOKk82Io1EyjW_l4X25Min3NRyU0hgiY=',
+            'user_role': 'admin-manager'
+        }
+    )
+
+    def setup(self):
+        with mock.patch('flask.app.create_logger', return_value=mock.Mock(spec=Logger('flask.app'), handlers=[])):
```
risicle: Ahh I've not seen this trick before - do we use it elsewhere?

benvand: Yeah it's one of the ways we have in utils. It's good in a way because it's completely handler agnostic

https://github.com/alphagov/digitalmarketplace-utils/blob/master/tests/conftest.py#L43

PR #968, opened 2019-09-13
update_service_status endpoint: add wait-for-index option

risicle:
Similar to https://github.com/alphagov/digitalmarketplace-apiclient/pull/223, when attempting to implement https://trello.com/c/LABe8BKA it would be helpful to add these options to the actual api calls that the admin view in question uses in the first place.

So this adds a `wait-for-index` option to the `update_service_status` endpoint, which in turn will use the `client_wait_for_response` argument on `SearchAPIClient`'s `delete()` or `index()` methods, depending on whether the service is being published or unpublished.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -335,12 +335,13 @@ def update_service_status(service_id, status):
                                            'new_status': status})
 
     if prior_status != status:
-        # If it's being unpublished, delete it from the search api.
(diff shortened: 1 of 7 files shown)


lfdebrux:
Few tests need updating but otherwise seems fine

risicle:
Damn yeah missed a test.

PR #970, opened 2019-09-18
Fix logging function

jonodrew:
A small bug ([ticket](https://trello.com/c/knYixsFf/302-bug-in-prod-api-publishdraftservice)) that meant the logger was failing silently. This should fix up that issue.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -418,10 +418,8 @@ def publish_draft_service(draft_id):
         db.session.rollback()
         action = 'update' if draft.status == 'submitted' else 'delete'
         current_app.logger.warning(
-            'Failed to {action} draft {draft_id} after publishing service {service_id}: {error}'.format(
-                extra=dict(
-                    action=action, draft_id=draft_id, service_id=service_from_draft.service_id, error=e.message)))
-
+            f'Failed to {action} draft {draft_id} after publishing service {service_from_draft.service_id}: {e}'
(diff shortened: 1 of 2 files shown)


app/main/views/drafts.py
```diff
@@ -418,10 +418,9 @@ def publish_draft_service(draft_id):
         db.session.rollback()
         action = 'update' if draft.status == 'submitted' else 'delete'
         current_app.logger.warning(
-            'Failed to {action} draft {draft_id} after publishing service {service_id}: {error}'.format(
-                extra=dict(
-                    action=action, draft_id=draft_id, service_id=service_from_draft.service_id, error=e.message)))
-
+            f'Failed to {action} draft {draft_id} '
+            f'after publishing service {service_from_draft.service_id}: {e.statement}'
```
katstevens: Just spotted this - why `e.statement` instead of `e.message`?

jonodrew: As far as I can tell [from the docs](https://docs.sqlalchemy.org/en/13/core/exceptions.html#sqlalchemy.exc.IntegrityError) that's how SQLAlchemy has written the class 🤔

katstevens: I think we can just pass `e` if the string is going to be formatted anyway - we do this in `views.suppliers`

jonodrew: Done!

PR #973, opened 2019-09-25
list_brief_responses: add "with-data" query parameter 

risicle:
https://trello.com/c/2FBIg5LP

This is one of those things that reveals the farce of living with an "API" - we're really not supposed to acknowledge the existence of e.g. `essentialRequirementsMet` or any explicit key in `.data`, but the specific frontend view we've got in mind for this feature _does_ use `essentialRequirementsMet`, so we can't omit it. Whereas the other members of `.data` are where a lot of the bulk data we don't need resides.

### DIFF
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -1848,6 +1848,13 @@ class BriefResponse(db.Model):
             raise ValidationError(errs)
 
     def serialize(self, with_data: bool = True):
+        """
+            :param with_data: allows serialization to be produced while omitting the majority of the content
+            that comes from the `data` field. This tends to constitute the bulk of the volume of the serialized
+            result, so this is useful in cases where response size is becoming an issue. the
+            `essentialRequirementsMet` key (present in DOS2 BRs onwards)is pragmatically included anyway because it
+            is referenced in some important listing views.
+        """
         data = {k: v for k, v in self.data.items() if with_data or k == "essentialRequirementsMet"}
         parent_brief = self.brief.serialize()
         parent_brief_fields = ['id', 'title', 'status', 'applicationsClosedAt', 'framework']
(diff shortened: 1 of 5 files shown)


tests/main/views/test_brief_response.py
```diff
@@ -688,16 +688,19 @@ def test_list_empty_brief_responses(self):
         assert data['briefResponses'] == []
         assert 'self' in data['links'], data
 
-    def test_list_brief_responses(self):
+    @pytest.mark.parametrize("with_data_false", (False, True))
```
katstevens: This var name is hurting my brain a bit with the prospect of `with_data_false` equalling `True` - how about `data_not_included`?

app/models/main.py
```diff
@@ -1847,8 +1847,8 @@ def validate(self, enforce_required=True, required_fields=None, max_day_rate=Non
         if errs:
             raise ValidationError(errs)
 
-    def serialize(self):
-        data = self.data.copy()
+    def serialize(self, with_data: bool = True):
+        data = {k: v for k, v in self.data.items() if with_data or k == "essentialRequirementsMet"}
```
katstevens: I think a comment here about when this field may or may not appear might be useful for future detectives, complementing the commit message saying why it's needed in the response. Something like `# 'essentialRequirementsMet' is present in brief responses from DOS2 onwards`?

risicle: Yes - well, `with_data` is a bit of a dumb argument name anyway, but it seemed like the most systematic name and I couldn't think of anything better. (After all I wasn't the one that started this whole thing of calling this and that important variable/field `data`)

katstevens: I think it's fine for the arg itself, just the parametrised var name in the test was confusing me (about what we were expecting in each case)

PR #974, opened 2019-09-26
Add supplierId to audit events in brief responses

jonodrew:
We found that it was a real struggle to find brief responses by supplier ID. This capability is useful when we receive inquiries from customers about when a response to a brief was submitted. I've added supplier ID in all three audit events created in the brief_responses view

Before merging this branch I'll squash these two commits together. [Ticket](https://trello.com/c/K7idSJht/320-add-supplierid-to-auditevents-for-newly-created-and-submitted-briefresponses)


katstevens:
@jonodrew this should be OK to merge now 👍

PR #976, opened 2019-10-01
Experimental unit test speedup using unlogged tables

risicle:
I was just having a play at speeding up the unit tests and came up with this only to find out it has negligible impact on my machine.

But before I bin it, I thought I'd throw it up here and ask people if it has any impact on _their_ machines, having different balances of processing power & IO speed. Please try (attempt a "fair test" please) & report back.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -33,6 +33,24 @@ def db_migration(request):
 
     with app.app_context():
         upgrade(config, 'head')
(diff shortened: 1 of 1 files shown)


katstevens:
For me `pytest` took 438 seconds on master, 464 seconds on this branch...

risicle:
Well ... that's pretty conclusive.

PR #979, opened 2019-10-14
list_briefs view: selectively compress response if draft & withdrawn briefs are filtered out

risicle:
https://trello.com/c/d1jpHXSp

Provided we modify the indexing script to actually *filter out* these briefs (which it doesn't need anyway), this should give us a far lower likelihood of incomplete responses during the indexing job.

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -154,10 +154,14 @@ def list_briefs():
             Lot.slug.in_(lot_slug.strip() for lot_slug in request.args["lot"].split(","))
         ))
 
+    response_headers = {"X-Compression-Safe": "0"}
     if request.args.get('status'):
-        briefs = briefs.has_statuses(
-            *(status.strip() for status in request.args['status'].split(','))
-        )
+        statuses = frozenset(status.strip() for status in request.args['status'].split(','))
+        briefs = briefs.has_statuses(*statuses)
+
+        if frozenset(("draft", "withdrawn",)).isdisjoint(statuses) and not with_users:
+            # then all data we're returning is fairly public anyway
+            response_headers["X-Compression-Safe"] = "1"
 
     status_filters = ['created', 'published', 'withdrawn', 'cancelled', 'unsuccessful', 'closed']
     temporal_filters = ['on', 'before', 'after']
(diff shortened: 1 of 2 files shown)


app/main/views/briefs.py
```diff
@@ -154,10 +154,13 @@ def list_briefs():
             Lot.slug.in_(lot_slug.strip() for lot_slug in request.args["lot"].split(","))
         ))
 
+    response_headers = {"X-Compression-Safe": "0"}
     if request.args.get('status'):
-        briefs = briefs.has_statuses(
-            *(status.strip() for status in request.args['status'].split(','))
-        )
+        statuses = frozenset(status.strip() for status in request.args['status'].split(','))
+        briefs = briefs.has_statuses(*statuses)
```
benvand: What are we inferring from this condition? The likelihood of it being an indexing job?

If so I think it would be safer to have a parameter on the view which would allow the client to decide whether they expect the response compressed.

`allow_compression` then add the compression header regardless of the request 🤷‍♂ In fact; that would be a good thing for all our api views to have.

risicle: > What are we inferring from this condition? The likelihood of it being an indexing job?

No, we're actually deciding based on what we think is safe to output. Drafts aren't safe because they're not meant to be public. Withdrawn briefs aren't public - they were withdrawn for a reason. User data isn't generally public. So if the endpoint can be sure that none of this data is going to be output it'll allow it to be compressed.

The client can always opt out by not including `Accept-Encoding gzip`.

app/main/views/briefs.py
```diff
@@ -154,10 +154,13 @@ def list_briefs():
...
+
+        if frozenset(("draft", "withdrawn",)).isdisjoint(statuses) and not with_users:
+            response_headers["X-Compression-Safe"] = "1"
```
katstevens: Will we need to also turn it off if the `with_users=True` param has been supplied?

risicle: > `and not with_users:`

Indeed.

app/main/views/briefs.py
risicle: However, the point stands that this could do with a comment or two.

app/main/views/briefs.py
katstevens: 🤦‍♀ Clearly not drunk enough coffee yet

app/main/views/briefs.py
benvand: 👍 Ah fairs, it's _extra_ security. Sounds good to me

app/main/views/briefs.py
benvand: 😂 😁

PR #986, opened 2019-10-31
Change the default port for the search api

lfdebrux:
2nd line ticket: https://trello.com/c/cwjzIU8U/1212-stop-search-api-from-using-port-5001

See the PR on the search api that adds `DM_SEARCH_API_PORT` for more details: https://github.com/alphagov/digitalmarketplace-search-api/pull/241

Note that this PR replaces the use of Flask-Script, our previous way of running the development server, but does not remove it. This can be done at a later date, but I wanted to see first if there were any workflows that are not possible with the new system. The major change from Flask-Script to Flask CLI is that instead of running `python application.py <command>` you run `flask <command>`. `flask shell` should work the same as `python application.py shell`, `flask routes` should work the same as `python application.py list_routes`. We can add the `list_external_routes` to the Flask CLI at a later date. [Flask-Migrate] commands such as `db show` will work the same as before.

[Flask-Migrate]: https://flask-migrate.readthedocs.io/en/latest/


config.py
```diff
@@ -91,7 +91,7 @@ class Development(Config):
     DM_API_AUTH_TOKENS = 'myToken'
     DM_API_CALLBACK_AUTH_TOKENS = 'myToken'
     DM_SEARCH_API_AUTH_TOKEN = 'myToken'
-    DM_SEARCH_API_URL = 'http://localhost:5001'
+    DM_SEARCH_API_URL = f"http://localhost:{os.getenv('DM_SEARCH_API_PORT', 5001)}"
```
katstevens: I think we need to pick a default port and stick with it across the repos/manual etc - 5001 or 5009.

lfdebrux: I think we should change the default to 5009, although not necessarily using this method (depending on what the team thinks).

@benvand has already made a PR for the manual https://github.com/alphagov/digitalmarketplace-manual/pull/273

katstevens: It's more about removing 5001 across our repos to avoid confusion (and it matching what's in the manual - good work @benvand!)

katstevens: I've spotted one other repo where we use port 5001 - the router's `make test-nginx` command. It's just passing in dummy env variables to the docker image before building it so not a blocker, but if we're removing 5001 everywhere that's another one for the list 😸

lfdebrux: I hadn't done all the repos thoroughly as was waiting for reviews, I will create a PR for the above when I get a moment though

PR #993, opened 2019-12-04
create_user view: add qualifyingBuyerEmailDomain to audit data

risicle:
https://trello.com/c/j6YDMslP

Required an amount of refactoring of the validation functions. Also set me off on a mission to make the `create_user` tests actually assert resulting database state following the request.

~Something I haven't done (but feel I probably should) is something equivalent for creation of admin accounts...~

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -29,6 +29,7 @@ from ...utils import (
 from ...validation import (
     admin_email_address_has_approved_domain,
     buyer_email_address_has_approved_domain,
+    buyer_email_address_first_approved_domain,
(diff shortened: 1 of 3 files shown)


risicle:
> Something I haven't done (but feel I probably should) is something equivalent for creation of admin accounts...

Hold on, having just implemented 90% of this I realize it's pointless, because admin domain restrictions are exact. So, the matching domain entry in the list is... the domain.

katstevens:
Think `flake8` is unhappy... Passes for me otherwise.

risicle:
"one last fix before I push" always gets me. Should be better now...

PR #999, opened 2019-12-17
192-cause-additionalchecks-to-always-be-run-regardless-of-ignore-dependencies-on-status-endpoint

benvand:
Update api to use new utils including causing run of additional_checks regardless of `include-dependencies` parameter

See:
https://trello.com/c/B2GfZgfO/192-cause-additionalchecks-to-always-be-run-regardless-of-ignore-dependencies-on-status-endpoint
https://github.com/alphagov/digitalmarketplace-utils/blob/master/CHANGELOG.md#5100

### DIFF
--- a/requirements-app.txt
+++ b/requirements-app.txt
@@ -1,7 +1,7 @@
 # Run `make freeze-requirements` to update requirements.txt
 # with package version changes made in requirements-app.txt
 
-git+https://github.com/alphagov/digitalmarketplace-utils.git@50.0.1#egg=digitalmarketplace-utils==50.0.1
+git+https://github.com/alphagov/digitalmarketplace-utils.git@51.0.0#egg=digitalmarketplace-utils==51.0.0
 git+https://github.com/alphagov/digitalmarketplace-apiclient.git@21.4.1#egg=digitalmarketplace-apiclient==21.4.1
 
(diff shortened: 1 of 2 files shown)


benvand:
The major version bumps stem from `jsonschema` which requires `importlib_metadata` 
(https://github.com/Julian/jsonschema/blob/master/setup.cfg#L32)
which requires `zipp` 
(https://gitlab.com/python-devs/importlib_metadata/blob/master/setup.cfg#L21)
which requires `more_itertools`
(https://github.com/jaraco/zipp/blob/master/setup.cfg#L25)

The breaking change for `importlib_metadata` is:
```
V1.0.0
* Project adopts semver for versioning.
* Removed compatibility shim introduced in 0.23.
* For better compatibility with the stdlib implementation and to avoid the same distributions being discovered by the stdlib and backport implementations, the backport now disables the stdlib DistributionFinder during initialization (import time). Closes #91 and closes #100.
```

The breaking change for `more-itertools` is:
https://pypi.org/project/more-itertools/
Dropping python 3.4 support (we're running 3.6)
the only functions used in our venv `unique_everseen`, `collapse` and `always_iterable` did not have signature changes. `unique_everseen` had an internals refactor to presumably make it quicker

PR #1000, opened 2020-01-08
DraftService: enforce lot's one_service_limit using database constraint

risicle:
~**I'm not planning on merging this PR directly, at merge time I will take the first two commits and submit them separately.** This is just here to allow people to review it conveniently.~ This _is_ now the second half., the first half having been merged as #1005 

https://trello.com/c/bqmbNLBu

Please read commit comments for more detailed information - I've tried to document it step by step.

In short, it's actually implemented largely in the abstract `ServiceTableMixin` and only _enabled_ when the concrete model contains a `lot_one_service_limit` field. For now, that's only `DraftService`.

It would be good to test the first two commits work fine for you on their own before testing the full branch...

### DIFF
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -1040,9 +1040,30 @@ class ServiceTableMixin(object):
 
     @declared_attr
     def __table_args__(cls):
-        return (db.ForeignKeyConstraint([cls.framework_id, cls.lot_id],
-                                        ['framework_lots.framework_id', 'framework_lots.lot_id']),
-                {})
+        return (
+            db.ForeignKeyConstraint(
+                (cls.framework_id, cls.lot_id,),
+                ("framework_lots.framework_id", "framework_lots.lot_id",),
+            ),
+        ) + (() if not hasattr(cls, "lot_one_service_limit") else (
+            # if the model has a field named lot_one_service_limit, we enforce the one service limit where appropriate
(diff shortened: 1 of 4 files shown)


app/models/main.py
```diff
@@ -1266,6 +1274,7 @@ def copy(self):
             supplier=self.supplier,
             data=data,
             status='not-submitted',
+            lot_one_service_limit=self.lot.one_service_limit,
```
katstevens: This covers both copying from a live Service, and copying from another DraftService, correct?

risicle: I think so, yes.

risicle: Well, it's `from_service` that does the former.

katstevens:
First two commits work fine for me locally (unit tests, draft-service related functional tests for G11 and DOS4, some manual testing) 👍

risicle:
Yippee thanks. Weirdly travis is failing but only for python 3.7. Not 3.6, not 3.8.

benvand:
Does this rely on the current duplication being sorted out?

risicle:
> Does this rely on the current duplication being sorted out?

YES.

It will be unhappy if applied to a non-compliant database.

katstevens:
We can use the `delete_draft_service` API method to address the inconsistency on the DraftService table.

PR #1001, opened 2020-01-10
Add ability to filter audit events by date range

lfdebrux:
I did a firebreak thing!

I thought it would be useful to be able to have the admin frontend show changes to a service over a certain time period; however I didn't want the frontend to be making a huge number of requests to the API, so I thought I would add the ability to filter audit events by a date range.

This PR adds a helper function `sql_compare_date_string` (I'm not wedded to the name) that parses a string such as `>=2001-04-13` or `1999-01-01..1999-12-31` and turns it into a comparison operation. I've used this to extend the `/audit-events` endpoint so the `audit-date` query string can be a date range.

The syntax for date ranges is based on that for GitHub's API: https://help.github.com/en/github/searching-for-information-on-github/understanding-the-search-syntax#query-for-dates

I'm haven't got a version of the service changes page for a certain date range that I'm happy with yet, but as firebreak is nearly over I don't think I will get that done so I thought I would see whether they think the API functionality would be useful.

### DIFF
--- a/app/main/views/audits.py
+++ b/app/main/views/audits.py
@@ -1,4 +1,4 @@
-from datetime import datetime, timedelta
+from datetime import datetime
 
 from flask import jsonify, abort, request, current_app
 from sqlalchemy.exc import IntegrityError
@@ -6,18 +6,18 @@ from sqlalchemy.sql.expression import true, false, func
 from sqlalchemy.orm import class_mapper
 from dmapiclient.audit import AuditTypes
 from dmutils.config import convert_to_boolean
-from dmutils.formats import DATE_FORMAT
 
 from .. import main
 from ... import db, models
 from ...models import AuditEvent
-from ...validation import is_valid_date, is_valid_acknowledged_state
+from ...validation import is_valid_acknowledged_state
 from ...utils import (
     get_json_from_request,
     get_valid_page_or_1,
     json_has_required_keys,
     paginated_result_response,
     single_result_response,
+    compare_sql_datetime_with_string,
     validate_and_return_updater_request,
 )
 
@@ -64,13 +64,12 @@ def list_audits():
 
     audit_date = request.args.get('audit-date', None)
     if audit_date:
-        if is_valid_date(audit_date):
-            audit_datetime = datetime.strptime(audit_date, DATE_FORMAT)
-            audits = audits.filter(
-                AuditEvent.created_at.between(audit_datetime, audit_datetime + timedelta(days=1))
-            )
-        else:
+        try:
+            filter_test = compare_sql_datetime_with_string(AuditEvent.created_at, audit_date)
+        except ValueError:
             abort(400, 'invalid audit date supplied')
+
+        audits = audits.filter(filter_test)
 
     audit_type = request.args.get('audit-type')
     if audit_type:
--- a/app/utils.py
+++ b/app/utils.py
@@ -1,8 +1,11 @@
+import datetime
 import random
 
 from flask import url_for as base_url_for
 from flask import abort, current_app, request, jsonify
 from werkzeug.exceptions import BadRequest
+
+from dmutils.formats import DATE_FORMAT
 
 from .validation import validate_updater_json_or_400
 from . import search_api_client, dmapiclient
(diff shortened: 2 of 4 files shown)


risicle:
Was wondering for a second what the point of `sql_compare_date_string` was but I see it's so we can accept several different comparison types on the rest api under one parameter name. Pretty neat.

There's probably a more object orientated way of exposing it by subclassing the sqlalchemy datetime type to give this function as a `.stringcompare(...)` method or whatever you'd want to call it...

tests/main/views/test_audits.py
```diff
@@ -687,6 +709,31 @@ def test_should_not_get_audit_event_for_date_with_no_events(self):
...
+
+        response = self.client.get(
+            "/audit-events",
+            query_string={"audit-date": date_range_query},
+        )
+        assert response.status_code == 200
+        events = response.json["auditEvents"]
+        assert len(events) == number_of_events
+
     def test_should_reject_invalid_audit_dates(self):
```
katstevens: Can we add some parametrized cases to this test as well e.g. `"<=invalid"`? Either here or in the `test_it_should_raise_value_error_if_date_string_is_not_valid` test below

tests/test_utils.py
```diff
@@ -299,3 +301,63 @@ def check(data, data_required_keys, data_optional_keys, result):
...
+        assert self.Timestamp(2000, 1, 1) <= datetime.datetime(2000, 1, 1)
+        assert self.Timestamp(2000, 1, 1) > datetime.datetime(1999, 1, 1)
+        assert self.Timestamp(2000, 1, 1) < datetime.datetime(2001, 1, 1)
+        assert self.Timestamp(2000, 1, 1, 12) >= datetime.datetime(2000, 1, 1)
+        assert self.Timestamp(2000, 1, 1, 12) <= datetime.datetime(2000, 1, 2)
+        assert self.Timestamp(2000, 1, 1, 12).between(datetime.datetime(2000, 1, 1), datetime.datetime(2000, 1, 2))
+
+    def test_can_test_whether_timestamp_is_during_day(self):
+        timestamp = self.Timestamp(2010, 12, 2, 15, 34)
+        assert sql_compare_date_string(timestamp, "2010-12-02")
```
katstevens: I think the function name gets confusing when it's just a straight 'is it this date, yes/no'. How about `compare_sql_datetime_with_date_string`?

app/utils.py
```diff
@@ -231,3 +234,69 @@ def index_object(framework, doc_type, object_id, serialized_object, wait_for_res
...
+    ====================== ========================================
+     Query                  Description
+    ====================== ========================================
+    YYYY-MM-DD             Matches dates on day
+    >YYYY-MM-DD            Matches dates after day
+    >=YYYY-MM-DD           Matches dates on or after day
+    <YYYY-MM-DD            Matches dates before day
+    <=YYYY-MM-DD           Matches dates on or before day
+    YYYY-MM-DD..YYYY-MM-DD Matches dates between days (inclusively)
+    ====================== ========================================
```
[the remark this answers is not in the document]
lfdebrux: Ah! I did think about doing that actually, but I didn't want to have to change the parameters too much haha

app/main/views/audits.py
```diff
@@ -64,12 +64,11 @@ def list_audits():
 
     audit_date = request.args.get('audit-date', None)
     if audit_date:
-        if is_valid_date(audit_date):
-            audit_datetime = datetime.strptime(audit_date, DATE_FORMAT)
+        try:
             audits = audits.filter(
-                AuditEvent.created_at.between(audit_datetime, audit_datetime + timedelta(days=1))
+                sql_compare_date_string(AuditEvent.created_at, audit_date)
```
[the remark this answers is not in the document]
lfdebrux: Yeah that makes sense

tests/test_utils.py
lfdebrux: Yes, the name is definitely the thing I spent the longest on...

I like your suggestion, what about (along the same lines) `compare_sql_datetime_with_string`?

lfdebrux:
> Was wondering for a second what the point of `sql_compare_date_string` was but I see it's so we can accept several different comparison types on the rest api under one parameter name. Pretty neat.

Yes, I basically thought the idea of having `audit-date-from`, `audit-date-before`, etc. would be tedious to write

> There's probably a more object orientated way of exposing it by subclassing the sqlalchemy datetime type to give this function as a `.stringcompare(...)` method or whatever you'd want to call it...

That's an interesting idea, not something I had considered. I think having it as a function makes it a little easier to test though; I found it hard to work out how to isolate the proper part of the SQLAlchemy class hierarchy for type hints and mocking, which is why I gave up and made a mock out of a datetime in the tests.

app/main/views/audits.py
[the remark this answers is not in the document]
lfdebrux: Done

tests/main/views/test_audits.py
lfdebrux: Done

tests/test_utils.py
katstevens: Yep - let's go with that 👍

lfdebrux: Done

PR #1003, opened 2020-01-10
Add test of email addresses to `/users` endpoint

lfdebrux:
We had an issue where a request to `/users?email_address=...` with a NUL character in the query string caused a number of HTTP `500`s. See the [2nd line ticket](https://trello.com/c/bcchmats/1353-bad-user-input-on-login-form-causing-500s) for details.

@katstevens convinced me that an invalid email address should result in a `400`, so I've added a validation helper function and a test to the endpoint.

Note that the pattern used to check whether the email address is valid is NOT the same as that used in our JSON schemas for authentication; this because the regex used in our JSON schemas will still let NUL characters through (and thus cause a 500). I'm not sure what the implications of changing our schema are, so I haven't done it yet.

### DIFF
--- a/app/validation.py
+++ b/app/validation.py
@@ -405,3 +405,11 @@ def admin_email_address_has_approved_domain(email_address):
     :return: boolean
     """
     return email_address.split('@')[-1] in current_app.config.get('DM_ALLOWED_ADMIN_DOMAINS', [])
+
+
+def is_valid_email_address(email_address: str) -> bool:
+    "Check the email address is valid"
+    # regex from Mozilla Developer Network
+    # https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email#Validation
+    pattern = r"^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$"  # noqa: E501
+    return bool(re.fullmatch(pattern, email_address))
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -30,6 +30,7 @@ from ...validation import (
     admin_email_address_has_approved_domain,
     buyer_email_address_has_approved_domain,
     buyer_email_address_first_approved_domain,
+    is_valid_email_address,
     validate_user_auth_json_or_400,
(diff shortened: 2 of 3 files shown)


app/validation.py
```diff
@@ -405,3 +405,11 @@ def admin_email_address_has_approved_domain(email_address):
     :return: boolean
     """
     return email_address.split('@')[-1] in current_app.config.get('DM_ALLOWED_ADMIN_DOMAINS', [])
+
+
+def is_valid_email_address(email_address: str) -> bool:
+    "Check the email address is valid"
+    # regex from Mozilla Developer Network
+    # https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email#Validation
+    pattern = r"^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$"  # noqa: E501
```
katstevens: What do we think is the risk of an existing email address in our DB (perhaps entered by mistake) not matching this regex? (I would say low, and the likelihood of the user retyping the mistake when e.g. resetting a pw probably lower...?)

lfdebrux: Presumably any that didn't match wouldn't be a valid email address and therefore would be an inactive account?

risicle: Unless something in our email sending pipeline silently stripped nulls - I think it's worth checking before rollout...

risicle: It should be easy to check using `api-clients-shell.py` on a connection to production...

benvand: As long as this is more strict than:
https://trello.com/c/3plghQzC/427-our-wtforms-email-validation-is-less-strict-than-notifys
https://github.com/alphagov/notifications-utils/blob/67889886ec1476136d12e7f32787a7dbd0574cc2/notifications_utils/recipients.py#L439

All good 👍

lfdebrux: Currently running a test against production

```
import re
email_address_pattern = ...
invalid_email_addresses = [print(user["id"]) or user for user in data.find_users_iter() if not user["personalDataRemoved"] and not re.fullmatch(email_address_pattern, user["emailAddress"])]
```

lfdebrux: Looks like there aren't any active users whose email address don't fit the pattern

risicle: :100:

benvand: Nice one 👍

PR #1004, opened 2020-01-16
Allow update_service view to move service to different supplier

katstevens:
https://trello.com/c/mLU7vty9/1343-inconsistency-with-supplier-on-the-dos4-framework-amaze-realize

Depends on https://github.com/alphagov/digitalmarketplace-apiclient/pull/228.

We currently have no auditable way of moving a service from one supplier to another. This will be useful for any future novations as well as addressing an inconsistency on one of our supplier's services.

The `update_service` view uses a new `AuditType` to allow us to link the old and new supplier IDs on the same audit event. There's a bit of stickiness around fetching the supplier_id from the new object before we do the commit, any suggestions for improvements welcome. 

The new audit event type will also mean these changes shouldn't appear on the service edits view for Category Admins, but I haven't tested that yet...

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -159,8 +159,10 @@ def update_service(service_id):
             abort(400, "Invalid value for 'copiedToFollowingFramework' supplied")
         service.copied_to_following_framework = update['copiedToFollowingFramework']
 
-    if 'supplierId' in update:
+    if 'supplierId' in update and update['supplierId'] != service.supplier_id:
         audit_type = AuditTypes.update_service_supplier
+        if len(update.keys()) > 1:
+            abort(400, "Cannot update supplierID and other fields at the same time")
         # Discard any other updates
         update = {
             'supplierId': update['supplierId']
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -159,13 +159,17 @@ def update_service(service_id):
             abort(400, "Invalid value for 'copiedToFollowingFramework' supplied")
         service.copied_to_following_framework = update['copiedToFollowingFramework']
 
-    updated_service = update_and_validate_service(service, update)
     if 'supplierId' in update:
         audit_type = AuditTypes.update_service_supplier
+        # Discard any other updates
+        update = {
+            'supplierId': update['supplierId']
+        }
     else:
         audit_type = (
             AuditTypes.update_service_admin if request.args.get('user-role') == 'admin' else AuditTypes.update_service
         )
+    updated_service = update_and_validate_service(service, update)
 
     commit_and_archive_service(updated_service, update_details, audit_type)
     index_service(updated_service, wait_for_response=convert_to_boolean(request.args.get("wait-for-index", "true")))
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -160,9 +160,12 @@ def update_service(service_id):
         service.copied_to_following_framework = update['copiedToFollowingFramework']
 
     updated_service = update_and_validate_service(service, update)
-    audit_type = (
-        AuditTypes.update_service_admin if request.args.get('user-role') == 'admin' else AuditTypes.update_service
-    )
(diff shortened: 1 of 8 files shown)


app/main/views/services.py
```diff
@@ -160,9 +160,12 @@ def update_service(service_id):
...
     updated_service = update_and_validate_service(service, update)
-    audit_type = (
-        AuditTypes.update_service_admin if request.args.get('user-role') == 'admin' else AuditTypes.update_service
-    )
+    if 'supplierId' in update:
+        audit_type = AuditTypes.update_service_supplier
+    else:
+        audit_type = (
+            AuditTypes.update_service_admin if request.args.get('user-role') == 'admin' else AuditTypes.update_service
+        )
```
risicle: Is the user allowed to both update the contents of the service and change the supplier at the same time?

We're starting to pack an awful lot of information into this "audit type" field.

katstevens: Hmm good point.

risicle: I don't have an _answer_ though of course.

risicle:
> There's a bit of stickiness around fetching the supplier_id from the new object before we do the commit, any suggestions for improvements welcome.

Am I missing something? All looks fairly straightforward to me.

katstevens:
> > There's a bit of stickiness around fetching the supplier_id from the new object before we do the commit, any suggestions for improvements welcome.
> 
> Am I missing something? All looks fairly straightforward to me.

Using `'supplierId': updated_service.supplier_id` and `'supplierId': updated_service.supplier.supplier_id` for the different types of audit events felt a bit icky...

risicle:
Yeeeeah I wouldn't say massively so. You can always leave a little comment to stop someone coming along and "tidying it up".

risicle:
> Ignore other update fields if supplierId given

Cool. Should it silently ignore them or throw an error? Or should we just leave it possible to update and I'm making an unnecessary fuss?

katstevens:
@risicle I've made it strict to disallow updates for both `supplierId` and anything else. Also added basic unit tests for the validator function where this happens.

app/main/views/services.py
```diff
@@ -159,8 +159,10 @@ def update_service(service_id):
             abort(400, "Invalid value for 'copiedToFollowingFramework' supplied")
         service.copied_to_following_framework = update['copiedToFollowingFramework']
 
-    if 'supplierId' in update:
+    if 'supplierId' in update and update['supplierId'] != service.supplier_id:
```
risicle: Does the endpoint allow you to assign a string and it'll automagically convert it to an int? If so you'd need to wrap the left hand side in `int()` too to be sure.

tests/main/views/test_services.py
```diff
@@ -754,6 +755,27 @@ def test_can_change_the_supplier_id_for_a_service(self, index_service):
...
+        data = json.loads(response.get_data())
+        # No updates made
+        assert data['services']['supplierId'] == 1
+        assert 'foo' not in data['services']
+
+        # No audit events created
+        audit_response = self.client.get('/audit-events')
+        data = json.loads(audit_response.get_data())
+        assert len(data['auditEvents']) == 0
+
```
risicle: This matches the rest of the tests around here so it's fine, but I think we're trying to move more towards direct db checking for this kind of thing else we're testing 3 different endpoints in each of these tests which adds up to be slow if nothing else.

risicle:
:+1:

app/main/views/services.py
katstevens: Good shout.

tests/main/views/test_services.py
katstevens: Yeah, agree it's not ideal. I didn't want to go too far down a refactoring hole on this ticket though.

risicle: Yeah no please don't.

PR #1005, opened 2020-01-17
DraftService: enforce lot's one_service_limit using database constraint (part 1)

risicle:
This is part 1 of the for-real version of #1000. Please see discussion there.

https://trello.com/c/bqmbNLBu


risicle:
Here we go...

PR #1014, opened 2020-03-12
Update SQLAlchemy from 1.2.18 to 1.3.15

lfdebrux:
Closes https://trello.com/c/Z9YDGp8w/1527-sql-injection-cves

One breaking change in SQLAlchemy 1.3 is that too long identifiers cause an exception (where previously they were passed on to PostgresSQL to be silently truncated) [1]. We don't have any too long identifiers in our model but we did in our migrations, which caused tests to error out; this PR includes a tweak to that migration.

Another issue is the increased use of deprecation warnings by SQLAlchemy (to the point where Travis breaks from too many log messages). Most of the deprecated uses seem to be from Flask-SQLAlchemy; to fix this we have to upgrade Flask-SQLAlchemy as well.

#### SQLAlchemy from 1.2.18 to 1.3.15

https://docs.sqlalchemy.org/en/13/changelog/changelog_13.html

#### Flask-SQLAlchemy from 2.1 to 2.4.1

https://flask-sqlalchemy.palletsprojects.com/en/2.x/changelog/#version-2-4-1

[1]: https://docs.sqlalchemy.org/en/13/changelog/migration_13.html#new-multi-column-naming-convention-tokens-long-name-truncation

### DIFF
--- a/migrations/versions/1210_add_outcome_model.py
+++ b/migrations/versions/1210_add_outcome_model.py
@@ -18,7 +18,7 @@ down_revision = '1200'
 def upgrade():
     op.create_unique_constraint('uq_brief_responses_id_brief_id', 'brief_responses', ['id', 'brief_id'])
     op.create_unique_constraint(
-        'uq_direct_award_search_result_entries_archived_service_id_search_id',
+        'uq_direct_award_search_result_entries_archived_service_id_searc',
         'direct_award_search_result_entries',
         ['archived_service_id', 'search_id'],
     )
@@ -119,7 +119,7 @@ def downgrade():
     op.drop_table('outcomes')
     op.drop_constraint('uq_direct_award_searches_id_project_id', 'direct_award_searches', type_='unique')
     op.drop_constraint(
-        'uq_direct_award_search_result_entries_archived_service_id_search_id',
+        'uq_direct_award_search_result_entries_archived_service_id_searc',
         'direct_award_search_result_entries',
         type_='unique',
     )
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -45,7 +45,7 @@ requests-mock==1.7.0      # via -r requirements-dev.in
 requests==2.23.0          # via -c requirements.txt, coveralls, requests-mock
 six==1.14.0               # via -c requirements.txt, freezegun, isodate, jsonschema, packaging, pip-tools, python-dateutil, requests-mock
 sortedcontainers==2.1.0   # via hypothesis
-sqlalchemy==1.2.18        # via -c requirements.txt, alchemyjsonschema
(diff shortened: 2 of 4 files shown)


agalamatis:
I think you are missing an "h" at the end of the index' name in the file: "migrations/versions/1210_add_outcome_model.py"

lfdebrux:
> I think you are missing an "h" at the end of the index' name in the file: "migrations/versions/1210_add_outcome_model.py"

Good spot 😁 The truncation is intentional; see https://github.com/alphagov/digitalmarketplace-api/blob/master/app/models/direct_award.py#L164-L167

lfdebrux:
🤦‍♂ 

line 36546: the job exceeded the maximum log length and has been terminated

lfdebrux:
I'm going to look at how much work it would be to update our tests to use Hypothesis's `@given` instead of `.example()` as mentioned in the warning and in the [changelog](https://hypothesis.readthedocs.io/en/latest/changes.html#v4-37-0).

lfdebrux:
> I'm going to look at how much work it would be to update our tests to use Hypothesis's `@given` instead of `.example()` as mentioned in the warning and in the [changelog](https://hypothesis.readthedocs.io/en/latest/changes.html#v4-37-0).

Rewriting the tests that use hypothesis `.example()` looks like a pain, so I've turned it into a tech debt ticket instead [1].

I tried suppressing the errors from hypothesis to fix the Travis issue, but for some reason that meant that I then saw the errors from SQLAlchemy... 64129 of them.

```
tests/models/test_main.py::TestArchivedService::test_archived_service_serialize_matches_api_stub_response
tests/models/test_main.py::TestArchivedService::test_archived_service_serialize_matches_api_stub_response
tests/models/test_main.py::TestArchivedService::test_archived_service_serialize_matches_api_stub_response
tests/models/test_main.py::TestArchivedService::test_archived_service_serialize_matches_api_stub_response
tests/models/test_main.py::TestArchivedService::test_archived_service_serialize_matches_api_stub_response
tests/models/test_main.py::TestArchivedService::test_archived_service_serialize_matches_api_stub_response
tests/models/test_main.py::TestArchivedService::test_archived_service_serialize_matches_api_stub_response
  /Users/laurencedebruxelles/Code/digitalmarketplace/digitalmarketplace-api/venv/lib/python3.6/site-packages/flask_sqlalchemy/__init__.py:169: SADeprecationWarning: Use .persist_selectable
    info = getattr(mapper.mapped_table, 'info', {})

-- Docs: https://docs.pytest.org/en/latest/warnings.html
======================================== 1627 passed, 64129 warnings in 547.20s (0:09:07) ========================================
```

[1]: https://trello.com/c/22dAhqCx/1304-replace-uses-of-hypothesis-example-with-given

lfdebrux:
Okay tests are passing now!

PR #1030, opened 2020-05-11
List draft services by framework and status

katstevens:
https://trello.com/c/kCUjDnH4/48-3-3-jenkins-job-for-publish-services

One of our long-standing problems when managing draft services, is that we can currently only fetch them by supplier. 

This PR adds a new paginated endpoint to list draft services by framework, optionally filtering by status.

This would allow us to:
- easily see what a draft service response looks like, without needing a supplier ID (I've been annoyed by this several times in the past)
- determine how many `not-submitted` / `submitted` services there are for each framework
- get an iterable list of drafts, ready for publishing en-masse (for example, if we have a list of failed supplier IDs, we can run through each draft and decide whether to publish it without needing an extra API call to check the supplier status)

I've also renamed the old `list_draft_services` endpoint to 1) make it clear that the view needs a supplier ID 2) differentiate it from the new `find_draft_services_by_framework` endpoint.

If this seems like a good idea to people, I'll create a corresponding API client method for the new endpoint.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -288,11 +288,16 @@ def find_draft_services_by_framework(framework_slug):
     framework = Framework.query.filter(
         Framework.slug == framework_slug
     ).first()
+    if not framework:
+        abort(404, "Framework '{}' not found".format(framework_slug))
 
     draft_service_query = DraftService.query.order_by(asc(DraftService.id))
     draft_service_query = draft_service_query.filter(DraftService.framework_id == framework.id)
 
     if status:
+        # Filter by 'submitted', 'not-submitted' only
+        if status not in ('submitted', 'not-submitted'):
+            abort(400, "Invalid argument: status must be 'submitted' or 'not-submitted'")
         draft_service_query = draft_service_query.filter(DraftService.status == status)
 
     pagination_params = request.args.to_dict()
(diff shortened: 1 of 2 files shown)


app/main/views/drafts.py
```diff
@@ -277,6 +279,40 @@ def list_draft_services():
     return list_result_response(RESOURCE_NAME, services), 200
 
 
+@main.route('/draft-services/framework/<string:framework_slug>', methods=['GET'])
```
daniele-occhipinti, line 282: I feel a more RESTful approach would be to have this new resource attached to /frameworks/<string:framework_slug>/draft-services
But, even if true, it would not be very practical?

katstevens: Ideally we'd like a single paginated `/draft-services` endpoint, with multiple filter options such as `supplier_id`, `framework`, `status` etc. However this would introduce a breaking change to the current non-paginated endpoint, which is definitely to be avoided. 

However... if I added `supplier_id` as an optional filter to this endpoint, we could move towards deprecating the other endpoint?

daniele-occhipinti: That may be a good idea. However, if the deprecation would take much effort (as you need to change invocations from the frontend apps) it should probably be a tech debt ticket (or at least a separate PR). That's a call of judgement.

katstevens: Commit added!

PR #1036, opened 2020-06-03
Add Python version file for Dependabot

lfdebrux:
https://trello.com/c/dwYU1Hi3/1681-stop-dependabot-removing-importlib-metadata

We have been having issues where Dependabot has been removing
importlib-metadata and dependencies from our requirements files, because
jsonschema only requires it for python_version<3.8[[1]].

Our apps still use Python 3.6, so we want to run Dependabot on this
version of Python, to get requirements for that version of Python. It
looks like the way to do this is to create a .python-version file in the
repo[[2]].

[1]: Julian/jsonschema@0e6c684
[2]: dependabot/dependabot-core#1455


gidsg:
looks good to me in theory @lfdebrux can you take a look at why the build failed?

lfdebrux:
Not sure why it failed, couldn't reproduce locally, so re-ran and it passed.

PR #1037, opened 2020-06-09
add POST method to send email address in JSON body

gidsg:
part of https://trello.com/c/CUik4BTJ/103-3-we-store-email-addresses-of-users-who-try-to-create-a-buyer-account-in-our-logs-in-plaintext

Adds POST method in addition to GET method for backwards compatibility. Until client library and all clients can be updated to use new method.

### DIFF
--- a/tests/main/views/test_users.py
+++ b/tests/main/views/test_users.py
@@ -2107,23 +2107,54 @@ class TestUsersExport(BaseUserTest, FixtureMixin, PutDeclarationAndDetailsAndSer
 
 
 class TestUsersEmailCheck(BaseUserTest):
-    def test_valid_email_is_ok_if_domain_found_in_database(self):
+    def _setup(self):
         # Create a domain that isn't in the buyer-email-domains.txt file
         db.session.add(BuyerEmailDomain(domain_name="bananas.org"))
         db.session.commit()
 
-        response = self.client.get('/users/check-buyer-email', query_string={'email_address': 'buyer@bananas.org'})
-        assert response.status_code == 200
-        assert json.loads(response.get_data())['valid'] is True
+    # Deprecated
+    def test_get_valid_email_is_ok_if_domain_found_in_database(self):
+        self._setup()
+        get_response = self.client.get('/users/check-buyer-email', query_string={'email_address': 'buyer@bananas.org'})
+        assert get_response.status_code == 200
+        assert json.loads(get_response.get_data())['valid'] is True
 
-    def test_invalid_email_is_not_ok(self):
-        response = self.client.get('/users/check-buyer-email', query_string={'email_address': 'someone@notgov.uk'})
-        assert response.status_code == 200
-        assert json.loads(response.get_data())['valid'] is False
+    def test_post_valid_email_is_ok_if_domain_found_in_database(self):
+        self._setup()
+        post_response = self.client.post('/users/check-buyer-email',
+                                         data=json.dumps({'emailAddress': 'buyer@bananas.org'}),
+                                         content_type='application/json')
+        assert post_response.status_code == 200
+        assert json.loads(post_response.get_data())['valid'] is True
 
-    def test_email_address_is_required(self):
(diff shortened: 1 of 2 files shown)


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally
- [ ] run functional tests locally

---

Looks good to me, I haven't tested locally yet.

I know you said you didn't want to make the PR too big, but I think it would be good to also change the `/users/valid-admin-domain` route in this PR, as it is in the same file and all.

gidsg:
> it would be good to also change the /users/valid-admin-domain route in this PR

@lfdebrux OK I have done this now in https://github.com/alphagov/digitalmarketplace-api/pull/1037/commits/8dca2675b324cf6f77074d7fead2f81ebb366b84

lfdebrux:
I've tested these new endpoints locally using HTTPie and they work for me

gidsg:
@lfdebrux force pushed as some unwanted refactoring snuck into the last commit.

tests/main/views/test_users.py
```diff
@@ -2107,23 +2107,54 @@ def test_published_service_count_with_different_statuses(self):
 
 
 class TestUsersEmailCheck(BaseUserTest):
-    def test_valid_email_is_ok_if_domain_found_in_database(self):
+    def _setup(self):
```
katstevens, line 2110: A small thing but can we either give this method a more descriptive name (e.g. `def _setup_buyer_domain`) or make it a proper `def setup(self)` method that applies to all tests in the class?

gidsg: Good spot! Have done latter suggestion in https://github.com/alphagov/digitalmarketplace-api/pull/1037/commits/15603b993ea2aff5b90e5411fa767f0a8302f1c5

tests/main/views/test_users.py
```diff
@@ -2107,23 +2107,52 @@ def test_published_service_count_with_different_statuses(self):
...
+        assert json.loads(post_response.get_data())['valid'] is False
+
+    # deprecated
+    def test_get_email_address_is_required(self):
+        get_response = self.client.get('/users/check-buyer-email')
+        assert get_response.status_code == 400
+
+    def test_post_email_address_is_required(self):
+        post_response = self.client.post('/users/check-buyer-email')
+        assert post_response.status_code == 400
```
katstevens, line 2155: Ah - this test needs the `content-type` header in the request, otherwise it's always going to 400.

It'd be great if we could also assert the error message as well here, as it will highlight the eventual change to the contract that we'll be introducing with the `json_has_required_keys` function (some of our frontend apps rely on the content of the error message returned by the API, so it's good to know what that'll be). 

With the content type included, for the test above we can assert the value of  `json.loads(post_response.get_data())['error']`.

tests/main/views/test_users.py
```diff
@@ -2132,26 +2161,58 @@ def setup(self):
...
-    def test_email_address_is_required(self):
+    # deprecated
+    def test_email_address_is_required_get(self):
         response = self.client.get('/users/valid-admin-email')
         assert response.status_code == 400
 
-    def test_invalid_email_is_not_ok(self):
+    def test_email_address_is_required_post(self):
+        response = self.client.post('/users/valid-admin-email')
+        assert response.status_code == 400
```
katstevens, line 2171: This test will also need a content type.

gidsg: @katstevens good spot. This is fixed in https://github.com/alphagov/digitalmarketplace-api/pull/1037/commits/6456d091ca3b62f4e53b6133ff0ee61c16066c3b.

tests/main/views/test_users.py
gidsg: Great suggestion @katstevens and this also flushed out issues with the code and the tests. Have updated now in https://github.com/alphagov/digitalmarketplace-api/pull/1037/commits/0bf90778add324e0048d674ab15cf179d5e2caa9.

PR #1038, opened 2020-06-11
Exclude questions when copying draft services

katstevens:
https://trello.com/c/WBjRZd14/80-2-frameworks-repo-copyservices-list-exclude-rather-than-include-fields

When copying draft services from a previous framework, we currently specify a list of questions to include. This list is hardcoded in the frameworks repo, and is proving difficult to manage between iterations (introducing a serious bug for G11 applications, and we nearly had the same issue for G12). To make this easier to manage, we'd like to specify a list of fields to exclude instead. These might be questions where the type has changed (from boolean to text, say), or it's been removed entirely. The vast majority of questions on a service are copiable, so the exclude lists should be shorter, and it should be easier to see what can be reintroduced in the following iteration.

However given G12 applications are still open, we want to ensure the current 'include' system is still supported for now. 

There are two service copying endpoints: one for copying a single service, one for bulk-copying all services on a particular lot. I've split the changes for these into separate commits, but they're pretty much doing the same thing. (Saying that, the test setup for the bulk-copying is a bit odd - I haven't tried to refactor that too much in this PR though, due to time constraints). 

Once this PR is merged, we'll need:
- changes to API client
- changes to supplier FE 
- example change in frameworks repo for DOS5/G13?

### DIFF
--- a/app/draft_utils.py
+++ b/app/draft_utils.py
@@ -8,3 +8,24 @@ def validate_and_return_draft_request(draft_id=0):
     if draft_id:
         json_has_matching_id(json_payload['services'], draft_id)
     return json_payload['services']
+
+
+def get_copiable_service_data(service, questions_to_exclude=None, questions_to_copy=None):
+    """
+    Filter out any service data that shouldn't be copied to a new draft service, by either including
+    copyable questions or excluding non-copyable questions.
+    There is validation at view-level to prevent both `to_copy` and `to_exclude` lists being supplied, so
+    this function should only receive one or the other. However we want to deprecate use of `to_copy`, so
+    if both are somehow supplied, we use the `to_exclude` list.
+    If neither list is provided, then all data fields on the service are included.
+
+    :param service: service object with a JSON dict `data` attribute (required)
+    :param questions_to_exclude: iterable of question IDs that must not be copied
+    :param questions_to_copy: iterable of question IDs that may be copied (to be deprecated)
+    :return: JSON dict of service data
+    """
+    if questions_to_exclude:
+        return {key: value for key, value in service.data.items() if key not in questions_to_exclude}
+    if questions_to_copy:
+        return {key: value for key, value in service.data.items() if key in questions_to_copy}
+    return service.data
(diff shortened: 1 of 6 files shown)


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally
- [ ] run functional tests locally

tests/main/views/test_drafts.py
```diff
@@ -362,10 +407,13 @@ def post_to_copy_published_from_framework(
...
             pass
-        elif override_questions_to_copy:
-            data.update(questionsToCopy=override_questions_to_copy)
+        elif override_questions_to_copy or override_questions_to_exclude:
+            if override_questions_to_copy:
+                data.update(questionsToCopy=override_questions_to_copy)
+            if override_questions_to_exclude:
+                data.update(questionsToExclude=override_questions_to_exclude)
         else:
             data.update(self.basic_questions_json)
```
lfdebrux, line 418: I found this logic quite hard to untangle.

I think what is happening is that we include `questionsToCopy` (from `self.basic_questions.json`) in the request unless either `override_questions_to_copy` or `override_questions_to_exclude` is specified.

Is that what we want?

lfdebrux:
Copying services from G-Cloud 11 to G-Cloud 12 seems to work still

tests/main/views/test_drafts.py
katstevens: For now, yes. This test setup a bit of a mess really. I considered refactoring, but didn't want to yank the bit of string and unravel the whole jumper 😹

katstevens:
FYI I'm currently writing some functional tests for the current service copying function, which we have zero coverage for at the moment.

app/draft_utils.py
```diff
@@ -8,3 +8,24 @@ def validate_and_return_draft_request(draft_id=0):
     if draft_id:
         json_has_matching_id(json_payload['services'], draft_id)
     return json_payload['services']
+
+
+def get_copiable_service_data(service, questions_to_exclude=None, questions_to_copy=None):
```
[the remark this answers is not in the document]
katstevens: Good shout - will commit

katstevens:
Yikes - something is very unhappy after I rebased against master...

PR #1040, opened 2020-06-15
remove deprecated GET endpoints

gidsg:
part of https://trello.com/c/CUik4BTJ/103-3-we-store-email-addresses-of-users-who-try-to-create-a-buyer-account-in-our-logs-in-plaintext

Please see previous PRs for context:
https://github.com/alphagov/digitalmarketplace-apiclient/pull/231
https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/655
https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/322

remove deprecated GET endpoints now all clients are using POST endpoints ﻿

(I have confirmed in Kibana the GET endpoints are getting no traffic)

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -430,17 +430,6 @@ def export_users_for_framework(framework_slug):
     return jsonify(users=user_rows), 200
 
 
-# deprecated
-@main.route("/users/check-buyer-email", methods=["GET"])
-def email_has_valid_buyer_domain():
-    email_address = request.args.get('email_address')
-    if not email_address:
-        abort(400, "'email_address' is a required parameter")
-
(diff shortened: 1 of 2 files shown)


app/main/views/users.py
```diff
@@ -450,17 +439,6 @@ def email_has_valid_buyer_domain_post():
     return jsonify(valid=domain_ok), 200
 
 
-# deprecated
-@main.route("/users/valid-admin-email", methods=["GET"])
-def email_is_valid_for_admin_user():
```
agalamatis, line 455: This endpoint is used still in the [API client](https://github.com/alphagov/digitalmarketplace-apiclient/blob/master/dmapiclient/data.py#L541), which is used in two tests in admin FE (test_form_validators.py and test_admin_manager.py). I think we need to remove the tests first, then the client and finally this endpoint.

PR #1041, opened 2020-06-15
document the test-bootstrap command

gidsg:
document previously undocumented command which I found useful for troubleshooting.﻿

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,13 @@ To run the `flake8` linter:
 make test-flake8
 ```
 
+To re-create the test database (this is a useful troubleshooting step if you are having issues with the test database):
+
+```
+make test-bootstrap
+```
+
+
 ### Updating Python dependencies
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,6 +65,13 @@ To run the `flake8` linter:
 make test-flake8
 ```
 
+To re-create the test database (this is a useful troubleshooting step if you are having issues with the test database):
+
+```
+make test-bootstrap
```
katstevens, line 71: I'd add a note clarifying that this command doesn't run migrations on the test db - that's handled in `tests/conftest.py`.

gidsg: @katstevens OK I clarified the text a bit in https://github.com/alphagov/digitalmarketplace-api/pull/1041/commits/79d46cf9f780474fe4ba49e959bece2b1e6d5306. Does that look OK?

PR #1045, opened 2020-07-02
Remove hypothesis from API tests

katstevens:
https://trello.com/c/22dAhqCx/1304-2-spike-1d-use-hypothesis-given-instead-of-example-in-tests-in-digitalmarketplace-api

I did a short investigation (1 day, as per spike) into whether it was worth switching to `@given` instead of `example()`. I couldn't get `@given` to work without some major upheaval, and I don't think the benefit of having the semi-random data outweighs the complexity it brings. So I thought I'd just get rid of it, and use static test data containing the usual problematic encoding/characters etc.

Removing the `example_listings` module also reduces confusion with the `tests.helpers.load_example_listing` function, which gets data from JSON files (and has nothing to do with `hypothesis`).


tests/main/views/test_brief_response.py
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
```
gidsg, line 1: do we need this line https://stackoverflow.com/a/58652741/238230
seems to suggest you don't need it in Python 3?

katstevens: You're right, we don't need it for our Python 3 apps (my IDE is still set up to complain about Python 2 violations). Will remove.

lfdebrux:
🙌

PR #1053, opened 2020-09-07
Allow countersignature update without approval for G12 onwards

katstevens:
https://trello.com/c/UfmJ2EOd/308-adapt-countersigning-jenkins-job-for-e-signature-flow

In the pre-esignature world, CCS admins would approve agreements manually. This approval action would set the `countersigned_agreement_returned_at` date on the agreement object in the API. The countersignature details (including the link to the countersigned PDF) could not be updated without that date set.

Now the countersigning is to be automated, we won't have this date set in the same way. This PR removes the restriction for G12 onwards. I used a hardcoded datetime object, as our reference point so far has been the `frameworkLiveAtUTC` on the framework object - which is what we're checking here! Alternative suggestions welcome.

I also added a note on the approval endpoint, to clarify it's only used for earlier frameworks (this may still be required if a DOS4 supplier needs to change their agreement, so we can't delete it just yet!).

### DIFF
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -19,6 +19,8 @@ from ...supplier_utils import validate_agreement_details_data
 
 RESOURCE_NAME = "agreement"
 
+E_SIGNATURE_LIVE_DATE = datetime(2020, 9, 28)
+
 
 @main.route('/agreements', methods=['POST'])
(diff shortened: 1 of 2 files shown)


gidsg (approved):
LGTM 👍

PR #1057, opened 2020-09-25
Revert "Temporary change in logic for G12"

lfdebrux:
Reverts alphagov/digitalmarketplace-api#1054


agalamatis (approved):
thank you

bjgill (commented):
The PR this is reverting doesn't have much detail. Why did we do this originally, and why do we now need to revert?

lfdebrux:
> The PR this is reverting doesn't have much detail. Why did we do this originally, and why do we now need to revert?

Let's not go there.

PR #1059, opened 2020-10-07
Update lot singular and plural

gidsg:
https://trello.com/c/T70XvJ70/437-update-save-and-return-to-lab-summary

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ After editing `models.py` to add/edit/remove models for the database, you'll nee
 The easiest way to do this is to run
 
 ```
-flask db migrate --rev-id <revision_id> -m '<description'>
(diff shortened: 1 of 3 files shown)


lfdebrux:
I don't think utils needs to be updated for this PR, looking at the test you've changed the lots are derived from an actual database rather than from the stubs, and the stubs that are used are from [dmtestutils](https://github.com/alphagov/digitalmarketplace-test-utils/) rather than dmutils.

gidsg:
@lfdebrux you were right! I've forced pushed now without those other changes.

lfdebrux (approved):

lfdebrux:
Woops, sorry I accidentally clicked merge 😅

PR #1068, opened 2020-12-01
Distinguish G12 recovery suppliers

bjgill:
Trello: https://trello.com/c/KikQx3R2/593-as-a-g12-recovery-supplier-i-can-see-something-other-suppliers-cannot

We had an outage during G12 closing. Some suppliers will be involved in a recovery process to account for the outage causing them to fail to submit their services in time.

We have previously provided a list of the affected supplier IDs via the environment - see alphagov/digitalmarketplace-credentials#359 and alphagov/digitalmarketplace-aws#797. This PR allows suppliers who are part of the G12 recovery to create G12 draft services when G12 is not 'open'. Tests show that this is working correctly.

We don't expect real suppliers to use this API endpoint - we hope to have loaded all their drafts into the system already. However, we will need the ability to create draft services for testing with.

This also shows that we've wired everything up correctly.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -8,6 +8,7 @@ from datetime import datetime
 
 from .. import main
 from ... import db
+from ...supplier_utils import is_g12_recovery_supplier
 from ...validation import is_valid_service_id_or_400
 from ...models import Service, DraftService, Supplier, AuditEvent, Framework, Lot
 from ...utils import (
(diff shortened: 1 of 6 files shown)


agalamatis (approved):

bjgill:
Force pushed to fix flake 8 error. Will merge once Travis wakes up - or once we've switched to Github actions.

PR #1069, opened 2020-12-02
Add CI using GitHub Actions

bjgill:
Trello: https://trello.com/c/cnNEUjhz/1399-epic-migrate-travis-ci-usages-to-github-actions

Following Travis' recent changes, our Travis builds now regularly take much more than 30 minutes to even start. So switch across to GitHub actions, which is much faster. It's been approved by IA, and we aren't giving it any secrets in any case.

Modelled after https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1289


bjgill:
Ah. This is non-trivial. We're going to need to add postgres...

PR #1071, opened 2020-12-16
Switch to Github Action with Postgres service

gidsg:
closes #1069 

Trello: https://trello.com/c/cnNEUjhz/1399-epic-migrate-travis-ci-usages-to-github-actions

Adds Github Action including Postgres service modelled after https://docs.github.com/en/free-pro-team@latest/actions/guides/creating-postgresql-service-containers#running-jobs-directly-on-the-runner-machine

Builds on work by @bjgill in #1069


.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
...
+      matrix:
+        python-version: [ 3.6, 3.7, 3.8, 3.9 ]
+    services:
+      postgres:
+        image: postgres:9.5
+        env:
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: digitalmarketplace_test
+        # Set health checks to wait until postgres has started
+        options: >-
```
lfdebrux, line 17: I've not seen `>-` before, what does the dash do?

.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
...
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          # Maps tcp port 5432 on service container to the host
+          - 5432:5432
```
lfdebrux, line 24: We should make this a string, YAML sometimes treats numbers with colons weirdly... https://docs.docker.com/compose/compose-file/#ports

```suggestion
          - "5432:5432"
```

lfdebrux (approved):
Magic 🎉

.github/workflows/test.yml
gidsg: I _think_ this is about allowing the expression to continue on to the next line.

PR #1073, opened 2020-12-24
Run CI against postgres 10

gidsg:
Trello: https://trello.com/c/WdULXylT/2141-upgrade-postgresql

We need to upgrade to Postgres 12 by 11 February. Let's start by seeing what breaks if we start with Postgres 10.

We'll first upgrade our pre-merge CI. Once this is gone in, I'll test against Postgres 10 locally. Then we can think about upgrading Preview.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
       postgres:
         image: postgres:${{ matrix.postgres-version }}
         env:
-          POSTGRES_PASSWORD: postgres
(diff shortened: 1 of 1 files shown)


lfdebrux (approved):

lfdebrux:
Looks like we need to update the required checks in the branch protection rules.

bjgill:
Yes. I'll sort that out and then merge. I'll mark the two 3.6 tests as the required ones.

PR #1083, opened 2021-01-21
Allow DMp team to create G12 drafts

bjgill:
Trello: https://trello.com/c/13Ryt5o5

For the G12 recovery, suppliers will only be able to submit the drafts that are part of the G12 recovery. We won't be allowing them to create any extra. This is already disabled in supplier-frontend, so we can remove it here as well.

Instead, we need to be able to create arbitrary G12 drafts. This is because some of the G12 recovery services don't have corresponding drafts, so [we need to create them](https://github.com/alphagov/digitalmarketplace-scripts/pull/616). And we don't want to add the suppliers to the list of G12 recovery suppliers until the recovery process opens, because it makes user-visible changes to the website.

This is a terrible hack, and I feel bad about introducing it to the codebase. However, we will be deleting all the G12 recovery code in a month's time, so I think it's OK.

I've tested this manually locally, and it seems to work just fine.

### DIFF
--- a/app/main/views/drafts.py
+++ b/app/main/views/drafts.py
@@ -8,7 +8,6 @@ from datetime import datetime
 
 from .. import main
 from ... import db
-from ...supplier_utils import is_g12_recovery_supplier
 from ...validation import is_valid_service_id_or_400
 from ...models import Service, DraftService, Supplier, AuditEvent, Framework, Lot
 from ...utils import (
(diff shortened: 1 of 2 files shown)


lfdebrux (approved):
😆 it's by no means the worse thing I've seen

bjgill:
Ha. I'll merge this tomorrow, run the scripts, and then erase all traces of this nonsense.

PR #1089, opened 2021-02-17
Add API endpoint to un-award a brief

bjgill:
https://trello.com/c/rELvaDeA/2174-add-a-way-for-developers-to-award-a-brief-to-a-different-brief-response
https://govuk.zendesk.com/agent/tickets/4467615

We need the ability to un-award briefs. This will allow us to fix a brief that was accidentally awarded to the wrong supplier. This API method will not be exposed via any frontend app, so will be accessible to developers only.

Once a developer has un-awarded the brief using this new method, the buyer can then follow the usual process to award the brief to the correct supplier.

Update the validation rules on `awarded_at` accordingly.

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -360,6 +360,43 @@ def award_brief_details(brief_id, brief_response_id):
     return single_result_response(RESOURCE_NAME, brief), 200
 
 
+@main.route('/briefs/<int:brief_id>/award/<int:brief_response_id>/contract-details', methods=['DELETE'])
+def un_award_brief_details(brief_id, brief_response_id):
+    updater_json = validate_and_return_updater_request()
+
+    brief_response = BriefResponse.query.filter(
+        BriefResponse.brief_id == brief_id,
+        BriefResponse.id == brief_response_id
+    ).first_or_404()
+
+    if brief_response.status != 'awarded':
+        abort(400, "Can only un-award an awarded brief response")
+
+    # Set status back to pending-awarded
+    brief_response.awarded_at = None
+    brief_response.award_details = {'pending': True}
+
+    audit_event = AuditEvent(
+        audit_type=AuditTypes.update_brief_response,
+        user=updater_json['updated_by'],
+        data={
+            'briefId': brief_id,
+            'briefResponseId': brief_response_id,
+            'briefResponseAwardedValue': True,
+            'unAwardBriefDetails': True
+        },
+        db_object=brief_response
+    )
+
+    db.session.add_all([brief_response, audit_event])
+    db.session.commit()
+
+    brief = Brief.query.filter(
+        Brief.id == brief_id
+    ).first_or_404()
+    return single_result_response(RESOURCE_NAME, brief), 200
+
+
 @main.route('/briefs/<int:brief_id>/copy', methods=['POST'])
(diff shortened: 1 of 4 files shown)


app/main/views/briefs.py
```diff
@@ -360,6 +360,43 @@ def award_brief_details(brief_id, brief_response_id):
...
+        },
+        db_object=brief_response
+    )
+
+    db.session.add_all([brief_response, audit_event])
+    db.session.commit()
+
+    brief = Brief.query.filter(
+        Brief.id == brief_id
+    ).first_or_404()
```
lfdebrux, line 396: I'm a bit confused, shouldn't this check be made before updating the database?

app/main/views/briefs.py
```diff
@@ -360,6 +360,43 @@ def award_brief_details(brief_id, brief_response_id):
...
+    brief_response.awarded_at = None
+    brief_response.award_details = {'pending': True}
+
+    audit_event = AuditEvent(
+        audit_type=AuditTypes.update_brief_response,
+        user=updater_json['updated_by'],
+        data={
+            'briefId': brief_id,
+            'briefResponseId': brief_response_id,
+            'briefResponseAwardedValue': True,
```
lfdebrux, line 385: I'm not sure what this piece of information is for, could you explain for me?

lfdebrux (changes requested):

app/main/views/briefs.py
gidsg: Hmm, I'm not sure about this either. I think the idea was that it was a similar action to this https://github.com/alphagov/digitalmarketplace-api/blob/acc24c649d316e37059c51e336c58978f2ca028f/app/main/views/briefs.py#L299 which sets the status to pending, but I agree it is confusing. Do you have any further thoughts @bjgill?

bjgill: Yes. I think we're just doing it to match the method that sets the brief response to `pending-awarded`, which is also what we're doing here.

But happy to remove `briefResponseAwardedValue` if you think it's likely to cause confusion?

lfdebrux: Hmm, I guess I'd just like to know what it means, if anything 😅

lfdebrux: Searching our code for the key name does not elucidate anything.

bjgill: I'm not sure it means anything :shrug:. It's used in `/briefs/<int:brief_id>/award` when a brief response is moved to `pending-awarded`, so we're just doing it for consistency.

lfdebrux: I guess it doesn't matter too much.

app/main/views/briefs.py
lfdebrux: My concern with this is, could we get into a situation where the database is updated but the caller gets a 404?

That would mean a brief response to a deleted brief? In which case something is broken, and it should probably be a 5xx?

I don't know maybe I'm over thinking it

bjgill: Yes - we're going to move it earlier.

gidsg: addressed in 7c5847c

lfdebrux (approved):
Tested locally and works for me 👍

PR #1094, opened 2021-02-23
Add endpoint to delete buyer email domains

bjgill:
https://govuk.zendesk.com/agent/tickets/4475776

The buyer email domain list acts as an allow-list controlling who can sign up as a buyer. Currently, admins are able to view and add domains to the allow list.

A domain was recently mistakenly added to the allow-list. So we need to create a new API endpoint to allow us to remove it from the list. This is the first time this has happened. So we're not going to do the work to expose this API method to admins - it'll be available to developers only.

Uses a [new audit event type](https://github.com/alphagov/digitalmarketplace-apiclient/pull/245).


lfdebrux (approved):
Works for me 👍

PR #1096, opened 2021-02-26
Replace make with invoke

lfdebrux:
Ticket: https://trello.com/c/hCCo28Bc/

We want to be able to reduce duplication of code across our repos; one of the places we have a lot of duplicated code is in our Makefiles; all of our repos have very similar Makefiles with small historical differences.

This commit replaces Make with invoke, using the tasks from alphagov/digitalmarketplace-developer-tools.


gidsg (approved):
👍 🏒

PR #1100, opened 2021-03-09
Add 'deleted' service status

alex9smith:
This status is used for when we want to remove a service page completely and show a 404, rather than the `disabled` status which leaves the page up with a warning message. The user-facing change to allow this behaviour was added in https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/1150 - this PR adds the new status as an allowed option in the API.

Alembic won't autodetect migrations like this (see [the docs](https://alembic.sqlalchemy.org/en/latest/autogenerate.html#what-does-autogenerate-detect-and-what-does-it-not-detect)) so I wrote the migration script myself. I have tested upgrades and downgrades locally and it appears to be working, but I'd encourage the reviewer to do the same.

Team manual update with documentation for the new status & when to use it is coming soon.

https://trello.com/c/qqL3sqWV/2182-remove-supplier-92526-from-g-cloud

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -325,7 +325,8 @@ def update_service_status(service_id, status):
     valid_statuses = [
         "published",
         "enabled",
-        "disabled"
+        "disabled",
+        "deleted",
     ]
 
     is_valid_service_id_or_400(service_id)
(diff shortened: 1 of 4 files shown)


gidsg (approved):
Have tested the Up and Down of the migration and that calling the method in the API client works as expected after the migration has run. 👍 🍡

PR #1101, opened 2021-03-10
Make e-signature determination stable

bjgill:
Previously, we decided whether a framework supported esignature or not based on the date it went live. This worked. However, it means that if you need to make a non-esignature framework live now for testing, it breaks because it erroneously thinks it should support esignature.

Instead hard-code a list of the framework slugs that do not support esignature. This should be safer because all future frameworks will support esignature.

### DIFF
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -19,8 +19,6 @@ from ...supplier_utils import validate_agreement_details_data
 
 RESOURCE_NAME = "agreement"
 
-E_SIGNATURE_LIVE_DATE = datetime(2020, 9, 28)
-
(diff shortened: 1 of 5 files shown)


gidsg (changes requested):
This should be updated to use the new logic: https://github.com/alphagov/digitalmarketplace-api/blob/fb8cdd80119d5233648af486620c540e34d84353/app/models/main.py#L233

bjgill:
Excellent point - and that inspired me to put all the logic in the model, where it looks more natural

gidsg (approved):
Nice one 📅 👍

PR #1102, opened 2021-03-10
Add missing manual signing framework

bjgill:
I forgot that the universe began with G6: https://technology.blog.gov.uk/2016/05/24/digital-marketplace-moving-from-portal-towards-a-platform/

Missed from https://github.com/alphagov/digitalmarketplace-api/pull/1101/, this should fix the current test failures.

### DIFF
--- a/app/models/main.py
+++ b/app/models/main.py
@@ -140,6 +140,7 @@ class Framework(db.Model):
         'digital-outcomes-and-specialists',
     )
     MANUAL_SIGNATURE_FRAMEWORK_SLUGS = [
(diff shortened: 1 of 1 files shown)


alex9smith:
Bad news - we have `g-cloud-5` as well

bjgill:
Oh no. That makes no sense. But that's going to cause problems. So much for my cunning plan - it turns out that trying to enumerate all old frameworks is hard.

This is causing more trouble than it's worth. I'll revert the whole thing :crying_cat_face:

PR #1105, opened 2021-03-15
E-signature boolean from denylist 

gidsg:
Reprises #1101 with a more comprehensive list of frameworks. I based the possible frameworks list on querying the `frameworks` database table in different environments.

This is useful as we have found when editing framework statuses for testing it can change the `frameworkLiveAtUtc` date causing unexpected behaviours.

I have run the functional tests locally against this branch, I did get some failures locally but they seemed unrelated e.g. Notify, so I _hope_ this should work on Preview.

### DIFF
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -19,8 +19,6 @@ from ...supplier_utils import validate_agreement_details_data
 
 RESOURCE_NAME = "agreement"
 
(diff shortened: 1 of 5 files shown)


alex9smith (approved):
I've also run the functional tests locally and only had the 'expected' failures, so I think this should work

PR #1107, opened 2021-03-18
Log all notify callbacks

lfdebrux:
This is important, write it down.

There shouldn't be any PII though.

### DIFF
--- a/app/callbacks/views/notify.py
+++ b/app/callbacks/views/notify.py
@@ -19,9 +19,25 @@ def callbacks_root():
 def notify_callback():
     notify_data = get_json_from_request()
 
-    if notify_data['status'] == 'permanent-failure':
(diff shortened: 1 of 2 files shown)


bjgill (approved):
:tada:

For reference, https://docs.notifications.service.gov.uk/rest-api.html#delivery-receipts is the documentation for callbacks. To is the only field that should contain PII, so this looks fine.

bjgill:
For future reference, this was triggered by https://govuk.zendesk.com/agent/tickets/4501738, where the lack of logging was a bit of a pain.

Also, is this possibly something worth recording in the team manual?

lfdebrux:
> Also, is this possibly something worth recording in the team manual?

Definitely. I will add something in tomorrow.

PR #1108, opened 2021-03-19
Change how Notify delivery receipts are logged

lfdebrux:
Putting the Notify delivery receipt object straight in `extra` isn't great, because the delivery receipt has a field `status` that our ELK logging stack doesn't like (Logit expects fields called `status` to be an integer number, i.e. a HTTP status code).

This commit nests the Notify delivery receipt in it's own key, so there shouldn't be any collisions now.

### DIFF
--- a/app/callbacks/views/notify.py
+++ b/app/callbacks/views/notify.py
@@ -32,7 +32,7 @@ def notify_callback():
 
     current_app.logger.info(
         f"Notify callback: {status}: {reference} to {hashed_email}",
-        extra=clean_notify_data,
+        extra={"notify_delivery_receipt": clean_notify_data},
     )
 
     if status == "permanent-failure":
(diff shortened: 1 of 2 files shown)


lfdebrux:
Here's what the JSON log message looks like for `/callbacks/notify` with this change:

```
{"name": "flask.app", "levelname": "INFO", "message": "Notify callback: test: test to n4bQgYhMfWWaL-qgxVrQFaO_TxsrC4Is0V1sFbDwCgg=",
15:15:35                      api | "pathname": "/Users/laurencedebruxelles/Code/digitalmarketplace/digitalmarketplace-api/app/callbacks/views/notify.py", "lineno": 35,
15:15:35                      api | "notify_delivery_receipt": {"status": "test", "reference": "test"}, "time": "2021-03-19 15:15:35,754", "requestId":
15:15:35                      api | "71e2fe9bb7c8104245a99fb875d38107", "spanId": null, "parentSpanId": null, "application": "api", "isSampled": "0", "debugFlag": "0",
15:15:35                      api | "logType": "application"}

```

bjgill (approved):

PR #1113, opened 2021-03-22
Add Docker to Dependabot configuration

lfdebrux:
We want to make sure we are using the latest version of our Docker base images, let's use Dependabot to help us with that.


gidsg (approved):
👍 🏬

PR #1128, opened 2021-04-30
change database connection URL to use postgresql:// protocol

gidsg:
https://trello.com/c/UIj9AODz/2212-upgrade-sqlalchemy-in-digitalmarketplace-api

The postgres:// protocol has been [deprecated in sqlalchemy since 0.6](https://github.com/sqlalchemy/sqlalchemy/blob/8fc5005dfe3eb66a46470ad8a8c7b95fc4d6bdca/lib/sqlalchemy/dialects/postgres.py): 
It is [removed in the 1.4 release](https://help.heroku.com/ZKNTJQSK/why-is-sqlalchemy-1-4-x-not-connecting-to-heroku-postgres) but we can change the protocol now.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -38,7 +38,8 @@ def create_app(config_name):
 
     if application.config['VCAP_SERVICES']:
         cf_services = json.loads(application.config['VCAP_SERVICES'])
-        application.config['SQLALCHEMY_DATABASE_URI'] = cf_services['postgres'][0]['credentials']['uri']
+        application.config['SQLALCHEMY_DATABASE_URI'] = (cf_services['postgres'][0]['credentials']['uri']
+                                                         .replace('postgres://', "postgresql://", 1))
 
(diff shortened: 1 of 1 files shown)


bjgill (approved):
A bit ugly, but understandable. Do you know if PaaS have any plans to change the form of the URI to match?

Is this going to affect running postgres locally?

gidsg:
@bjgill 

> Do you know if PaaS have any plans to change the form of the URI to match?

I don't believe so, as it presumably would break other clients which rely on the current URL. There is some internal discussion on this: https://gds.slack.com/archives/CADHV9267/p1619615769060100?thread_ts=1619615613.060000&cid=CADHV9267 

In any case, if they were to do so I believe the code in this PR would still work as the replace would no longer match.

> Is this going to affect running postgres locally?

No, this is all guarded behind `VCAP_SERVICES` env var which only exists on PaaS.

PR #1131, opened 2021-05-05
bump alchemyjsonschema

gidsg:
https://trello.com/c/UIj9AODz/2212-upgrade-sqlalchemy-in-digitalmarketplace-api

We need to update `alchemyjsonschema` to a version which supports sqlalchemy 1.4 as part of the work to upgrade to that version.

In this version it no longer reports one boolean field as 'required', this is due to a change in the definition of 'required' by the library, see 06c65a8 for more details.

### DIFF
--- a/tests/test_generate_model_schemas.py
+++ b/tests/test_generate_model_schemas.py
@@ -62,7 +62,6 @@ def test_generate_schema():
         "required": [
             "id",
             "framework_id",
-            "is_a_copy",
             "lot_id"
         ]
     }
--- a/requirements-dev.in
+++ b/requirements-dev.in
@@ -3,7 +3,7 @@
 pip-tools>=5.4.0
 
 # For schema generation
-alchemyjsonschema==0.5.0
+alchemyjsonschema==0.7.1
 
 # For tests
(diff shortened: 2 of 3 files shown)


tests/test_generate_model_schemas.py
```diff
@@ -62,7 +62,6 @@ def test_generate_schema():
         "required": [
             "id",
             "framework_id",
-            "is_a_copy",
```
bjgill, line 65: Why is this unexpected change safe?

gidsg: @bjgill I have done some more research into this, with assistance from @alex9smith please see the extended commit message in 06c65a8 for more details. See also definition of `is_a_copy` from https://github.com/alphagov/digitalmarketplace-api/blob/e781077b52797a963a2a0663bf8e2a5463813cb5/migrations/versions/900_add_brief_is_a_copy.py#L18

bjgill: That makes sense - thanks for the explanation

bjgill (approved):
:shipit:

PR #1132, opened 2021-05-05
Remove Python 2 import compatibility

bjgill:
We can simplify the imports - we now only support Python 3.6+, which has builtins. __builtins__ was a Python 2 thing.


gidsg (approved):
🆙 👍

PR #1143, opened 2021-06-03
Upgrade Flask to 1.1.4

alex9smith:
Pin requirements to 1.1 or greater, but less than 2. Flask 1.1 is the last planned v1 release before 2.0.0, so take patch versions of 1.1 for security but not 2.0 which we're not ready for yet.

Take new versions of utils and content-loader which have been updated for Flask 1.1

Also remove the pin for `itsdangerous` - Flask now controls the supported versions. See https://github.com/pallets/flask/blob/1.1.4/setup.py#L58

Also remove pyup comments - we don't need them now we're using dependabot.

Essentially the same as https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1436

https://trello.com/c/obh5gpQp/2250-update-to-flask-114

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -4,11 +4,11 @@
 digitalmarketplace-apiclient
 digitalmarketplace-utils
 
-Flask==1.0.4
+Flask>=1.1,<2
 Flask-Bcrypt==0.7.1
 Flask-Migrate==3.0.1
 Flask-SQLAlchemy>=2.4.1,<2.6.0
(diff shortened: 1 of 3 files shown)


alex9smith:
Utils had some breaking changes to logging in the tests, looks like these are the same thing https://github.com/alphagov/digitalmarketplace-utils/pull/650

bjgill (approved):
All looks sensible

PR #1146, opened 2021-06-10
Add optional Employment Status question to brief journey

alex9smith:
Reverts alphagov/digitalmarketplace-api#1145

Readd the employment status question that was added in #1145 but broke the functional tests. This time, however, make it optional temporarily while we release everything.

Without this change we can't release the new question to production without causing a period of time where buyers can't create new briefs. This is because both the briefs FE and the API require the changes in the other app to have already been deployed for everything to work.

With the question as optional in the API (but not in the briefs frontend), we can release the API first, then the briefs FE, then update the API to make the question required. This will allow us to release everything without breaking things for our buyers.

https://trello.com/c/l9lnd1GD/14-2-add-new-ir35-question-to-the-create-an-opportunity-journey

### DIFF
--- a/json_schemas/brief-responses-digital-outcomes-and-specialists-5-digital-outcomes.json
+++ b/json_schemas/brief-responses-digital-outcomes-and-specialists-5-digital-outcomes.json
@@ -3,8 +3,7 @@
   "additionalProperties": false,
   "properties": {
     "availability": {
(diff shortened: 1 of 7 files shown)


gidsg (approved):
Looks good, maybe change PR title to reflect question has been made optional

PR #1148, opened 2021-06-11
Make employment status question required

alex9smith:
In #1146 we added a new question to the brief creation journey. This question is required, but we left it as optional to make the process of releasing the question and updating the tests easier.

All that work's now done, so we can mark the question as required so the validation schema in the API now matches what's generated from the frameworks repo.

https://trello.com/c/l9lnd1GD/14-2-add-new-ir35-question-to-the-create-an-opportunity-journey


bjgill (approved):
:tada:

I assume this is safe for existing briefs that haven't answered the question?

alex9smith:
Yes, it only affects creating new briefs and those will have an answer because the briefs frontend requires an answer to this question before the new brief can be published.

PR #1156, opened 2021-06-30
Add error handling for social value 

gidsg:
The social weighting question introduces a new validation case which was not previously handled. We can differentiate between the other `oneOf` error handling as only one item is returned in the error path.  We then add some new logic to the `_translate_json_schema_error` function to translate this validation error to the desired error code.


domoscargin (approved):
LGTM!

PR #1161, opened 2021-07-20
Add initial G-Cloud 13 validation schemas

alex9smith:
These will change, but add in the initial schemas so we can work on G-Cloud 13 locally and test the changes we want to make.

Schemas generated from the script after https://github.com/alphagov/digitalmarketplace-frameworks/pull/689

https://trello.com/c/3SULJbtw/78-1-g-cloud-13-filter-a-person-from-lots-1-2-3-pricing-question


bjgill (approved):
This is basically unreviewable. Skimmed through it and didn't notice anything amiss. You used a script to generate it, so it should be safe to merge.

PR #1199, opened 2022-02-25
Add ability to move a brief from unsuccessful to closed

tim-s-ccs:
Add route which will move the supplier from unsuccessful
This is required as we cannot update unsuccessful Briefs through the API

Add tests for the various scenarios that can occur when the request is made

This ticket was created due to a [request in the service desk](https://crowncommercial.zendesk.com/agent/tickets/34755)

### DIFF
--- a/app/main/views/briefs.py
+++ b/app/main/views/briefs.py
@@ -546,3 +546,37 @@ def list_brief_services(brief_id):
 
     services = services.filter(Service.supplier_id == supplier.supplier_id)
     return list_result_response("services", services), 200
+
+
+@main.route('/briefs/<int:brief_id>/move-unsuccessful-brief-to-closed', methods=['POST'])
+def move_unsuccessful_brief_to_closed(brief_id):
+    """
+    Status changes:
+        Brief: unsuccessful -> closed
+    """
+    updater_json = validate_and_return_updater_request()
+
+    brief = Brief.query.filter(
+        Brief.id == brief_id
+    ).first_or_404()
+
+    if brief.status != 'unsuccessful':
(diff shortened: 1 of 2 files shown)


app/main/views/briefs.py
```diff
@@ -546,3 +546,37 @@ def list_brief_services(brief_id):
...
+    updater_json = validate_and_return_updater_request()
+
+    brief = Brief.query.filter(
+        Brief.id == brief_id
+    ).first_or_404()
+
+    if brief.status != 'unsuccessful':
+        abort(400, "Can only move an unsuccessful brief to close")
+
+    # Set status back to pending-awarded
```
bjgill, line 566: Is this true? It doesn't match with the method name and doc comment.

Otherwise, looks good.

tim-s-ccs: I'm not sure I understand what you mean? How does it not match the method name and doc comment, I'm not sure what I'm missing.

bjgill: Are we setting the status back to `pending-awarded` or `closed`? Or to those two mean the same thing?

bjgill: Just looked it up. `pending-awarded` is not a Brief status - it's a Brief Response status.

tim-s-ccs: Fortunately, we don't have to worry about brief responses with this as an unsuccessful brief will not have any

tberey (approved):

app/main/views/briefs.py
bjgill: Sorry - I should have been clearer. 

```suggestion
    # Set status back to closed
```

PR #1215, opened 2022-05-10
HM-5113: New seller application fix

dta-tyler:
This pull request removes a restriction which prevented new sellers from submitting an application.


dta-tyler:
This pull request was incorrectly created.

